PR #4, opened 2020-08-23
JMX metric gatherer

rmfitzpatrick:
**Description:**
Feature addition - Provides a new utility for running groovy scripts for the creation of custom, JMX attribute-based metrics.  It currently supports otlp, prometheus, logging, and inmemory (testing) exporters.

**Testing:**
Added basic unit and integration tests w/ associated make target and properties and circleci job.

**Documentation:**
Library has own documentation and link provided in project readme.

**Outstanding items:**
This addition doesn't provided asynchronous instrument support, and this will need to be added in a subsequent PR.
Publishing isn't included, and neither are report or ci artifacts.  These will be addressed in another PR, potentially w/ github actions adoption.
There is also a likely a need to be able to read configuration from stdin when this is used by an upcoming Collector extension.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,12 +2,18 @@
 #
 # Check https://circleci.com/docs/2.0/language-java/ for more details
 #
-version: 2
-jobs:
-  build:
+version: 2.1
+
+executors:
+  openjdk8:
     docker:
       - image: circleci/openjdk:8-jdk
-
+  integration:
+    machine:
+      image: ubuntu-1604:201903-01
+jobs:
+  build:
+    executor: openjdk8
       # Specify service dependencies here if necessary
       # CircleCI maintains a library of pre-built images
       # documented at https://circleci.com/docs/2.0/circleci-images/
@@ -51,3 +57,45 @@ jobs:
       - run:
           name: Build validation
           command: make build
+
+  integration-tests:
+    executor: integration
+      # Specify service dependencies here if necessary
+      # CircleCI maintains a library of pre-built images
+      # documented at https://circleci.com/docs/2.0/circleci-images/
+    # - image: circleci/postgres:9.4
+
+    working_directory: ~/repo
+
+    environment:
+      # Customize the JVM maximum heap limit
+      JVM_OPTS: -Xmx1g
+      GRADLE_OPTS: "-Dorg.gradle.workers.max=4"
+      TERM: dumb
+
+    steps:
+      - checkout
+
+      # Download and cache dependencies
+      - restore_cache:
+          keys:
+            - gradle-{{ checksum "build.gradle" }}
+            # fallback to using the latest cache if no exact match is found
+            - gradle-
+
+      - run:
+          name: Integration Tests
+          command: make integration-test
+
+      - save_cache:
+          paths:
+            - ~/.gradle
+          key: gradle-{{ checksum "build.gradle" }}
+workflows:
+  version: 2
+  verification:
+    jobs:
+      - build
+      - integration-tests:
+          requires:
+            - build
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -59,6 +59,7 @@ jobs:
           command: make build
 
   integration-tests:
+    # machine executor required for testcontainers docker access
     executor: integration
       # Specify service dependencies here if necessary
       # CircleCI maintains a library of pre-built images
--- a/contrib/jmx-metrics/README.md
+++ b/contrib/jmx-metrics/README.md
@@ -1,0 +1,112 @@
+# JMX Metric Gatherer
+
+This utility provides an easy framework for gathering and reporting metrics based on queried
+MBeans from a JMX server.  It loads a custom Groovy script and establishes a helpful, bound `otel`
+object with methods for obtaining MBeans and constructing synchronous OpenTelemetry instruments:
+
+### Usage
+
+```bash
+$ java -D<otel.jmx.metrics.property=value> -jar io.opentelemetry.contrib.jmx-metrics-<version>-all.jar [-config ./optional_config.properties]
+```
+
+##### `optional_config.properties` example
+
+```properties
+otel.jmx.metrics.service.url = service:jmx:rmi:///jndi/rmi://<my-jmx-host>:<my-jmx-port>/jmxrmi
+otel.jmx.metrics.groovy.script = /opt/script.groovy
+otel.jmx.metrics.interval.seconds = 5
+otel.jmx.metrics.exporter.type = otlp
+otel.jmx.metrics.exporter.otlp.endpoint = my-opentelemetry-collector:55680
+otel.jmx.metrics.username = my-username
+otel.jmx.metrics.password = my-password
+```
+
+##### `script.groovy` example
+
+```groovy
+import io.opentelemetry.common.Labels
+
+def loadMatches = otel.queryJmx("org.apache.cassandra.metrics:type=Storage,name=Load")
+def load = loadMatches.first()
+
+def lvr = otel.longValueRecorder(
+        "cassandra.storage.load",
+        "Size, in bytes, of the on disk data size this node manages",
+        "By", [myConstantLabelKey:"myConstantLabelValue"]
+)
+
+lvr.record(load.Count, Labels.of("myKey", "myVal"))
+```
+
+As configured in the example, this metric gatherer will configure an otlp gRPC metric exporter
+at the `otel.jmx.metrics.exporter.otlp.endpoint` and establish an MBean server connection using the
+provided `otel.jmx.metrics.service.url`. After loading the Groovy script whose path is specified
+via `otel.jmx.metrics.groovy.script`, it will then run the script on the specified
+`otel.jmx.metrics.interval.seconds` and export the resulting metrics.
+
+### JMX Query Helpers
+
+- `otel.queryJmx(String objectNameStr)`
+   - This method will query the connected JMX application for the given `objectName`, which can
+   include wildcards.  The return value will be a `List<GroovyMBean>` of zero or more
+   [`GroovyMBean` objects](http://docs.groovy-lang.org/latest/html/api/groovy/jmx/GroovyMBean.html),
+   which are conveniently wrapped to make accessing attributes on the MBean simple.
+   See http://groovy-lang.org/jmx.html for more information about their usage.
+
+- `otel.queryJmx(javax.management.ObjectName objectName)`
+   - This helper has the same functionality as its other signature, but takes an `ObjectName`
+   instance if constructing raw names is undesired.
+
+### OpenTelemetry Synchronous Instrument Helpers
+
+- `otel.doubleCounter(String name, String description, String unit, Map<String, String> labels)`
+
+- `otel.longCounter(String name, String description, String unit, Map<String, String> labels)`
+
+- `otel.doubleUpDownCounter(String name, String description, String unit, Map<String, String> labels)`
+
+- `otel.longUpDownCounter(String name, String description, String unit, Map<String, String> labels)`
+
+- `otel.doubleValueRecorder(String name, String description, String unit, Map<String, String> labels)`
+
+- `otel.longValueRecorder(String name, String description, String unit, Map<String, String> labels)`
+
+These methods will return a new or previously registered instance of the applicable metric
+instruments.  Each one provides three additional signatures  where labels, unit, and description
+aren't desired upon invocation.
+
+- `otel.<meterMethod>(String name, String description, String unit)` - `labels` are empty map.
+
+- `otel.<meterMethod>(String name, String description)` - `unit` is "1" and `labels` are empty map.
+
+- `otel.<meterMethod>(String name)` - `description` is empty string, `unit` is "1" and `labels` are empty map.
+
+### Compatibility
+
+This metric extension supports Java 7+, though SASL is only supported where
+`com.sun.security.sasl.Provider` is available.
+
+### Configuration
+
+The following properties are supported via the command line or specified config properties file `(-config)`.
+Those provided as command line properties take priority of those contained in a properties file.
+
+| Property | Required | Description |
+| ------------- | -------- | ----------- |
+| `otel.jmx.metrics.service.url` | **yes** | The service URL for the JMX RMI/JMXMP endpoint (generally of the form `service:jmx:rmi:///jndi/rmi://<host>:<port>/jmxrmi` or `service:jmx:jmxmp://<host>:<port>`).|
+| `otel.jmx.metrics.groovy.script` | **yes** | The path for the desired Groovy script. |
+| `otel.jmx.metrics.interval.milliseconds` | no | How often, in milliseconds, the Groovy script should be run and its resulting metrics exported. 10 by default. |
+| `otel.jmx.metrics.exporter.type` | no | The type of `io.opentelemetry.sdk.metrics.export.MetricExporter` to use: (`otlp`, `prometheus`, `inmemory`, `logging`).  `logging` by default. |
+| `otel.jmx.metrics.exporter.otlp.endpoint` | no | The otlp exporter endpoint to use, Required for `otlp`.  |
+| `otel.jmx.metrics.exporter.prometheus.host` | no | The prometheus collector server host. Default is `localhost`.  |
+| `otel.jmx.metrics.exporter.prometheus.port` | no | The prometheus collector server port. Default is `9090`.  |
+| `otel.jmx.metrics.username` | no | Username for JMX authentication, if applicable. |
+| `otel.jmx.metrics.password` | no | Password for JMX authentication, if applicable. |
+| `otel.jmx.metrics.keystore.path` | no | The key store path is required if client authentication is enabled on the target JVM. |
+| `otel.jmx.metrics.keystore.password` | no | The key store file password if required. |
+| `otel.jmx.metrics.keystore.type` | no | The key store type. |
+| `otel.jmx.metrics.truststore.path` | no | The trusted store path if the TLS profile is required. |
+| `otel.jmx.metrics.truststore.password` | no | The trust store file password if required. |
+| `otel.jmx.metrics.remote.profiles` | no | Supported JMX remote profiles are TLS in combination with SASL profiles: SASL/PLAIN, SASL/DIGEST-MD5 and SASL/CRAM-MD5. Thus valid `jmxRemoteProfiles` values are: `SASL/PLAIN`, `SASL/DIGEST-MD5`, `SASL/CRAM-MD5`, `TLS SASL/PLAIN`, `TLS SASL/DIGEST-MD5` and `TLS SASL/CRAM-MD5`. |
+| `otel.jmx.metrics.realm` | no | The realm is required by profile SASL/DIGEST-MD5. |
--- a/contrib/jmx-metrics/README.md
+++ b/contrib/jmx-metrics/README.md
@@ -13,13 +13,13 @@ $ java -D<otel.jmx.metrics.property=value> -jar io.opentelemetry.contrib.jmx-met
 ##### `optional_config.properties` example
 
 ```properties
-otel.jmx.metrics.service.url = service:jmx:rmi:///jndi/rmi://<my-jmx-host>:<my-jmx-port>/jmxrmi
-otel.jmx.metrics.groovy.script = /opt/script.groovy
-otel.jmx.metrics.interval.seconds = 5
-otel.jmx.metrics.exporter.type = otlp
-otel.jmx.metrics.exporter.otlp.endpoint = my-opentelemetry-collector:55680
-otel.jmx.metrics.username = my-username
-otel.jmx.metrics.password = my-password
+otel.jmx.service.url = service:jmx:rmi:///jndi/rmi://<my-jmx-host>:<my-jmx-port>/jmxrmi
+otel.jmx.groovy.script = /opt/script.groovy
+otel.jmx.interval.milliseconds = 5000
+otel.exporter = otlp
+otel.otlp.endpoint = my-opentelemetry-collector:55680
+otel.jmx.username = my-username
+otel.jmx.password = my-password
 ```
 
 ##### `script.groovy` example
@@ -40,10 +40,10 @@ lvr.record(load.Count, Labels.of("myKey", "myVal"))
 ```
 
 As configured in the example, this metric gatherer will configure an otlp gRPC metric exporter
-at the `otel.jmx.metrics.exporter.otlp.endpoint` and establish an MBean server connection using the
-provided `otel.jmx.metrics.service.url`. After loading the Groovy script whose path is specified
-via `otel.jmx.metrics.groovy.script`, it will then run the script on the specified
-`otel.jmx.metrics.interval.seconds` and export the resulting metrics.
+at the `otel.otlp.endpoint` and establish an MBean server connection using the
+provided `otel.jmx.service.url`. After loading the Groovy script whose path is specified
+via `otel.jmx.groovy.script`, it will then run the script on the specified
+`otel.jmx.interval.milliseconds` and export the resulting metrics.
 
 ### JMX Query Helpers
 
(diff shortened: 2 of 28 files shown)


rmfitzpatrick:
@keitwb, @jrcamp, @asuresh4, your feedback would be appreciated.

.circleci/config.yml
```diff
@@ -2,12 +2,18 @@
...
+version: 2.1
+
+executors:
+  openjdk8:
     docker:
       - image: circleci/openjdk:8-jdk
-
+  integration:
+    machine:
+      image: ubuntu-1604:201903-01
```
anuraaga, line 13: Can you add a comment on why we need a machine instead of a docker image? I guess it's because of testcontainers?

contrib/jmx-metrics/README.md
```diff
@@ -0,0 +1,112 @@
...
+        "cassandra.storage.load",
+        "Size, in bytes, of the on disk data size this node manages",
+        "By", [myConstantLabelKey:"myConstantLabelValue"]
+)
+
+lvr.record(load.Count, Labels.of("myKey", "myVal"))
+```
+
+As configured in the example, this metric gatherer will configure an otlp gRPC metric exporter
+at the `otel.jmx.metrics.exporter.otlp.endpoint` and establish an MBean server connection using the
```
anuraaga, line 43: We have standard environment variables for defining things like otlp endpoint, I guess we don't need to define custom ones?

contrib/jmx-metrics/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+### Configuration
+
+The following properties are supported via the command line or specified config properties file `(-config)`.
+Those provided as command line properties take priority of those contained in a properties file.
+
+| Property | Required | Description |
+| ------------- | -------- | ----------- |
+| `otel.jmx.metrics.service.url` | **yes** | The service URL for the JMX RMI/JMXMP endpoint (generally of the form `service:jmx:rmi:///jndi/rmi://<host>:<port>/jmxrmi` or `service:jmx:jmxmp://<host>:<port>`).|
+| `otel.jmx.metrics.groovy.script` | **yes** | The path for the desired Groovy script. |
```
anuraaga, line 98: It doesn't block initial submission but I think we do need to follow up with a configuration file approach to allow using without running scripts. I remember elasticsearch dropping support for Groovy scripts because it's too hard to sandbox it. Their use case was more dangerous because scripts could be populated in requests themselves, but there is definitely a lot more surface than just a yaml file. I think there will be many users who would love to have this functionality but not ok with executing arbitrary groovy files when scraping, especially if we allow the file to be updated without restart which I think currently we don't allow but would be a nice follow up.

Also, noticed elasticsearch supports mustache files which is interesting - can we expect much utility out of arbitrary groovy compared to a templating language?
https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-scripting.html

Food for thought, supporting groovy and hopefully additionally files seems fine though.

contrib/jmx-metrics/jmx-metrics.gradle
```diff
@@ -0,0 +1,79 @@
...
+sourceSets.main.java.srcDirs = []
+sourceSets.test.java.srcDirs = []
+
+jar {
+    manifest {
+        attributes('Main-Class': 'io.opentelemetry.contrib.jmxmetrics.JmxMetrics')
+    }
+}
+
+ext {
```
anuraaga, line 30: I guess it's similar to just use `def` instead of `ext` for a leaf project?

contrib/jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/OtelHelper.groovy
```diff
@@ -0,0 +1,157 @@
...
+
+import javax.management.MBeanServerConnection
+import javax.management.ObjectName
+
+class OtelHelper {
+
+    private final JmxClient jmxClient
+    private final GroovyUtils groovyUtils
+
+    final String scalar = '1'
```
anuraaga, line 34:
```suggestion
    private static final String scalar = '1'
```

contrib/jmx-metrics/src/main/java/io/opentelemetry/contrib/jmxmetrics/ClientCallbackHandler.java
```diff
@@ -0,0 +1,62 @@
...
+
+import javax.annotation.Nullable;
+import javax.security.auth.callback.Callback;
+import javax.security.auth.callback.CallbackHandler;
+import javax.security.auth.callback.NameCallback;
+import javax.security.auth.callback.PasswordCallback;
+import javax.security.auth.callback.UnsupportedCallbackException;
+import javax.security.sasl.RealmCallback;
+
+public class ClientCallbackHandler implements CallbackHandler {
```
anuraaga, line 27: Instead of adding our own options for authentication, can we just let users use the standard JMX pattern like `com.sun.management.jmxremote.password.file`? Less surface for security that we have to think about, the better, even if the current implementation is sane.

contrib/jmx-metrics/src/main/java/io/opentelemetry/contrib/jmxmetrics/GroovyRunner.java
```diff
@@ -0,0 +1,98 @@
...
+
+public class GroovyRunner {
+  private static final Logger logger = Logger.getLogger(GroovyRunner.class.getName());
+
+  private final GroovyShell gshell = new GroovyShell();
+
+  private final Script script;
+  private final GroovyUtils gutil;
+
+  GroovyRunner(final String groovyScript, final JmxClient jmxClient, final GroovyUtils gutil) {
```
anuraaga, line 41: Looks like we don't handle dynamic updates to the file - this is probably safer for groovy anyways. But it'd be nice in the future if we can handle a yaml file with dynamic updates.

contrib/jmx-metrics/jmx-metrics.gradle
```diff
@@ -0,0 +1,79 @@
...
+sourceSets.test.groovy.srcDirs = [
+    "src/test/java",
+    "src/test/groovy"
+]
+sourceSets.main.java.srcDirs = []
+sourceSets.test.java.srcDirs = []
+
+jar {
+    manifest {
+        attributes('Main-Class': 'io.opentelemetry.contrib.jmxmetrics.JmxMetrics')
```
anuraaga, line 26: I think instead we can add the application plugin to also have run task right?

Similarly, since this is an application, not an SDK, no reason to leave it Java 7, Java 8 should be good, or even Java 11

contrib/jmx-metrics/src/main/java/io/opentelemetry/contrib/jmxmetrics/GroovyUtils.java
```diff
@@ -0,0 +1,257 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opentelemetry.contrib.jmxmetrics;
+
+import io.opentelemetry.OpenTelemetry;
+import io.opentelemetry.common.Labels;
+import io.opentelemetry.common.Labels.Builder;
```
anuraaga, line 21: Avoid importing generic names like Builder

contrib/jmx-metrics/src/main/java/io/opentelemetry/contrib/jmxmetrics/JmxClient.java
```diff
@@ -0,0 +1,130 @@
...
+    this.url = new JMXServiceURL(config.serviceUrl);
+    this.username = config.username;
+    this.password = config.password;
+    this.realm = config.realm;
+    this.jmxRemoteProfiles = config.jmxRemoteProfiles;
+  }
+
+  private static void setSystemProperties(final JmxConfig config) {
+    if (!JmxConfig.isBlank(config.keyStorePath)) {
+      System.setProperty("javax.net.ssl.keyStore", config.keyStorePath);
```
anuraaga, line 56: If we're only forwarding the properties anyways, I think we don't need to define our own property names, users should generally be more familiar with the standard props anyways. And we're not preventing collission if we copy them in so there doesn't seem to be value in the otel versions.

contrib/jmx-metrics/src/main/java/io/opentelemetry/contrib/jmxmetrics/JmxMetrics.java
```diff
@@ -0,0 +1,132 @@
...
+    JmxClient jmxClient;
+    try {
+      jmxClient = new JmxClient(config);
+    } catch (MalformedURLException e) {
+      throw new ConfigureError("Malformed serviceUrl: ", e);
+    }
+
+    runner = new GroovyRunner(config.groovyScript, jmxClient, new GroovyUtils(config));
+
+    timer.scheduleAtFixedRate(
```
anuraaga, line 52: `scheduleWithFixedDelay` since it's an interval

anuraaga: Let's use singlethread executor instead of timer, no need for the wrapping

contrib/jmx-metrics/src/main/java/io/opentelemetry/contrib/jmxmetrics/JmxMetrics.java
```diff
@@ -0,0 +1,132 @@
...
+    logger.info("Shutting down JmxMetrics Groovy runner and exporting final metrics.");
+    timer.cancel();
+    runner.flush();
+  }
+
+  private static JmxConfig getConfigFromArgs(final String[] args) {
+    if (args.length != 0 && (args.length != 2 || !args[0].equalsIgnoreCase("-config"))) {
+      System.out.println(
+          "Usage: java io.opentelemetry.contrib.jmxmetrics.JmxMetrics "
+              + "-config <path_to_config.properties>");
```
anuraaga, line 88: In most of our configuration we support all three of environment variables, system properties, and config file. I think if we wanted to keep it simple for now, we should start with system properties to reduce the number of parts.

contrib/jmx-metrics/src/main/java/io/opentelemetry/contrib/jmxmetrics/JmxMetrics.java
```diff
@@ -0,0 +1,132 @@
...
+  public static void main(final String[] args) {
+    JmxConfig config = getConfigFromArgs(args);
+    config.validate();
+
+    final JmxMetrics jmxMetrics = new JmxMetrics();
+    try {
+      jmxMetrics.start(config);
+    } catch (ConfigureError e) {
+      logger.severe(e.getMessage());
+      System.exit(1);
```
anuraaga, line 120: Think it's fine to throw the exception instead of logging / exiting

contrib/jmx-metrics/README.md
anuraaga: Also related to this is, is there any default export we can do that makes sense when no groovy script or even config file is provided? It'd be sweet to be able to get the jar, set the system property to point at service.url, and get something without messing around with anything else.

.circleci/config.yml
anuraaga: Also FWIW, testcontainers work well on github actions :) We should migrate this build sooner than latter

rmfitzpatrick: updated integration-test job w/ comment.  I agree and won't plan on any additional major CI additions without having moved to actions beforehand.

contrib/jmx-metrics/README.md
rmfitzpatrick: This seems like the right approach so I've adopted otlp properties directly.

contrib/jmx-metrics/src/main/java/io/opentelemetry/contrib/jmxmetrics/JmxClient.java
rmfitzpatrick: This is much more straightforward, and I've adopted this request.

contrib/jmx-metrics/src/main/java/io/opentelemetry/contrib/jmxmetrics/JmxMetrics.java
rmfitzpatrick: I'll need to add properties file content via stdin next, and think it's better ux to accept a file than to require everything (including password) over the command line.  Are you ok w/ me leaving as is?  It's not a required option as is.

contrib/jmx-metrics/README.md
rmfitzpatrick: A standardized config format - yes, this would be helpful and you've made a strong case for this as potentially a more accessible use case for several scenarios.  In general security concerns are obviously a high priority and planning some key or integrity checks will likely be a requirement for any runtime configuration or script update features.

default export - this would be handy and I think porting elements of the runtime metrics extension could be a reasonable candidate: https://github.com/open-telemetry/opentelemetry-java/tree/master/extensions/runtime_metrics/src/main/java/io/opentelemetry/extensions/metrics/runtime

contrib/jmx-metrics/jmx-metrics.gradle
rmfitzpatrick: removed

contrib/jmx-metrics/src/main/java/io/opentelemetry/contrib/jmxmetrics/JmxMetrics.java
rmfitzpatrick: Updated

contrib/jmx-metrics/src/main/java/io/opentelemetry/contrib/jmxmetrics/ClientCallbackHandler.java
rmfitzpatrick: I'm not certain, but I believe that file is only for setting up the mbean server, similar to the access file, and establishing a client connection requires the connector to use the `jmx.remote.credentials` (not a property?) string array entry.

rmfitzpatrick: https://docs.oracle.com/cd/E19698-01/816-7609/6mdjrf86t/index.html

contrib/jmx-metrics/jmx-metrics.gradle
rmfitzpatrick: Updated w/ application plugin.

As to java version, would it be ok to defer this decision w/ a tracking issue?

contrib/jmx-metrics/src/main/java/io/opentelemetry/contrib/jmxmetrics/ClientCallbackHandler.java
```diff
@@ -0,0 +1,62 @@
...
+  private final String username;
+  @Nullable private char[] password;
```
anuraaga, line 29:
```suggestion
  @Nullable private final char[] password;
```

contrib/jmx-metrics/src/main/java/io/opentelemetry/contrib/jmxmetrics/JmxConfig.java
```diff
@@ -0,0 +1,136 @@
...
+
+  private static final String PROMETHEUS_HOST = "prometheus.host";
+  private static final String PROMETHEUS_PORT = "prometheus.port";
+
+  private static final String JMX_USERNAME = "jmx.username";
+  private static final String JMX_PASSWORD = "jmx.password";
+  private static final String JMX_REMOTE_PROFILES = "jmx.remote.profiles";
+  private static final String JMX_REALM = "jmx.realm";
+
+  public String serviceUrl;
```
anuraaga, line 38: Looks like all can be `final`

contrib/jmx-metrics/src/main/java/io/opentelemetry/contrib/jmxmetrics/JmxMetrics.java
```diff
@@ -0,0 +1,120 @@
...
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+public class JmxMetrics {
+  private static final Logger logger = Logger.getLogger(JmxMetrics.class.getName());
+
+  private final ScheduledExecutorService exec = Executors.newSingleThreadScheduledExecutor();
+  private GroovyRunner runner;
```
anuraaga, line 34:
```suggestion
  private final GroovyRunner runner;
```

anuraaga (approved):
Thanks! Just small nits noticed some fields can probably be final

contrib/jmx-metrics/src/main/java/io/opentelemetry/contrib/jmxmetrics/JmxConfig.java
rmfitzpatrick: updated

rmfitzpatrick:
Thanks for the detailed feedback @anuraaga!  Keeping this open for others to chime in.

settings.gradle
```diff
@@ -1,6 +1,7 @@
 rootProject.name = 'opentelemetry-java-contrib'
 
 include ':example'
+include ':jmx-metrics'
```
pmcollins, line 4: I cloned this branch and opened it in IntelliJ, and things appear to be mostly fine, but `contrib/jmx-metrics` isn't recognized out of the box as a project module by IntelliJ even though its sibling directory, `contrib/example`, is. Not sure if this is a quirk in IntelliJ and/or if there's a Gradle (which I'm not an expert in) config that can be tweaked.

rmfitzpatrick: I've found enabling auto reload can be helpful, but haven't had any issues personally (can run all expected tasks): https://www.jetbrains.com/help/idea/work-with-gradle-projects.html#auto_reload

Any errors or logs pop up?

pmcollins: That fixed it, thanks!

contrib/jmx-metrics/src/main/java/io/opentelemetry/contrib/jmxmetrics/GroovyUtils.java
```diff
@@ -0,0 +1,255 @@
...
+   * @param instrumentationName - meter's instrumentationName
+   * @param instrumentationVersion - meter's instrumentationVersion
+   */
+  public GroovyUtils(
+      final JmxConfig config,
+      final String instrumentationName,
+      final String instrumentationVersion) {
+    meter = OpenTelemetry.getMeter(instrumentationName, instrumentationVersion);
+
+    switch (config.exporterType.toLowerCase()) {
```
pmcollins, line 61: Maybe exporterType should be an enum?

contrib/jmx-metrics/src/main/java/io/opentelemetry/contrib/jmxmetrics/GroovyUtils.java
```diff
@@ -0,0 +1,255 @@
...
+import io.opentelemetry.sdk.metrics.export.MetricExporter;
+import io.opentelemetry.sdk.metrics.export.MetricProducer;
+import io.prometheus.client.exporter.HTTPServer;
+import java.io.IOException;
+import java.util.Collection;
+import java.util.Map;
+
+public class GroovyUtils {
+
+  public Meter meter;
```
pmcollins, line 43: private final?

contrib/jmx-metrics/src/main/java/io/opentelemetry/contrib/jmxmetrics/GroovyUtils.java
```diff
@@ -0,0 +1,255 @@
...
+import io.opentelemetry.sdk.OpenTelemetrySdk;
+import io.opentelemetry.sdk.metrics.data.MetricData;
+import io.opentelemetry.sdk.metrics.export.MetricExporter;
+import io.opentelemetry.sdk.metrics.export.MetricProducer;
+import io.prometheus.client.exporter.HTTPServer;
+import java.io.IOException;
+import java.util.Collection;
+import java.util.Map;
+
+public class GroovyUtils {
```
pmcollins, line 41: Wondering if there's a name that better indicates what this class does.

contrib/jmx-metrics/src/main/java/io/opentelemetry/contrib/jmxmetrics/GroovyRunner.java
```diff
@@ -0,0 +1,98 @@
...
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import org.codehaus.groovy.control.CompilationFailedException;
+
+public class GroovyRunner {
+  private static final Logger logger = Logger.getLogger(GroovyRunner.class.getName());
+
+  private final GroovyShell gshell = new GroovyShell();
```
pmcollins, line 36: Looks like this can be a local variable.

contrib/jmx-metrics/src/main/java/io/opentelemetry/contrib/jmxmetrics/GroovyRunner.java
```diff
@@ -0,0 +1,98 @@
...
+    Binding binding = new Binding();
+    binding.setVariable("log", logger);
+
+    OtelHelper otelHelper = new OtelHelper(jmxClient, gutil);
+    binding.setVariable("otel", otelHelper);
+
+    script.setBinding(binding);
+  }
+
+  static String getFileAsString(final String fileName) throws IOException {
```
pmcollins, line 68: Looks like Guava is available, so you can use `Files.readlines`

contrib/jmx-metrics/src/main/java/io/opentelemetry/contrib/jmxmetrics/ConfigureError.java
```diff
@@ -0,0 +1,29 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opentelemetry.contrib.jmxmetrics;
+
+public class ConfigureError extends RuntimeException {
```
pmcollins, line 19: ConfigurationException?

contrib/jmx-metrics/src/main/java/io/opentelemetry/contrib/jmxmetrics/JmxConfig.java
```diff
@@ -0,0 +1,135 @@
...
+    password = getProperty(JMX_PASSWORD, "");
+    remoteProfiles = getProperty(JMX_REMOTE_PROFILES, "");
+    realm = getProperty(JMX_REALM, "");
+  }
+
+  public JmxConfig() {
+    this(new Properties());
+  }
+
+  private String getProperty(final String key, final String dfault) {
```
pmcollins, line 83: This method may not be needed because there's already a `getProperty(String key, String defaultValue)` method in the `Properties` class.

contrib/jmx-metrics/src/main/java/io/opentelemetry/contrib/jmxmetrics/JmxClient.java
```diff
@@ -0,0 +1,111 @@
...
+    this.realm = config.realm;
+    this.remoteProfiles = config.remoteProfiles;
+  }
+
+  private MBeanServerConnection ensureConnected() {
+    if (jmxConn != null) {
+      try {
+        return jmxConn.getMBeanServerConnection();
+      } catch (IOException e) {
+        // Attempt to connect with authentication below.
```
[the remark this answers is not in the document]
rmfitzpatrick: If subsequent auth-based connection fails there will be a logged exception.

contrib/jmx-metrics/src/main/java/io/opentelemetry/contrib/jmxmetrics/JmxConfig.java
```diff
@@ -0,0 +1,135 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opentelemetry.contrib.jmxmetrics;
+
+import java.util.Properties;
+
+public class JmxConfig {
```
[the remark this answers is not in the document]
rmfitzpatrick: I think it's likely that subsequent utilities may want to extend this, but in this stage it's fine for default access?

contrib/jmx-metrics/src/main/java/io/opentelemetry/contrib/jmxmetrics/JmxConfig.java
```diff
@@ -0,0 +1,135 @@
...
+  protected static final String PREFIX = "otel.";
```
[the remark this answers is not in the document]
rmfitzpatrick: updated

contrib/jmx-metrics/src/main/java/io/opentelemetry/contrib/jmxmetrics/JmxConfig.java
```diff
@@ -0,0 +1,135 @@
...
+
+  protected static final String PROMETHEUS_HOST = PREFIX + "prometheus.host";
+  protected static final String PROMETHEUS_PORT = PREFIX + "prometheus.port";
+
+  protected static final String JMX_USERNAME = PREFIX + "jmx.username";
+  protected static final String JMX_PASSWORD = PREFIX + "jmx.password";
+  protected static final String JMX_REMOTE_PROFILES = PREFIX + "jmx.remote.profiles";
+  protected static final String JMX_REALM = PREFIX + "jmx.realm";
+
+  public final String serviceUrl;
```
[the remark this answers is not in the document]
rmfitzpatrick: updated

contrib/jmx-metrics/src/main/java/io/opentelemetry/contrib/jmxmetrics/JmxConfig.java
```diff
@@ -0,0 +1,135 @@
...
+    }
+  }
+
+  /**
+   * Will determine if parsed config is complete, setting any applicable defaults.
+   *
+   * @throws ConfigureError - Thrown if a configuration value is missing or invalid.
+   */
+  public void validate() throws ConfigureError {
+    if (isBlank(this.serviceUrl)) {
```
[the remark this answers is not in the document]
rmfitzpatrick: I think it improves readability but updated since it's likely more conventional not to use it.

contrib/jmx-metrics/src/main/java/io/opentelemetry/contrib/jmxmetrics/JmxMetrics.java
```diff
@@ -0,0 +1,117 @@
...
+import java.io.InputStream;
+import java.net.MalformedURLException;
+import java.util.Properties;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+public class JmxMetrics {
```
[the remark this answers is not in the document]
rmfitzpatrick: made default since nothing but the lib as it stands really needs it and no subclasses are anticipated at this time.

contrib/jmx-metrics/src/main/java/io/opentelemetry/contrib/jmxmetrics/GroovyUtils.java
rmfitzpatrick: `GroovyMetricEnvironment` seems more accurate.

contrib/jmx-metrics/src/main/java/io/opentelemetry/contrib/jmxmetrics/GroovyRunner.java
rmfitzpatrick: This was in anticipation of reading from stdin, and we're not actually using that lib (porting remnant).  Removed the dep.

PR #9, opened 2020-09-14
jmx metrics: Allow properties file contents from stdin

rmfitzpatrick:
**Description:**
Feature addition - These changes introduce the ability to provide jmx metric gatherer configuration properties over stdin when using `-` as the file identifier.

**Testing:**

Parameterized jmx integration tests to use properties file directly and piped.

**Documentation:**

Updated jmx metric gatherer readme.

**Outstanding items:**

These changes don't introduce any dynamic property configuration mechanism and the functionality of the utility is still determined wholly at startup.

### DIFF
--- a/contrib/jmx-metrics/src/main/java/io/opentelemetry/contrib/jmxmetrics/JmxMetrics.java
+++ b/contrib/jmx-metrics/src/main/java/io/opentelemetry/contrib/jmxmetrics/JmxMetrics.java
@@ -16,6 +16,7 @@
 
 package io.opentelemetry.contrib.jmxmetrics;
 
+import java.io.DataInputStream;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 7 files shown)


contrib/jmx-metrics/src/main/java/io/opentelemetry/contrib/jmxmetrics/JmxMetrics.java
```diff
@@ -75,18 +76,27 @@ private static JmxConfig getConfigFromArgs(final String[] args) {
...
 
     Properties props = new Properties();
     if (args.length == 2) {
-      try (InputStream is = new FileInputStream(args[1])) {
-        props.load(is);
-      } catch (IOException e) {
-        System.out.println(
-            "Failed to read config properties file at '" + args[1] + "': " + e.getMessage());
-        System.exit(1);
+      if (args[1].trim().equals("-")) { // parse from stdin
```
pmcollins, line 85: Nit: a possibility here is to put these blocks into functions with descriptive names that also replace their comments: e.g. `parseFromStdin` and `parseFromPath` below.

pmcollins (approved):

contrib/jmx-metrics/src/main/java/io/opentelemetry/contrib/jmxmetrics/JmxMetrics.java
rmfitzpatrick: good idea, updated

Issue #8, opened 2020-09-14
Versioning


rmfitzpatrick:
We need to establish versioning and release support practices.  At this time it seems* tracking w/ opentelemetry-java versions similar to opentelemetry-java-instrumentation is likely not appropriate given our reliance on their released, stable versions, as well as our probable release cadence.  Basing our releases on their latest versions in some capacity could provide advantages in transparency and expected sdk functionality though, so this approach may still be desired.

When we begin releasing contributions this or whatever convention is desired should be adopted, though initial 0.0.x may be necessary.

PR #10, opened 2020-09-18
Add local maven and oss-snapshot-local artifactory publishing

rmfitzpatrick:
**Description:**

Feature addition - These changes add maven and artifactory publishing task configurations for contrib projects, and source them in the jmx-metrics project.  They are based on current otel java and java-instrumentation publishing and include:

* Setting 0.0.1-SNAPSHOT version
* Local maven publishing via `gradle publish`, which publishes artifacts to `build/repo` directory.
* Artifactory snapshot publishing via `gradle artifactoryPublish`, which publishes to https://oss.jfrog.org/artifactory.

**Testing:**

Manually tested: https://oss.jfrog.org/artifactory/oss-snapshot-local/io/opentelemetry/contrib/opentelemetry-java-contrib-jmx-metrics/0.0.1-SNAPSHOT/

**Documentation:**

Added a RELEASE.md file w/ details.

**Outstanding items:**

No automation has been added to use these tasks, and I plan on moving to gh actions before this is implemented.  Also does not resolve versioning approach - https://github.com/open-telemetry/opentelemetry-java-contrib/issues/8.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,3 +27,11 @@ test:
 .PHONY: integration-test
 integration-test:
 	./gradlew -Pojc.integration.tests=true test
+
+.PHONY: maven-publish
+maven-publish:
+	./gradlew publish
+
+.PHONY: oss-snapshot
+oss-snapshot:
+	./gradlew artifactoryPublish
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,42 @@
+# Releasing OpenTelemetry Java Contrib Artifacts
+
+This project currently has two make targets capable of preparing and releasing artifacts: `maven-publish`
+and `oss-snapshot`.  In order for you to register your contributed project to be published by these commands,
+you must apply the provided publish script plugin in your subproject's gradle file:
+
+```groovy
+apply from: project.publish
+```
+
+To set the desired library name and description for your artifact's pom file, you can set the `libraryName`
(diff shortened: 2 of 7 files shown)


rmfitzpatrick:
@iNikem if interested, your expertise would be appreciated.

@asuresh4 fyi

Makefile
```diff
@@ -27,3 +27,11 @@ test:
 .PHONY: integration-test
 integration-test:
 	./gradlew -Pojc.integration.tests=true test
+
+.PHONY: maven-publish
+maven-publish:
```
iNikem, line 32: Not specific for this PR, but why do you use make? :)

iNikem (approved):

Makefile
iNikem: it is certainly not "more trouble", just IMO, if "users should be using `./gradlew -PmagicFlag=magic -x check -x randomTask assemble specialTest ` instead of `./gradlew build`", then we should fix gradle build, not introduce another layer on top of it :)

trask (approved):

Makefile
rmfitzpatrick: We're hopefully not already at diminishing returns, but I appreciate your perspective: https://github.com/open-telemetry/opentelemetry-java-contrib/issues/11.

Issue #11, opened 2020-09-21
Make v. Gradle tasks


rmfitzpatrick:
Re: https://github.com/open-telemetry/opentelemetry-java-contrib/pull/10#discussion_r491269068 the adoption of Make in this project may be a point of contention or usability hindrance.  Usage of Make isn't standardized across Otel, but is used.

Available:
https://github.com/open-telemetry/opentelemetry-java
https://github.com/open-telemetry/opentelemetry-proto
https://github.com/open-telemetry/opentelemetry-go
https://github.com/open-telemetry/opentelemetry-go-contrib
https://github.com/open-telemetry/opentelemetry-php
https://github.com/open-telemetry/opentelemetry-collector
https://github.com/open-telemetry/opentelemetry-collector-contrib
https://github.com/open-telemetry/opentelemetry.io

Unavailable:
https://github.com/open-telemetry/opentelemetry-java-instrumentation
https://github.com/open-telemetry/opentelemetry-js
https://github.com/open-telemetry/opentelemetry-js-contrib
https://github.com/open-telemetry/opentelemetry-python (except docs)
https://github.com/open-telemetry/opentelemetry-ruby
https://github.com/open-telemetry/opentelemetry-cpp
https://github.com/open-telemetry/opentelemetry-rust
https://github.com/open-telemetry/opentelemetry-dotnet
https://github.com/open-telemetry/opentelemetry-dotnet-contrib
https://github.com/open-telemetry/opentelemetry-erlang
https://github.com/open-telemetry/opentelemetry-erlang-api

Of the projects using it, four are in golang and two others are polyglot, where Make could be a natural choice.  Java projects are split in usage.

Should we continue to use Make or should we only adopt direct `gradlew` usage?  In the hope to avoid bikeshedding, we should be mainly concerned where the stability, maintainability, and general accessibility of the project are in potential jeopardy.

My 2¢:  I like the relatively intuitive nature of Make as a standardize interface and don't see its usage as an issue.  It also plays nicely w/ my shell's completion/suggestion plugin where gradle does not.  If gradle is adopted, then at the least we should define and document high level tasks similar to the current make targets to prevent loaded invocations.

trask:
> Available:
> https://github.com/open-telemetry/opentelemetry-java

I'm surprised Makefile is used in Java repo, @jkwatson / @anuraaga?

jkwatson:
> > Available:
[...]

PR #15, opened 2020-10-14
Adopt Gradle over Make

rmfitzpatrick:
**Description:**
Feature addition - Removes Make usage from project in favor of gradle.

**Existing Issue(s):**

https://github.com/open-telemetry/opentelemetry-java-contrib/issues/11

**Testing:**

Should be picked up by existing checks

**Documentation:**

Updated existing docs as necessary


carlosalberto (approved):

build.gradle
```diff
@@ -21,10 +21,19 @@ allprojects {
         jcenter()
         mavenCentral()
     }
+}
+
+task lint(dependsOn: ['spotlessApply'])
```
anuraaga, line 26: Is it just `lint` => `spotlessCheck`? I'm not sure I've seen lint refer to a fixing step before.

anuraaga (approved):
Thanks!

build.gradle
rmfitzpatrick: Thanks, this is term misusage on my end.  I've associated linting w/ the fix features of eslint and flake8.  Corrected for `lint` to be the check and `format` for spotlessApply.

rmfitzpatrick:
Bypassing EasyCLA, which appears to not be working across otel projects atm.

PR #18, opened 2020-11-04
Add asynchronous instruments to JMX Metric Gatherer

rmfitzpatrick:
**Description:**
Feature addition - Adds missing asynchronous instrument support to JMX Metric Gatherer and associated `otel` helper.

**Testing:**

Added unit test suite and additional parameterization to instrument helper test suite.

**Documentation:**

Updated readme

**Outstanding items:**

None of the existing target scripts adopt these changes, as this should be done in a subsequent PR.

### DIFF
--- a/contrib/jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/InstrumentHelper.groovy
+++ b/contrib/jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/InstrumentHelper.groovy
@@ -17,9 +17,16 @@
 package io.opentelemetry.contrib.jmxmetrics
 
 import io.opentelemetry.metrics.DoubleCounter
+import io.opentelemetry.metrics.DoubleSumObserver
 import io.opentelemetry.metrics.DoubleUpDownCounter
+import io.opentelemetry.metrics.DoubleUpDownSumObserver
+import io.opentelemetry.metrics.DoubleValueObserver
 import io.opentelemetry.metrics.LongCounter
+import io.opentelemetry.metrics.LongSumObserver
 import io.opentelemetry.metrics.LongUpDownCounter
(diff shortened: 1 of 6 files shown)


contrib/jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/InstrumentHelper.groovy
```diff
@@ -89,21 +96,48 @@ class InstrumentHelper {
...
             if (value instanceof CompositeData) {
                 value.getCompositeType().keySet().each { key ->
                     def val = value.get(key)
                     def updatedInstrumentName = "${instrumentName}.${key}"
                     def labels = getLabels(mbean, labelFuncs)
                     def inst = instrument(updatedInstrumentName, description, unit)
+                    println "InstrumentHelper.update (composite) - ${inst}"
                     logger.fine("Recording ${updatedInstrumentName} - ${inst} w/ ${val} - ${labels}")
-                    updateInstrumentWithValue(inst, val, labels)
+                    if (!instToUpdates.containsKey(inst)) {
```
pmcollins, line 113: Nit: looks like you could factor out these four lines with line 123-126 if you wanted.

pmcollins (approved):
LGTM, just one small nit.

contrib/jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/InstrumentHelper.groovy
rmfitzpatrick: Since this section is in an inner iterator I think I'd have to store everything in an additional structure for final `instToUpdates` modification at the end and I'm not sure that's worth the reduced redundancy.  I'm generally a "rule of three" follower though...

Issue #24, opened 2020-11-13
JMX: Better modularize integration testing patterns


rmfitzpatrick:
**Is your feature request related to a problem? Please describe.**
Currently adding integration tests to the JMX Metric Gatherer consists largely of repeating patterns and expanding the `configureContainers()` method of https://github.com/open-telemetry/opentelemetry-java-contrib/blob/master/contrib/jmx-metrics/src/test/groovy/io/opentelemetry/contrib/jmxmetrics/IntegrationTest.groovy.

**Describe the solution you'd like**
There should be a clear format for expected metrics and a more intuitive and maintainable system of registering expected target test containers and determining their startup behavior.  Ideally this will allow easier test authoring and maintenance as more target systems are added.

**Additional context**
The `assertTraces` and related helpers in [opentelemetry-java-instrumentation](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/c6eee706604a9edf35758d90dc974e4ee82a0f0a/instrumentation/kafka-clients-0.11/src/test/groovy/KafkaClientTest.groovy#L186) could be one inspiration for this effort.

rmfitzpatrick:
I think this has effectively been resolved thanks to the refactoring efforts of @anuraaga.

PR #31, opened 2021-04-21
Move from CircleCI to GitHub Actions

rmfitzpatrick:
**Description:**
These changes remove the CircleCI config and add an equivalent GitHub Actions workflow.

**Existing Issue(s):**
resolves https://github.com/open-telemetry/opentelemetry-java-contrib/issues/30

**Testing:**
Some integration test flake needed to be addressed for the seemingly underpowered GH runners compared to Circle's executor.  Also appears we're not able to use the default rmi port on their runners.

**Documentation:**
Updated build shield on main readme.

**Outstanding items:**
These changes are lateral, but release actions should be added in the future.


rmfitzpatrick:
@bogdandrutu any chance you'd be able to disable/stop building Circle CI for this project when this lands?  I don't have admin access.

README.md
```diff
@@ -1,5 +1,5 @@
 # OpenTelemetry Java Contrib
-[![Build status](https://img.shields.io/circleci/build/github/open-telemetry/opentelemetry-java-contrib?style=for-the-badge)](https://app.circleci.com/pipelines/github/open-telemetry/opentelemetry-java-contrib?branch=master)
+[![Build](https://github.com/open-telemetry/opentelemetry-java-contrib/actions/workflows/pr-build.yml/badge.svg)](https://github.com/open-telemetry/opentelemetry-java-contrib/actions/workflows/pr-build.yml)
```
breedx-splk, line 2: 👍🏻

contrib/jmx-metrics/src/test/groovy/io/opentelemetry/contrib/jmxmetrics/OtlpIntegrationTests.groovy
```diff
@@ -25,13 +25,15 @@ import io.opentelemetry.proto.metrics.v1.DoubleHistogramDataPoint
...
+import spock.lang.Retry
 import spock.lang.Timeout
 import spock.lang.Unroll
 
 @Requires({
     System.getProperty('ojc.integration.tests') == 'true'
 })
-@Timeout(60)
+@Timeout(90)
+@Retry
```
breedx-splk, line 36: Why the `@Retry`?  This makes me wonder if a bad/flaky test could be hiding among the retries...

(Same question applies a few more times below)

contrib/jmx-metrics/src/test/resources/otlp_config.properties
```diff
@@ -1,4 +1,4 @@
-otel.jmx.interval.milliseconds = 3000
+otel.jmx.interval.milliseconds = 5000
```
breedx-splk, line 1: Is this a polling interval?  3s is already pretty slow...was this increase absolutely necessary?

breedx-splk (approved):
This seems good to me.  I had a couple of small questions, but no show stoppers.  Thanks!

contrib/jmx-metrics/src/test/groovy/io/opentelemetry/contrib/jmxmetrics/OtlpIntegrationTests.groovy
rmfitzpatrick: You're right, this is hiding flake where only partial sets of expected metrics are received and the test fails [here](https://github.com/open-telemetry/opentelemetry-java-contrib/blob/main/contrib/jmx-metrics/src/test/groovy/io/opentelemetry/contrib/jmxmetrics/target-systems/CassandraIntegrationTests.groovy#L68).  There is no "wait for desired metrics" feature in our mock collector at the moment.  I had wanted to add this functionality as part of addressing https://github.com/open-telemetry/opentelemetry-java-contrib/issues/24.

contrib/jmx-metrics/src/test/resources/otlp_config.properties
rmfitzpatrick: No, but it was helpful w/ https://github.com/open-telemetry/opentelemetry-java-contrib/pull/31#discussion_r617864459.  Reverting.

.github/workflows/pr-build.yml
```diff
@@ -0,0 +1,98 @@
...
+          remote-build-cache-proxy-enabled: false
+      - name: ./gradlew test
+        uses: burrunan/gradle-cache-action@v1
+        id: gradle-test
+        timeout-minutes: 5
+        with:
+          arguments: --stacktrace --max-workers 2 test
+          gradle-version: wrapper
+          job-id: java-contrib
+          remote-build-cache-proxy-enabled: false
```
trask, line 48: I'm don't know GitHub Actions well, so likely don't know what I'm talking about 😁, but you might be able to have just one `uses: burrunan/gradle-cache-action@v1`, and still have multiple commands after that, e.g.

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/f956a58c0abd44391a5df2bf54748cfa65277e07/.github/workflows/pr.yaml#L124-L130

trask (approved):

.github/workflows/pr-build.yml
rmfitzpatrick: I thought so too, but didn't see `UP-TO-DATE` markers in the logs for repeated tasks without doing so.  I'm very new to gh actions so will keep looking into it.

.github/workflows/pr-build.yml
```diff
@@ -0,0 +1,98 @@
...
+jobs:
+  build:
+    name: build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - id: setup-java-8
+        name: Setup Java 8
```
anuraaga, line 20: We've been generally building with Java 11 (though if the build script isn't forcing Java 8 bytecode that would need to be fixed)

.github/workflows/pr-build.yml
```diff
@@ -0,0 +1,98 @@
...
+        name: Save unit test results
+        if: always()
+        with:
+          name: test-results
+          path: contrib/jmx-metrics/build/reports/tests/test
+      - name: ./gradlew build
+        uses: burrunan/gradle-cache-action@v1
+        id: gradle-build
+        timeout-minutes: 5
+        with:
```
anuraaga, line 59: build runs assemble and test, we'd generally just have one step running `build`

.github/workflows/pr-build.yml
```diff
@@ -0,0 +1,98 @@
...
+          fetch-depth: 0
+      - id: setup-java-8
+        name: Setup Java 8
+        uses: actions/setup-java@v2
+        with:
+          distribution: adopt
+          java-version: 8
+      - uses: burrunan/gradle-cache-action@v1
+        name: restore gradle cache
```
anuraaga, line 26: We don't need this step, the other ones use cache transparently without explicit restore

.github/workflows/pr-build.yml
rmfitzpatrick: I liked the distinction between assemble and test to reduce the surface area of issues, with a final build verification step.  I've combined the test and build steps, if the remaining distinction is ok with you (can update otherwise).

.github/workflows/pr-build.yml
rmfitzpatrick: Removed

.github/workflows/pr-build.yml
rmfitzpatrick: Updated, and the target version is 8: https://github.com/open-telemetry/opentelemetry-java-contrib/blob/3de42160e92e0bd0ab876d835ee2273e11de959c/gradle/contrib.gradle#L10

PR #32, opened 2021-04-26
Adopt OpenTelemetry 1.0.0-alpha

rmfitzpatrick:
**Description:**
These change update the project and JMX Metric Gatherer OTel dependencies to use 1.0.0(-alpha).  They are largely transparent signature change adoptions, and reliance on the Autoconfigure sdk extension, but also include readme improvements for the utility.  The largest changes are in the observer instrument helpers to account for only being able to specify the updater on the instrument builder in the api.

**Testing:**

All existing tests have been updated with some basic expansion for coverage.

**Documentation:**

Metric gatherer docs have been updated to account for signature changes, with additional readability/accuracy improvements.

**Outstanding items:**

Updating the release tasks to use Sonatype and upgrading to 1.1.0(-alpha).

PR #36, opened 2021-05-21
Initial 1.0.0-alpha release

rmfitzpatrick:
**Description:**

EDIT: After [feedback from @anuraaga](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/36#issuecomment-846338604) I've decided to not implement the versioning scheme proposed below to avoid any [potential fallout](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/36#pullrequestreview-665878215) from using the build number in an unanticipated fashion.  This project will maintain the opentelemetry-java major and minor release versions but use the patch release as an incremental signifier of its own instead of the initially proposed build number.

~These changes propose an initial release version of `1.0.0-alpha-0` to track with the minimum OpenTelemetry Java API/SDK dependency version and use a build number for project-specific releases*.  This system will allow us to clearly document which version of the OTel libraries we are using and allow additional maintenance releases to be pushed out to a build number:~

~`1.0.0-alpha-0` -> `1.0.0-alpha-1-SNAPSHOT` -> `1.1.0-alpha-0-SNAPSHOT` -> `1.1.0-alpha-0` -> `1.2.0-alpha-0-SNAPSHOT` -> `1.2.0-alpha-0` -> `1.2.0-alpha-1-SNAPSHOT` etc.~

~This way as new features and bug fixes are added to our projects we aren't forced to move to a version that obscures our OTel dependencies.  As alpha specifiers are removed when metrics are stable they will be removed from these versions as well.~

**Existing Issue(s):**

Resolves https://github.com/open-telemetry/opentelemetry-java-contrib/issues/8

**Testing:**

~Tested using maven-artifact's `ComparableVersion`, which I believe is the basis for version comparisons.  A relevant article is a bit vague to me (https://docs.oracle.com/middleware/1212/core/MAVEN/maven_version.htm#MAVEN8855) but I believe the proposed system is acceptable and doesn't fall into the string comparison trap of nonstandard versions.~

```groovy
   def 'proposed versions are ordered as expected'() {
        when: 'versions use build numbers'
        def numMajor = 10
        def numMinor = 10
        def numMaintenance = 10
        def numBuild = 10
        def versions = []
        (0..numMajor).each{ major ->
            (0..numMinor).each{ minor ->
                (0..numMaintenance).each { maintenance ->
                    (0..numBuild).each { build ->
                        versions.add(new ComparableVersion("${major}.${minor}.${maintenance}-alpha-${build}-SNAPSHOT"))
                        versions.add(new ComparableVersion("${major}.${minor}.${maintenance}-alpha-${build}"))
                    }
                    (0..numBuild).each { build ->
                        versions.add(new ComparableVersion("${major}.${minor}.${maintenance}-${build}-SNAPSHOT"))
                        versions.add(new ComparableVersion("${major}.${minor}.${maintenance}-${build}"))
                    }
                }
            }
        }

        def cloned = versions.clone()
        Collections.sort(cloned)

        then:
        versions == cloned
    }
```

**Documentation:**

Added initial change log since this will be first version tagged and published to Maven Central.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,11 @@
+# Changelog
+
+## Unreleased
+
+## 1.0.0-alpha
+
+* Adopt OpenTelemetry 1.0.0(-alpha) dependencies (#32)
+  * Update JMX Metric Gatherer to use 1.0.0(-alpha) proto, API, SDK, and exporters.
+  * Update JMX Metric Gatherer to use Autoconfigure SDK extension properties*
+* JMX Metric Gatherer - Handle missing MBean attributes without failing (#39)
+  * Thanks to @dehaansa.
--- a/build.gradle
+++ b/build.gradle
@@ -8,7 +8,7 @@ description = 'OpenTelemetry Contrib libraries and utilities for the JVM'
 
 allprojects {
     group = 'io.opentelemetry.contrib'
-    version = '1.0.0-alpha-SNAPSHOT'
+    version = '1.0.0-alpha'
 
     apply from: "$rootDir/gradle/spotless.gradle"
     apply from: "$rootDir/gradle/dependencies.gradle"


CHANGELOG.md
```diff
@@ -0,0 +1,9 @@
+# Changelog
+
+## Unreleased
+
+## 1.0.0-alpha-0
+
+* Adopt OpenTelemetry 1.0.0(-alpha) (#32)
```
jkwatson, line 7: I don't know what this sentence means. Is it saying that you're adopting the version format, or that you're depending on OTel v1.0.0 ?

breedx-splk (approved):
I think this is ok to proceed with.  I have seen cases in the past where gradle gets confused or struggles with changing the "qualifier" (is that what you call the thing after the dash after the version?).

I don't know the build system of this repo well enough to predict, but I can say that if it does become a problem we might be able to change `alpha-0` to `alpha0`...

CHANGELOG.md
rmfitzpatrick: It refers to the dependency adoption: https://github.com/open-telemetry/opentelemetry-java-contrib/pull/32.  Clarified a bit if that helps.

anuraaga:
I tried a similar versioning scheme for the aws distro of the agent but decided to instead tie the major and minor to upstream and not have the patch number 1:1. One reason is the relative lack of independent changes in that repo. But it did make the version much cleaner. What do you think about that for this repo?

jkwatson (approved):

rmfitzpatrick:
> I tried a similar versioning scheme for the aws distro of the agent but decided to instead tie the major and minor to upstream and not have the patch number 1:1. One reason is the relative lack of independent changes in that repo. But it did make the version much cleaner. What do you think about that for this repo?

I think this is the approach used by java-instrumentation and custom distributions for other otel projects I've seen.  It's helpful in its general cue for the major/minor dependency version but does require checking dependency files or using a manager feature to determine the actual one when verifying if a particular dep update is provided or available to the project.  This project has definitely demonstrated it has a low likelihood of coincidental deviation though so it's not a concern for frequent ambiguity.

It's my preference we'd always be clear in our adopted version but wouldn't want to do so at the risk of unexpected fallout.

CHANGELOG.md
jkwatson: I guess I'm mostly confused by the usage of the word "adopt" here. Why not just say "Depend on"?

rmfitzpatrick:
Updated to use `1.0.0-alpha` per @anuraaga's feedback.  Will move forward in this way without potentially taxing version compatibilities of relevant tools.

anuraaga (approved):

Issue #38, opened 2021-05-28
Missing Attributes on existing MBean causes partial collection failure


dehaansa:
**Description**
If an MBean exists but the requested attribute is missing, an AttributeNotFoundException is thrown and only the metrics collected before that exception are returned.

**Steps to reproduce**
This was observed in a Windows system that lacked the OpenFileDescriptorCount metric of the java.lang:type=OperatingSystem MBean.

**Expectation**
The metric requested would be missing, but metrics listed after it in the groovy script would still be collected.

**What applicable config did you use?**
`def os = otel.mbean("java.lang:type=OperatingSystem")
otel.instrument(os, "jvm.os.files.open", "open file descriptors",
        "1", "OpenFileDescriptorCount", otel.&longValueObserver)`

rmfitzpatrick:
Resolved by #39.

PR #39, opened 2021-05-28
Handle missing attributes without failing

dehaansa:
**Description:** Enhance InstrumentHelper & MBeanHelper to handle AttributeNotFoundException cleanly. Prior to this PR, if a groovy script included an instrument where the MBean existed but the Attribute was missing, it would fail and the remainder of the metrics listed afterwards in the groovy script would not be collected. This PR makes the AttributeNotFoundException cause that metric to be missing, but the remainder of the script's requested metrics are still collected.

**Existing Issue(s):** #38 

**Testing:** Enhanced groovy tests to capture this.


rmfitzpatrick (approved):
Thanks for the fix!

PR #40, opened 2021-06-25
Follow standard Gradle patterns for project folders / build files.

anuraaga:
@jkwatson Is kicking out one of my modules from SDK repo to contrib :) So I'd like to get this repo to match the styles, including updating to some recent build file refactoring I've done in both SDK and instrumentation repos.

First is to reogranize the project folder / files. There doesn't seem to be a need for a `contrib` folder in a repo called `-contrib`. And sticking to the default build file names just seems more consistent with the ecosystem (I don't believe `github.com/gradle/gradle` used the default names at some point before but seem to have changed to it, presumably for similar reasons, seems like a good repo to follow ;)


trask:
@anuraaga I think something went wrong here 😂 

image

jkwatson:
I mean, deleting code is good, right?

Right?

anuraaga:
Yeah I think our original goal here was reducing maintenance burden right? :-D

trask (approved):

anuraaga:
Oh hmm - I thought java and or instrumentation maintainers were supposed to be admins in this repo but doesn't seem to be the case

anuraaga:
Ah looks like we're approvers - @rmfitzpatrick can you add @trask and me to java-contrib-maintainers team?

https://github.com/orgs/open-telemetry/teams/opentelemetry-java-contrib-maintainers/members

Both since a repo should have more than one maintainer to prevent single point of failure, and that we'll be using this repo more.

rmfitzpatrick:
@anuraaga apologies for the delay, I've been on a short vacation.  Added you and @trask to maintainers.

Mind updating the artifact path for https://github.com/open-telemetry/opentelemetry-java-contrib/blob/main/.github/workflows/pr-build.yml#L41 ?

rmfitzpatrick:
Looks like a readme link also needs updating https://github.com/open-telemetry/opentelemetry-java-contrib/blob/main/README.md#provided-libraries

anuraaga:
@rmfitzpatrick Yup - thanks!

PR #41, opened 2021-06-29
Build logic changes

anuraaga:
For consistency with other repos

- Java, Groovy, Spotless convention moved to buildSrc plugins
  - Will follow up for publish
- Used same copyright header
- Java uses toolchain and release flag instead of sourceCompatibility
- Shared dependencies moved to dependencyManagement project
- Removed lint / format task aliases
- PR build just runs one task


.github/workflows/pr-build.yml
```diff
@@ -22,17 +22,10 @@ jobs:
           distribution: adopt
           java-version: 11
       - uses: burrunan/gradle-cache-action@v1
-        name: ./gradlew lint
+        name: Build
         with:
-          arguments: --stacktrace lint
-          gradle-version: wrapper
```
anuraaga, line 28: These two flags are basically the defaults

buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
```diff
@@ -0,0 +1,54 @@
+plugins {
+    `java-library`
+
+    id("otel.spotless-conventions")
+}
+
+group = "io.opentelemetry.contrib"
+version = "1.0.0-alpha"
+
+base.archivesBaseName = "${rootProject.name}-${project.name}"
```
anuraaga, line 10: Prepending with rootProject looks a bit weird to me. Will check this in a followup

buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
```diff
@@ -0,0 +1,54 @@
...
+            release.set(8)
+        }
+    }
+
+    val integrationTest by registering {
+        dependsOn(test)
+    }
+
+    test {
+        if (gradle.startParameter.taskNames.contains(integrationTest.name)) {
```
anuraaga, line 33: I think we'll want to separate out test-sets instead of this property mechanism

jmx-metrics/build.gradle
```diff
@@ -38,25 +39,36 @@ def deps = [
     testcontainers : "org.testcontainers:testcontainers:1.15.1",
 ]
 
+repositories {
+    mavenCentral()
+    maven {
+        setUrl("http://www.terracotta.org/download/reflector/releases")
```
anuraaga, line 45: Didn't want to include the mostly dead JCenter so found this

jmx-metrics/build.gradle
```diff
@@ -38,25 +39,36 @@ def deps = [
...
+            "io.opentelemetry:opentelemetry-sdk-extension-autoconfigure",
+            "io.opentelemetry:opentelemetry-sdk-testing",
+            "io.opentelemetry:opentelemetry-exporter-logging",
+            "io.opentelemetry:opentelemetry-exporter-otlp-metrics",
+            "io.opentelemetry:opentelemetry-exporter-prometheus",
             deps.slf4j,
             dependencies.create(group: 'org.slf4j', name: 'slf4j-simple', version: versions.slf4j)
 
-    runtime "org.terracotta:jmxremote_optional-tc:1.0.5"
+    runtime "org.terracotta:jmxremote_optional-tc:1.0.8"
```
anuraaga, line 71: New repo only had 1.0.6+

jmx-metrics/src/test/groovy/io/opentelemetry/contrib/jmxmetrics/InstrumentHelperTest.groovy
```diff
@@ -40,7 +28,7 @@ import spock.lang.Unroll
 
 class InstrumentHelperTest extends Specification {
 
-    @Rule public final TestRule name = new TestName()
+    @Rule public TestRule name = new TestName()
```
anuraaga, line 31: Looks like the dependency management may have caused an update to spock version

README.md
```diff
@@ -18,7 +18,7 @@ feature or via instrumentation, this project is hopefully for you.
 
 ```bash
 # Apply formatting
-$ ./gradlew format
+$ ./gradlew spotlessApply
```
rmfitzpatrick, line 21: The aim here was to not tie closely to a dependency's task in our workflow but if spotless and this usage are in the other otel projects it seems ok to me.

.github/workflows/pr-build.yml
```diff
@@ -53,13 +46,9 @@ jobs:
           distribution: adopt
           java-version: 11
       - uses: burrunan/gradle-cache-action@v1
-        name: ./gradlew integrationTest
-        id: integration-test
-        timeout-minutes: 10
+        name: Integration Tests
         with:
-          arguments: --stacktrace --max-workers 2 integrationTest
```
rmfitzpatrick, line 60: Because we're setting the parallel gradle property I found not limiting the max workers to overrun the actions runner and cause flake.  I'm not sure how many cores are advertised but in circle ci the host machine's are instead of what's granted to the executing container.

jkwatson: We should get out of circle and use GHA, correct?

buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
rmfitzpatrick: Since this was designed for the jmx metric gatherer that runs as an uber jar having the name `opentelemetry-java-contrib-jmx-metrics.jar` seemed more informative and clear to me than just `jmx-metrics`.  This doesn't have to translate to all contributed projects so I'm open to migrating it there.

.github/workflows/pr-build.yml
rmfitzpatrick: This has been done (https://github.com/open-telemetry/opentelemetry-java-contrib/issues/30) and I'm referring to an issue I observed in actions.

jmx-metrics/build.gradle
rmfitzpatrick: On a clean build I see 

```
Could not determine the dependencies of task ':jmx-metrics:shadowJar'.
> Could not resolve all dependencies for configuration ':jmx-metrics:runtimeClasspath'.
   > Could not resolve org.terracotta:jmxremote_optional-tc:1.0.8.
     Required by:
         project :jmx-metrics
      > Could not resolve org.terracotta:jmxremote_optional-tc:1.0.8.
         > Could not parse POM http://www.terracotta.org/download/reflector/releases/org/terracotta/jmxremote_optional-tc/1.0.8/jmxremote_optional-tc-1.0.8.pom
            > Could not find org.terracotta.forge:forge-parent:4.12.
              Searched in the following locations:
                - https://repo.maven.apache.org/maven2/org/terracotta/forge/forge-parent/4.12/forge-parent-4.12.pom
```

.github/workflows/pr-build.yml
anuraaga: We haven't had much issue even in our huge instrumentation repo build :P I've looked at some build scans and it seems to be using about 2 workers by default on GHA so looks like that issue was probably fixed by now

buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
anuraaga: Yeah thinking of making this `opentelemetry-jmx-metrics`

jmx-metrics/build.gradle
anuraaga: Oops accidentally reverted a change for this

rmfitzpatrick (approved):
Thanks!

PR #43, opened 2021-06-29
Handle list

### LINKED ISSUE #42: Add Support for Multiple ObjectNames
**Is your feature request related to a problem? Please describe.**
There is an issue when multiple mbeans are under different object names but are similar enough that they could be given the same metric name and differentiated through labels. This isn't possible currently because when a second call to instrument with the same metric name happens, it overrides the previous call to instrument.

EX:
```
def beanActiveTasks = otel.mbeans("org.apache.cassandra.metrics:type=ThreadPools,path=*,scope=*,name=PendingTasks")
otel.instrument(beanActiveTasks, "cassandra.current_tasks", "Number of tasks in queue with the given task status.", "1", 
    ["stage_name":{ mbean -> mbean.name().getKeyProperty("scope")},
    "task_status":{"pending"}],
    "Value", otel.&doubleValueObserver)

def beanPendingTasks = otel.mbeans("org.apache.cassandra.metrics:type=ThreadPools,path=*,scope=*,name=ActiveTasks")
otel.instrument(beanPendingTasks, "cassandra.current_tasks", "Number of tasks in queue with the given task status.", "1", 
    ["stage_name":{ mbean -> mbean.name().getKeyProperty("scope")},
    "task_status":{"active"}],
    "Value", otel.&doubleValueObserver)
```
Here is where there are two different calls to instrument with the same metric name. They end up overriding each other when passed through.

**Describe the solution you'd like**
A way to make this possible is by adding an additional signature to `otel.mbeans( )` that allows multiple object names to be passed in.

Mrod1598:
**Description:**
 Resolves #42 through the proposed solution of adding the ability to pass in multiple Object Names to better support adding additional mbeans under the same metric name.

**Existing Issue(s):**
#42

**Testing:**
Added test that defines multiple mbeans and then pulls those mbeans out using multiple object names.

**Documentation:**
Documented the new signature of `otel.mbeans`

### DIFF
--- a/contrib/jmx-metrics/README.md
+++ b/contrib/jmx-metrics/README.md
@@ -102,7 +102,7 @@ mutually exclusive with `otel.jmx.groovy.script`. The currently supported target
    instances (available via `otel.instrument()`) as described below.  It is intended to be used in cases
    where your given `objectNameStr` can return a multiple element `List<GroovyMBean>`.
 - `otel.mbeans(List<String> objectNameStrs)`
-   - This method is equavalant to the above method except, it adds support for multiple ObjectNames. This support is meant for when there are multiple mbeans that relate to the same metric and can be seperated using labels in `otel.instrument()`.
+   - This method is equivalent to the above method except, it adds support for multiple ObjectNames. This support is meant for when there are multiple mbeans that relate to the same metric and can be seperated using labels in `otel.instrument()`.
 
 - `otel.instrument(MBeanHelper mBeanHelper, String instrumentName, String description, String unit, Map<String, Closure> labelFuncs, String attribute, Closure instrument)`
    - This method provides the ability to easily create and automatically update instrument instances from an
--- a/contrib/jmx-metrics/README.md
+++ b/contrib/jmx-metrics/README.md
@@ -101,6 +101,8 @@ mutually exclusive with `otel.jmx.groovy.script`. The currently supported target
    but returns an `MBeanHelper` instance representing all matching MBeans for usage by subsequent `InstrumentHelper`
    instances (available via `otel.instrument()`) as described below.  It is intended to be used in cases
    where your given `objectNameStr` can return a multiple element `List<GroovyMBean>`.
+- `otel.mbeans(List<String> objectNameStrs)`
+   - This method is equavalant to the above method except, it adds support for multiple ObjectNames. This support is meant for when there are multiple mbeans that relate to the same metric and can be seperated using labels in `otel.instrument()`.
 
 - `otel.instrument(MBeanHelper mBeanHelper, String instrumentName, String description, String unit, Map<String, Closure> labelFuncs, String attribute, Closure instrument)`
    - This method provides the ability to easily create and automatically update instrument instances from an
--- a/contrib/jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/MBeanHelper.groovy
+++ b/contrib/jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/MBeanHelper.groovy
@@ -43,19 +43,19 @@ class MBeanHelper {
 
     private final JmxClient jmxClient
     private final boolean isSingle
-    private List<String> objectNames
+    private final List<String> objectNames
 
     private List<GroovyMBean> mbeans
 
     MBeanHelper(JmxClient jmxClient, String objectName, boolean isSingle) {
         this.jmxClient = jmxClient
-        this.objectNames = [objectName]
+        this.objectNames = Collections.unmodifiableList([objectName])
         this.isSingle = isSingle
     }
 
     MBeanHelper(JmxClient jmxClient, List<String> objectNames) {
         this.jmxClient = jmxClient
-        this.objectNames = objectNames
+        this.objectNames = Collections.unmodifiableList(objectNames)
         this.isSingle = false
     }
 
@@ -72,12 +72,11 @@ class MBeanHelper {
     void fetch() {
         this.mbeans = []
         for(objectName in objectNames){
-            println objectName
             def tmpMbeans = queryJmx(jmxClient, objectName)
-            this.mbeans.addAll(tmpMbeans)
             if (tmpMbeans.size() == 0) {
                 logger.warning("Failed to fetch MBean ${objectName}.")
             } else {
+                this.mbeans.addAll(tmpMbeans)
                 logger.fine("Fetched ${tmpMbeans.size()} MBeans - ${tmpMbeans}")
             }
         }
--- a/contrib/jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/MBeanHelper.groovy
+++ b/contrib/jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/MBeanHelper.groovy
@@ -43,14 +43,13 @@ class MBeanHelper {
 
     private final JmxClient jmxClient
     private final boolean isSingle
-    private final String objectName
+    private List<String> objectNames
 
     private List<GroovyMBean> mbeans
 
     MBeanHelper(JmxClient jmxClient, String objectName, boolean isSingle) {
         this.jmxClient = jmxClient
-        List<String> objectNames = [objectName]
-        this.objectNames = objectNames
+        this.objectNames = [objectName]
         this.isSingle = isSingle
     }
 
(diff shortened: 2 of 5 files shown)


rmfitzpatrick:
Provided a comment on the parent issue https://github.com/open-telemetry/opentelemetry-java-contrib/issues/42#issuecomment-870981841 that could effect approach.  Overall this seems ok to me though depending on your thoughts.

Would you be willing to modify this test and its groovy script to utilize your thread pools mbean use case? https://github.com/open-telemetry/opentelemetry-java-contrib/blob/main/jmx-metrics/src/test/groovy/io/opentelemetry/contrib/jmxmetrics/OtlpIntegrationTests.groovy

Mrod1598:
> Provided a comment on the parent issue [#42 (comment)](https://github.com/open-telemetry/opentelemetry-java-contrib/issues/42#issuecomment-870981841) that could effect approach. Overall this seems ok to me though depending on your thoughts.
> 
> Would you be willing to modify this test and its groovy script to utilize your thread pools mbean use case? https://github.com/open-telemetry/opentelemetry-java-contrib/blob/main/jmx-metrics/src/test/groovy/io/opentelemetry/contrib/jmxmetrics/OtlpIntegrationTests.groovy

@rmfitzpatrick I've updated the test and moved it up to the newest version. I was wondering if you could approve the CI. Thanks!

jmx-metrics/src/test/groovy/io/opentelemetry/contrib/jmxmetrics/OtlpIntegrationTests.groovy
```diff
@@ -58,27 +60,21 @@ class OtlpIntegrationTests extends OtlpIntegrationTest {
...
+        metric.description == 'Number of tasks in queue with the given task status.'
+        metric.unit == '1'
+        metric.hasGauge()
 
         when: 'we examine the datapoints'
-        Summary datapoints = metric.summary
+        Gauge datapoints = metric.gauge
         then: 'they are of the expected size'
-        datapoints.dataPointsCount == 1
+        datapoints.dataPointsCount == 44
```
rmfitzpatrick, line 71: I'm not sure if this is related to intermediary proto changes but the number of datapoints is now more manageable.  Would you mind adding more complete content validation?

rmfitzpatrick: It may be preferable overall to keep the existing integration tests as they were and add another for this feature.

Mrod1598: moved to a separate test but had to modify the signature for `configureContainers` to allow me to pass in my own script.

Mrod1598:
Altered script and test to make metrics more predictable. It was passing locally but not on actions.

Mrod1598:
After testing it seems that the actions are failing through a lack of resources. The tests pass locally if docker is given enough resources but I get the same errors if I reduce the amount of memory given to docker down to 2GB. I was wondering if you had any ideas to help reduce parallelism so that the tests aren't using all of the resources available. @rmfitzpatrick

rmfitzpatrick:
@Mrod1598, could you try restricting the worker count for the integration test action like was done previously? https://github.com/open-telemetry/opentelemetry-java-contrib/pull/41#discussion_r660603803.  Overall better helper logic around waiting for expected metrics is needed and I will try to take care of this in the near future, but that shouldn't block this.

I think it would also be helpful if you incorporated main changes/rebased to ensure this PR is still compatible with recent updates.

Mrod1598:
> @Mrod1598, could you try restricting the worker count for the integration test action like was done previously? [#41 (comment)](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/41#discussion_r660603803). Overall better helper logic around waiting for expected metrics is needed and I will try to take care of this in the near future, but that shouldn't block this.
> 
> I think it would also be helpful if you incorporated main changes/rebased to ensure this PR is still compatible with recent updates.

Sorry, forgot to mention that I tried restricting the worker count and it doesn't seem to help locally either. any other suggestions?

rmfitzpatrick:
@Mrod1598 thanks for your patience and I saw that it wasn't apparently effective.  I'm hoping to be able to add better waiting helpers shortly and investigate this issue.

djaglowski:
@rmfitzpatrick Is there anything that could be done to help move this forward?

jmx-metrics/src/integrationTest/resources/multi_obj_script.groovy
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ */
+
+def helper = otel.mbeans([
+    "org.apache.cassandra.metrics:type=ThreadPools,path=*,scope=MemtablePostFlush,name=PendingTasks",
+    "org.apache.cassandra.metrics:type=ThreadPools,path=*,scope=MemtablePostFlush,name=ActiveTasks"
+])
+otel.instrument(helper, "cassandra.current_tasks", "Number of tasks in queue with the given task status.", "1",
+        ["stage_name":{ mbean -> mbean.name().getKeyProperty("scope")},
+            "task_status":{ mbean -> mbean.name().getKeyProperty("name")}],
```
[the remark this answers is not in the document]
Mrod1598: definitely have some ideas moving forward.

.github/workflows/pr-build.yml
```diff
@@ -44,7 +44,7 @@ jobs:
       - uses: burrunan/gradle-cache-action@v1
         name: Integration Tests
         with:
-          arguments: --stacktrace integrationTest
+          arguments: --stacktrace --max-workers 1 integrationTest
```
anuraaga, line 47: Can you revert this? Don't think it's needed anymore

jmx-metrics/src/integrationTest/java/io/opentelemetry/contrib/jmxmetrics/MultiObjTest.java
```diff
@@ -0,0 +1,73 @@
...
+
+import java.time.Duration;
+import org.junit.jupiter.api.Test;
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.containers.Network;
+import org.testcontainers.containers.wait.strategy.Wait;
+import org.testcontainers.junit.jupiter.Container;
+import org.testcontainers.utility.MountableFile;
+
+class MultiObjTest extends AbstractIntegrationTest {
```
anuraaga, line 18: Since this looks like a new feature of the helper rather than a new backend integration I think it's better not to add yet another integration test to make sure they don't get too slow.

How about adding unit test to instrumenthelpertest and modifying the existing Cassandra test to exercise this?

anuraaga (commented):
@rmfitzpatrick Can you take another look at this? Would be good to verify the new groovy API

jmx-metrics/src/integrationTest/java/io/opentelemetry/contrib/jmxmetrics/MultiObjTest.java
Mrod1598: [here](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/43#discussion_r690663935) I had originally setup a modified version of the cassandra test and was told to revert it and add an additional test.

rmfitzpatrick: My second request to keep the single mbean integration test as is and add a new one with multiples was to preserve the datapoint validation that was being removed in favor of a simple size check that was apparently flakey.

rmfitzpatrick: I'm personally a fan of integration tests but can understand not want to balloon CI and dev times.  If there is to be a unit test instead of the current approach I think checking the functionality of the InstrumentHelper with multi-mbean MBeanHelper would be a good way forward.

Mrod1598: So, can this be merged? or are the changes required to be unit tests? @rmfitzpatrick

rmfitzpatrick: @anuraaga are you ok with the changes as is?  Improving coverage could be a future item.

anuraaga: I'd like to have the test change made yes - it's why I left a comment :P The integration tests are already up to 5 min compared to 2 for unit tests so I don't want to continue trending up.

djaglowski:
@rmfitzpatrick Could this be merged?

Mrod1598:
> @Mrod1598 thanks for updating. Would you be open to expanding the instrument helper tests a bit as mentioned earlier? [observIQ#2](https://github.com/observIQ/opentelemetry-java-contrib/pull/2)

Yes, I appreciate that

anuraaga:
Was hoping to merge main for you but looks like I don't have access to update the branch

PR #44, opened 2021-07-01
Update to OTel 1.3.0

anuraaga:


jmx-metrics/src/test/groovy/io/opentelemetry/contrib/jmxmetrics/OtlpIntegrationTests.groovy
```diff
@@ -4,24 +4,23 @@
...
 import spock.lang.Requires
-import spock.lang.Retry
 import spock.lang.Timeout
 import spock.lang.Unroll
 
 @Requires({
     System.getProperty('ojc.integration.tests') == 'true'
 })
 @Timeout(90)
-@Retry
```
anuraaga, line 24: After being stuck for a long time I realized that this annotation is dangerous - it would retry on assertion failure, and eventually hit the Timeout, throwing a Timeout error while starting testcontainers and not showing anything about the assertion. Better if we can make the test bodies more robust, I did that in one place

jmx-metrics/src/test/groovy/io/opentelemetry/contrib/jmxmetrics/target-systems/CassandraIntegrationTests.groovy
```diff
@@ -31,170 +30,164 @@ class CassandraIntegrationTests extends OtlpIntegrationTest  {
         Testcontainers.exposeHostPorts(otlpPort)
         configureContainers('target-systems/cassandra.properties',  otlpPort, 0, false)
 
-        expect:
-        when: 'we receive metrics from the JMX metrics gatherer'
-        List<ResourceMetrics> receivedMetrics = collector.receivedMetrics
-        then: 'they are of the expected size'
-        receivedMetrics.size() == 1
+        ArrayList<Metric> metrics
+        await().atMost(30, TimeUnit.SECONDS).untilAsserted {
```
anuraaga, line 34: It seems like this made the tests fairly reliable. If some other flakes come up we can probably tweak those tests too

dependencyManagement/build.gradle.kts
```diff
@@ -12,8 +12,8 @@ val dependencyVersions = hashMapOf<String, String>()
 rootProject.extra["versions"] = dependencyVersions
 
 val DEPENDENCY_BOMS = listOf(
-    "io.opentelemetry:opentelemetry-bom:1.0.0",
-    "io.opentelemetry:opentelemetry-bom-alpha:1.0.0-alpha"
+    "io.opentelemetry:opentelemetry-bom:1.3.0",
+    "io.opentelemetry:opentelemetry-bom-alpha:1.3.0-alpha"
```
rmfitzpatrick, line 16: Mind updating the project version to 1.3.0-alpha-SNAPSHOT with these changes if you think it's the right time?

rmfitzpatrick (approved):
Thanks @anuraaga

jmx-metrics/src/test/groovy/io/opentelemetry/contrib/jmxmetrics/target-systems/CassandraIntegrationTests.groovy
```diff
@@ -31,170 +30,164 @@ class CassandraIntegrationTests extends OtlpIntegrationTest  {
...
-        when: 'we examine the instrumentation library'
-        InstrumentationLibraryMetrics ilMetric = ilMetrics.get(0)
-        InstrumentationLibrary il = ilMetric.instrumentationLibrary
-        then: 'it is of the expected content'
-        il.name  == 'io.opentelemetry.contrib.jmxmetrics'
-        il.version == '1.0.0-alpha'
+            InstrumentationLibraryMetrics ilMetric = ilMetrics.get(0)
+            InstrumentationLibrary il = ilMetric.instrumentationLibrary
+            assert il.name  == 'io.opentelemetry.contrib.jmxmetrics'
+            assert il.version == '1.0.0-alpha'
```
rmfitzpatrick, line 46: I think this should be updated so untilAsserted may be silencing the failures:

```
NOTE: Be VERY careful so that you're not using this method incorrectly in languages (like Kotlin and Groovy) that doesn't disambiguate between a ThrowingRunnable that doesn't return anything (void) and Callable that returns a value. For example in Kotlin you can do like this:
 
       await().untilAsserted { true == false }
       
and the compiler won't complain with an error (as is the case in Java). If you were to execute this test in Kotlin it'll pass!
```

rmfitzpatrick: I'm sorry, forgot this wasn't set by the sdk: https://github.com/open-telemetry/opentelemetry-java-contrib/blob/4bfd97dfff1ad7d58898a9e0aa2a43ce82794f3a/jmx-metrics/src/main/java/io/opentelemetry/contrib/jmxmetrics/GroovyMetricEnvironment.java#L102

rmfitzpatrick: I can take care these when adopting Awaitility.await in another PR.

anuraaga: Ah hehe yeah let's do in another PR then best would be to fill the version into a file and read it similar to what the SDK does

PR #46, opened 2021-07-02
Move in x-ray extensions from java repo

anuraaga:
The Java code is copied in with modifications in https://github.com/open-telemetry/opentelemetry-java-contrib/commit/23fdb1d30dc25d7125886f4f35d982beea26076c to remove some internal usage, only other significant changes are in build logic


buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
```diff
@@ -7,7 +9,7 @@ plugins {
 group = "io.opentelemetry.contrib"
 version = "1.3.0-alpha"
 
-base.archivesBaseName = "${rootProject.name}-${project.name}"
+base.archivesBaseName = "opentelemetry-${project.name}"
```
anuraaga, line 12: Results in `io.java.contrib:opentelemetry-jmx-metrics` and `io.java.contrib:opentelemetry-aws-xray` which seem good to me

rmfitzpatrick:
Would you be able to add a readme and link it in the main provided libraries section? https://github.com/open-telemetry/opentelemetry-java-contrib#provided-libraries

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsXrayIdGenerator.java
```diff
@@ -0,0 +1,54 @@
...
+import io.opentelemetry.api.trace.TraceId;
+import io.opentelemetry.sdk.trace.IdGenerator;
+import java.util.Random;
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Generates tracing ids compatible with the AWS X-Ray tracing service. In the X-Ray system the
+ * first 32 bits of the trace id are the Unix epoch time in secords. Spans (AWS calls them segments)
+ * submit with trace id timestamps outside of the last 30 days are rejected.
```
rmfitzpatrick, line 16: nit
```suggestion
 * submitted with trace id timestamps outside of the last 30 days are rejected.
```

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/GetSamplingRulesResponse.java
```diff
@@ -0,0 +1,109 @@
...
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.auto.value.AutoValue;
+import java.util.List;
+import java.util.Map;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+@AutoValue
+abstract class GetSamplingRulesResponse {
```
rmfitzpatrick, line 15: Not really an issue but I wonder if the aws sdk provides any of this and if adopting the dependency is undesirable overall.

anuraaga: I wish the models were published separately but they are tied to the real SDK so would end up relatively heavyweight, and a particular problem is users using AWS SDK v1 prefer a library not to bring on v1 and vice versa. We got this report on the AWS x-ray SDK where we do use the AWS sdk for the models.

Stubs do get published which we may be able to autogen these classes with if we wanted to but not worth it for just these I guess.

rmfitzpatrick (approved):
lgtm with a readme entry

anuraaga:
@rmfitzpatrick Thanks - will flesh out the submodule README in a followup

PR #47, opened 2021-07-02
Update to Gradle 7.1

anuraaga:
And copies in a couple settings from the other repos


jmx-metrics/build.gradle
```diff
@@ -42,7 +42,7 @@ def deps = [
 repositories {
     mavenCentral()
     maven {
-        setUrl("http://www.terracotta.org/download/reflector/releases")
+        setUrl("https://repo.terracotta.org/maven2")
```
anuraaga, line 45: This isn't documented on terracotta's site but seems to be the only way to use HTTPS. It seemed to have worked

rmfitzpatrick (approved):

PR #50, opened 2021-07-06
Change initial sampler to 5% random to match x-ray docs and add jitte…

anuraaga:
…r to polling.

### DIFF
--- a/aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsXrayRemoteSampler.java
+++ b/aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsXrayRemoteSampler.java
@@ -25,6 +25,8 @@ import org.checkerframework.checker.nullness.qual.Nullable;
 /** Remote sampler that gets sampling configuration from AWS X-Ray. */
 public final class AwsXrayRemoteSampler implements Sampler, Closeable {
 
+  private static final Random RANDOM = new Random();
+
   private static final Logger logger = Logger.getLogger(AwsXrayRemoteSampler.class.getName());
 
   private static final String WORKER_THREAD_NAME =
@@ -37,8 +39,10 @@ public final class AwsXrayRemoteSampler implements Sampler, Closeable {
   private final Sampler initialSampler;
   private final XraySamplerClient client;
   private final ScheduledExecutorService executor;
-  private final ScheduledFuture<?> pollFuture;
+  private final long pollingIntervalNanos;
+  private final int jitterNanos;
 
+  @Nullable private volatile ScheduledFuture<?> pollFuture;
   @Nullable private volatile GetSamplingRulesResponse previousRulesResponse;
   private volatile Sampler sampler;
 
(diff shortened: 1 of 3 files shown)


aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsXrayRemoteSampler.java
```diff
@@ -106,11 +113,20 @@ private void getAndUpdateSampler() {
     } catch (Throwable t) {
       logger.log(Level.FINE, "Failed to update sampler", t);
     }
+    scheduleSamplerUpdate();
```
anuraaga, line 116: I didn't follow the current pattern of xray sdk which schedules a fixed rate with jitter

https://github.com/aws/aws-xray-sdk-java/blame/5143932f0da7d921c07fc8149d14c3f767cb09ee/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/strategy/sampling/pollers/RulePoller.java#L80

It's unfair scheduling, as some processes will forever have a lower polling rate and some will forever have a higher one. The intent should always be to make sure over the long term the semantics are the same for any process, so we need to apply the jitter per-schedule for it to be fair.

Of course, the downside is at any point of time the randoms can align to a thundering herd. I don't think it can be avoided though and is why backends need (and I figure have) rate limits anyways.

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsXrayRemoteSampler.java
```diff
@@ -72,9 +76,12 @@ public static AwsXrayRemoteSamplerBuilder newBuilder(Resource resource) {
 
     sampler = initialSampler;
 
-    pollFuture =
-        executor.scheduleAtFixedRate(
-            this::getAndUpdateSampler, 0, pollingIntervalNanos, TimeUnit.NANOSECONDS);
+    this.pollingIntervalNanos = pollingIntervalNanos;
+    // Add ~1% of jitter. Truncating to int is safe for any practical polling interval.
```
anuraaga, line 80: Currently 5 sec of jitter is 1.6666% while 10ms for the target poller is 1%. 3s seems fine to me, especially since we compute the jitter with nanosecond precision instead of seconds precision like the xray sdk

trask:
Closing and re-opening to try to trigger updated CODEOWNERS reviewers

trask:
That didn't work, but was able to manually add @open-telemetry/opentelemetry-java-contrib-approvers as a reviewer (and the lock icon showed up)

trask:
I'm not sure why lock icon is not showing up for @willarmiros though now that he is in CODEOWNERS for `aws-xray/`

willarmiros:
@trask we ran into a similar issue in JS repo, it turns out you can only be a codeowner if you have at a minimum write permission on the repo to be an official codeowner per the first couple paragraphs here: https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/creating-a-repository-on-github/about-code-owners

Very unfortunate limitation on GitHub's part, we still don't have a concrete decision for how to address it on the js-contrib repo.

willarmiros (approved):
lgtm

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsXrayRemoteSampler.java
willarmiros: Yep, I prefer your approach here.

trask:
@anuraaga do you know how collector-contrib repo is handling this? if they are giving out write permissions on the repo to solve this, then seems reasonable to do the same here

anuraaga:
collector-contrib still only has approvers with the ability to mark green and maintainers with the ability to merge. I think giving more approval access in contrib makes sense and we could probably do it here but don't know if there is another example of it.

anuraaga:
@trask In the meantime do you think you can do a quick skim / approve? My next PR has quite some overlap with this so it being merged would be helpful.

trask (approved):

trask:
> @anuraaga do you know how collector-contrib repo is handling this? if they are giving out write permissions on the repo to solve this, then seems reasonable to do the same here

I pinged the collector-contrib team to see how they are handling this. In the meantime, I went ahead and gave @willarmiros write permission to this repo to unblock.

PR #53, opened 2021-07-07
Fetch and update rules with sampling targets.

anuraaga:
Logic should be basically complete, want to get an initial review of this hairy logic before jumping into hairier tests.

### DIFF
--- a/aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/GetSamplingTargetsResponse.java
+++ b/aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/GetSamplingTargetsResponse.java
@@ -51,6 +51,8 @@ abstract class GetSamplingTargetsResponse {
     @Nullable
     abstract Integer getReservoirQuota();
 
+    // Careful that this is a timestamp when the quota expires, not a duration as we'd normally
+    // expect for a Time to live.
     @Nullable
     abstract Date getReservoirQuotaTtl();
 
--- a/aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/GetSamplingTargetsResponse.java
+++ b/aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/GetSamplingTargetsResponse.java
@@ -35,8 +35,8 @@ abstract class GetSamplingTargetsResponse {
     @JsonCreator
     static SamplingTargetDocument create(
         @JsonProperty("FixedRate") double fixedRate,
-        @JsonProperty("Interval") int intervalSecs,
-        @JsonProperty("ReservoirQuota") int reservoirQuota,
+        @JsonProperty("Interval") @Nullable Integer intervalSecs,
+        @JsonProperty("ReservoirQuota") @Nullable Integer reservoirQuota,
         @JsonProperty("ReservoirQuotaTTL") @Nullable Date reservoirQuotaTtl,
         @JsonProperty("RuleName") String ruleName) {
       return new AutoValue_GetSamplingTargetsResponse_SamplingTargetDocument(
@@ -45,9 +45,11 @@ abstract class GetSamplingTargetsResponse {
 
     abstract double getFixedRate();
 
-    abstract int getIntervalSecs();
+    @Nullable
+    abstract Integer getIntervalSecs();
 
-    abstract int getReservoirQuota();
+    @Nullable
+    abstract Integer getReservoirQuota();
 
     @Nullable
     abstract Date getReservoirQuotaTtl();
--- a/aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/SamplingRuleApplier.java
+++ b/aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/SamplingRuleApplier.java
@@ -8,6 +8,9 @@ import io.opentelemetry.api.common.AttributeKey;
 import io.opentelemetry.api.common.Attributes;
 import io.opentelemetry.api.trace.SpanKind;
 import io.opentelemetry.context.Context;
+import io.opentelemetry.contrib.awsxray.GetSamplingTargetsRequest.SamplingStatisticsDocument;
+import io.opentelemetry.contrib.awsxray.GetSamplingTargetsResponse.SamplingTargetDocument;
+import io.opentelemetry.sdk.common.Clock;
 import io.opentelemetry.sdk.resources.Resource;
 import io.opentelemetry.sdk.trace.data.LinkData;
 import io.opentelemetry.sdk.trace.samplers.Sampler;
@@ -15,11 +18,13 @@ import io.opentelemetry.sdk.trace.samplers.SamplingDecision;
 import io.opentelemetry.sdk.trace.samplers.SamplingResult;
 import io.opentelemetry.semconv.resource.attributes.ResourceAttributes;
 import io.opentelemetry.semconv.trace.attributes.SemanticAttributes;
+import java.time.Duration;
 import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.LongAdder;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
@@ -44,7 +49,9 @@ final class SamplingRuleApplier {
 
   private final String clientId;
   private final String ruleName;
+  private final Clock clock;
   private final Sampler reservoirSampler;
+  private final long reservoirEndTimeNanos;
   private final Sampler fixedRateSampler;
   private final boolean borrowing;
 
@@ -58,20 +65,30 @@ final class SamplingRuleApplier {
 
   private final Statistics statistics;
 
-  SamplingRuleApplier(String clientId, GetSamplingRulesResponse.SamplingRule rule) {
+  private final long nextSnapshotTimeNanos;
+
+  SamplingRuleApplier(String clientId, GetSamplingRulesResponse.SamplingRule rule, Clock clock) {
     this.clientId = clientId;
+    this.clock = clock;
     ruleName = rule.getRuleName();
+
+    // We don't have a SamplingTarget so are ready to report a snapshot right away.
+    nextSnapshotTimeNanos = clock.nanoTime();
+
+    // We either have no reservoir sampling or borrow until we get a quota so have no end time.
+    reservoirEndTimeNanos = Long.MAX_VALUE;
+
     if (rule.getReservoirSize() > 0) {
       // Until calling GetSamplingTargets, the default is to borrow 1/s if reservoir size is
       // positive.
-      reservoirSampler = new RateLimitingSampler(1);
+      reservoirSampler = createRateLimited(1);
       borrowing = true;
     } else {
       // No reservoir sampling, we will always use the fixed rate.
       reservoirSampler = Sampler.alwaysOff();
       borrowing = false;
     }
-    fixedRateSampler = Sampler.parentBased(Sampler.traceIdRatioBased(rule.getFixedRate()));
+    fixedRateSampler = createFixedRate(rule.getFixedRate());
 
     if (rule.getAttributes().isEmpty()) {
       attributeMatchers = Collections.emptyMap();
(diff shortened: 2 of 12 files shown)


aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/SamplingRuleApplier.java
```diff
@@ -138,9 +190,12 @@ SamplingResult shouldSample(
       List<LinkData> parentLinks) {
     // Incrementing requests first ensures sample / borrow rate are positive.
     statistics.requests.increment();
+    boolean reservoirExpired = System.nanoTime() >= reservoirEndTimeNanos;
     SamplingResult result =
-        reservoirSampler.shouldSample(
-            parentContext, traceId, name, spanKind, attributes, parentLinks);
+        !reservoirExpired
+            ? reservoirSampler.shouldSample(
+                parentContext, traceId, name, spanKind, attributes, parentLinks)
```
willarmiros, line 197: What is the purpose of checking for the reservoir being expired here? Maybe I'm misunderstanding the meaning of expired, but shouldn't the reservoirSampler itself understand whether the amount requests to sample in a given second are used up or not?

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/GetSamplingTargetsResponse.java
```diff
@@ -45,9 +45,11 @@ static SamplingTargetDocument create(
 
     abstract double getFixedRate();
 
-    abstract int getIntervalSecs();
+    @Nullable
+    abstract Integer getIntervalSecs();
```
willarmiros, line 49: What is the benefit of this change? Is this just how Jackson works with optional fields?

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/SamplingRuleApplier.java
willarmiros: I believe I answered my own question - expired here refers to the reservoir quota's TTL, not to the 1-second window for which the quota applies. Leaving the comment so you can confirm.

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/SamplingRuleApplier.java
```diff
@@ -171,6 +230,71 @@ SamplingResult shouldSample(
...
+  SamplingRuleApplier withTarget(SamplingTargetDocument target, Date now) {
+    Sampler newFixedRateSampler = createFixedRate(target.getFixedRate());
+    Sampler newReservoirSampler = Sampler.alwaysOff();
+    long newReservoirEndTimeNanos = System.nanoTime();
+    // Not well documented but a quota should always come with a TTL
+    if (target.getReservoirQuota() != null && target.getReservoirQuotaTtl() != null) {
+      newReservoirSampler = createRateLimited(target.getReservoirQuota());
+      newReservoirEndTimeNanos =
+          clock.nanoTime()
+              + Duration.between(now.toInstant(), target.getReservoirQuotaTtl().toInstant())
```
willarmiros, line 246: Is it typical that a TTL is actually an expiration timestamp and not a duration of time? If not, maybe worth referencing the API docs to clarify the TTL is in fact a timestamp to avoid confusion: https://docs.aws.amazon.com/xray/latest/api/API_SamplingTargetDocument.html

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/XrayRulesSampler.java
```diff
@@ -70,4 +96,46 @@ public SamplingResult shouldSample(
...
+    return Arrays.stream(ruleAppliers)
+        .map(rule -> rule.snapshot(now))
+        .filter(Objects::nonNull)
+        .collect(Collectors.toList());
+  }
+
+  long nextTargetFetchTimeNanos() {
+    return Arrays.stream(ruleAppliers)
+        .mapToLong(SamplingRuleApplier::getNextSnapshotTimeNanos)
+        .min()
```
willarmiros, line 110: Does this do away with the constant-interval polling?

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/XrayRulesSampler.java
```diff
@@ -70,4 +96,46 @@ public SamplingResult shouldSample(
...
+      Map<String, SamplingTargetDocument> ruleTargets,
+      Set<String> requestedTargetRuleNames,
+      Date now) {
+    long defaultNextSnapshotTimeNanos =
+        clock.nanoTime() + AwsXrayRemoteSampler.DEFAULT_TARGET_INTERVAL_NANOS;
+    SamplingRuleApplier[] newAppliers =
+        Arrays.stream(ruleAppliers)
+            .map(
+                rule -> {
+                  SamplingTargetDocument target = ruleTargets.get(rule.getRuleName());
```
willarmiros, line 125: Does this constrain the amount of rule appliers to always be the same as the starting amount of rule appliers? For example, if a customer adds a new rule, and we fetch it, I believe there should now be 1 additional applier (because appliers are 1:1 with customer rules). But in this map, the `newAppliers` is always the same size as the current appliers. Does this get updated elsewhere? And the use I described of adding a new rule will still work with this logic?

willarmiros (commented):
Hairy indeed! Tough to reason through everything in my head, so just want to sanity check some things, but for the most part I see everything working as it should.

anuraaga (commented):
Oops forgot to hit submit, here are some explanatory comments I had written yesterday

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/GetSamplingTargetsResponse.java
anuraaga: It defaults to "default" if not present, `0` for an int or `null` for objects. As `0` doesn't seem like a valid interval, I guess comparing to 0 instead of comparing to null would work too but it's not quite as straightforward (though does allocate a bit less memory I guess). For what it's worth, AWS SDK uses non-primitive types for all of these to indicate they could be null

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/SamplingRuleApplier.java
anuraaga: Yup, after the TTL we switch back to fully fixed-rate sampling

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/SamplingRuleApplier.java
anuraaga: This was indeed very confusing, linking seems good

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/XrayRulesSampler.java
anuraaga: Yeah - given we have intervals for the targets in the responses themselves there doesn't seem to be a need to fixed-poll targets

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/XrayRulesSampler.java
anuraaga: Rule updates are handled by `AwsXrayRemoteSampler.getAndUpdateSampler`. If rules changed, any pending target fetches are cancelled and they'll be fetched again for the new rules. So basically "updating targets" indeed can never increase the number of rules, but "fetching rules" can and will invalidate targets if rules change.

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/GetSamplingTargetsResponse.java
willarmiros: I see - and the problem *could be* that if 0 was a valid interval, we'd be treating a not-present interval the same as a provided 0-value interval.

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/GetSamplingTargetsResponse.java
```diff
@@ -35,8 +35,8 @@ static GetSamplingTargetsResponse create(
     @JsonCreator
     static SamplingTargetDocument create(
         @JsonProperty("FixedRate") double fixedRate,
```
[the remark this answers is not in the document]
willarmiros: I think it's safe to assume this will always be present, since when creating a sampling rule `fixedRate` is required: https://docs.aws.amazon.com/xray/latest/api/API_SamplingRule.html

willarmiros (approved):
LGTM on the initial implementation. In addition to unit tests, one other thing that would be really nice is debug statements. Some ideas are logging:
* When we match a rule and what matched
* When we're polling for targets/rules, and maybe some details about the response
* When the reservoir expires and we're switching to fixed-rate
* What type of sampling is being used for a sample decision (borrow vs. reservoir vs. fixed rate)
* Maybe even logging the sampling statistics when we snapshot them?

Just some thoughts to help our future selves as someone who's spent more than a few hours debugging weird sampling behavior ;)

anuraaga:
@willarmiros PTAL

willarmiros (commented):
I think just some last clarifications needed on the logic and/or tests then lgtm

aws-xray/src/test/java/io/opentelemetry/contrib/awsxray/SamplingRuleApplierTest.java
```diff
@@ -585,9 +529,130 @@ void borrowing() {
...
+        .isEqualTo(SamplingResult.create(SamplingDecision.RECORD_AND_SAMPLE));
+    assertThat(doSample(applier)).isEqualTo(SamplingResult.create(SamplingDecision.DROP));
+    clock.advance(Duration.ofSeconds(3));
+    now = Instant.ofEpochSecond(0, clock.now());
+    // Statistics expired, snapshot requests are served.
+    assertThat(doSample(applier))
+        .isEqualTo(SamplingResult.create(SamplingDecision.RECORD_AND_SAMPLE));
+    assertThat(doSample(applier))
+        .isEqualTo(SamplingResult.create(SamplingDecision.RECORD_AND_SAMPLE));
+    assertThat(doSample(applier)).isEqualTo(SamplingResult.create(SamplingDecision.DROP));
```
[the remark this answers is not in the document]
anuraaga: The expiration of the reservoir (TTL) is 10s, not 5s (`now.plusSeconds(10)`). 5s is the time until next target poll (Interval)

aws-xray/src/test/java/io/opentelemetry/contrib/awsxray/SamplingRuleApplierTest.java
```diff
@@ -585,9 +529,130 @@ void borrowing() {
...
+    assertThat(applier.snapshot(Date.from(now))).isNotNull();
```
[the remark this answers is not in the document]
anuraaga: We only request targets for expired rules, presumably to reduce backend load? So null means that it's not sent in the GetSamplingTargets request. It's not the clearest thing but it's an internal implementation detail and similar to the xray sdk java so figure it's ok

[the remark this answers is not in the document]
willarmiros: cool makes sense

aws-xray/src/test/java/io/opentelemetry/contrib/awsxray/SamplingRuleApplierTest.java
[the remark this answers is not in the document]
willarmiros: Got it, so my comment would be valid if you had advanced 10 seconds instead of 5 right?

willarmiros (approved):
Thanks for the clarifications!

aws-xray/src/test/java/io/opentelemetry/contrib/awsxray/SamplingRuleApplierTest.java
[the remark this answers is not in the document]
anuraaga: Yup - so we advance another 5s below, where the comment says "No more reservoir quota"

PR #56, opened 2021-07-26
Add manually run test code to exercise AWS X-Ray remote sampler end-t…

anuraaga:
…o-end.

Confirmed that I can influence the sampling rate printing by editing rules in the console. It's hard to confirm targets are working since can't really reason about them though.

### DIFF
--- a/aws-xray/src/awsTest/resources/otel-collector.yml
+++ b/aws-xray/src/awsTest/resources/otel-collector.yml
@@ -1,0 +1,15 @@
+receivers:
+  awsxray:
+
+extensions:
+  health_check:
+
+exporters:
+  logging:
+
+service:
+  extensions: [health_check]
+  pipelines:
+    traces:
+      receivers: [awsxray]
+      exporters: [logging]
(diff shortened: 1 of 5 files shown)


aws-xray/src/awsTest/resources/otel-collector.yml
```diff
@@ -0,0 +1,15 @@
+receivers:
+  awsxray:
```
anuraaga, line 2: Yup, think this does need to be migrated to an `extension`

rmfitzpatrick: Is this receiver different than other trace receivers such that it would apply to all or multiple pipelines?

rmfitzpatrick (approved):

willarmiros (approved):
Hoping we can get AWS creds in the near future - for this and automated performance tests.

aws-xray/src/awsTest/resources/otel-collector.yml
anuraaga: It currently is indeed - it opens a separate port which acts as a proxy for sampling requests outside the pipeline. This follows the design of the x-ray daemon but feels weird due to being outside the pipeline

PR #57, opened 2021-07-27
Don't use static client ID to allow multiple xray samplers in same pr…

anuraaga:
In practice, there is only one initialized SDK in an app with a sampler. But there could be use cases for having multiple client IDs in the same process (like the integration test we have) and it doesn't really make sense to forbid this by using a static.

Also fix serialization of Timestamp - I thought it is String based on the AWS CLI output in the [doc](https://docs.aws.amazon.com/xray/latest/devguide/xray-api-sampling.html) but the CLI just formats that for humans but must be sending milliseconds in the actual API call. The Jackson default is milliseconds so no need for anything special.

Was able to confirm targets acting something like targets - with 2 r/s centralized reservoir, it often allocated 1/s for each sampler, but sometimes ended up with 2/s for one or both, with a period of 0/s presumably as it compensated for the oversampling, and back to 1/s, etc. Seems like what one might expect.

### DIFF
--- a/aws-xray/src/awsTest/java/io/opentelemetry/contrib/aws/xray/AwsXrayRemoteSamplerIntegrationTest.java
+++ b/aws-xray/src/awsTest/java/io/opentelemetry/contrib/aws/xray/AwsXrayRemoteSamplerIntegrationTest.java
@@ -10,6 +10,8 @@ import io.opentelemetry.context.Context;
 import io.opentelemetry.contrib.awsxray.AwsXrayRemoteSampler;
 import io.opentelemetry.sdk.resources.Resource;
 import io.opentelemetry.sdk.trace.IdGenerator;
+import io.opentelemetry.sdk.trace.samplers.Sampler;
+import io.opentelemetry.sdk.trace.samplers.SamplingResult;
 import java.time.Duration;
 import java.util.Collections;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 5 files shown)


aws-xray/src/awsTest/java/io/opentelemetry/contrib/aws/xray/AwsXrayRemoteSamplerIntegrationTest.java
```diff
@@ -49,29 +51,38 @@
...
+            .build();
+
+    AwsXrayRemoteSampler sampler2 =
         AwsXrayRemoteSampler.newBuilder(Resource.getDefault())
             .setEndpoint("http://localhost:" + otelCollector.getMappedPort(2000))
             .setPollingInterval(Duration.ofSeconds(5))
             .build();
 
     try {
       while (true) {
```
willarmiros, line 68: Should've asked the first time around - but when does this loop exit? Or is it not intended to exit, because it's only manually runnable so the human is meant to terminate it?

willarmiros (approved):
LGTM, glad to hear the targets are being targeted :)

aws-xray/src/awsTest/java/io/opentelemetry/contrib/aws/xray/AwsXrayRemoteSamplerIntegrationTest.java
anuraaga: Right since it's just for running manually this doesn't exit until user kills the test

PR #61, opened 2021-08-02
Separate out workflow for main branch push and publish snapshots.

anuraaga:


anuraaga:
I removed the integration-test on build, don't think it's necessary.

trask (approved):

PR #62, opened 2021-08-03
Add release workflows

anuraaga:


trask (approved):

.github/workflows/patch-release-build.yml
```diff
@@ -0,0 +1,121 @@
...
+# release tag.
+name: Patch Release Build
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: The version to tag the release with, e.g., 1.2.1, 1.2.2
+        required: true
+      commits:
+        description: Comma separated list of commit shas to cherrypick
```
mateuszrzeszutek, line 11: We're not doing any cherrypicks in this PR yet, is that correct?

.github/workflows/release-build.yml
```diff
@@ -0,0 +1,80 @@
...
+        name: Integration Tests
+        with:
+          arguments: --stacktrace integrationTest
+          remote-build-cache-proxy-enabled: false
+      - uses: actions/upload-artifact@v2
+        name: Save integrationTest results
+        if: always()
+        with:
+          name: integration-test-results
+          path: jmx-metrics/build/reports/tests/test
```
mateuszrzeszutek, line 56: Just checking: is the report path for `integrationTest` the same as for `test`?

.github/workflows/release-build.yml
```diff
@@ -0,0 +1,80 @@
...
+        name: Publish
+        with:
+          arguments: --stacktrace final closeAndReleaseSonatypeStagingRepository -Prelease.version=${{ github.event.inputs.version }}
+        env:
+          SONATYPE_USER: ${{ secrets.SONATYPE_USER }}
+          SONATYPE_KEY: ${{ secrets.SONATYPE_KEY }}
+          GRGIT_USER: ${{ github.actor }}
+          GRGIT_PASS: ${{ secrets.GITHUB_TOKEN }}
+          GPG_PRIVATE_KEY: ${{ secrets.GPG_PRIVATE_KEY }}
+          GPG_PASSWORD: ${{ secrets.GPG_PASSWORD }}
```
mateuszrzeszutek, line 80: Nit: should we create a GH release here?

.github/workflows/patch-release-build.yml
anuraaga: Nah we're not doing any releasing in this PR, just adding workflows

.github/workflows/release-build.yml
anuraaga: Right we currently override the default `test` task. I want to change to test-sets eventually

.github/workflows/release-build.yml
anuraaga: Oh yeah, added

mateuszrzeszutek (approved):

PR #67, opened 2021-08-06
Allow configuring endpoint with environment variable.

anuraaga:
While I didn't want to introduce too much environment variable configuration initially, especially when even the sampler name is stuck in spec review, I realized it's too important to be able to change endpoint for docker use.


trask (approved):

willarmiros (approved):
lgtm

PR #68, opened 2021-08-19
First try rate limited sampler for initialSampler of XRay remote samp…

anuraaga:
…ler.

To match behavior of X-Ray SDK and because testing apps locally can be harder if the backend is required just to see traces, 1 trace per second ensures that traces are created regardless of backend.


willarmiros (approved):
lgtm

PR #73, opened 2021-09-02
Update to OTel 1.5

### LINKED ISSUE #72: Latest JMX gatherer broken with collector JMX receiver
JMX Gatherer 1.4.x release is unable to communicate with the collector (last collector release doesn't support latest OTLP metric attributes).

I think this could be fixed by bumping to 1.5.x where both labels + attributes are sent (or we can wait for the next collector version which will support full OTLP 0.9).

anuraaga:
Fixes #72?

### DIFF
--- a/aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsXrayRemoteSamplerProvider.java
+++ b/aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsXrayRemoteSamplerProvider.java
@@ -6,7 +6,7 @@ package io.opentelemetry.contrib.awsxray;
 
 import com.google.auto.service.AutoService;
 import io.opentelemetry.sdk.autoconfigure.ConfigProperties;
-import io.opentelemetry.sdk.autoconfigure.OpenTelemetrySdkAutoConfiguration;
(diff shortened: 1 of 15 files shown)


jsuereth (approved):
LGTM, barring I didn't look into details on the proxy-callback semantics (and those actually concern me.  I'm wondering if there's an API issue we need to address rather than whatever is being done there)

anuraaga:
Yeah my understanding of the comments is there is something specific to jmx that is needing the pattern but not sure the specifics. @rmfitzpatrick anything the API could change to remove the proxy callbacks?

trask (approved):

rmfitzpatrick:
@anuraaga, @jsuereth sorry I missed this.  The proxy helper is to allow updating the functionality of the callback with each script interval while accounting for the enforced single registration.  The api changed sometime ago but `setUpdater()` used to take effect with each call and this system was needed as a workaround to allow detecting changes to mbeans with the same meter name but differing datapoint attributes.  I'm not sure if any changes are needed to the actual api since this solves the problem, but enabling callback updates would be nice if still not possible.

Thank you for taking care of these updates, @anuraaga.

Issue #78, opened 2021-09-10
Donating a Maven OpenTelemetry extension to the OpenTelemetry project?


cyrille-leclerc:
Noter that this message is a cross post of https://github.com/open-telemetry/community/discussions/832 as suggested by @trask .

---

Dear community,

We would like to donate to the OpenTelemetry project a [Maven OpenTelemetry extension](https://github.com/elastic/opentelemetry-maven-extension) that we have created to observe the execution of Maven builds.

This Maven instrumentation integrates with other OpenTelemetry capable CI/CD tools such as Jenkins pipelines ([Jenkins OpenTelemetry plugin](https://github.com/jenkinsci/opentelemetry-plugin)), Ansible deployments ([Ansible OpenTelemetry plugin](https://github.com/ansible-collections/community.general/pull/3091)) or the [otel-cli](https://github.com/equinix-labs/otel-cli).
 
Here is an example of distributed trace providing visibility of a Jenkins / Maven CI pipeline.

![](https://raw.githubusercontent.com/cyrille-leclerc/opentelemetry-maven-extension/main/docs/images/jenkins-maven-execution-trace-jaeger.png)

It's very straightforward to test the extension on any existing Maven project as we have documented [here](https://github.com/elastic/opentelemetry-maven-extension#getting-started).

We have very positive feedback from CI/CD practitioners when we show the capabilities and the extension is already used by Elastic on its production CI platform in conjunction with the Jenkins OpenTelemetry plugin.

We would like to donate the Maven OpenTelemetry extension to the OpenTelemetry project / CNCF foundation because we feel that it would be the best place to develop this integration and because we see the opportunity to create an entire ecosystem of  OpenTelemetry capable CI/CD tools.

Note that we preferred to build this integration as a Maven extension rather than as an auto instrumentation that would be part of the [OpenTelemetry Java agent](https://github.com/open-telemetry/opentelemetry-java-instrumentation) because the Maven community is used to define its builds adding plugins and extension to build definitions (pom.xml) rather than wiring a Java auto-instrumentation agent when launching the "mvn" command.

**Would the OpenTelemetry Java Contrib project be interested in hosting such [Maven OpenTelemetry integration](https://github.com/elastic/opentelemetry-maven-extension)?**

Following the [donations guidelines](https://github.com/open-telemetry/community/blob/main/CONTRIBUTING.md#donations), I feel this donation can be qualified as a "small donation" and a simple Pull Request would then be enough.

[...]

PR #79, opened 2021-09-12
Maven OpenTelemetry Extension initial commit

cyrille-leclerc:
**Description:**

As discussed on [Donating a Maven OpenTelemetry extension to the OpenTelemetry project? #78](https://github.com/open-telemetry/opentelemetry-java-contrib/issues/78) with @kenfinnigan and @jkwatson, here is the PR to donate a Maven extension to observe Maven build executions with OpenTelemetry and distributed traces.

**Existing Issue(s):**

- https://github.com/open-telemetry/opentelemetry-java-contrib/issues/78

**Testing:**

Some unit tests and manual end to end tests

**Documentation:**

See `README.md` file added to the submodule: https://github.com/cyrille-leclerc/opentelemetry-java-contrib/blob/add-maven-extension-v2/maven-extension/README.md

**Outstanding items:**

* Introduce a subproject `maven-extension` to the `opentelemetry-java-contrib` project
* To test, as documented on https://github.com/cyrille-leclerc/opentelemetry-java-contrib/blob/add-maven-extension-v2/maven-extension/README.md :
   * Build opentelemetry-java-contrib with `./gradlew build`
   * Start an OpenTelemetry endpoint, e.g. an OpenTelemetry Collector listening on http://localhost:4317 that would forward to Jaeger
   * Build Maven project with
```
      
mvn -Dmaven.ext.class.path=/path/to/opentelemetry-java-contrib/maven-extension/build/libs/opentelemetry-maven-extension-1.6.0-SNAPSHOT.jar verify
```
   * Verify that the build have been captured in Jaeger


maven-extension/README.md
```diff
@@ -0,0 +1,95 @@
...
+![](https://raw.githubusercontent.com/cyrille-leclerc/opentelemetry-java-contrib/add-maven-extension-v2/maven-extension/docs/images/maven-execution-trace-jaeger.png)
+
+## Example of a distributed trace of a Jenkins pipeline executing a Maven build
+
+Distributed trace combining the OpenTelemetry Maven Extension with the [Jenkins OpenTelemetry plugin](https://plugins.jenkins.io/opentelemetry/).
+
+### Trace visualized with [Elastic Observability](https://www.elastic.co/observability)
+
+![](https://raw.githubusercontent.com/cyrille-leclerc/opentelemetry-java-contrib/add-maven-extension-v2/maven-extension/docs/images/jenkins-maven-execution-trace-elastic.png)
+
```
trask, line 80: we try not to include vendor-specific references in general purpose docs

trask (commented):
thx @cyrille-leclerc, this looks really useful!

maven-extension/src/main/java/io/opentelemetry/maven/OtelLifecycleParticipant.java
```diff
@@ -0,0 +1,45 @@
...
+import org.apache.maven.execution.ExecutionListener;
+import org.apache.maven.execution.MavenSession;
+import org.codehaus.plexus.component.annotations.Component;
+import org.codehaus.plexus.component.annotations.Requirement;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Add the {@link OtelExecutionListener} to the lifecycle of the MAven execution */
+@Component(role = AbstractMavenLifecycleParticipant.class)
+public class OtelLifecycleParticipant extends AbstractMavenLifecycleParticipant {
```
anuraaga, line 18: If I'm not mistaken, this is the entry-point so hopefully this is the only class we need to make `public`

maven-extension/src/main/java/io/opentelemetry/maven/OtelExecutionListener.java
```diff
@@ -0,0 +1,248 @@
...
+ * execution of the last project ({@link #projectSucceeded(ExecutionEvent)} and {@link
+ * #projectFailed(ExecutionEvent)}) rather than on the end of the Maven session {@link
+ * #sessionEnded(ExecutionEvent)} because OpenTelemetry and GRPC classes are unloaded by the Maven
+ * classloader before {@link #sessionEnded(ExecutionEvent)} causing {@link NoClassDefFoundError}
+ * messages in the logs.
+ */
+@Component(role = ExecutionListener.class, hint = "otel-execution-listener")
+public class OtelExecutionListener extends AbstractExecutionListener {
+
+  private final Logger logger = LoggerFactory.getLogger(getClass());
```
anuraaga, line 39:
```suggestion
  private final Logger logger = LoggerFactory.getLogger(OtelExecutionListener.class);
```

maven-extension/src/main/java/io/opentelemetry/maven/OtelExecutionListener.java
```diff
@@ -0,0 +1,248 @@
...
+   *
+   * @see org.apache.maven.execution.MavenExecutionRequest#setExecutionListener(ExecutionListener)
+   */
+  public static void registerOtelExecutionListener(
+      @Nonnull MavenSession session, @Nonnull OtelExecutionListener otelExecutionListener) {
+    @Nullable
+    ExecutionListener initialExecutionListener = session.getRequest().getExecutionListener();
+    if (initialExecutionListener instanceof ChainedExecutionListener) {
+      // already initialized
+      LoggerFactory.getLogger(OtelExecutionListener.class)
```
anuraaga, line 57:
```suggestion
      logger
```

maven-extension/src/main/java/io/opentelemetry/maven/OtelExecutionListener.java
```diff
@@ -0,0 +1,248 @@
...
+          .debug(
+              "OpenTelemetry: OpenTelemetry extension already registered as execution listener, skip.");
+    } else {
+      session
+          .getRequest()
+          .setExecutionListener(
+              new ChainedExecutionListener(otelExecutionListener, initialExecutionListener));
+      LoggerFactory.getLogger(OtelExecutionListener.class)
```
anuraaga, line 65:
```suggestion
      logger
```

maven-extension/src/main/java/io/opentelemetry/maven/OtelExecutionListener.java
```diff
@@ -0,0 +1,248 @@
...
+    if (initialExecutionListener instanceof ChainedExecutionListener) {
+      // already initialized
+      LoggerFactory.getLogger(OtelExecutionListener.class)
+          .debug(
+              "OpenTelemetry: OpenTelemetry extension already registered as execution listener, skip.");
+    } else {
+      session
+          .getRequest()
+          .setExecutionListener(
+              new ChainedExecutionListener(otelExecutionListener, initialExecutionListener));
```
anuraaga, line 64: Instead of filtering null it's clearer (and probably a bit more performant) to only use the chained when initialExecutionListener isn't null

maven-extension/src/main/java/io/opentelemetry/maven/OtelExecutionListener.java
```diff
@@ -0,0 +1,248 @@
...
+
+  /**
+   * Register in given {@link OtelExecutionListener} to the lifecycle of the given {@link
+   * MavenSession}
+   *
+   * @see org.apache.maven.execution.MavenExecutionRequest#setExecutionListener(ExecutionListener)
+   */
+  public static void registerOtelExecutionListener(
+      @Nonnull MavenSession session, @Nonnull OtelExecutionListener otelExecutionListener) {
+    @Nullable
```
anuraaga, line 53: We don't use Nullable annotations on local variables.

maven-extension/src/main/java/io/opentelemetry/maven/OtelExecutionListener.java
```diff
@@ -0,0 +1,248 @@
...
+  @Requirement private OpenTelemetrySdkService openTelemetrySdkService;
+
+  /**
+   * Register in given {@link OtelExecutionListener} to the lifecycle of the given {@link
+   * MavenSession}
+   *
+   * @see org.apache.maven.execution.MavenExecutionRequest#setExecutionListener(ExecutionListener)
+   */
+  public static void registerOtelExecutionListener(
+      @Nonnull MavenSession session, @Nonnull OtelExecutionListener otelExecutionListener) {
```
anuraaga, line 52: We don't use `Nonnull` annotations, provide a package-info instead similar to

https://github.com/open-telemetry/opentelemetry-java-contrib/blob/main/aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/package-info.java#L6

maven-extension/src/main/java/io/opentelemetry/maven/OtelExecutionListener.java
```diff
@@ -0,0 +1,248 @@
...
+    @Nullable
+    ExecutionListener initialExecutionListener = session.getRequest().getExecutionListener();
+    if (initialExecutionListener instanceof ChainedExecutionListener) {
```
anuraaga, line 55: IIUC, due to a limitation of maven's extensions, there isn't really a way to guarantee a different extension didn't overwrite our execution listener after we run. This is OK, we should add to the README if there's any POM ordering constraint.

maven-extension/src/main/java/io/opentelemetry/maven/OtelExecutionListener.java
```diff
@@ -0,0 +1,248 @@
...
+          .debug(
+              "OpenTelemetry: OpenTelemetry extension registered as execution listener. InitialExecutionListener: "
+                  + initialExecutionListener);
+    }
+  }
+
+  @Override
+  public void sessionStarted(ExecutionEvent executionEvent) {
+    MavenProject project = executionEvent.getSession().getTopLevelProject();
+    TextMapGetter<Map<String, String>> getter =
```
anuraaga, line 75: Let's go ahead and extract a constant for this

maven-extension/src/main/java/io/opentelemetry/maven/OtelExecutionListener.java
```diff
@@ -0,0 +1,248 @@
...
+          @Nullable
+          @Override
+          public String get(@Nullable Map<String, String> environmentVariables, String key) {
+            return environmentVariables == null
+                ? null
+                : environmentVariables.get(key.toUpperCase(Locale.ROOT));
+          }
+        };
+    io.opentelemetry.context.Context context =
+        W3CTraceContextPropagator.getInstance()
```
anuraaga, line 91: Use `OpenTelemetry.getPropagators` instead of hard-coding the propagation format.

maven-extension/src/main/java/io/opentelemetry/maven/OtelExecutionListener.java
```diff
@@ -0,0 +1,248 @@
...
+            .extract(io.opentelemetry.context.Context.current(), System.getenv(), getter);
+    try (Scope scope = context.makeCurrent()) {
```
anuraaga, line 93: A scope is a threadlocal that is generally meant to be accessed in some downstream code (e.g. in a servlet filter type of api, `.next()` would be called within the scope). There's no reason to make the scope current with this sort of pattern since there is no downstream code being called.

It seems that instead of threadlocal context propagation, we get around it by always parenting projects to the root span, and always parenting mojos to the project. This is OK and we could consider improving in the future.

The main issue is that other mojos won't be able to add their own data to the trace if they ever get instrumented. If Maven's executor _always_ processes a project in a single thread (e.g., it gets multi-threading by dispatching projects across threads, and it's guaranteed to be a contiguous start - finish of a single project on a single thread), then in `projectStarted` we would want to `makeCurrent`() without using try / resources (highly unsafe which is why we need to know the threading model correctly) and close the scope in the projectSucceeded / Failed.

Unless you're sure of the threading model, let's definitely not try that in this PR

maven-extension/src/main/java/io/opentelemetry/maven/SpanRegistry.java
```diff
@@ -0,0 +1,245 @@
...
+import java.util.Objects;
+import java.util.stream.Collectors;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import org.apache.maven.model.Plugin;
+import org.apache.maven.plugin.MojoExecution;
+import org.apache.maven.project.MavenProject;
+import org.codehaus.plexus.component.annotations.Component;
+
+/** Holds the state of the spans in progress FIXME support multi module projects */
```
anuraaga, line 19: What's needed for multi module (I thought project execution is tied to a module). Just curious

maven-extension/src/main/java/io/opentelemetry/maven/SpanRegistry.java
```diff
@@ -0,0 +1,245 @@
...
+@Component(role = SpanRegistry.class)
+public class SpanRegistry {
+
+  private final Map<MojoExecutionKey, Span> mojoExecutionKeySpanMap = new HashMap<>();
```
anuraaga, line 23: Perhaps these aren't ConcurrentHashMap due to not supporting multi module yet

maven-extension/src/main/java/io/opentelemetry/maven/SpanRegistry.java
```diff
@@ -0,0 +1,245 @@
...
+  public Span removeSpan(@Nonnull MojoExecution mojoExecution) throws IllegalStateException {
+    MojoExecutionKey key = MojoExecutionKey.fromMojoExecution(mojoExecution);
+    Span span = mojoExecutionKeySpanMap.remove(key);
+    if (span == null) {
+      throw new IllegalStateException();
+    }
+    return span;
+  }
+
+  private static class MavenProjectKey {
```
anuraaga, line 113: We can use auto-value for these

https://github.com/open-telemetry/opentelemetry-java-contrib/blob/main/aws-xray/build.gradle.kts#L27
https://github.com/open-telemetry/opentelemetry-java-contrib/blob/ed31fbaac9b936e0b68152f3c74412ef8c8d08c2/aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/GetSamplingTargetsRequest.java#L13

anuraaga (commented):
Thanks @cyrille-leclerc, looking forward to this! Did a first pass

cyrille-leclerc:
Many thanks for your warm welcome and for your lengthy review, I'll fix all this ASAP, mid or end of this week.

maven-extension/build.gradle.kts
```diff
@@ -0,0 +1,58 @@
...
+
+description = "Maven extension to observe Maven builds with distributed traces using OpenTelemetry SDK"
+
+// FIXME disable ext.libraryName due to gradle error "Unresolved reference: libraryName"
+// ext.libraryName = 'OpenTelemetry Maven Extension'
+
+dependencies {
+  implementation(platform("io.opentelemetry:opentelemetry-bom:1.5.0"))
+  implementation(platform("io.opentelemetry:opentelemetry-bom-alpha:1.5.0-alpha"))
+  implementation(platform("io.grpc:grpc-bom:1.39.0"))
```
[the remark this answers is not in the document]
anuraaga: BTW after you merge main and remove this, you'll be on OTel 1.6 so hopefully can remove those workarounds :)

maven-extension/src/main/java/io/opentelemetry/maven/ChainedExecutionListener.java
```diff
@@ -0,0 +1,224 @@
...
+import java.util.stream.Collectors;
+import org.apache.maven.execution.ExecutionEvent;
+import org.apache.maven.execution.ExecutionListener;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Util class to chain multiple {@link ExecutionListener} as Maven APIs don't offer this capability.
+ */
+public class ChainedExecutionListener implements ExecutionListener {
```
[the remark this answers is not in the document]
cyrille-leclerc: Thanks @anuraaga . I applied your recommendations wherever it was easy. I'll have to see how I can combine this programming pattern with the rule of Maven Plexus dependency injection. Note that I'm not familiar with the Plexus dependency injection framework

maven-extension/src/main/java/io/opentelemetry/maven/ChainedExecutionListener.java
```diff
@@ -0,0 +1,224 @@
...
+
+  @Override
+  public void projectDiscoveryStarted(ExecutionEvent event) {
+    for (ExecutionListener listener : this.listeners) {
+      try {
+        listener.projectDiscoveryStarted(event);
+      } catch (RuntimeException e) {
+        logger.error("Silently skip exception", e);
+      }
+    }
```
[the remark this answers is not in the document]
cyrille-leclerc: It was a kind of paranoid protection. I remove it.

maven-extension/src/main/java/io/opentelemetry/maven/SpanRegistry.java
cyrille-leclerc: I guess it was more a multi threading bug as I implemented multi module support a while ago 👼 good catch thanks!

maven-extension/src/main/java/io/opentelemetry/maven/SpanRegistry.java
cyrille-leclerc: I forgot to remove this comment when I added the capability with https://github.com/elastic/opentelemetry-maven-extension/commit/ac37870a32b0bec6ea06066a531ef64d50fd0048 

I primarily had to introduce a "Maven project span" layer between the "Maven build span" and the "Mojo spans".
I removed this outdated comment.

maven-extension/src/main/java/io/opentelemetry/maven/OtelExecutionListener.java
```diff
@@ -0,0 +1,248 @@
...
+      final String spanName =
+          "Build: " + project.getGroupId() + ":" + project.getArtifactId(); // TODO find better name
```
kenfinnigan, line 95: Would `project.getName()` return the name of the Maven module? If so, that's probably a good name

maven-extension/src/main/java/io/opentelemetry/maven/SpanRegistry.java
```diff
@@ -0,0 +1,245 @@
...
+                  .map(MojoExecutionKey::toString)
+                  .collect(Collectors.joining(", ")));
+    }
+    return rootSpan;
+  }
+
+  public void putSpan(@Nonnull Span span, @Nonnull MavenProject mavenProject) {
+    MavenProjectKey key = MavenProjectKey.fromMavenProject(mavenProject);
+    Span previousSpanForKey = mavenProjectKeySpanMap.put(key, span);
+    if (previousSpanForKey != null) {
```
kenfinnigan, line 80: What is being prevented with this check? Is there a situation that another span for the same `MavenProject` could be created?

maven-extension/src/main/java/io/opentelemetry/maven/StringUtils.java
```diff
@@ -0,0 +1,37 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package io.opentelemetry.maven;
+
+final class StringUtils {
+
+  private StringUtils() {}
+
+  /** Copy org.apache.commons.lang3.StringUtils#isBlank(java.lang.CharSequence) */
+  public static boolean isBlank(final CharSequence cs) {
```
kenfinnigan, line 12: Any reason not to use the `String` class methods for these?

maven-extension/build.gradle.kts
```diff
@@ -0,0 +1,58 @@
...
+  id("otel.java-conventions")
+  id("otel.publish-conventions")
+}
+
+// Note: handle `META-INF/plexus/components.xml` is manually handled under src/main/resources because there is not Gradle equivalent to `plexus-component-metadata:generate-metadata`
+
+description = "Maven extension to observe Maven builds with distributed traces using OpenTelemetry SDK"
+
+// FIXME disable ext.libraryName due to gradle error "Unresolved reference: libraryName"
+// ext.libraryName = 'OpenTelemetry Maven Extension'
```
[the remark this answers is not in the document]
cyrille-leclerc: It was a copy paste of an example to produce a Maven friendly artifact when I was learning Gradle, I removed it

maven-extension/build.gradle.kts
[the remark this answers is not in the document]
cyrille-leclerc: Thanks.
> BTW after you merge main and remove this, you'll be on OTel 1.6 so hopefully can remove those workarounds :)

I'll have to merge the main branch on my PR branch, I'll have to re-learn how to do it 👼 I've left development for product management for almost 10 years ago

maven-extension/src/main/java/io/opentelemetry/maven/OpenTelemetrySdkService.java
```diff
@@ -0,0 +1,201 @@
...
+
+/**
+ * Service to configure the {@link OpenTelemetry} instance.
+ *
+ * <p>Mimic the <a
+ * href="https://github.com/open-telemetry/opentelemetry-java/tree/main/sdk-extensions/autoconfigure">OpenTelemetry
+ * SDK Autoconfigure</a> that could not be used as is due to class loading issues when declaring the
+ * Maven OpenTelemetry extension using the pom.xml {@code <extension>} declaration.
+ *
+ * <p>Exception example: https://gist.github.com/cyrille-leclerc/57903e63d1f162969154eec3bb82a576
```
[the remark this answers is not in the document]
cyrille-leclerc: will go away with 1.6.0

maven-extension/src/main/java/io/opentelemetry/maven/StringUtils.java
cyrille-leclerc: It's the copy/paste of Commons Lang3 [StringUtils.isBlank(s)](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#isBlank-java.lang.CharSequence- ) with minimum changes. Shall I make it more consistent and use String?

maven-extension/src/main/java/io/opentelemetry/maven/OtelExecutionListener.java
cyrille-leclerc: Comment added in the code to explain the rationale on the current choice

maven-extension/src/main/java/io/opentelemetry/maven/SpanRegistry.java
cyrille-leclerc: I have clarified with an IllegalStateException message. It should not happen, it would mean we have misunderstood the lifecycle of a Maven build.

maven-extension/src/main/java/io/opentelemetry/maven/OtelExecutionListener.java
cyrille-leclerc: Comment added, I use `rootSpan.makeCurrent()` so it is defined as the parent span of the `projectSpan` that is created just after.

maven-extension/src/main/java/io/opentelemetry/maven/OtelExecutionListener.java
cyrille-leclerc: @anuraaga as the class is now `final`, is there is reason to hardcode the classname here?

maven-extension/src/main/java/io/opentelemetry/maven/OtelExecutionListener.java
cyrille-leclerc: I didn't identify any constraint on the ordering of dependencies in the pom.xml. The order in which Maven loads `LifecycleParticipant`s is not specified, if this Otel lifecycle participant doesn't wrap all the extensions, then the measurement mays be slightly less accurate but I don't expect this to be an issue as it's uncommon in the Maven ecosystem to have extensions hooking in the lifecycle.

maven-extension/src/main/java/io/opentelemetry/maven/OtelExecutionListener.java
cyrille-leclerc: Please ignore, I got your point, I'm fixing this.

maven-extension/src/main/java/io/opentelemetry/maven/OtelExecutionListener.java
cyrille-leclerc: done above ^^

cyrille-leclerc:
I have squashed all the comments for better git history. I did my best to answer to every feedback, I hope I didn't miss any.

maven-extension/src/main/java/io/opentelemetry/maven/SpanRegistry.java
kenfinnigan: ok thanks

maven-extension/src/main/java/io/opentelemetry/maven/StringUtils.java
kenfinnigan: Possibly, unless there's a reason to use the Commons Lang3 version I'm not aware of

cyrille-leclerc: Fixed

maven-extension/build.gradle.kts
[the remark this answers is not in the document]
cyrille-leclerc: done

maven-extension/src/main/java/io/opentelemetry/maven/MyOtlpGrpcSpanExporter.java
```diff
@@ -0,0 +1,95 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Wrapper of {@link OtlpGrpcSpanExporter} to temporarily fix <a
+ * href="https://github.com/cyrille-leclerc/opentelemetry-maven-extension/issues/1">NoClassDefFoundError
+ * on GRPC classes after the Maven build is finished #1</a> thanks to a workaround of <a
+ * href="https://github.com/open-telemetry/opentelemetry-java/issues/3521">OtlpGrpcExporter/Netty
+ * still active after SdkTracerProvider#shutdown() #3521</a>
+ */
```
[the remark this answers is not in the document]
cyrille-leclerc: done

maven-extension/src/main/java/io/opentelemetry/maven/ChainedExecutionListener.java
```diff
@@ -0,0 +1,150 @@
...
+  }
+
+  @Override
+  public void sessionStarted(ExecutionEvent event) {
+    for (ExecutionListener listener : this.listeners) {
+      try {
+        listener.sessionStarted(event);
+      } catch (RuntimeException e) {
+        logger.error("Silently skip exception", e);
```
anuraaga, line 41:
```suggestion
        logger.error("Error starting session", e);
```

It's clearly not silent if we're logging :P

maven-extension/src/main/java/io/opentelemetry/maven/OpenTelemetrySdkService.java
```diff
@@ -0,0 +1,192 @@
...
+ * SDK Autoconfigure</a> that can't be used due to class loading issues when declaring the Maven
+ * OpenTelemetry extension using the pom.xml {@code <extension>} declaration: the OpenTelemetry SDK
+ * Autoconfigure registers a JVM shutdown hook that is incompatible with the fact that Maven
+ * extensions are unloaded before the JVM shuts down, causing {@link NoClassDefFoundError} when the
+ * shutdown hook is invoked.
+ */
+@Component(role = OpenTelemetrySdkService.class, hint = "opentelemetry-service")
+public final class OpenTelemetrySdkService implements Initializable, Disposable {
+
+  private final Logger logger = LoggerFactory.getLogger(getClass());
```
anuraaga, line 50:
```suggestion
  private static final Logger logger = LoggerFactory.getLogger(OpenTelemetrySdkService.class);
```

We always use static for loggers, and hence we never use getClass

maven-extension/src/main/java/io/opentelemetry/maven/OpenTelemetrySdkService.java
```diff
@@ -0,0 +1,192 @@
...
+  private Tracer tracer;
+
+  private SpanExporter spanExporter;
+
+  @Override
+  public synchronized void dispose() {
+    logger.debug("OpenTelemetry: dispose OpenTelemetrySdkService...");
+    if (this.openTelemetrySdk != null) {
+      logger.debug("OpenTelemetry: Shutdown SDK Trace Provider...");
+      long before = System.currentTimeMillis();
```
anuraaga, line 65: Always use `System.nanoTime()` to compute durations (though not sure how important this duration tracking for the log message actually is consider removing it)

maven-extension/src/main/java/io/opentelemetry/maven/OpenTelemetrySdkService.java
```diff
@@ -0,0 +1,192 @@
...
+      this.openTelemetrySdk = null;
+    }
+    logger.debug("OpenTelemetry: OpenTelemetrySdkService disposed");
+  }
+
+  /** TODO add support for `OTEL_EXPORTER_OTLP_CERTIFICATE` */
+  @Override
+  public void initialize() throws InitializationException {
+    logger.debug("OpenTelemetry: initialize OpenTelemetrySdkService...");
+    // OTEL_EXPORTER_OTLP_ENDPOINT
```
anuraaga, line 94: Is there a reason not to use the autoconfigure module by the way?

maven-extension/src/main/java/io/opentelemetry/maven/OtelExecutionListener.java
anuraaga: Ah you should just use `SpanBuilder.setParent`, no need to make current just to set the parent.

maven-extension/src/main/java/io/opentelemetry/maven/SpanRegistry.java
```diff
@@ -0,0 +1,183 @@
...
+    return span;
+  }
+
+  @AutoValue
+  abstract static class MavenProjectKey {
+    abstract String groupId();
+
+    abstract String artifactId();
+
+    public static Builder builder() {
```
anuraaga, line 109: It looks like my example may have caused confusion - the usages have the `from` pattern so I think the Builder is unnecessary. You can remove the Builder class, which will allow calling the autovalue constructor directly instead

```java
public static MavenProjectKey.fromMavenProject(MavenProject mavenProject)) {
  return new AutoValue_SpanRegistry_MavenProjectKey(mavenProject.getGroupId(), mavenProject.getArtifactId());
}
```

maven-extension/src/main/java/io/opentelemetry/maven/SpanRegistry.java
```diff
@@ -0,0 +1,183 @@
...
+    abstract String pluginGroupId();
+
+    abstract String pluginArtifactId();
+
+    public static Builder builder() {
+      return new AutoValue_SpanRegistry_MojoExecutionKey.Builder();
+    }
+
+    @AutoValue.Builder
+    public abstract static class Builder {
```
anuraaga, line 148: Ditto

maven-extension/README.md
```diff
@@ -0,0 +1,89 @@
...
+```xml
+<project>
+  ...
+  <build>
+    <extensions>
+      <extension>
+          <groupId>io.opentelemetry.contrib</groupId>
+          <artifactId>opentelemetry-maven-extension</artifactId>
+          <version>1.6.0-SNAPSHOT</version>
```
anuraaga, line 39: By the way we'll be releasing 1.6.0 this week probably before this PR is merged. I suspect there is no rush to get this into the release but just want to confirm :)

maven-extension/src/main/java/io/opentelemetry/maven/OpenTelemetrySdkService.java
cyrille-leclerc: I explained the reason in the javadoc of the OpenTelemetrySdkService class, it due to the fact that the Autoconfigure uses a JVM shutdown hook to close the Trace Provider which is incompatible with the lifecycle of a Maven `<extension>` that is unloaded from the classloader way before the JVM shuts down, causing a `NoClassDefFoundError` in the shutdown hook

See 
https://github.com/open-telemetry/opentelemetry-java/blob/v1.6.0/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/TracerProviderConfiguration.java#L58
```java
Runtime.getRuntime().addShutdownHook(new Thread(tracerProvider::close));
```

maven-extension/README.md
cyrille-leclerc: I can't say there is a rush but 
* I'll be happy to see this capability released
* We want to archive as soon as possible https://github.com/elastic/opentelemetry-maven-extension

cyrille-leclerc:
@anuraaga thanks for your review, I have answered all your points, the code is ready for review.
Note that I forced pushed an amended commit for better git history.

trask:
> Note that I forced pushed an amended commit for better git history.

we use GitHub's "Squash and merge" across all the OTel repos, so no need for nice git history in PRs 👍

trask:
🥳

maven-extension/src/main/java/io/opentelemetry/maven/OpenTelemetrySdkService.java
trask: I opened an issue to track this #82

PR #80, opened 2021-09-13
Update to OTel 1.6

anuraaga:
Hate Groovy (/end monthly rant)

### DIFF
--- a/jmx-metrics/src/test/groovy/io/opentelemetry/contrib/jmxmetrics/InstrumentHelperTest.groovy
+++ b/jmx-metrics/src/test/groovy/io/opentelemetry/contrib/jmxmetrics/InstrumentHelperTest.groovy
@@ -7,9 +7,9 @@ package io.opentelemetry.contrib.jmxmetrics
 import static io.opentelemetry.api.common.AttributeKey.stringKey
 import static io.opentelemetry.sdk.metrics.data.MetricDataType.DOUBLE_GAUGE
 import static io.opentelemetry.sdk.metrics.data.MetricDataType.DOUBLE_SUM
+import static io.opentelemetry.sdk.metrics.data.MetricDataType.HISTOGRAM
 import static io.opentelemetry.sdk.metrics.data.MetricDataType.LONG_GAUGE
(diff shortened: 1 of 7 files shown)


jmx-metrics/src/test/groovy/io/opentelemetry/contrib/jmxmetrics/InstrumentHelperTest.groovy
```diff
@@ -140,13 +140,10 @@ class InstrumentHelperTest extends Specification {
             metric.data.points.eachWithIndex { point, i ->
                 assert point.attributes == Attributes.of(stringKey("labelOne"), "labelOneValue", stringKey("labelTwo"), "${i}".toString())
 
-                if (metricType == SUMMARY) {
+                if (metricType == HISTOGRAM) {
                     assert point.count == 1
                     assert point.sum == value
-                    assert point.percentileValues[0].percentile == 0
```
anuraaga, line 146: Buckets are set up by the SDK so doesn't seem much point in asserting them. In general, I've loosened some assertions but they could probably be loosened even further to reduce coupling to SDK implementation details

jmx-metrics/src/test/groovy/io/opentelemetry/contrib/jmxmetrics/PrometheusIntegrationTests.groovy
```diff
@@ -41,19 +41,19 @@ class PrometheusIntegrationTests extends IntegrationTest {
         def scraped = receivedMetrics()
 
         then: 'they are of the expected format'
-        scraped.size() == 6
+        scraped.size() == 19
         scraped[0].contains(
                 '# HELP cassandra_storage_load Size, in bytes, of the on disk data size this node manages')
         scraped[1].contains(
-                '# TYPE cassandra_storage_load summary')
+                '# TYPE cassandra_storage_load histogram')
```
anuraaga, line 48: IIUC, the integration tests seem to just verify the link between SDK and backend without anything particularly JMX specific (we've already verified the mapping to SDK data in unit tests) - we have integration tests in the SDK for this, we could probably remove them from this artifact.

mateuszrzeszutek (approved):

Issue #93, opened 2021-09-24
[maven extension] Exception if OTLP Exporter not configured properly


kenfinnigan:
**Description**
`NullPointerException` if OTLP Exporter endpoint is not set.

```
[ERROR] Internal error: java.lang.NullPointerException -> [Help 1]
org.apache.maven.InternalErrorException: Internal error: java.lang.NullPointerException
    at org.apache.maven.DefaultMaven.execute (DefaultMaven.java:120)
    at org.apache.maven.cli.MavenCli.execute (MavenCli.java:957)
    at org.apache.maven.cli.MavenCli.doMain (MavenCli.java:289)
    at org.apache.maven.cli.MavenCli.main (MavenCli.java:193)
    at jdk.internal.reflect.NativeMethodAccessorImpl.invoke0 (Native Method)
    at jdk.internal.reflect.NativeMethodAccessorImpl.invoke (NativeMethodAccessorImpl.java:62)
    at jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke (DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke (Method.java:566)
    at org.codehaus.plexus.classworlds.launcher.Launcher.launchEnhanced (Launcher.java:282)
    at org.codehaus.plexus.classworlds.launcher.Launcher.launch (Launcher.java:225)
    at org.codehaus.plexus.classworlds.launcher.Launcher.mainWithExitCode (Launcher.java:406)
    at org.codehaus.plexus.classworlds.launcher.Launcher.main (Launcher.java:347)
Caused by: java.lang.NullPointerException
    at io.opentelemetry.maven.OpenTelemetrySdkService.getPropagators (OpenTelemetrySdkService.java:188)
    at io.opentelemetry.maven.OtelExecutionListener.sessionStarted (OtelExecutionListener.java:80)
    at io.opentelemetry.maven.ChainedExecutionListener.sessionStarted (ChainedExecutionListener.java:38)
    at org.apache.maven.lifecycle.internal.DefaultExecutionEventCatapult.fire (DefaultExecutionEventCatapult.java:61)
    at org.apache.maven.lifecycle.internal.DefaultExecutionEventCatapult.fire (DefaultExecutionEventCatapult.java:42)
    at org.apache.maven.lifecycle.internal.LifecycleStarter.execute (LifecycleStarter.java:74)
[...]

PR #94, opened 2021-09-24
Modify maven-extension to retrieve propagators from GlobalOpenTelemetry

kenfinnigan:
**Description:**

Change the `getPropagators()` helper to use `GlobalOpenTelemetry.getPropagators()` to prevent an NPE if there is no endpoint set.

**Existing Issue(s):**

#93 

**Testing:**

Performed manual verification with a local Maven build

**Documentation:**

N/A

**Outstanding items:**

N/A

### DIFF
--- a/maven-extension/src/main/java/io/opentelemetry/maven/OpenTelemetrySdkService.java
+++ b/maven-extension/src/main/java/io/opentelemetry/maven/OpenTelemetrySdkService.java
@@ -4,7 +4,6 @@
  */
 package io.opentelemetry.maven;
 
-import io.opentelemetry.api.GlobalOpenTelemetry;
 import io.opentelemetry.api.OpenTelemetry;
 import io.opentelemetry.api.common.Attributes;
 import io.opentelemetry.api.common.AttributesBuilder;
@@ -55,6 +54,7 @@ public final class OpenTelemetrySdkService implements Initializable, Disposable 
 
   @Requirement private RuntimeInformation runtimeInformation;
 
+  private OpenTelemetry openTelemetry;
   private OpenTelemetrySdk openTelemetrySdk;
 
   private Tracer tracer;
@@ -77,10 +77,10 @@ public final class OpenTelemetrySdkService implements Initializable, Disposable 
                 + sdkProviderShutdown.isDone()
                 + ")");
       }
-      // reset GlobalOpenTelemetry as we have just shutdown its TraceProvider
-      GlobalOpenTelemetry.resetForTest();
       this.openTelemetrySdk = null;
     }
+    this.openTelemetry = null;
+
     logger.debug("OpenTelemetry: OpenTelemetrySdkService disposed");
   }
 
@@ -95,7 +95,7 @@ public final class OpenTelemetrySdkService implements Initializable, Disposable 
     if (StringUtils.isBlank(otlpEndpoint)) {
       logger.debug(
           "OpenTelemetry: No -Dotel.exporter.otlp.endpoint property or OTEL_EXPORTER_OTLP_ENDPOINT environment variable found, use a NOOP tracer");
-      GlobalOpenTelemetry.set(OpenTelemetry.noop());
+      this.openTelemetry = OpenTelemetry.noop();
     } else {
       // OtlpGrpcSpanExporterBuilder spanExporterBuilder = OtlpGrpcSpanExporter.builder();
       OtlpGrpcSpanExporterBuilder spanExporterBuilder = OtlpGrpcSpanExporter.builder();
@@ -155,9 +155,10 @@ public final class OpenTelemetrySdkService implements Initializable, Disposable 
           OpenTelemetrySdk.builder()
               .setTracerProvider(sdkTracerProvider)
               .setPropagators(ContextPropagators.create(W3CTraceContextPropagator.getInstance()))
-              .buildAndRegisterGlobal();
+              .build();
+      this.openTelemetry = this.openTelemetrySdk;
     }
-    this.tracer = GlobalOpenTelemetry.getTracer("io.opentelemetry.contrib.maven");
+    this.tracer = this.openTelemetry.getTracer("io.opentelemetry.contrib.maven");
   }
 
   public Tracer getTracer() {
@@ -184,6 +185,6 @@ public final class OpenTelemetrySdkService implements Initializable, Disposable 
 
   /** Returns the {@link ContextPropagators} for this {@link OpenTelemetry}. */
   public ContextPropagators getPropagators() {
-    return openTelemetrySdk.getPropagators();
+    return openTelemetry.getPropagators();
   }
 }


cyrille-leclerc (approved):

cyrille-leclerc:
Thanks @kenfinnigan , I don't understand how I didn't catch this weakness after I changed the pattern to retrieve the propagator.

maven-extension/src/main/java/io/opentelemetry/maven/OpenTelemetrySdkService.java
```diff
@@ -184,6 +184,6 @@ protected Resource getMavenResource() {
 
   /** Returns the {@link ContextPropagators} for this {@link OpenTelemetry}. */
   public ContextPropagators getPropagators() {
-    return openTelemetrySdk.getPropagators();
+    return GlobalOpenTelemetry.getPropagators();
```
jkwatson, line 187: why is this better than using a configured instance? What is setting the global?

jkwatson: It seems like relying on the global instance, and the initialization thereof would be a bad idea in almost every case.

kenfinnigan: As the same instance is set globally, it seemed preferable to use that approach.

The problem I found occurs when `openTelemetrySdk` is null because the endpoint was not specified on the command line.

Happy to resolve with an alternative approach, such as a null check and returning the noop propagators

jkwatson: you could pre-initialize it to the no-op implementation : `OpenTelemetry.noop()`

kenfinnigan: ah, will try that. thanks

kenfinnigan: I've updated it with that approach

maven-extension/src/main/java/io/opentelemetry/maven/OpenTelemetrySdkService.java
```diff
@@ -95,7 +95,8 @@ public void initialize() throws InitializationException {
     if (StringUtils.isBlank(otlpEndpoint)) {
       logger.debug(
           "OpenTelemetry: No -Dotel.exporter.otlp.endpoint property or OTEL_EXPORTER_OTLP_ENDPOINT environment variable found, use a NOOP tracer");
-      GlobalOpenTelemetry.set(OpenTelemetry.noop());
+      this.openTelemetrySdk = (OpenTelemetrySdk) OpenTelemetry.noop();
```
jkwatson, line 98: that cast will fail. :( `noop()` is an instance of `DefaultOpenTelemetry`, not `OpenTelemetrySdk`. 

In general, I recommend keeping one instance variable of the API (of type `OpenTelemetry`) for doing actual instrumentation, and a 2nd instance of the SDK (which will be set the API instance when initialized) which can be used for SDK-concerns, like shutdown handling, etc. Keeping the concerns separate makes it clear what's to be used for instrumentation, and what is an operational concern.

kenfinnigan: Thanks @jkwatson, not sure what I was thinking there.

Revised PR coming shortly

kenfinnigan:
I have no idea why jmx_metrics integration test failed.

Anyone have ideas?

trask:
looks like a sporadic failure, I opened #99 to track and hit re-run

kenfinnigan:
Thanks @trask

maven-extension/src/main/java/io/opentelemetry/maven/OpenTelemetrySdkService.java
```diff
@@ -77,10 +78,12 @@ public synchronized void dispose() {
                 + sdkProviderShutdown.isDone()
                 + ")");
       }
-      // reset GlobalOpenTelemetry as we have just shutdown its TraceProvider
-      GlobalOpenTelemetry.resetForTest();
       this.openTelemetrySdk = null;
     }
+    // reset GlobalOpenTelemetry as we have just shutdown its TraceProvider
+    GlobalOpenTelemetry.resetForTest();
```
jkwatson, line 84: I would also add a comment about this being a hack, since this is definitely not intended for anything exception unit testing purposes. It's one of the reasons that I strongly recommend avoiding using the Global instance if you possibly can.

anuraaga: Good callout - I guess everything should still work even if we remove `GlobalOpenTelemetry.set` and use `build()` instead of `buildAndRegisterGlobal` below?

cyrille-leclerc: I was interested in enabling `GlobalOpenTelemetry` as the static helper methods are commonly used when using Otel in Java and I was influenced by the fact that the AutoConfigure module was relying on `buildAndRegisterGlobal`.

I'm happy to move away from `GlobalOpenTelemetry`, it should be straightforward as the usage of the Otel APIs in Maven are limited, for the moment, to the code of the OpenTelemetry Maven Extension. Maybe we will have to evolve when we want to use the Otel APIs in other Maven mojos.

kenfinnigan: Is the preference to remove use of the Global OTeL instance completely and just rely on the local one?

cyrille-leclerc: @kenfinnigan in fact, we are not using GlobalOpenTelemetry at the moment, the primary reason being that I was interested in standardising the name of the tracer so the [OtelExecutionListener](https://github.com/open-telemetry/opentelemetry-java-contrib/blob/v1.6.0/maven-extension/src/main/java/io/opentelemetry/maven/OtelExecutionListener.java) just relies on [OpenTelemetrySdkService#getTracer()](https://github.com/open-telemetry/opentelemetry-java-contrib/blob/v1.6.0/maven-extension/src/main/java/io/opentelemetry/maven/OpenTelemetrySdkService.java).

I'm confident that removing the `registerGlobal` part will have no impact on the existing code.

Does this clarify?

kenfinnigan: Sure, will make the change

kenfinnigan: @cyrille-leclerc I've pushed a change to remove the use of `GlobalOpenTelemetry`

jkwatson (approved):

cyrille-leclerc (approved):
I have successfully tested the removal of `GlobalOpenTelemetry` on a test and I have review the code that looks intuitive to me.

trask (approved):
thx everyone 👍

PR #95, opened 2021-09-24
Allowing programmatic attach of Javaagent

iNikem:
This is a first stab on https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/691. I would like to merge it and then give to end-users for testing.


runtime-attach/src/main/java/io/opentelemetry/contrib/attach/AgentFileLocator.java
```diff
@@ -0,0 +1,35 @@
...
+  static File locateAgentFile() {
+    CodeSource codeSource = OpenTelemetryAgent.class.getProtectionDomain().getCodeSource();
+
+    if (codeSource == null) {
+      throw new IllegalStateException("could not get agent jar location");
+    }
+
+    URL codeSourceLocation = codeSource.getLocation();
+    try {
+      File javaagentFile = new File(codeSourceLocation.toURI());
```
anuraaga, line 24: Is there anything to worry about for a fat jar?

iNikem: I haven't try, but I believe `io.opentelemetry.javaagent.OpenTelemetryAgent#verifyJarManifestMainClassIsThis` will "take care" of it. Meaning that fat jar will not be supported for now.

anuraaga (approved):

trask (approved):

PR #102, opened 2021-09-29
Move in disruptor processor from SDK repo.

anuraaga:
Same as in SDK repo except

- Renamed artifact to `disruptor-processor`, if we add a logs one I think it should go in the same artifact
- Removed `Async` wording - it seemed noisy to me, BSP is similarly async as far as the app is concerned so that's probably not what matters here


disruptor-processor/build.gradle.kts
```diff
@@ -0,0 +1,14 @@
...
+  id("otel.publish-conventions")
+}
+
+description = "OpenTelemetry Disruptor Processor"
+
+dependencies {
+  api("io.opentelemetry:opentelemetry-api")
+  api("io.opentelemetry:opentelemetry-sdk")
+
+  implementation("com.google.guava:guava")
```
mateuszrzeszutek, line 12: Is guava really used here?

disruptor-processor/src/main/java/io/opentelemetry/contrib/disruptor/trace/DisruptorEventQueue.java
```diff
@@ -0,0 +1,231 @@
...
+  private final boolean blocking;
+
+  private enum EventType {
+    ON_START,
+    ON_END,
+    ON_SHUTDOWN,
+    ON_FORCE_FLUSH
+  }
+
+  // Creates a new EventQueue. Private to prevent creation of non-singleton instance.
```
mateuszrzeszutek, line 56: The whole class is package-private, this comment could probably be removed.

mateuszrzeszutek (approved):

disruptor-processor/build.gradle.kts
anuraaga: Nope thanks

trask (approved):

PR #103, opened 2021-09-29
Remove Prometheus IT for jmx metrics.

anuraaga:
I originally had an idea of removing all the integration tests for jmx metrics thinking they overlap with the SDK's exporter integration tests. I realize now that the jmx-metrics module itself has logic specific to targets like cassandra and kafka, so we do need these target-specific integration tests. But I guess we can remove a bit with this as the conversion of OTel to Prometheus seems far-removed form this module.


mateuszrzeszutek (approved):

rmfitzpatrick:
I agree the intended components to test here are now provided by the autoconfigure extension so this test is a bit superfluous.  You can also delete https://github.com/open-telemetry/opentelemetry-java-contrib/blob/main/jmx-metrics/src/test/resources/prometheus_config.properties

anuraaga:
Thanks for the pointer!

rmfitzpatrick (approved):

PR #105, opened 2021-09-30
Migrate most jmx metrics unit tests to Java.

anuraaga:
Mostly to avoid extreme maintenance pain when trying to update the many not-statically-compilable / IDE supportable lines. One last test actually needs to execute some Groovy code. Will take another look if it's worth calling into GroovyEngine from Java or not but still a good improvement for me.

image


jmx-metrics/src/test/java/io/opentelemetry/contrib/jmxmetrics/OtelHelperSynchronousMetricTest.java
```diff
@@ -0,0 +1,466 @@
+/*
+ * Copyright The OpenTelemetry Authors
```
anuraaga, line 2: Figured it's ok not to copy in the memoization tests here since we don't do anything special to memoize sync measurements

rmfitzpatrick (approved):

PR #106, opened 2021-10-01
Migrate JMX metrics last test to Java

anuraaga:


rmfitzpatrick:
Overall the changes look great, but the test names in the gradle provided results don't seem to work well with the nested tests and their inner results are merged with the other suites.  Maybe using `@DisplayName` would help?

anuraaga:
@rmfitzpatrick Thanks a lot for finding that! It seems to be a long standing feature request https://github.com/gradle/gradle/issues/15075, JUnit5 has been out for a long time for this to not be natively rendered... Used DisplayName to make it look a bit better

rmfitzpatrick (approved):

Issue #97, opened 2021-09-24
[maven-extension] Disable Mojo spans


kenfinnigan:
**Is your feature request related to a problem? Please describe.**
In playing with the extension today on a build of Quarkus, it generates 17400 spans as each Mojo/plugin in a module build has its own span.

Though this information is beneficial in diagnosing a problem it does add a lot more information than is probably necessary for regular project builds. It might also be interesting to provide a list of mojo names that should have spans and ignore the rest, so known plugins that are troublesome can always have spans created?

**Describe the solution you'd like**
Ability to turn off span creation at the mojo level. Which option should be the default is possibly another discussion.

**Describe alternatives you've considered**
It would be possible to provide a sampling strategy to exclude them, but I think it's better to prevent their creation in the first place if that's what is desired.

**Additional context**
Happy to work on a fix for this if there's interest and agreement

cyrille-leclerc:
@v1v @kuisathaverat it reminds me the problem we faced on Jenkins observability. 17k span is one order of magnitude higher than the problem we faced in CI pipeline traces :-)
I like the idea to be incremental and first enable disabling bluntly the mojo spans. Later, we will be able to refine the feature to support smarter filtering if it makes sense.
Note that this telles me that the challenge will come when we add unit test visibility. One span per unit test?

kenfinnigan:
Ok, I will start playing around with this when I have a moment.

Whether we need a span per unit test would depend on the purpose of the spans.

If it's purely as a view on how long a build is taking, does it need that really fine-grained level of a particular test? Or only all the tests and then recognize if a set of tests takes too long, to evaluate the logs for how long each test actually took to find the culprit?

PR #108, opened 2021-10-08
[maven-extension] Option to disable the creation of spans per mojo goal execution

cyrille-leclerc:
**Description:**

Option to disable the creation of spans per mojo goal execution

**Existing Issue(s):**

- https://github.com/open-telemetry/opentelemetry-java-contrib/issues/97

**Testing:**

Unfortunately I don't know yet how to add integration tests for Maven extensions. To manually test

```
export OTEL_MAVEN_INSTRUMENTATION_MOJOS_ENABLED=true
mvn verify

# verify spans for modules and spans where created

export OTEL_MAVEN_INSTRUMENTATION_MOJOS_ENABLED=false
mvn verify

# verify spans spans where NOT created, only spans for modules where created
```

**Documentation:**

Documented the new configuration parameter `OTEL_MAVEN_INSTRUMENTATION_MOJOS_ENABLED=true/false` / `-Dotel.maven.instrumentation.mojos.enabled=true/false`

**Outstanding items:**

Adding integration tests. I plan to engage with the Maven community to get guidance on this.

### DIFF
--- a/maven-extension/README.md
+++ b/maven-extension/README.md
@@ -63,7 +63,7 @@ The Maven OpenTelemetry Extension supports a subset of the [OpenTelemetry auto c
 | otel.exporter.otlp.headers   | OTEL_EXPORTER_OTLP_HEADERS  | Key-value pairs separated by commas to pass as request headers on OTLP trace and metrics requests.        |
 | otel.exporter.otlp.timeout   | OTEL_EXPORTER_OTLP_TIMEOUT  | The maximum waiting time, in milliseconds, allowed to send each OTLP trace and metric batch. Default is `10000`.  |
 | otel.resource.attributes | OTEL_RESOURCE_ATTRIBUTES | Specify resource attributes in the following format: key1=val1,key2=val2,key3=val3 |
-| otel.instrumentation.maven-mojos.enabled | OTEL_INSTRUMENTATION_MAVEN_MOJOS_ENABLED | Whether to create spans for mojo goal executions, `true` or `false`. Can be configured to reduce the number of spans created for large builds. Default value: `true` |
+| otel.instrumentation.maven.mojo.enabled | OTEL_INSTRUMENTATION_MAVEN_MOJO_ENABLED | Whether to create spans for mojo goal executions, `true` or `false`. Can be configured to reduce the number of spans created for large builds. Default value: `true` |
 
 
 ℹ️ The `service.name` is set by default to `maven`, it can be overwritten specifying resource attributes.
--- a/maven-extension/src/main/java/io/opentelemetry/maven/OpenTelemetrySdkService.java
+++ b/maven-extension/src/main/java/io/opentelemetry/maven/OpenTelemetrySdkService.java
@@ -164,8 +164,8 @@ public final class OpenTelemetrySdkService implements Initializable, Disposable 
 
     String mojosInstrumentationEnabledAsString =
         System.getProperty(
-            "otel.instrumentation.maven-mojos.enabled",
-            System.getenv("OTEL_INSTRUMENTATION_MAVEN_MOJOS_ENABLED"));
+            "otel.instrumentation.maven.mojo.enabled",
+            System.getenv("OTEL_INSTRUMENTATION_MAVEN_MOJO_ENABLED"));
     this.mojosInstrumentationEnabled =
         Boolean.parseBoolean(
             StringUtils.defaultIfBlank(mojosInstrumentationEnabledAsString, "true"));


cyrille-leclerc:
@kenfinnigan please look at this proposal to solve the problem you raised to disable the creation of spans per mojo goal execution for large builds.

I introduced the configuration parameter `OTEL_MAVEN_INSTRUMENTATION_MOJOS_ENABLED=true/false` / `-Dotel.maven.instrumentation.mojos.enabled=true/false`.
The naming convention is aligned on the configuration params to disable instrumentations on the Java agent: 

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/docs/suppressing-instrumentation.md

> You can suppress agent instrumentation of specific libraries by using
`-Dotel.instrumentation.[name].enabled=false` (or using the equivalent environment
variable `OTEL_INSTRUMENTATION_[NAME]_ENABLED`) where `name` (`NAME`) is the
corresponding instrumentation `name`

maven-extension/README.md
```diff
@@ -63,6 +63,7 @@ The Maven OpenTelemetry Extension supports a subset of the [OpenTelemetry auto c
 | otel.exporter.otlp.headers   | OTEL_EXPORTER_OTLP_HEADERS  | Key-value pairs separated by commas to pass as request headers on OTLP trace and metrics requests.        |
 | otel.exporter.otlp.timeout   | OTEL_EXPORTER_OTLP_TIMEOUT  | The maximum waiting time, in milliseconds, allowed to send each OTLP trace and metric batch. Default is `10000`.  |
 | otel.resource.attributes | OTEL_RESOURCE_ATTRIBUTES | Specify resource attributes in the following format: key1=val1,key2=val2,key3=val3 |
+| otel.maven.instrumentation.mojos.enabled | OTEL_MAVEN_INSTRUMENTATION_MOJOS_ENABLED | Disable the creation of spans for each mojo goal execution, `true` or `false`. Convenient to reduce the number of spans for large builds. Default value: `true` |
```
kenfinnigan, line 66: The description confused me as it implies disabling, but based on the code I think it's meant to be enabling.

Is that correct?

cyrille-leclerc: I think that the proposed code keeps mojo spans enabled by default (see [L171](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/108/files#diff-8625ba4b8e68496923fa886962656dc739e75194535a5d9a9c7a81e555fff294R171) and code snippet below).

I'm wondering if the naming convention of to disable instrumentations on the Otel Java Agent, `OTEL_INSTRUMENTATION_[NAME]_ENABLED` , is intuitive enough for our use case on the Otel Maven extension. 

**❓  Would it be more intuitive to suffix the config param by `_DISABLED` (ie `OTEL_MAVEN_INSTRUMENTATION_MOJOS_DISABLED`) than to follow the `_ENABLED` convention of the Otel Java Agent?**




```
    this.mojosInstrumentationEnabled =
        Boolean.parseBoolean(
            StringUtils.defaultIfBlank(mojosInstrumentationEnabledAsString, "true"));
```

anuraaga: Following the convention of the instrumentation repo, for better or worse, is probably a good idea. I guess @kenfinnigan's pointing to the comment starting with `Disable the creation` which makes it seem like this flags disables when it's true. I think this would read fine

"Whether to create spans for each mojo goal execution"

cyrille-leclerc: Thanks @anuraaga , I have improved the documentation based on your recommendation.

maven-extension/README.md
```diff
@@ -63,9 +63,10 @@ The Maven OpenTelemetry Extension supports a subset of the [OpenTelemetry auto c
 | otel.exporter.otlp.headers   | OTEL_EXPORTER_OTLP_HEADERS  | Key-value pairs separated by commas to pass as request headers on OTLP trace and metrics requests.        |
 | otel.exporter.otlp.timeout   | OTEL_EXPORTER_OTLP_TIMEOUT  | The maximum waiting time, in milliseconds, allowed to send each OTLP trace and metric batch. Default is `10000`.  |
 | otel.resource.attributes | OTEL_RESOURCE_ATTRIBUTES | Specify resource attributes in the following format: key1=val1,key2=val2,key3=val3 |
+| otel.maven.instrumentation.mojos.enabled | OTEL_MAVEN_INSTRUMENTATION_MOJOS_ENABLED | Whether to create spans for mojo goal executions, `true` or `false`. Can be configured to reduce the number of spans created for large builds. Default value: `true` |
```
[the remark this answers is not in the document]
kenfinnigan: Based on the ones I see here: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/docs/config/common.md

I would agree it should probably be `OTEL_INSTRUMENTATION_MAVEN_MOJOS_ENABLED`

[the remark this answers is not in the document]
kenfinnigan: Bit of a nitpick, but can it be `maven.mojos` instead of a hyphen separator?

If additional properties are added in the future it's good to have them all under `maven.*` and it wouldn't require a rename of this one to do that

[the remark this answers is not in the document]
kenfinnigan: I understand the rationale of the `-` over `.` based on the linked document, but I wonder if it applies given there isn't a separate library of instrumentation for Maven vs Maven Mojo. I consider it to be a Maven instrumentation and not a Maven Mojo one, meaning `maven` is the instrumentation and not `maven-mojo`.

I could easily be splitting hairs on this, so please ignore me if that's the case.

PR #111, opened 2021-10-11
Update Cassandra units for latency counters

JonathanWamsley:
**Description:**
Update Cassandra metrics latency counter units to better reflect what is being collected.

**Testing:**
Updated Cassandra Integration Test

**Documentation:**
Updated Cassandra Docs


jmx-metrics/src/main/resources/target-systems/cassandra.groovy
```diff
@@ -30,7 +30,7 @@ otel.instrument(clientRequestRangeSliceLatency,
 
 otel.instrument(clientRequestRangeSliceLatency,
         "cassandra.client.request.range_slice.latency.count",
-        "Total token range read request latency", "µs", "Count",
+        "Total token range read request latency", "1", "Count",
```
anuraaga, line 33: Is it a good time to update the description too? Seems to not be accurate - I'd expect "total latency" to be temporal since it's still latency

Issue #82, opened 2021-09-21
[maven-extension] Figure out how to leverage standard autoconfigure


trask:
See comment in the original maven-extension PR about the technical challenges faced with using standard autoconfigure:

* https://github.com/open-telemetry/opentelemetry-java-contrib/pull/79#discussion_r712689149

cyrille-leclerc:
I faced 2 distinct families of challenges:

* The Otel Auto Configurer assumes that Otel will be shutdown at the shutdown of the JVM which is not true when Otel is managed as an extension or a plugin of the system it instruments. The assumption of the Otel Auto Configurer is materialized by JVM shutdown hooks:
   * https://github.com/open-telemetry/opentelemetry-java/blob/v1.6.0/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/TracerProviderConfiguration.java#L58
```java
Runtime.getRuntime().addShutdownHook(new Thread(tracerProvider::close)); 
```
   * https://github.com/open-telemetry/opentelemetry-java/blob/v1.6.0/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/MetricExporterConfiguration.java#L151
```java
```


* registers JVM shutdown hooks that are incompatible

cyrille-leclerc:
FYI we are migrating the Jenkins OpenTelemetry Plugin to the Otel SDK Auto. Configuration Extension.
It's working well so far. Some lessons learned.

* The pattern to do integration tests with autoconfigured in memory span and metrics exporters is not easy
* We need for the configuration to combine parameters from
   * Sensible defaults (e.g. service.name= "jenkins" or "maven")
   * Standard Auto Configuration System Properties and Environment variables
   * Configuration settings passed via the configuration mechanism of the integrated system (e.g. standard Jenkins plugin configuration mechanisms).

https://github.com/jenkinsci/opentelemetry-plugin/pull/185

I'll share more as we progress

PR #112, opened 2021-10-12
[maven-extension] Use the Otel SDK Auto Configuration Extension

cyrille-leclerc:
**Description:**

Use the Otel SDK Auto Configuration Extension instead of reinventing all the configuration boilerplate.

ℹ️  Note that this PR doesn't extend the supported exporter (e.g. doesn't add support for OTLP HTTP...), we only have OTLP GRPC for the moment, it should be easy to extend as all the logic is provided by the Otel SDK Auto Configuration Extension now.

**Existing Issue(s):**

- https://github.com/open-telemetry/opentelemetry-java-contrib/issues/82

**Testing:**

We don't have auto integration tests yet so tests have to be done manually.

```
mvn verify
```

```
export OTEL_RESOURCE_ATTRIBUTES="foo=bar"
mvn verify
```

```
export OTEL_RESOURCE_ATTRIBUTES="service.name=MyMavenService"
mvn verify
```

**Documentation:**

No documentation added because the user experience didn't change.

**Outstanding items:**

Both items below are tracked in the code as `TODO`.

* Create an enhancement request on the Otel SDK Auto Configuration Extension to not register the shut down hook during initialisation if the system that embeds the Otel SDK manages the lifecycle. It doesn't cause problems with the Maven Plexus lifecycle but there is a risk.
* Explore how to use `io.opentelemetry.sdk.autoconfigure.spi.ConfigProperties` to pass configuration parameters instead of shoving them in system properties.

### DIFF
--- a/jmx-metrics/docs/target-systems/cassandra.md
+++ b/jmx-metrics/docs/target-systems/cassandra.md
@@ -18,8 +18,8 @@ These metrics are sourced from Cassandra's exposed Dropwizard Metrics for each n
 * Instrument Type: DoubleValueObserver
 
 * Name: `cassandra.client.request.range_slice.latency.count`
-* Description: Total token range read request latency
-* Unit: `µs`
+* Description: Number of token range read request operations
+* Unit: `1`
 * Instrument Type: LongSumObserver
 
 * Name: `cassandra.client.request.range_slice.latency.max`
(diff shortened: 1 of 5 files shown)


maven-extension/src/main/java/io/opentelemetry/maven/OtelUtils.java
```diff
@@ -7,25 +7,32 @@
...
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
+import javax.annotation.CheckForNull;
+import javax.annotation.Nullable;
 
 final class OtelUtils {
 
+  public static String getComaSeparatedString(Map<String, String> keyValuePairs) {
```
anuraaga, line 20:
```suggestion
  public static String getCommaSeparatedString(Map<String, String> keyValuePairs) {
```

kenfinnigan (approved):

anuraaga:
Yeah ready to merge this after fixing the build in #114

maven-extension/src/main/java/io/opentelemetry/maven/OtelUtils.java
cyrille-leclerc: Fixed

PR #115, opened 2021-10-16
Add agent to bridge JFR Streaming into metrics 

kittylyst:
**Description:**

Adds a an agent that listens for JFR events (using the Streaming API) and converts them to OpenTelemetry metrics on a dedicated background thread.

**Testing:**

Currently just 1 unit test file, which uses `InMemoryMetricReader` - more to follow

**Documentation:**

Documentation contained in `jfr-streaming/README.md`

**Outstanding items:**

* More coverage of events
* Thread grouping for high cardinality JFR events (e.g. allocation)
* More test cases
* Integration tests (both TestContainers and with various vendor backends)


kittylyst:
The build task is failing, because the tests require an already-built agent in `jfr-streaming.jar`

What's the best way to fix this? Do a priming build that builds the jar via `-x test` and then re-run the build? Or something else?

jfr-streaming/build.gradle.kts
```diff
@@ -0,0 +1,62 @@
+val gsonVersion: String by project
+val slf4jVersion: String by project
+val otelVersion: String by project
+val otelAlphaVersion: String by project
+
+
+plugins {
+    id("com.github.johnrengelman.shadow") version "5.2.0"
```
anuraaga, line 8:
```suggestion
    id("com.github.johnrengelman.shadow")
```

Version already defined in settings.gradle

jfr-streaming/build.gradle.kts
```diff
@@ -0,0 +1,62 @@
...
+    id("java")
```
anuraaga, line 9:
```suggestion
    id("otel.java-conventions")
```

This applies our standard conventions - much of the boilerplate like repositories you'll be able to remove

jfr-streaming/build.gradle.kts
```diff
@@ -0,0 +1,62 @@
...
+}
+
+repositories {
+    mavenCentral()
+    mavenLocal()
+}
+
+java {
+    toolchain {
+        languageVersion.set(JavaLanguageVersion.of(17))
```
anuraaga, line 19: You'll still need this even with the conventions, but I'll work on upgrading our build's toolchain java to 17 (Groovy...)

Either way, add something like

https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk-extensions/jfr-events/build.gradle.kts#L15

to make sure it targets java 17

jfr-streaming/build.gradle.kts
```diff
@@ -0,0 +1,62 @@
...
+    }
+}
+
+tasks.test {
+  useJUnitPlatform()
+  jvmArgs("-Djdk.attach.allowAttachSelf=true")
+}
+
+dependencies {
+    implementation(platform("io.opentelemetry:opentelemetry-bom:${otelVersion}"))
```
anuraaga, line 29: The conventions automatically apply dependency management, it looks like you can remove all boms and versions from this file

jfr-streaming/build.gradle.kts
```diff
@@ -0,0 +1,62 @@
...
+    implementation("io.opentelemetry:opentelemetry-api")
+    implementation("io.opentelemetry:opentelemetry-sdk")
+    implementation("io.opentelemetry:opentelemetry-exporter-otlp")
+    implementation("io.opentelemetry:opentelemetry-sdk-metrics")
+
+    implementation("io.opentelemetry:opentelemetry-exporter-otlp-metrics:${otelAlphaVersion}")
+    implementation(platform("io.grpc:grpc-bom:1.34.1"))
+    implementation("io.grpc:grpc-netty-shaded")
+
+    testImplementation("org.junit.jupiter:junit-jupiter-api:5.7.1")
```
anuraaga, line 39: These non-otel test dependencies are automatically applied so can remove them

jfr-streaming/gradle.properties
```diff
@@ -0,0 +1,6 @@
+version=0.0.1
```
anuraaga, line 1: Remove these three files to let it slot into the normal build

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/Agent.java
```diff
@@ -0,0 +1,126 @@
...
+    private static Consumer<RecordedEventHandler> eventEnablerFor(RecordingStream recordingStream) {
+        return handler -> {
+            EventSettings eventSettings = recordingStream.enable(handler.getEventName());
+            handler.getPollingDuration().ifPresent(eventSettings::withPeriod);
+            handler.getThreshold().ifPresent(eventSettings::withThreshold);
+            recordingStream.onEvent(handler.getEventName(), handler);
+        };
+    }
+
+    static void configureOpenTelemetry() {
```
anuraaga, line 81: You should use `sdk-extension-autoconfigure` to allow configuring OTel with our standard variables instead of configuring manually (you can use a meter provider configurer to add the aggregators)

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/Agent.java
```diff
@@ -0,0 +1,126 @@
...
+
+    public static void premain(String agentArgs, Instrumentation inst) {
+      meterProvider = GlobalMeterProvider.get();
+      if (meterProvider == null) {
+        configureOpenTelemetry();
+      }
+      startJfrService(agentArgs, inst);
+    }
+
+    public static void startJfrService(String agentArgs, Instrumentation inst) {
```
anuraaga, line 56: Looks like `Instrumentation` isn't used - can we extract into a non-agent entrypoint, probably `JfrMetrics.enable()` (and ideally `JfrMetrics.disable()` if it's supportable) to allow use programatically? We can just call it from the agent entrypoint if the agent seems useful (and let's publish the shadow jar with `-agent` classifier).

But I'm a bit wary of publishing another javaagent - is it better to add this to the normal opentelemetry-javaagent instead of having yet another agent?

It might keep this PR simpler to remove the agent stuff for now and add it later, either here, or in the otel javaagent.

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/Constants.java
```diff
@@ -0,0 +1,10 @@
+package io.opentelemetry.contrib.jfr;
+
+public class Constants {
```
anuraaga, line 3: If what I said about exposing a programmatic entrypoint makes sense, then the API matters so let's only expose `public` what's needed (I'm guessing just the one entrypoint class)

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/Constants.java
```diff
@@ -0,0 +1,10 @@
...
+    private Constants() {}
+
+    public static final String THREAD_NAME = "thread.name";
```
anuraaga, line 6: Use an `AttributeKey` for the constant rather than creating one every time

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/container/ContainerConfigurationHandler.java
```diff
@@ -0,0 +1,42 @@
...
+
+    private final Meter otelMeter;
+    private volatile long value = 0L;
+
+
+    public ContainerConfigurationHandler(Meter otelMeter) {
+        this.otelMeter = otelMeter;
+    }
+
+    public ContainerConfigurationHandler init() {
```
anuraaga, line 22: We'd generally set up metrics, etc in the constructor I don't think we need init methods

jfr-streaming/build.gradle.kts
```diff
@@ -0,0 +1,62 @@
...
+        attributes(
+                "Premain-Class" to "io.opentelemetry.contrib.jfr.Agent",
+                "Agent-Class" to "io.opentelemetry.contrib.jfr.Agent",
+                "Implementation-Version" to project.version,
+                "Implementation-Vendor" to "Open Telemetry"
+        )
+    }
+}
+
+tasks.named("build") {
```
anuraaga, line 60:
```suggestion
tasks.named("assemble") {
```

jfr-streaming/build.gradle.kts
```diff
@@ -0,0 +1,62 @@
...
+    dependsOn("shadowJar")
```
anuraaga, line 61: My comment about considering adding the agent in a follow up might invalidate it, but otherwise we'd want `test` to depend on `shadowJar`, and add the agent to the test's jvm args

Something like this (with far fewer flags needed)

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/conventions/src/main/kotlin/io.opentelemetry.instrumentation.javaagent-testing.gradle.kts#L99

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/container/ContainerConfigurationHandler.java
```diff
@@ -0,0 +1,42 @@
...
+
+        @Override
+    public String getEventName() {
+        return EVENT_NAME;
+    }
+
+    @Override
+    public void accept(RecordedEvent ev) {
+        if (ev.hasField(EFFECTIVE_CPU_COUNT)) {
+            value = ev.getLong(EFFECTIVE_CPU_COUNT);
```
anuraaga, line 39: Just want to confirm, are these metrics not present in management beans? Is there anything conceptually different between metrics on jfr events vs management beans?

I ask mostly because if we end up memoizing to a field anyways, then there wouldn't be a real difference between using jfr metrics vs reading from a bean I guess, assuming the metric is actually available in a bean

settings.gradle.kts
```diff
@@ -44,3 +44,4 @@ include(":example")
 include(":jmx-metrics")
 include(":maven-extension")
 include(":runtime-attach")
+include(":jfr-streaming")
```
anuraaga, line 47: Alphabetical order

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/Agent.java
```diff
@@ -0,0 +1,126 @@
...
+
+        var exporterBuilder = OtlpGrpcMetricExporter.builder()
+                .setEndpoint(otlpUrl);
+
+        if (apiKey != null) {
+            exporterBuilder.addHeader(API_KEY_HEADER, apiKey);
+        }
+        var meterProviderBuilder = SdkMeterProvider.builder().setResource(resource);
+        // Set delta aggregation
+        setAggregatorFactory(meterProviderBuilder, COUNTER, Aggregation.sum(DELTA));
```
anuraaga, line 105: My understanding is generally users should be setting aggregations, not instrumentation. Do we need to set these?

/cc @jsuereth

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/container/ContainerConfigurationHandler.java
kittylyst: JFR events are not exactly the same as JMX values. They are basically better quality data, but I think a broader discussion of how to handle the multiplicity of data sources is required. Let's have that wider discussion after this initial agent is landed.

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/Agent.java
jsuereth: +1, was going to make the same comment.  As much as possible, we should rely on autoconfigure and push users to a common configuration component.

Again, if there were exposed as a library, you could assume OTel is initialized prior to wiring in your instrumentation (and accept a MeterProvider as your input).

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/Agent.java
```diff
@@ -0,0 +1,126 @@
...
+        setAggregatorFactory(meterProviderBuilder, UP_DOWN_COUNTER, Aggregation.sum(DELTA));
+        setAggregatorFactory(meterProviderBuilder, OBSERVABLE_SUM, Aggregation.histogram());
```
jsuereth, line 107: This one seems a bit odd to me.  Why are you converting observable sums to histograms?   Theoretically there should only be 1 point per recording interval in an observable_sum.   Similarly for up-down-sum.

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/Agent.java
```diff
@@ -0,0 +1,126 @@
...
+        setAggregatorFactory(meterProviderBuilder, OBSERVABLE_UP_DOWN_SUM, Aggregation.histogram());
+
+        var factory =
+          PeriodicMetricReader.create(exporterBuilder.build(),
+              Duration.ofMillis(EXPORT_MILLIS));
```
jsuereth, line 112: Nit:  Use Duration.ofSeconds(2)

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/AbstractThreadDispatchingHandler.java
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  public abstract String getEventName();
+
+  public abstract RecordedEventHandler createPerThreadSummarizer(String threadName);
+
+  @Override
+  public void accept(RecordedEvent ev) {
+    final Optional<String> possibleGroupedThreadName = grouper.groupedName(ev);
+    possibleGroupedThreadName.ifPresent(
```
jsuereth, line 29: so I understand, if the thread doesn't have a group name, we don't do anything with it?

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/HandlerRegistry.java
```diff
@@ -0,0 +1,56 @@
...
+import io.opentelemetry.contrib.jfr.metrics.memory.ObjectAllocationInNewTLABHandler;
+
+
+import java.util.*;
+import java.util.stream.Stream;
+
+public class HandlerRegistry {
+  private static final String SCHEMA_URL = "https://opentelemetry.io/schemas/1.6.1";
+  private static final String INSTRUMENTATION_NAME = "io.opentelemetry.contrib.jfr";
+  private static final String INSTRUMENTATION_VERSION = "1.7.0-SNAPSHOT";
```
jsuereth, line 22: Nit:  Maybe open a bug about pulling this version from a generated resources file in the build.

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/HandlerRegistry.java
```diff
@@ -0,0 +1,56 @@
...
+
+  private final List<RecordedEventHandler> mappers;
+
+  private HandlerRegistry(List<? extends RecordedEventHandler> mappers) {
+    this.mappers = new ArrayList<>(mappers);
+  }
+
+  public static HandlerRegistry createDefault(MeterProvider meterProvider) {
+    var otelMeter = meterProvider.get(INSTRUMENTATION_NAME, INSTRUMENTATION_VERSION, null);
```
jsuereth, line 31: Should this be using the SCHEMA_URL from above?

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/container/ContainerConfigurationHandler.java
```diff
@@ -0,0 +1,42 @@
+package io.opentelemetry.contrib.jfr.metrics.container;
+
+import io.opentelemetry.api.metrics.Meter;
+import io.opentelemetry.contrib.jfr.Constants;
+import io.opentelemetry.contrib.jfr.metrics.RecordedEventHandler;
+import jdk.jfr.consumer.RecordedEvent;
+
+public class ContainerConfigurationHandler implements RecordedEventHandler {
+    public static final String EVENT_NAME = "jdk.ContainerConfiguration";
+    public static final String JFR_CONTAINER_CONFIGURATION = "jfr.ContainerConfiguration";
```
jsuereth, line 10: This doesn't really abide by [our metric name conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/README.md#instrument-naming).

The `upDownCounterBuilder` name defines the metric name.   Likely this should be something like `cpu.usage` or `jvm.cpu.usage` or `jfr.cpu.usage`.    While metric names aren't stable, if we stick with existing conventions, then based on [existing system conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/system-metrics.md#metric-instruments) and guidance around runtime, I'd suggest:


Name: `runtime.jvm.cpu.utilization` 
Unit: `1` (guidance for all percentages are to use this unit).

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/cpu/ContextSwitchRateHandler.java
```diff
@@ -0,0 +1,44 @@
...
+
+  private final Meter otelMeter;
+  private volatile double value = 0;
+
+  public ContextSwitchRateHandler(Meter otelMeter) {
+    this.otelMeter = otelMeter;
+  }
+
+  public ContextSwitchRateHandler init() {
+    otelMeter.upDownCounterBuilder("jfr.ThreadContextSwitchRate")
```
jsuereth, line 23: Same guidance as CPU.

Now, if the context-switch-rate is actually PART of the cpu usage that's dedicated to thread-context-switching, then I'd suggest using `runtime.jvm.cpu.utilization` with an appropriate Attribute denoting `state=thread.context_switch`.

MIght be worth documenting the metrics we plan to expose in their own Markdown file, similar to what the collector does here: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/hostmetricsreceiver/internal/scraper/cpuscraper/metadata.yaml

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/cpu/LongLockHandler.java
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  @Override
+  public String getEventName() {
+    return EVENT_NAME;
+  }
+
+  @Override
+  public RecordedEventHandler createPerThreadSummarizer(String threadName) {
+    var ret = new PerThreadLongLockHandler(otelMeter, threadName);
```
jsuereth, line 27: Depending on what you're recording, you might be able to pass down a `bound` Histogram or Counter instrument here.

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/cpu/ContextSwitchRateHandler.java
```diff
@@ -0,0 +1,44 @@
...
+                                     .ofDoubles()
+                                     .setUnit(Constants.PERCENTAGE)
+                                     .buildWithCallback(codm -> codm.observe(value));
+    return this;
+  }
+
+  @Override
+  public void accept(RecordedEvent ev) {
+    value = ev.getDouble("switchRate");
```
jsuereth, line 32: Instead of building async instruments, you should be able to just create an UpDownCounter (or Histogram) instrument and directly call "record" or "add" here.

Given this is really the full value, UpDownCounter doesn't make much sense, but perhaps Histogram would.

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/cpu/OverallCPULoadHandler.java
```diff
@@ -0,0 +1,74 @@
...
+import java.time.temporal.ChronoUnit;
+import java.util.Optional;
+
+public class OverallCPULoadHandler implements RecordedEventHandler {
+  public static final String SIMPLE_CLASS_NAME = OverallCPULoadHandler.class.getSimpleName();
+  public static final String EVENT_NAME = "jdk.CPULoad";
+  public static final String JVM_USER = "jvmUser";
+  public static final String JVM_SYSTEM = "jvmSystem";
+  public static final String MACHINE_TOTAL = "machineTotal";
+  public static final String JFR_CPU_LOAD_JVM_USER = "jfr.CPULoad.jvmUser";
```
jsuereth, line 18: Instead of three different metrics, these should be one metric:  `runtime.jvm.cpu.utilization`, with attributes for `user` or `system`, etc.

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/cpu/PerThreadLongLockHandler.java
```diff
@@ -0,0 +1,58 @@
...
+  public static final String SIMPLE_CLASS_NAME = PerThreadLongLockHandler.class.getSimpleName();
+  public static final String MONITOR_CLASS = "monitorClass";
+  public static final String CLASS = "class";
+  public static final String THREAD_NAME = "thread.name";
+  public static final String EVENT_THREAD = "eventThread";
+  public static final String DURATION = "duration";
+  public static final String STACK_TRACE = "stackTrace";
+  public static final String JFR_JAVA_MONITOR_WAIT = "JfrJavaMonitorWait";
+
+  public static final String HISTOGRAM_NAME = "jfr.JavaMonitorWait.locktime";
```
jsuereth, line 21: Suggested name:   `runtime.jvm.lock.time` or `runtime.jvm.longlock.time`

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/memory/G1GarbageCollectionHandler.java
```diff
@@ -0,0 +1,42 @@
...
+import io.opentelemetry.api.metrics.Meter;
+import io.opentelemetry.contrib.jfr.Constants;
+import io.opentelemetry.contrib.jfr.metrics.RecordedEventHandler;
+import jdk.jfr.consumer.RecordedEvent;
+
+/** This class aggregates the duration of G1 Garbage Collection JFR events */
+public final class G1GarbageCollectionHandler implements RecordedEventHandler {
+  public static final String EVENT_NAME = "jdk.G1GarbageCollection";
+  public static final String JFR_G1_GARBAGE_COLLECTION_DURATION =
+      "jfr.G1GarbageCollection.duration";
```
jsuereth, line 13: Metric name suggestion:  jvm.runtime.gc.time

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/memory/GCHeapSummaryHandler.java
```diff
@@ -0,0 +1,111 @@
...
+import java.util.Map;
+
+/**
+ * This class handles GCHeapSummary JFR events. For GC purposes they come in
+ * pairs.
+ */
+public final class GCHeapSummaryHandler implements RecordedEventHandler {
+  public static final String SIMPLE_CLASS_NAME = GCHeapSummaryHandler.class.getSimpleName();
+  public static final String JFR_GC_HEAP_SUMMARY_DURATION = "jfr.GCHeapSummary.duration";
+  public static final String JFR_GC_HEAP_SUMMARY_HEAP_USED = "jfr.GCHeapSummary.heapUsed";
```
jsuereth, line 20: Same nit:  Used/Committed should be attributes of the same metric.

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/memory/PerThreadObjectAllocationInNewTLABHandler.java
```diff
@@ -0,0 +1,48 @@
...
+import io.opentelemetry.api.metrics.Meter;
+import io.opentelemetry.contrib.jfr.Constants;
+import io.opentelemetry.contrib.jfr.metrics.RecordedEventHandler;
+import jdk.jfr.consumer.RecordedEvent;
+
+/** This class aggregates all TLAB allocation JFR events for a single thread */
+public final class PerThreadObjectAllocationInNewTLABHandler implements RecordedEventHandler {
+
+  public static final String HISTOGRAM_NAME =
+          "jfr.ObjectAllocationInNewTLAB.allocation";
```
jsuereth, line 16: suggestion:   `runtime.jvm.tlab.allocations`, with attribute denoting "new"

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/memory/PerThreadObjectAllocationOutsideTLABHandler.java
```diff
@@ -0,0 +1,49 @@
...
+import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.api.metrics.BoundDoubleHistogram;
+import io.opentelemetry.api.metrics.Meter;
+import io.opentelemetry.contrib.jfr.Constants;
+import io.opentelemetry.contrib.jfr.metrics.RecordedEventHandler;
+import jdk.jfr.consumer.RecordedEvent;
+
+/** This class aggregates all non-TLAB allocation JFR events for a single thread */
+public final class PerThreadObjectAllocationOutsideTLABHandler implements RecordedEventHandler {
+  public static final String JFR_OBJECT_ALLOCATION_OUTSIDE_TLAB_ALLOCATION =
```
jsuereth, line 14: suggestion:   `runtime.jvm.tlab.allocations`, with attribute denoting "outside"

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/network/PerThreadNetworkReadHandler.java
```diff
@@ -0,0 +1,55 @@
...
+import io.opentelemetry.contrib.jfr.metrics.RecordedEventHandler;
+import jdk.jfr.consumer.RecordedEvent;
+
+public class PerThreadNetworkReadHandler implements RecordedEventHandler {
+  private static final String DESCRIPTION_BYTES = "Bytes Read";
+  private static final String DESCRIPTION_DURATION = "Read Duration";
+  private final String threadName;
+
+  public static final String JFR_SOCKET_READ_DURATION = "jfr.SocketRead.duration";
+  public static final String JFR_SOCKET_READ_BYTES_READ = "jfr.SocketRead.bytesRead";
```
jsuereth, line 18: Should stick to [system conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/system-metrics.md#systemnetwork---network-metrics) a bit closer.


Suggestion:   `runtime.jvm.network.io` with attribute denoting "read"

For duration, `runtime.jvm.network.duration`

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/network/PerThreadNetworkWriteHandler.java
```diff
@@ -0,0 +1,59 @@
...
+import io.opentelemetry.contrib.jfr.Constants;
+import io.opentelemetry.contrib.jfr.metrics.RecordedEventHandler;
+import jdk.jfr.consumer.RecordedEvent;
+
+public class PerThreadNetworkWriteHandler implements RecordedEventHandler {
+  public static final String SIMPLE_CLASS_NAME =
+      PerThreadNetworkWriteHandler.class.getSimpleName();
+  public static final String BYTES_WRITTEN = "bytesWritten";
+  public static final String THREAD_NAME = "thread.name";
+  public static final String JFR_SOCKET_WRITE_BYTES_WRITTEN = "jfr.SocketWrite.bytesWritten";
```
jsuereth, line 16: Same as for reads:

Suggestion: `runtime.jvm.network.io` with attribute denoting "write"

For duration, `runtime.jvm.network.duration` with attribute denoting "write"

jfr-streaming/src/test/java/io/opentelemetry/contrib/jfr/metrics/AbstractMetricsTest.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.Collection;
+import java.util.function.Consumer;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.awaitility.Awaitility.await;
+
+public class AbstractMetricsTest {
+
+  static SdkMeterProvider meterProvider;
+  static InMemoryMetricReader metricReader;
```
jsuereth, line 24: Note, post my latest PR you'll be able to create one of these that asks for DELTA on all metrics.

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/Constants.java
kittylyst: Fixed.

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/Agent.java
```diff
@@ -0,0 +1,126 @@
...
+    private static final String DEFAULT_SERVICE_NAME = "jfr-otlp-bridge";
+    private static final long EXPORT_MILLIS = 2000;
+
+    private static MeterProvider meterProvider;
+
+    public static void agentmain(String agentArgs, Instrumentation inst) {
+      premain(agentArgs, inst);
+    }
+
+    public static void premain(String agentArgs, Instrumentation inst) {
```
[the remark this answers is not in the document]
kittylyst: The idea is that we could do both. I.e. the main agent could call the premain of this one if running on Java 17

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/Agent.java
kittylyst: Thanks. Is there an example somewhere I can crib from?

jfr-streaming/README.md
```diff
@@ -0,0 +1,32 @@
+# jfrs-otlp
+JFR Streaming to OTLP Bridge
+
+* Java 17 only.
+
+* You need to run an initial bootstrapping build:
+
+```
+./gradlew :jfr-streaming:build -x test
```
anuraaga, line 9: Think we can update this to remove some stuff

jfr-streaming/build.gradle.kts
```diff
@@ -0,0 +1,63 @@
+plugins {
+  id("com.github.johnrengelman.shadow")
+  id("java")
```
anuraaga, line 3:
```suggestion

```
Handled by java-conventions

jfr-streaming/build.gradle.kts
```diff
@@ -0,0 +1,63 @@
...
+  mavenCentral()
+  mavenLocal()
+}
+
+java {
+  toolchain {
+    languageVersion.set(JavaLanguageVersion.of(17))
+  }
+}
+
```
anuraaga, line 17:
```suggestion
```

jfr-streaming/build.gradle.kts
```diff
@@ -0,0 +1,63 @@
...
+dependencies {
+  implementation("io.opentelemetry:opentelemetry-api")
+  implementation("io.opentelemetry:opentelemetry-sdk")
+  implementation("io.opentelemetry:opentelemetry-exporter-otlp")
+  implementation("io.opentelemetry:opentelemetry-sdk-metrics")
+  implementation("io.opentelemetry:opentelemetry-exporter-otlp-metrics")
+  implementation("io.grpc:grpc-netty-shaded")
```
anuraaga, line 24:
```suggestion
```

Now this library, like our instrumentation in general, only uses the API

jfr-streaming/build.gradle.kts
```diff
@@ -0,0 +1,63 @@
...
+}
+
+java {
+  toolchain {
+    languageVersion.set(JavaLanguageVersion.of(17))
+  }
+}
+
+dependencies {
+  implementation("io.opentelemetry:opentelemetry-api")
```
anuraaga, line 19:
```suggestion
  implementation("io.opentelemetry:opentelemetry-api-metrics")
```

jfr-streaming/build.gradle.kts
```diff
@@ -0,0 +1,63 @@
...
+  implementation("io.opentelemetry:opentelemetry-sdk")
+  implementation("io.opentelemetry:opentelemetry-exporter-otlp")
+  implementation("io.opentelemetry:opentelemetry-sdk-metrics")
+  implementation("io.opentelemetry:opentelemetry-exporter-otlp-metrics")
+  implementation("io.grpc:grpc-netty-shaded")
+
+  testImplementation("org.junit.jupiter:junit-jupiter-api")
+  testImplementation("org.junit.jupiter:junit-jupiter-engine")
+  testImplementation("org.assertj:assertj-core")
```
anuraaga, line 28:
```suggestion
```

jfr-streaming/build.gradle.kts
```diff
@@ -0,0 +1,63 @@
...
+  testImplementation("io.opentelemetry:opentelemetry-sdk-metrics-testing")
+  testImplementation("io.opentelemetry:opentelemetry-sdk-testing")
+
+  testImplementation("org.awaitility:awaitility")
```
anuraaga, line 32:
```suggestion
```

jfr-streaming/build.gradle.kts
```diff
@@ -0,0 +1,63 @@
+plugins {
+  id("com.github.johnrengelman.shadow")
```
anuraaga, line 2:
```suggestion
```

Not agenting for now

jfr-streaming/build.gradle.kts
```diff
@@ -0,0 +1,63 @@
...
+    }
+  }
+
+  test {
+    useJUnitPlatform()
+  }
+
+  named("assemble") {
+    dependsOn("shadowJar")
+  }
```
anuraaga, line 62:
```suggestion
```

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/HandlerRegistry.java
```diff
@@ -0,0 +1,60 @@
...
+  private static final String INSTRUMENTATION_NAME = "io.opentelemetry.contrib.jfr";
+  private static final String INSTRUMENTATION_VERSION = "1.7.0-SNAPSHOT";
+
+  private final List<RecordedEventHandler> mappers;
+
+  private HandlerRegistry(List<? extends RecordedEventHandler> mappers) {
+    this.mappers = new ArrayList<>(mappers);
+  }
+
+  public static HandlerRegistry createDefault(MeterProvider meterProvider) {
```
anuraaga, line 35:
```suggestion
 static HandlerRegistry createDefault(MeterProvider meterProvider) {
```

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/HandlerRegistry.java
```diff
@@ -0,0 +1,60 @@
...
+            new OverallCPULoadHandler(otelMeter),
+            new ContainerConfigurationHandler(otelMeter),
+            new LongLockHandler(otelMeter, grouper));
+    filtered.forEach(RecordedEventHandler::init);
+
+    return new HandlerRegistry(filtered);
+  }
+
+  /** @return a stream of all entries in this registry. */
+  public Stream<RecordedEventHandler> all() {
```
anuraaga, line 57:
```suggestion
 Stream<RecordedEventHandler> all() {
```

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/JfrMetrics.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.logging.Logger;
+import jdk.jfr.EventSettings;
+import jdk.jfr.consumer.RecordingStream;
+
+public final class JfrMetrics {
+  private JfrMetrics() {}
+
+  private static final Logger logger = Logger.getLogger(JfrMetrics.class.getName());
+
+  public static void enableJfr(MeterProvider meterProvider) {
```
anuraaga, line 22:
```suggestion
  public static void enable(MeterProvider meterProvider) {
```

Class has Jfr in it don't think method needs it

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/cpu/ContextSwitchRateHandler.java
```diff
@@ -0,0 +1,48 @@
...
+    value = ev.getDouble("switchRate");
+  }
+
+  public String getEventName() {
+    return EVENT_NAME;
+  }
+
+  @Override
+  public Optional<Duration> getPollingDuration() {
+    return Optional.of(Duration.of(1, ChronoUnit.SECONDS));
```
anuraaga, line 46: Would generally prefer the methods like `ofSeconds`

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/container/ContainerConfigurationHandler.java
```diff
@@ -0,0 +1,47 @@
...
+
+package io.opentelemetry.contrib.jfr.metrics.internal.container;
+
+import io.opentelemetry.api.metrics.Meter;
+import io.opentelemetry.contrib.jfr.metrics.internal.Constants;
+import io.opentelemetry.contrib.jfr.metrics.internal.RecordedEventHandler;
+import jdk.jfr.consumer.RecordedEvent;
+
+public final class ContainerConfigurationHandler implements RecordedEventHandler {
+  public static final String EVENT_NAME = "jdk.ContainerConfiguration";
```
anuraaga, line 14: Please go through all the static final fields in the PR to make private the ones that aren't used outside the class

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/RecordedEventHandler.java
```diff
@@ -0,0 +1,62 @@
...
+  default Optional<Duration> getThreshold() {
+    return Optional.empty();
+  }
+
+  /**
+   * Initialize the handler. Default implementation is a no-op
+   *
+   * @return
+   */
+  default RecordedEventHandler init() {
```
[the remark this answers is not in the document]
kittylyst: I think this should stay for now - it's related to the work I'm doing on making agents work in native binaries - separating construction from initialization is needed for that use case.

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/HandlerRegistry.java
kittylyst: That's why I put it in there, but @jkwatson is of the opinion it should be `null` for now.

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/JfrMetrics.java
```diff
@@ -0,0 +1,46 @@
...
+import io.opentelemetry.api.metrics.MeterProvider;
+import io.opentelemetry.contrib.jfr.metrics.internal.RecordedEventHandler;
+import java.util.concurrent.Executors;
+import java.util.function.Consumer;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import jdk.jfr.EventSettings;
+import jdk.jfr.consumer.RecordingStream;
+
+public final class JfrMetrics {
```
anuraaga, line 17: Would be good if you can add at least some simple javadoc to the class and enable method

anuraaga (approved):
LGTM would like a last look from @jsuereth too

PR #116, opened 2021-10-18
Update jmx-metrics to Groovy 3

anuraaga:
To allow building with Java 17

### DIFF
--- a/buildSrc/src/main/kotlin/otel.groovy-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/otel.groovy-conventions.gradle.kts
@@ -5,12 +5,6 @@ plugins {
     id("otel.spotless-conventions")
 }
 
-dependencies {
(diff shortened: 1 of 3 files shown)


buildSrc/src/main/kotlin/otel.groovy-conventions.gradle.kts
```diff
@@ -5,12 +5,6 @@ plugins {
     id("otel.spotless-conventions")
 }
 
-dependencies {
-    testImplementation("org.spockframework:spock-core:1.3-groovy-2.5") {
```
anuraaga, line 9: Actually I think the issue I had on Java 17 was just this one which I hadn't noticed when trying it. I could keep with Groovy 2, but figured it's a good time to upgrade anyways for support for modern syntax.

Integration tests still pass so presumably we're still compatible enough with existing scripts (though we don't guarantee it with the version being alpha anyways)

PR #121, opened 2021-10-21
Move some metrics over to histograms, tidy up

kittylyst:
After talking to Josh, we both feel like the majority of metrics from JFR should be histograms.

This PR moves a couple of the non-histograms over, and is really just a quick check that this the approach that we want to take.

No additional tests added. The CPU changes would be difficult to test, as they would be machine-dependent. We could try adding a GC test as well, though.

### DIFF
--- a/jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/network/PerThreadNetworkReadHandler.java
+++ b/jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/network/PerThreadNetworkReadHandler.java
@@ -22,7 +22,7 @@ public final class PerThreadNetworkReadHandler implements RecordedEventHandler {
   private final String threadName;
 
   private static final String METRIC_NAME_DURATION = "runtime.jvm.network.duration";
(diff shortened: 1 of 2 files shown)


jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/Constants.java
```diff
@@ -11,12 +11,26 @@
...
   public static final String WRITE = "write";
-  public static final AttributeKey<String> ATTR_THREAD_NAME = AttributeKey.stringKey("thread.name");
+  public static final String USER = "user";
+  public static final String SYSTEM = "system";
+  public static final String MACHINE = "machine_total";
+  public static final String G1 = "g1";
+  public static final String USED = "used";
+  public static final String COMMITTED = "committed";
+
+  public static final AttributeKey<String> ATTR_THREAD_NAME = AttributeKey.stringKey("thread_name");
```
jsuereth, line 26: In otel, we tend to use `.` for separators, even in attributes.  See: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/span-general.md#general-thread-attributes

PR #127, opened 2021-10-28
Tidy up jfr-streaming

jack-berg:
Was going through the the new jfr-streaming module and noticed some things I thought could use tidying up. In summary:
- Initialize instruments in constructors instead of `init()` method. 
- After all instrument initialization is done in constructors, can drop unused `init()` method
- Remove unused constants.
- Use java 11 `var` where appropriate. 
- Make classes static inner classes where appropriate. As a result, can make several constants `private`. 
- Change `AbstractThreadDispathingHandler#createPerThreadSummarizer` to return `Consumer<RecordedEvent>`. The difference in each thread is just in how they consumer recorded events - no need for each of these responses to require their own `getEventName()` implementation. 
- The logs told me that there was a conflict where multiple instruments for `runtime.jvm.network.duration` and `runtime.jvm.network.io`. Determined that the read and write handlers both were creating instruments with the same name. I differentiated them to include read / write in the metric name, eg, ``runtime.jvm.network.write.duration`. (On a related not, @jsuereth, these instruments had the same name but different descriptions, yet still conflicted. Does this match your expectations?)


jsuereth:
> The logs told me that there was a conflict where multiple instruments for runtime.jvm.network.duration and runtime.jvm.network.io. Determined that the read and write handlers both were creating instruments with the same name. I differentiated them to include read / write in the metric name, eg, ``runtime.jvm.network.write.duration`. (On a related not, @jsuereth, these instruments had the same name but different descriptions, yet still conflicted. Does this match your expectations?)

No, we should not differentiate that way.   Instead read/write need to be an attribute.   The conflict is likely due to the description/unit being different.   To the extent we register one instrument and "bind" those attributes, that'd be ideal.

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/network/NetworkWriteHandler.java
```diff
@@ -23,18 +38,30 @@
...
-import io.opentelemetry.contrib.jfr.metrics.internal.ThreadGrouper;
-
 public final class NetworkWriteHandler extends AbstractThreadDispatchingHandler {
-  static final String EVENT_NAME = "jdk.SocketWrite";
-  private final Meter otelMeter;
+  private static final String EVENT_NAME = "jdk.SocketWrite";
+  private static final String METRIC_NAME_DURATION = "runtime.jvm.network.write.duration";
+  private static final String METRIC_NAME_BYTES = "runtime.jvm.network.write.io";
+  private static final String DESCRIPTION_BYTES = "Bytes Written";
+  private static final String DESCRIPTION_DURATION = "Write Duration";
```
jsuereth, line 46: This is likely the issue between duration read.  This needs to be a more generic description about duration of a network operation.

jack-berg: Cool yeah that did the trick 👍

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/cpu/OverallCPULoadHandler.java
```diff
@@ -29,36 +28,32 @@
...
 
-  private final Meter otelMeter;
-  private BoundDoubleHistogram userHistogram;
-  private BoundDoubleHistogram systemHistogram;
-  private BoundDoubleHistogram machineHistogram;
+  private final BoundDoubleHistogram userHistogram;
+  private final BoundDoubleHistogram systemHistogram;
+  private final BoundDoubleHistogram machineHistogram;
 
   public OverallCPULoadHandler(Meter otelMeter) {
```
anuraaga, line 35: If we revert the removal of init I think we can merge this and follow up on that. Quarkus preinitializing classes was the reason to have an init method - ideally we can find some way to improve, possibly on the Quarkus side, to avoid forcing the pattern on us but I think we can try that later

jack-berg: I'm fine with doing that I just want to understand what problem it solves. It would make sense to me if the init methods accepted a `Meter`, so that the metric instruments could be initialized after the metric sdk was setup, but that wasn't happening.

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/cpu/OverallCPULoadHandler.java
```diff
@@ -30,36 +30,37 @@
...
+            .histogramBuilder(METRIC_NAME)
+            .setDescription(DESCRIPTION)
+            .setUnit(PERCENTAGE)
+            .build()
+            .bind(Attributes.of(ATTR_CPU_USAGE, USER));
+    systemHistogram =
+        meter
+            .histogramBuilder(METRIC_NAME)
+            .setDescription(DESCRIPTION)
+            .setUnit(ONE)
```
jack-berg, line 54: Noticed that user CPU usage unit is percentage while system and machine CPU usage unit is "1". Seems off.

trask: the unit for user CPU usage was just updated in #140, probably just missed it for system and machine

anuraaga (approved):

Issue #136, opened 2021-11-23
Add Support for Multiple Attributes


Mrod1598:
**Is your feature request related to a problem? Please describe.**
Add Ability to pass in multiple Attributes into one `otel.instrument()` call. 

**Describe the solution you'd like**
What I propose is adding additional signatures to otel.instrument that support attributes in the format of `Map<String, Map<String,String>>` where the strings represent `<Attribute, <LabelKey, LabelValue>>`

Example: 
currently what is required to collect this metric
```
otel.instrument(beantomcatconnectors, "tomcat.threads.idle", "description", "1",
    ["proto_handler" : { mbean -> mbean.name().getKeyProperty("name") }],
    "currentThreadCount", otel.&doubleValueObserver)
otel.instrument(beantomcatconnectors, "tomcat.threads.busy", "description", "1", 
    ["proto_handler" : { mbean -> mbean.name().getKeyProperty("name") }],
    "currentThreadsBusy", otel.&doubleValueObserver)
```
What it would like with the proposed solution
```
otel.instrument(beantomcatconnectors, "tomcat.threads", "description", "1",
    ["proto_handler" : { mbean -> mbean.name().getKeyProperty("name") }],
    ["currentThreadCount": ["Thread Type": "current"],
    "currentThreadsBusy": ["Thread Type": "busy"]], 
    otel.&doubleValueObserver)
```

rmfitzpatrick:
@Mrod1598 I can see the use of this feature and think it can be helpful.  Wouldn't we want the attributes to map to label functions to keep with the current convention?  Otherwise we seem to be limiting the capabilities out of the gate imo.

PR #137, opened 2021-11-23
Add multi attribute support

Mrod1598:
**Description:**
Adds support for multiple attributes through expanding on the `otel.instrument` signatures.  

**Existing Issue(s):**
https://github.com/open-telemetry/opentelemetry-java-contrib/issues/136

**Testing:**
Added to InstrumentHelperTest


jmx-metrics/src/test/java/io/opentelemetry/contrib/jmxmetrics/InstrumenterHelperTest.java
```diff
@@ -446,6 +446,64 @@ void doubleValueCallbackMultipleMBeans() throws Exception {
                       .satisfiesExactlyInAnyOrder(assertDoublePoints()));
     }
 
+    @Test
+    void doubleValueCallbackMultipleAttributes() throws Exception {
```
rmfitzpatrick, line 450: Could be a bit overkill, but would you mind adding a non-matching attribute name to this map or even a new case with all non-matching attributes (no valid values case)?

jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/MBeanHelper.groovy
```diff
@@ -95,4 +95,21 @@ class MBeanHelper {
...
+            return []
+        }
+
+        def ofInterest = isSingle ? [mbeans[0]]: mbeans
+        return [ofInterest, attributes].combinations().collect { pair ->
+            def (bean, attribute) = pair
+            try {
+                new Tuple3(bean, attribute, bean.getProperty(attribute))
+            } catch (AttributeNotFoundException e) {
+                logger.warning("Expected attribute ${attribute} not found in mbean ${bean.name()}")
```
rmfitzpatrick, line 110: It seems almost bound to happen by design since we are enabling a more aggregate metric type via filtering available attributes with the added interfaces. Maybe debug level? I think the above warning could be overzealous as well so even debug level could make sense there.

jmx-metrics/src/test/java/io/opentelemetry/contrib/jmxmetrics/InstrumenterHelperTest.java
Mrod1598: Added nonmatching attribute

jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/InstrumentHelper.groovy
```diff
@@ -57,21 +57,23 @@ class InstrumentHelper {
      * @param instrument - The {@link io.opentelemetry.api.metrics.Instrument}-producing {@link OtelHelper} method pointer:
      *        (e.g. new OtelHelper().&doubleValueRecorder)
      */
-    InstrumentHelper(MBeanHelper mBeanHelper, String instrumentName, String description, String unit, Map<String, Closure> labelFuncs, String attribute, Closure instrument) {
+    InstrumentHelper(MBeanHelper mBeanHelper, String instrumentName, String description, String unit, Map<String, Closure> labelFuncs, Map<String, Map<String, String>> attributes, Closure instrument) {
```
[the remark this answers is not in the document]
Mrod1598: I think the use of Closures is a good idea, added.

jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/MBeanHelper.groovy
Mrod1598: updated level

jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/InstrumentHelper.groovy
```diff
@@ -37,7 +37,7 @@ class InstrumentHelper {
     private final String instrumentName
     private final String description
     private final String unit
-    private final String attribute
+    private final Map<String, Map<String, Closure>> MBeanAttributes
```
rmfitzpatrick, line 40: nit 
```suggestion
    private final Map<String, Map<String, Closure>> mBeanAttributes
```

jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/InstrumentHelper.groovy
```diff
@@ -130,8 +135,11 @@ class InstrumentHelper {
         }
     }
 
-    private static Map<String, String> getLabels(GroovyMBean mbean, Map<String, Closure> labelFuncs) {
+    private static Map<String, String> getLabels(GroovyMBean mbean, Map<String, Closure> labelFuncs, Map<String, Closure> additionalLabels) {
         def labels = [:]
+        additionalLabels.each {label, labelFunc ->
```
rmfitzpatrick, line 140: I wonder if these should take priority over standard label functions since they are on an attribute basis.

rmfitzpatrick:
@Mrod1598 thanks for this. I think the implementation seems ok so it would be helpful to update the docstrings and readme to advertise the new signature(s) along with the changes.

jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/MBeanHelper.groovy
Mrod1598: wait actually i set the wrong level. https://docs.oracle.com/javase/7/docs/api/java/util/logging/Level.html these are the levels i see for the logger we're using here. I'm not seeing a debug level.

Mrod1598:
> @Mrod1598 thanks for this. I think the implementation seems ok so it would be helpful to update the docstrings and readme to advertise the new signature(s) along with the changes.

I'm wording if it's redundant to add all new signatures for all of the signatures added. Do you think it's fine to add a note about [here](https://github.com/open-telemetry/opentelemetry-java-contrib/blob/8992cf8dbda9173213868d0b9126034fe6ced79f/jmx-metrics/README.md?plain=1#L112) that says that attributes can be swapped out with the `Map<String,Map<String,Closure>>`?

jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/InstrumentHelper.groovy
Mrod1598: Makes sense to me

jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/InstrumentHelper.groovy
Mrod1598: updated

jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/MBeanHelper.groovy
rmfitzpatrick: Sorry, context switching.  Info or fine should be ok?

rmfitzpatrick:
> I'm wording if it's redundant to add all new signatures for all of the signatures added. Do you think it's fine to add a note about [here](https://github.com/open-telemetry/opentelemetry-java-contrib/blob/8992cf8dbda9173213868d0b9126034fe6ced79f/jmx-metrics/README.md?plain=1#L112) that says that attributes can be swapped out with the `Map<String,Map<String,Closure>>`?

I think it'd preferable to follow the existing pattern of the showing the instrument signature variations and introduce the new mbean attribute label function map with an example like is done with the labelFuncs.  I think the whole thing could be done in 5 lines new or so.

```markdown
In cases where you'd like to share instrument names while creating datapoints for multiple MBean attributes...

- `otel.instrument(MBeanHelper mBeanHelper, String instrumentName, String description, String unit, Map<String, Closure> labelFuncs, Map<String, Map<String, Closure>> attributeLabelFuncs, Closure instrument)`

<... description and example of mbeanAttributeFuncs with intended purpose... >

`otel.instrument()` provides additional signatures to allow this more expressive MBean attribute access:
- `otel.instrument(MBeanHelper mBeanHelper, String name, String description, String unit, Map<String, Map<String, Closure>> attributeLabelFuncs, Closure instrument)` - `labelFuncs` are empty map.
- `otel.instrument(MBeanHelper mBeanHelper, String name, String description, Map<String, Map<String, Closure>> attributeLabelFuncs, Closure instrument)` - `unit` is "1" and `labelFuncs` are empty map.
- `otel.instrument(MBeanHelper mBeanHelper, String name, Map<String, Map<String, Closure>> attributeLabelFuncs, Closure instrument)` - `description` is empty string, `unit` is "1" and `labelFuncs` are empty map

```

jmx-metrics/README.md
```diff
@@ -127,6 +127,25 @@ mutually exclusive with `otel.jmx.groovy.script`. The currently supported target
...
 - `otel.instrument(MBeanHelper mBeanHelper, String name, String attribute, Closure instrument)` - `description` is empty string, `unit` is "1" and `labelFuncs` are empty map.
 
+In cases where you'd like to share instrument names while creating datapoints for multiple MBean attributes:
+
+- `otel.instrument(MBeanHelper mBeanHelper, String instrumentName, String description, String unit, Map<String, Closure> labelFuncs, Map<String, Map<String, Closure>> attributeLabelFuncs, Closure instrument)`
+
+- An example of this would be with tomcat, To be able to consolidate different thread types into one metric:
+
+  ```groovy
+    otel.instrument(beantomcatconnectors, "tomcat.threads", "description", "1",
```
rmfitzpatrick, line 137: some suggested rewording after searching github for the connectors object:

```suggestion
- An example of this in Tomcat, is to consolidate different thread types into one `"tomcat.threads"` metric using both `currentThreadCount` and `currentThreadsBusy` MBean attributes, labeling with their applicable `"Thread Type"`:

  ```groovy
    otel.instrument(otel.mbean("Catalina:type=ThreadPool,name=*"), "tomcat.threads", "description", "1",
```

jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/OtelHelper.groovy
```diff
@@ -86,24 +86,40 @@ class OtelHelper {
...
-    InstrumentHelper instrument(MBeanHelper mBeanHelper, String instrumentName, String description, String unit, Map<String, Closure> labelFuncs, String attribute, Closure otelInstrument) {
-        def instrumentHelper = new InstrumentHelper(mBeanHelper, instrumentName, description, unit, labelFuncs, attribute, otelInstrument)
+    InstrumentHelper instrument(MBeanHelper mBeanHelper, String instrumentName, String description, String unit, Map<String, Closure> labelFuncs, Map<String, Map<String, Closure>> attributes, Closure otelInstrument) {
+        def instrumentHelper = new InstrumentHelper(mBeanHelper, instrumentName, description, unit, labelFuncs, attributes, otelInstrument)
         instrumentHelper.update()
         return instrumentHelper
     }
 
+    InstrumentHelper instrument(MBeanHelper mBeanHelper, String instrumentName, String description, String unit, Map<String, Closure> labelFuncs, String attribute, Closure otelInstrument) {
+      instrument(mBeanHelper, instrumentName, description, unit, labelFuncs, [(attribute): null], otelInstrument)
```
rmfitzpatrick, line 96: Since groovy was updated I guess explicit coercion is no longer required here but I wonder if this is preferable to using the NullObject:

```suggestion
      instrument(mBeanHelper, instrumentName, description, unit, labelFuncs, [(attribute): [:] as Map<String, Closure>], otelInstrument)
```

rmfitzpatrick (approved):

PR #138, opened 2021-11-24
Use service.name resource attribute instead of span name for service …

anuraaga:
I had misinterpreted the span name being more appropriate, remembering the transformation we do in the collector's xray exporter, but it should indeed by service name when sampling within OTel.

### DIFF
--- a/aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/SamplingRuleApplier.java
+++ b/aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/SamplingRuleApplier.java
@@ -144,7 +144,7 @@ final class SamplingRuleApplier {
     this.nextSnapshotTimeNanos = nextSnapshotTimeNanos;
   }
 
-  boolean matches(String name, Attributes attributes, Resource resource) {
+  boolean matches(Attributes attributes, Resource resource) {
     int matchedAttributes = 0;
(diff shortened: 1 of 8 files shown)


anuraaga (commented):
Note that SamplingRuleApplierTest was updated with the new matching semantics. The other tests were just changed to use a different per-span sampling criteria now that span name couldn't be used anymore.

laurit (approved):

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/SamplingRuleApplier.java
```diff
@@ -175,7 +175,7 @@ boolean matches(String name, Attributes attributes, Resource resource) {
     }
 
     return urlPathMatcher.matches(httpTarget)
```
willarmiros, line 177: What about updating this matcher to HTTP URL instead of target? Separate PR?

willarmiros (approved):
Code changes lgtm, one question about another attribute

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/SamplingRuleApplier.java
anuraaga: Yeah separate PR

PR #139, opened 2021-11-24
Remove usage of deprecated resource autoconfiguration

anuraaga:
The current method is deprecated, mostly because it's now possible to fully customize Resource customization out of band from any static context. New approach is hacky but should be fine until SDK provides a better solution.


laurit (approved):

willarmiros (approved):
Why is the current approach being deprecated if the only alternative is kind of a hack?

anuraaga:
The linked issue has some context, I'd say autoconfigure outgrew the current approach, which here is mostly working around not having a better way for samplers to use Resource in general.

PR #144, opened 2021-12-03
Update changelog and readme for 1.9

anuraaga:


trask (approved):
I think splitting out the updates by component was nice, but also support reducing release management overhead.

PR #146, opened 2021-12-06
[maven-extension] Capture details on mojo goal executions: `deploy:deploy`, `spring-boot:build-image`, `jib:build`, `snyk:test`, `snyk:monitor`

cyrille-leclerc:
**Description:**

Capture details on mojo goal executions: `deploy:deploy`, `spring-boot:build-image`, `jib:build`, `snyk:test`, and `snyk:monitor`

**Existing Issue(s):**

None

**Testing:**

Introduced unit tests.

**Documentation:**

See the updated README.md

**Outstanding items:**

**Span Attributes introduced in this PR**

Span attributes introduced in this PR.

Key namespaces:
* `maven.*`: anything related to Maven
   * `maven.build.*`: anything related to the artifacts produced by Maven builds
      * `maven.build.container`: producing docker/container images
         * `maven.build.container.image.*`: anything related to container images produced by Maven builds
         * `maven.build.container.registry.*`: attributes to describe container registry to which produced docker images are uploaded., describe the `url`of the registry and the `username` used for authentication.
      * `maven.build.repository`: uploading artifacts to a Maven repository
   * `maven.execution.*`: anything related to the Maven execution (e.g. the goal)
   * `maven.plugin.*`: anything related to the Maven plugin that is being executed
   * `maven.project.*`: anything related to the project/module that is building

**Noteworthy:** we capture the username used to upload to container registries and Maven repositories to help troubleshooting

| Span attribute                                 |  Type    | Description                                                          |
|------------------------------------------------|----------|----------------------------------------------------------------------|
| `maven.build.container.image.name`             | string   | Name of the produced Docker image                                                                          |
| `maven.build.container.image.tags`             | string[] | Tags of the produced Docker image                                                                          |
| `maven.build.container.registry.auth.username` | string   | Username used to upload the image to the container registry.                                               |
| `maven.build.container.registry.url`           | string   | URL of the container registry to which this image is uploaded.                                             |
| `maven.build.repository.auth.username`         | string   | Username used to upload artifacts to the Maven repository. See [Maven POM reference / Repository](https://maven.apache.org/pom.html#repository)     |
| `maven.build.repository.id`                    | string   | ID of the Maven repository to which the artifact is deployed. See [Maven POM reference / Repository](https://maven.apache.org/pom.html#repository)  |
| `maven.build.repository.url`                   | string   | URL of the Maven repository to which the artifact is deployed. See [Maven POM reference / Repository](https://maven.apache.org/pom.html#repository) |
| `maven.execution.goal`                         | string   | Goal that is being executed                                          |
| `maven.execution.id`                           | string   | ID of the execution                                                  |
| `maven.execution.lifecyclePhase`               | string   | Lifecycle phase to which belong the execution                        |
| `maven.plugin.artifactId`                      | string   | Artifact ID of the Maven plugin on which the Maven goal is executed  |
| `maven.plugin.groupId`                         | string   | Group ID of the Maven plugin on which the Maven goal is executed     |
| `maven.plugin.version`                         | string   | Version of the Maven plugin on which the Maven goal is executed      |
| `maven.project.artifactId`                     | string   | Artifact ID of the Maven project on which the Maven goal is executed |
| `maven.project.groupId`                        | string   | Group ID of the Maven project on which the Maven goal is executed    |
| `maven.project.version`                        | string   | Version of the Maven project on which the Maven goal is executed     |


maven-extension/src/main/java/io/opentelemetry/maven/OtelExecutionListener.java
```diff
@@ -200,11 +207,25 @@ public void mojoStarted(ExecutionEvent executionEvent) {
...
+    Optional<MojoGoalExecutionHandler> handler =
+        handlers.stream().filter(h -> h.supports(executionEvent)).findFirst();
+    logger.debug("OpenTelemetry: {} handler {}", executionEvent, handler);
+    handler.ifPresent(h -> h.enrichSpan(spanBuilder, executionEvent));
+
+    Span span = spanBuilder.startSpan();
     spanRegistry.putSpan(span, mojoExecution);
   }
 
+  /** TODO should we have a more generic and extensible mechanism to load handlers? */
```
kenfinnigan, line 222: I think we should, as ideally we only attach the handlers that are applicable for a given build.

maven-extension/src/main/java/io/opentelemetry/maven/handler/DeployDeployHandler.java
```diff
@@ -0,0 +1,88 @@
...
+import org.apache.maven.project.MavenProject;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Note: Later, we may prefer to have one span per actual upload than a generic span. We could
+ * achieve this instrumenting the <a
+ * href="https://projects.eclipse.org/projects/technology.aether">Aether library</a>
+ */
+public class DeployDeployHandler extends AbstractMojoGoalExecutionHandler
```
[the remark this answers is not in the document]
cyrille-leclerc: I changed to `MavenDeployDeployHandler, does it sound better?

maven-extension/src/main/java/io/opentelemetry/maven/OtelExecutionListener.java
cyrille-leclerc: I detailed more in the javadoc the options I identified. Questions I have are
* Will we ever have `MojoGoalExecutionHandlers` outside of this extension? Some could be added to the classpath but I wonder if it's realistic. I added such extension mechanisms on Jenkins plugins and it was never used outside of the library.
*  Will it have a performance impact? Probably not much.
* If we implement this, should we do it
   * "The Maven way" using the Maven Plexus container? I don't know how this would work and a quick Google search didn't help me
   * "The Otel way" using an SPI model with the `java.util.ServiceLoader`? Does it work well with the classloader isolation of Plexus? Probably yes.
*

maven-extension/src/main/java/io/opentelemetry/maven/handler/DeployDeployHandler.java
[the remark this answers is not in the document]
kenfinnigan: I see the name represents the action being performed, including plugin name, but is `MavenDeployHandler` sufficient?

The double `Deploy` messes with my brain.

maven-extension/src/main/java/io/opentelemetry/maven/OtelExecutionListener.java
kenfinnigan: I'd lean towards the service loader approach, unless there are issues with Plexus classloaders.

For adding all the handlers, although there may not be much of a performance impact in calling handlers that aren't needed, I would be concerned about unintended outcomes if we took that approach.

Maybe I'm overthinking it and it isn't a problem, but if we got to a point of more than a couple of the handlers, I could see there being an impact on performance as it's called for each Mojo start event.

maven-extension/src/main/java/io/opentelemetry/maven/handler/DeployDeployHandler.java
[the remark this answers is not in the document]
cyrille-leclerc: `MavenDeployHandler` works for me. I was focused on `mvn deploy:deploy` but 
* I forgot that there is a shortcut `mvn deploy` and 
* `MavenDeployHandler` will not conflict with the other goal of the Maven Deploy Plugin, `mvn deploy:deploy-file`

maven-extension/src/main/java/io/opentelemetry/maven/OtelExecutionListener.java
cyrille-leclerc: I'll propose this

maven-extension/src/main/java/io/opentelemetry/maven/handler/DeployDeployHandler.java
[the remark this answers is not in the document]
kenfinnigan: Thanks!

maven-extension/src/main/java/io/opentelemetry/maven/OtelExecutionListener.java
```diff
@@ -200,8 +226,16 @@ public void mojoStarted(ExecutionEvent executionEvent) {
                 MavenOtelSemanticAttributes.MAVEN_EXECUTION_ID, mojoExecution.getExecutionId())
             .setAttribute(
                 MavenOtelSemanticAttributes.MAVEN_EXECUTION_LIFECYCLE_PHASE,
-                mojoExecution.getLifecyclePhase())
-            .startSpan();
+                mojoExecution.getLifecyclePhase());
+    //  enrich spans with MojoGoalExecutionHandler
```
kenfinnigan, line 230: Thanks @cyrille-leclerc, looks good!

One question, as we're finding the first only, which I agree with, do we need to perform any validation earlier when retrieving the services to ensure there isn't a duplicate service handling the same goal and plugin combination?

maven-extension/src/main/java/io/opentelemetry/maven/handler/MojoGoalExecutionHandler.java
```diff
@@ -0,0 +1,25 @@
...
+import io.opentelemetry.api.trace.SpanBuilder;
+import org.apache.maven.execution.ExecutionEvent;
+
+public interface MojoGoalExecutionHandler extends Comparable<MojoGoalExecutionHandler> {
+
+  default void enrichSpan(SpanBuilder spanBuilder, ExecutionEvent executionEvent) {}
+
+  boolean supports(ExecutionEvent executionEvent);
+
+  default int ordinal() {
```
kenfinnigan, line 17: Doesn't look like this value is overridden in any of the implementations. Were you intending for it to be set and used in a comparision?

cyrille-leclerc: It's premature design 👼 
I used this concept in the Jenkins Pipeline Maven plugin ([here](https://github.com/jenkinsci/pipeline-maven-plugin/blob/pipeline-maven-parent-2.5.2/jenkins-plugin/src/main/java/org/jenkinsci/plugins/pipeline/maven/MavenPublisher.java#L83)).

It relates to your question above about the matching handlers, do we invoke just the first matching or all matching.

My algorithm is inconsistent. Either we just invoke the first matching and then sort by ordinal to have predictability or we invoke all matching handlers and don't necessarily need to sort the handler even though predictability is very valuable.

kenfinnigan: If we process multiple handlers, we'd likely end up with inconsistent, or possibly nonsensical, span attributes correct? It would be a mix of unique attributes from all handlers, with other values set based on the last one to set it.

Should probably only utilize one handler to enrich the span. So we either enforce distinctness on load, or provide an ordering and choose the first

maven-extension/src/main/java/io/opentelemetry/maven/OtelExecutionListener.java
cyrille-leclerc: done, please check the new code.

maven-extension/src/main/java/io/opentelemetry/maven/handler/MojoGoalExecutionHandler.java
cyrille-leclerc: Change done, please verify new code.
I break at load time if multiple handlers overlap handling the same goal.

maven-extension/README.md
```diff
@@ -57,42 +57,88 @@ mvn verify
...
+| `container.registry.username` | Username used to upload the image to the container registry.  Attribute only added when the `build-image` goal publishes the Docker image.                                               |
+| `peer.service`                | Docker Registry hostname. Attribute only added when the `build-image` goal publishes the Docker image.                                                                                   |
+| `http.url`                    | URL on the Docker registry deduced from the Docker image specified in the `build-image` goal configuration. Attribute only added when the `build-image` goal publishes the Docker image. |
+| `http.method`                 | `POST`. Attribute only added when the `build-image` goal publishes the Docker image.                                                                                                     |
+
+The `span.kind` is set to `client`
+
+TODO How can we report cases when multiple Docker image tags are published?  Introducing `container.image.tags`?
```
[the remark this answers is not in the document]
kenfinnigan: Looking at the semantic conventions, the `container.image` attributes are specific to a compute unit where an image is running. As we're dealing with creating images, they won't fit very well.

Maybe we need a `container.build.image.name`, to indicate it's a build vs a runtime? I'm also not sure how free we are to experiment with new attributes. @trask do you have any thoughts?

maven-extension/src/main/java/io/opentelemetry/maven/OtelExecutionListener.java
kenfinnigan: Thanks for the changes!

cyrille-leclerc:
@kenfinnigan the visibility OTel tracing brings on the flow of the build is amazing! It can greatly help teams optimize their builds.

I discover that even for pretty simple projects, it's easy to wrongly configure the chain of goals of a Maven build and have redundant executions of costly goals.

Here is an example below of a build just combining a Spring Boot app with Snyk checks, a deploy to a Maven repo, and and the publishing as a docker image.
I got some pom.xml fragments from snyk.io and Spring Boot web site to assemble a `mvn deploy` and the goals surefire:test, snyk:test are executed twice impacting seriously the duration ([pom.xml](https://github.com/cyrille-leclerc/ecommerce-antifraud/blob/8360baab884f7a8ac3628c6506a9beccdca5f8ad/pom.xml)).

maven-extension/README.md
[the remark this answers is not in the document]
cyrille-leclerc: Thanks @kenfinnigan! We enter an exciting conversation. I know that the Elastic Common Schema (ECS) philosophy is to try to reuse the same attribute names for the context and for the things produced. It's a difference we highlighted to the Otel Logs Sig when we presented ECS 18 months ago. It's the outcome of years and years of experience.

Following this philosophy of reuse, we would use `container.image.name` and `container.image.tag` when capturing docker image creations.
If we had to choose different attribute names, I see benefits in prefixing/suffixing the initial name rather than crafting a new name, I would prefer to look at `build.container.image.name` rather than `container.build.image.name`.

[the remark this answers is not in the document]
kenfinnigan: I was wondering about the "right" place to put `build`, and I take your point that using it as a prefix provides the ability to convert any runtime attribute to a build captured one.

I'd be good with adding `build` in front, but don't know if we need to get that into the spec or we can experiment here first?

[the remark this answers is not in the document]
trask: I'd recommend including the instrumentation name in the attribute until it gets spec'd to avoid confusion, e.g. a couple of examples from other instrumentations:

* `spring-webflux.handler.type`
* `rabbitmq.command`

so in this case something like

`maven....`

> Looking at the semantic conventions, the `container.image` attributes are specific to a compute unit where an image is running. As we're dealing with creating images, they won't fit very well.

ya, I would avoid re-using existing attribute names in a way that doesn't match the spec currently.

I like the general idea of `build.*` attributes / semantic conventions. Whenever you're ready, I think it's worth starting this discussion in the spec repo and proposing something based on what you've found useful here.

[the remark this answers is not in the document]
cyrille-leclerc: Note: as discussed with @v1v we will want to track "any" resource created by CI/CD/Deployment:
* binaries: jar file, executable file
* container images
* cloud resources produced by Ansible/Terraform/... like virtual machine, Kubernetes cluster, database instance, SNS topic...
* ...

Would we prefix by `build.*` ? or `produce.*` ?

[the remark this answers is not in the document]
kenfinnigan: Based on what @trask said, I think we need to have a name like `maven.spring.container.image`, or something equally wordy but it needs to start with `maven.*`.

However, for proposing something to the spec, we can come up with a good root level name to represent build outputs. For this, I like `build.*` as it enables capturing of details specific to a build process, and not just what was produced by a build, but that might depend on how much down that road the spec group wants to go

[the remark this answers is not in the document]
cyrille-leclerc: I've added handler for more Maven goals to better understand what attributes will look like.
We now handle `deploy:deploy`, `spring-boot:build-image`, `jib:build`, `snyk:test`, and `snyk:monitor` goals.
I still have to document this new goals in the README.md to clearly describe what will be the user experience and the attributes with this prefix `maven.build.*`

cyrille-leclerc:
@kenfinnigan @trask I have added in the description of this PR all the introduced span attributes with an explanation on the used namespaces (`container.registry.image.*` and `maven.*`) including the `maven.build.*` we have just discussed.
The detailed description of all the attribute per Maven Plugin Goal execution is available on the updated [README.md](https://github.com/open-telemetry/opentelemetry-java-contrib/blob/2021e0178e072bce3493940f21f8e02a5d1f4eb8/maven-extension/README.md).

I have implemented the enrichment for 5 plugin goals in order to have a broad set of use cases.

Even if @trask  suggested to subdivide the `maven.build.*` namespace per plugin that build things, I tried to use a common namespace for `maven.build.container.image.*` for Docker images produced either by the Spring Boot Maven Plugin or by the Google Jib Maven Plugin and IMO it works well.

I can easily implement @trask's suggestion with `maven.springboot.build.container.image.*` and `maven.googlejib.build.container.image.*` if we prefer.

trask:
> I tried to use a common namespace for `maven.build.container.image.*` for Docker images produced either by the Spring Boot Maven Plugin or by the Google Jib Maven Plugin and IMO it works well.

👍

on first glance, it may make sense to move `container.registry.url` and `container.registry.username` to `maven.build.container.registry.*`, which matches `maven.build.container.image.*`, and avoids some confusion with `container.*` resource attributes spec

maven-extension/src/main/java/io/opentelemetry/maven/OtelExecutionListener.java
```diff
@@ -200,8 +245,14 @@ public void mojoStarted(ExecutionEvent executionEvent) {
...
+    Optional<MojoGoalExecutionHandler> handler =
```
anuraaga, line 250: We don't use Optional much in OTel codebases in favor of Nullable. I just noticed I've been forgetting to add nullness checking to this repo like we have in SDK repo, will add that. Can you look into reducing Optional for more consistency with SDK repo? I don't think the usages here are all that different from null checks

maven-extension/src/main/java/io/opentelemetry/maven/handler/MavenDeployHandler.java
```diff
@@ -0,0 +1,97 @@
...
+
+  @Override
+  public void enrichSpan(SpanBuilder spanBuilder, ExecutionEvent execution) {
+    spanBuilder.setSpanKind(SpanKind.CLIENT);
+
+    MavenProject project = execution.getProject();
+    Optional<ArtifactRepository> optRepository =
+        Optional.ofNullable(project.getDistributionManagementArtifactRepository());
+
+    optRepository.ifPresent(
```
anuraaga, line 43: For example, `ofNullable` followed `ifPresent` like this is almost exactly the same as normal null checks, but adds verbose variables like `optRepository` and I think some extra indent too

maven-extension/src/main/java/io/opentelemetry/maven/MavenUtils.java
```diff
@@ -0,0 +1,37 @@
...
+   * <ul>
+   *   <li>maven-clean-plugin -&gt; clean
+   *   <li>sisu-maven-plugin -&gt; sisu
+   *   <li>spotbugs-maven-plugin -&gt; spotbugs
+   * </ul>
+   *
+   * @param pluginArtifactId the artifact ID of the mojo {@link MojoExecution#getArtifactId()}
+   * @return shortened name
+   */
+  public static String getPluginArtifactIdShortName(String pluginArtifactId) {
```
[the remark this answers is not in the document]
cyrille-leclerc: done

maven-extension/src/main/java/io/opentelemetry/maven/OtelExecutionListener.java
```diff
@@ -5,15 +5,22 @@
 
 package io.opentelemetry.maven;
 
+import com.google.common.annotations.VisibleForTesting;
```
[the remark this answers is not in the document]
cyrille-leclerc: done

maven-extension/src/main/java/io/opentelemetry/maven/handler/AbstractMojoGoalExecutionHandler.java
```diff
@@ -0,0 +1,14 @@
...
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.maven.handler;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public abstract class AbstractMojoGoalExecutionHandler implements MojoGoalExecutionHandler {
```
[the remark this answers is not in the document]
cyrille-leclerc: done

cyrille-leclerc:
> > I tried to use a common namespace for `maven.build.container.image.*` for Docker images produced either by the Spring Boot Maven Plugin or by the Google Jib Maven Plugin and IMO it works well.
> 
> 👍
> 
> on first glance, it may make sense to move `container.registry.url` and `container.registry.username` to `maven.build.container.registry.*`, which matches `maven.build.container.image.*`, and avoids some confusion with `container.*` resource attributes spec

@trask I followed your guidance and we now have `maven.build.container.*`and `maven.build.repository.*`

maven-extension/src/main/java/io/opentelemetry/maven/OtelExecutionListener.java
cyrille-leclerc: I like the principle _"When in Rome, do as the Romans do"_ :-). I'll replace the usage of `Optional` by using `@Nullable`

maven-extension/src/main/resources/META-INF/services/io.opentelemetry.maven.handler.MojoGoalExecutionHandler
```diff
@@ -0,0 +1,5 @@
+io.opentelemetry.maven.handler.GoogleJibBuildHandler
```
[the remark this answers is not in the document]
cyrille-leclerc: I introduced the SPI after discussing with @kenfinnigan . I see value in allowing Maven plugin authors to add instrumentation for the their plugins (plugin authors being vendors, oss projects or Maven practitioners).

What solution do you have in mind as an alternative to the SPI? 
Before implementing the SPI, I used a hardcoded list of handlers, could it be what you had in mind?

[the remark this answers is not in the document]
anuraaga: Ah I had meant a list of built in components that is appended to with SPI. I think it satisfies both itches.

anuraaga (commented):
So sorry I just merged in the change to error prone knowing that it is likely annoying for this PR. If you are able to merge main and fix issues that come up that would be great, otherwise I can do it to this PR tomorrow.

maven-extension/src/main/java/io/opentelemetry/maven/handler/MavenDeployHandler.java
cyrille-leclerc: Migration from `Optional` to `@Nullable` done

maven-extension/src/main/java/io/opentelemetry/maven/OtelExecutionListener.java
cyrille-leclerc: Migration from `Optional` to `@Nullable` done. Please verify

maven-extension/src/main/resources/META-INF/services/io.opentelemetry.maven.handler.MojoGoalExecutionHandler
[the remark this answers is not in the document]
cyrille-leclerc: Done through https://github.com/open-telemetry/opentelemetry-java-contrib/pull/146/commits/d393bea03468a3a1c2e8ef446d36a6ee96bc6946 . Please review

maven-extension/src/main/java/io/opentelemetry/maven/handler/MojoGoalExecutionHandlerConfiguration.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.ServiceLoader;
+import java.util.stream.Stream;
+
+public class MojoGoalExecutionHandlerConfiguration {
+
+  public static Map<MavenGoal, MojoGoalExecutionHandler> loadMojoGoalExecutionHandler(
+      ClassLoader classLoader) {
+
+    // built-in handlers
+    List<MojoGoalExecutionHandler> buildInHandlers =
```
anuraaga, line 23:
```suggestion
    List<MojoGoalExecutionHandler> builtInHandlers =
```

maven-extension/src/main/java/io/opentelemetry/maven/handler/MojoGoalExecutionHandlerConfiguration.java
```diff
@@ -0,0 +1,64 @@
...
+
+    Stream.concat(buildInHandlers.stream(), spiHandlers.stream())
+        .forEach(
+            handler ->
+                handler
+                    .getSupportedGoals()
+                    .forEach(
+                        goal -> {
+                          MojoGoalExecutionHandler previousHandler =
+                              mojoGoalExecutionHandlers.put(goal, handler);
```
anuraaga, line 49: Think you can use `toMap` since it will throw on duplicate keys

https://docs.oracle.com/javase/8/docs/api/java/util/stream/Collectors.html#toMap-java.util.function.Function-java.util.function.Function-

anuraaga (commented):
Thanks just small points

maven-extension/src/main/java/io/opentelemetry/maven/handler/MojoGoalExecutionHandlerConfiguration.java
cyrille-leclerc: done

maven-extension/src/main/java/io/opentelemetry/maven/handler/MojoGoalExecutionHandlerConfiguration.java
cyrille-leclerc: Thanks @anuraaga . Could you please give me more guidance on this one? 
I don't catch how I can use `stream.collect(Collectors.toMap())` as I have a double loop 👼

anuraaga: Oops - yeah didn't think enough about it, stream doesn't seem good for this case

anuraaga (approved):

cyrille-leclerc:
@ipleten  I see on https://github.com/jenkinsci/opentelemetry-plugin/issues/281 that you have CI pipelines build Java applications and packaging them as Docker images.

Could you please by any chance:
* Indicate us how you build Docker images in your CI pipelines? A Maven / Gradle plugin? A shell script invoking `docker build`? ...
* Have a look at the visualization of Maven + Docker builds we bring with this Pull Request and share feedback on this (screenshot [above](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/146#issuecomment-991115758) )?

Issue #160, opened 2022-01-03
[Maven Extension] Maven parallel / multithreaded builds fail


delanym:
My reactor fails with

```
java.lang.IllegalStateException: A span has already been started for com.github.ekryd.sortpom:sortpom-maven-plugin:3.0.0:sort {execution: project}
    at io.opentelemetry.maven.SpanRegistry.putSpan (SpanRegistry.java:90)
    at io.opentelemetry.maven.OtelExecutionListener.mojoStarted (OtelExecutionListener.java:205)                                                                                         
    at io.opentelemetry.maven.ChainedExecutionListener.mojoStarted (ChainedExecutionListener.java:88)
    at org.apache.maven.lifecycle.internal.DefaultExecutionEventCatapult.fire (DefaultExecutionEventCatapult.java:84)                                                                    
    at org.apache.maven.lifecycle.internal.DefaultExecutionEventCatapult.fire (DefaultExecutionEventCatapult.java:42)                                                                   
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:204)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:156)                                                                                                  
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:148)                                                                                                  
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:117)     
    at org.apache.maven.lifecycle.internal.builder.multithreaded.MultiThreadedBuilder$1.call (MultiThreadedBuilder.java:196)
    at org.apache.maven.lifecycle.internal.builder.multithreaded.MultiThreadedBuilder$1.call (MultiThreadedBuilder.java:186)
```

I effective ran `./mvnw validate -e --threads 1C` with an external OTEL_EXPORTER_OTLP_ENDPOINT

I disabled the sortpom, and then maven enforcer, and then it failed on clean plugin. Only way to survive the build is --threads 1.

cyrille-leclerc:
Thanks @delanym . Do you have a sample Maven project on which we could reproduce the problem?

delanym:
@cyrille-leclerc im afraid I don't, but I should think you could reproduce on any multithread build, no? It seems like a simple problem of not including the module name in the span name.

delanym:
```
otel-cli server tui
git clone https://github.com/apache/maven
[...]

PR #161, opened 2022-01-03
[Maven Extension] Support Maven Parallel Builds

cyrille-leclerc:
**Description:**

Support [Maven Parallel Builds](https://cwiki.apache.org/confluence/display/MAVEN/Parallel+builds+in+Maven+3).

**Existing Issue(s):**

* https://github.com/open-telemetry/opentelemetry-java-contrib/issues/160

**Testing:**

Test the following build with the modified opentelemetry-maven-extension
```
otel-cli server tui
git clone https://github.com/apache/maven
cd maven
mvn dependency:copy -Dartifact=io.opentelemetry.contrib:opentelemetry-maven-extension:1.9.0-alpha
cp target/dependency/opentelemetry-maven-extension-1.9.0-alpha.jar /tmp/
mvn clean -Dotel.traces.exporter=otlp -Dotel.exporter.otlp.endpoint=http://localhost:4317 -Dmaven.ext.class.path=/tmp/opentelemetry-maven-extension-1.9.0-alpha.jar -T3
```

**Documentation:**

No change needed

**Outstanding items:**

None

### DIFF
--- a/maven-extension/src/main/java/io/opentelemetry/maven/SpanRegistry.java
+++ b/maven-extension/src/main/java/io/opentelemetry/maven/SpanRegistry.java
@@ -16,10 +16,14 @@ import org.apache.maven.model.Plugin;
 import org.apache.maven.plugin.MojoExecution;
 import org.apache.maven.project.MavenProject;
 import org.codehaus.plexus.component.annotations.Component;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /** Hold the state of the spans in progress */
 @Component(role = SpanRegistry.class)
 public final class SpanRegistry {
+
+  private static final Logger logger = LoggerFactory.getLogger(SpanRegistry.class);
 
   private final Map<MojoExecutionKey, Span> mojoExecutionKeySpanMap = new ConcurrentHashMap<>();
   private final Map<MavenProjectKey, Span> mavenProjectKeySpanMap = new ConcurrentHashMap<>();
@@ -38,6 +42,7 @@ public final class SpanRegistry {
   }
 
   public Span getSpan(MavenProject mavenProject) {
+    logger.debug("OpenTelemetry: getSpan({}, {})", mavenProject, Thread.currentThread());
     final MavenProjectKey key = MavenProjectKey.fromMavenProject(mavenProject);
     final Span span = this.mavenProjectKeySpanMap.get(key);
     if (span == null) {
@@ -74,6 +79,7 @@ public final class SpanRegistry {
   }
 
   public void putSpan(Span span, MavenProject mavenProject) {
+    logger.debug("OpenTelemetry: putSpan({}, {})", mavenProject, Thread.currentThread());
     MavenProjectKey key = MavenProjectKey.fromMavenProject(mavenProject);
     Span previousSpanForKey = mavenProjectKeySpanMap.put(key, span);
     if (previousSpanForKey != null) {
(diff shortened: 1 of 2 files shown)


maven-extension/src/main/java/io/opentelemetry/maven/SpanRegistry.java
```diff
@@ -134,7 +148,9 @@ public static MavenProjectKey fromMavenProject(@Nonnull MavenProject mavenProjec
 
     abstract String pluginArtifactId();
 
-    static MojoExecutionKey fromMojoExecution(MojoExecution mojoExecution) {
+    abstract MavenProjectKey projectKey();
```
kenfinnigan, line 151: Where is `projectKey()` used?

maven-extension/src/main/java/io/opentelemetry/maven/SpanRegistry.java
```diff
@@ -114,9 +126,11 @@ public Span removeSpan(MojoExecution mojoExecution) {
 
     abstract String artifactId();
 
-    public static MavenProjectKey fromMavenProject(@Nonnull MavenProject mavenProject) {
+    abstract String version();
```
kenfinnigan, line 129: Where is `version()` used?

kenfinnigan (commented):
Looks reasonable, but had a question about a couple of new abstract methods

maven-extension/src/main/java/io/opentelemetry/maven/SpanRegistry.java
cyrille-leclerc: It is used by the Map of spans by `mojoExecutionKey` of the SpanRegistry. It's needed when using Maven Parallel Builds as the same mojo goal can be executed simultaneously on multiple Maven sub projects.

See https://github.com/open-telemetry/opentelemetry-java-contrib/blob/v1.9.0/maven-extension/src/main/java/io/opentelemetry/maven/SpanRegistry.java#L23

maven-extension/src/main/java/io/opentelemetry/maven/SpanRegistry.java
cyrille-leclerc: It's a protection for the map of spans by MojoExecutionKeys. I imagine 2 mojo executions can't execute simultaneously  on the same project with the different version of the Maven plugin but I felt it's a safe guardrail.

kenfinnigan (approved):

kenfinnigan:
@trask Is there a problem with the `CODEOWNERS` config as I expected to be one of the allowed approvers, which doesn't seem to be the case.

trask:
> @trask Is there a problem with the `CODEOWNERS` config as I expected to be one of the allowed approvers, which doesn't seem to be the case.

ya, the problem is that:

> The people you choose as code owners must have write permissions for the repository

(https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners)

I'm not sure what the best workaround is. For now I have given everyone in `CODEOWNERS` write permission to the repo, and I'll seek guidance in the otel community repo.

trask (approved):

maven-extension/src/main/java/io/opentelemetry/maven/SpanRegistry.java
```diff
@@ -38,6 +42,7 @@ public void setRootSpan(Span rootSpan) {
   }
 
   public Span getSpan(MavenProject mavenProject) {
+    logger.debug("OpenTelemetry: getSpan({}, {})", mavenProject, Thread.currentThread());
```
anuraaga, line 45: I don't think we should have `Thread.currentThread` in logging messages - most loggers will output the thread info if configured in the pattern anyways.

If we do have it we to check if debug level is enabled before calling the not trivially cheap method

cyrille-leclerc: Thanks @anuraaga , fixed in PR https://github.com/open-telemetry/opentelemetry-java-contrib/pull/163

> most loggers will output the thread info if configured in the pattern anyways.

Unfortunately it's not the case with the `mvn --debug` / `mvn -X` option but we can remove the thread name as we understood the behaviour of Maven Parallel Builds.

Issue #153, opened 2021-12-16
JMX receiver jar relies on setting otel.imr.export.interval


jsuereth:
See: https://github.com/open-telemetry/opentelemetry-java-contrib/blob/main/jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/JmxConfig.java#L23

We should have some kind of unit tests to make sure the pass-through enviornment variables line up.  This breaks using the JMX receiver using collector-contrib configuration.

rmfitzpatrick:
@jsuereth are you reporting that the `otel.jmx.interval.milliseconds` and `otel.imr.export.interval` are separate properties and should be unified? The first determines the groovy script running interval and the other is for the exporter. The JMX receiver could be updated to set both with the collection interval but I'm not immediately seeing another problem.

sophieyfang:
@rmfitzpatrick We are seeing the collection_interval not being respected, the streams are generated at a frequency that's often < 10s causing a problem in our side. (Points get denied by API side if it's being written too fast)

sophieyfang:
> @jsuereth are you reporting that* `otel.jmx.interval.milliseconds` and `otel.imr.export.interval` are separate properties and should be unified? The first determines the groovy script running interval and the other is for the exporter. The JMX receiver could be updated to set both with the collection interval but I'm not immediately seeing another problem.

I think Josh meant [otel.imr.export.interval](https://github.com/open-telemetry/opentelemetry-java-contrib/blob/v1.9.0/jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/JmxConfig.java#L23) and [otel.metric.export.interval](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/MetricExporterConfiguration.java#L158)?

cpheps:
So I've been researching this a bit. It appears that [otel.imr.export.interval](https://github.com/open-telemetry/opentelemetry-java-contrib/blob/v1.9.0/jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/JmxConfig.java#L23) has been deprecated since 1.9.0 according to the [opentelemetry-java changelog](https://github.com/open-telemetry/opentelemetry-java/blob/main/CHANGELOG.md?plain=1#L168-L170). 

[...]

PR #177, opened 2022-01-11
Changed EXPORTER_INTERVAL constant in JMX config to be metric.export.interval

cpheps:
**Description:**
Bug Fix: Changed deprecated value `otel.imr.export.interval` to new value `otel.metric.export.interval` for `EXPORTER_INTERVAL` in JMX Config.

**Existing Issue(s):**

References #153 

**Testing:**

Tested locally with configuration through collector-contrib.

**Documentation:**
None

**Outstanding items:**

It's not clear that this export interval is being fully respected. During testing if you don't set the input interval, which then sets the export interval in jmx, the default should be 1 minute per [this](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/MetricExporterConfiguration.java#L159-L161) . It doesn't seem like that's the case. Looking into what is happening and if it's expected behavior.

### DIFF
--- a/jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/JmxConfig.java
+++ b/jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/JmxConfig.java
@@ -20,7 +20,7 @@ class JmxConfig {
   static final String METRICS_EXPORTER_TYPE = PREFIX + "metrics.exporter";
   static final String EXPORTER = PREFIX + "exporter.";
 
-  static final String EXPORTER_INTERVAL = PREFIX + "imr.export.interval";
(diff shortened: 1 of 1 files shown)


cpheps:
I think this actually needs to be closed after doing further research. I missed that support for `otel.imr.export.interval` was being removed in 1.10 but there is still backwards compatibility in 1.9.1 [here](https://github.com/open-telemetry/opentelemetry-java/blob/v1.9.1/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/MetricExporterConfiguration.java#L155-L161) which is the current version on main.

Issue #183, opened 2022-01-12
[Maven extension] cant set service name


delanym:
I want to set the service.name to something other than Maven.

`.\mvnw.cmd  -e clean install --threads 3 -Dotel.traces.exporter=otlp -Dotel.exporter.otlp.endpoint=http://10.1.0.203:4317 -Dotel.instrumentation.maven.mojo.enabled=false -Dmaven.ext.class.path=c:\opt\opentelemetry-maven-extension-1.10.0-20220112.132157-1.jar -Dotel.resource.attributes=service.name=cats,service.namespace=mammals -Dotel.service.name=django`

In this line I've set the service name in 2 ways, but neither of them seems to have any effect
https://github.com/open-telemetry/opentelemetry-java/tree/main/sdk-extensions/autoconfigure

What am I doing wrong?

cyrille-leclerc:
Should be fixed by https://github.com/open-telemetry/opentelemetry-java-contrib/pull/184

delanym:
It looks fixed, but it also looks like there's no longer a default "maven".

cyrille-leclerc:
@delanym can you please verify it's fixed with v1.10.0?

delanym:
@cyrille-leclerc I've checked it on v1.10.0-alpha

delanym:
@cyrille-leclerc is it possible to set the service name to the runner when using -T3 etc? Similar to what you get with MAVEN_OPTS="-Dorg.slf4j.simpleLogger.showThreadName=true"

cyrille-leclerc:
> @cyrille-leclerc is it possible to set the service name to the runner when using -T3 etc? Similar to what you get with `MAVEN_OPTS="-Dorg.slf4j.simpleLogger.showThreadName=true"`

`service.name` should be settable using the `-Dotel.service.name=my-maven` system property

See https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk-extensions/autoconfigure/README.md#opentelemetry-resource

delanym:
@cyrille-leclerc im asking for something else #209

PR #187, opened 2022-01-13
Fix default `service.name` + simplify configuration using Otel AutoConfig SDK 1.10 ResourceProvider SPI improvements

cyrille-leclerc:
**Description:**

Fix default `service.name` + simplify configuration using Otel AutoConfig SDK 1.10 ResourceProvider SPI improvements (enable specifying the classloader making it compatible with Maven Plexus)

**Existing Issue(s):**

* https://github.com/open-telemetry/opentelemetry-java-contrib/issues/183

**Testing:**

Introduced tests with `OpenTelemetrySdkServiceTest`

**Documentation:**

NONE

**Outstanding items:**

NONE

### DIFF
--- a/maven-extension/src/main/java/io/opentelemetry/maven/resources/MavenResourceProvider.java
+++ b/maven-extension/src/main/java/io/opentelemetry/maven/resources/MavenResourceProvider.java
@@ -1,0 +1,27 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.maven.resources;
+
+import io.opentelemetry.maven.semconv.MavenOtelSemanticAttributes;
+import io.opentelemetry.sdk.autoconfigure.spi.ConfigProperties;
+import io.opentelemetry.sdk.autoconfigure.spi.ResourceProvider;
+import io.opentelemetry.sdk.resources.Resource;
+import io.opentelemetry.semconv.resource.attributes.ResourceAttributes;
+import org.apache.maven.rtinfo.RuntimeInformation;
+import org.apache.maven.rtinfo.internal.DefaultRuntimeInformation;
+
+public class MavenResourceProvider implements ResourceProvider {
+  @Override
+  public Resource createResource(ConfigProperties config) {
+    // TODO verify if there is solution to retrieve the RuntimeInformation instance loaded by the
+    //  Maven Plexus Launcher
+    RuntimeInformation runtimeInformation = new DefaultRuntimeInformation();
+    return Resource.builder()
+        .put(ResourceAttributes.SERVICE_NAME, MavenOtelSemanticAttributes.SERVICE_NAME_VALUE)
+        .put(ResourceAttributes.SERVICE_VERSION, runtimeInformation.getMavenVersion())
+        .build();
+  }
+}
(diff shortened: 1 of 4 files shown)


cyrille-leclerc:
@kenfinnigan  @trask sorry for the regression caused by the previous PR. I simplified the configuration code leveraging the AutoConfigure SDK 1.10 improvements and I introduced unit tests.

kenfinnigan (approved):

trask (approved):
> I simplified the configuration code leveraging the AutoConfigure SDK 1.10 improvements and I introduced unit tests.

👍

maven-extension/src/main/java/io/opentelemetry/maven/resources/MavenResourceProvider.java
```diff
@@ -0,0 +1,27 @@
...
+import io.opentelemetry.sdk.autoconfigure.spi.ResourceProvider;
+import io.opentelemetry.sdk.resources.Resource;
+import io.opentelemetry.semconv.resource.attributes.ResourceAttributes;
+import org.apache.maven.rtinfo.RuntimeInformation;
+import org.apache.maven.rtinfo.internal.DefaultRuntimeInformation;
+
+public class MavenResourceProvider implements ResourceProvider {
+  @Override
+  public Resource createResource(ConfigProperties config) {
+    // TODO verify if there is solution to retrieve the RuntimeInformation instance loaded by the
```
anuraaga, line 19: Hmm - the code is a bit simpler but the previous version seems to be more robust since it was using the real RuntimeInformation. It's one of the reasons for adding the resource customizer hook, for these situations where some class is used that's not available statically

cyrille-leclerc: I'll iterate on this ASAP. I have to understand the best pattern to support default values provided by runtime components with  overwrites by the environment variables

I see an interesting impedance question to integrate the Java ServiceLoader SPI with dependency injection frameworks like `@inject`. Do you see things the same way?

anuraaga: @cyrille-leclerc Yeah I don't think SPI works that well in the context of a system using dependency injection. So was hoping that the customizers on the builder will be able to solve those cases. Note that `service.name` is extremely special - it is one of few attributes automatically filled by the SDK, with the other ones having no use case for user customization.

https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java#L48

So some ugly code handling it should be fine, the attribute itself is an ugly duckling, probably something like

```java
if (oldResource.get(SERVICE_NAME).equals(Resource.getDefault().get(SERVICE_NAME)) {
  newResource.put(SERVICE_NAME, MAVEN_SERVICE_NAME);
}
```

Long term if this is tricky in many cases, we could possibly consider adding a `serviceName` field to the SDK autoconfiguration builder that takes care of that logic, but for now we'd like to see how things go.

cyrille-leclerc: Sorry @anuraaga I missed your comment, your point makes sense. Many thanks!

PR #190, opened 2022-01-13
jmx gatherer: Remove manual exporter flush 

rmfitzpatrick:
**Description:**
Bug fix  - More recent changes to the sdk have made manual meter provider collection and exporter functionality management unnecessary by the metric gatherer's groovy script runner. The existing runner functionality also now conflicts with the deprecated `otel.imr.export.interval` property that was being set.

These changes remove manual flushing and update the target property to `otel.metric.export.interval`, but still respect its value if set.

Credit to @cpheps and @sophieyfang for tracking this down.*

**Existing Issue(s):**
https://github.com/open-telemetry/opentelemetry-java-contrib/issues/153 

**Testing:**
Added applicable unit test and made required updates for runner functional change.

**Documentation:**
Documented functionality and `otel.metric.export.interval` property treatment.

Also adding unrelated tomcat target system link that was missing.

### DIFF
--- a/jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/GroovyRunner.java
+++ b/jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/GroovyRunner.java
@@ -135,15 +135,6 @@ public class GroovyRunner {
     for (final Script script : scripts) {
       script.run();
     }
-    flush();
-  }
-
-  public void flush() {
-    groovyMetricEnvironment.flush();
-  }
-
-  public void shutdown() {
-    flush();
   }
 
   // Visible for testing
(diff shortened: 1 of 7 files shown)


Mrod1598 (approved):
LGTM

jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/GroovyRunner.java
```diff
@@ -135,15 +135,6 @@ public void run() {
     for (final Script script : scripts) {
       script.run();
     }
-    flush();
-  }
-
-  public void flush() {
-    groovyMetricEnvironment.flush();
```
pmcollins, line 142: Looks like `groovyMetricEnvironment` can be converted to a local variable now that it's only used in the constructor.

pmcollins (approved):
Just one comment -- LGTM otherwise.

jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/GroovyRunner.java
rmfitzpatrick: I'm not able to push to this branch so will address in a followup: https://github.com/open-telemetry/opentelemetry-java-contrib/issues/197

Issue #152, opened 2021-12-16
[Maven Extension] Otel Maven Extension is not reporting its version as `otel.library.version`


cyrille-leclerc:
**Is your feature request related to a problem? Please describe.**

Otel Maven Extension is not reporting its version as `otel.library.version`

**Describe the solution you'd like**

Add the OTel Plugin version using `OpenTelemetry.getTracer(instrumentName, instrumentVersion)`.

**Describe alternatives you've considered**

N/A

**Additional context**

trask:
> Pending question: is there a standardised way to report this version number?

I believe the standard way would be to report this as the `instrumentationVersion` (when constructing the `Tracer`).

Check out

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/dba6e429e4ff437e386a07a1d101c57f2b232b44/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java#L111-L112

and

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/dba6e429e4ff437e386a07a1d101c57f2b232b44/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/InstrumentationVersion.java#L9-L10

(and https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#get-a-tracer)

PR #191, opened 2022-01-13
[Maven Extension] Add Tracer instrumentationVersion (ie `otel.library.version`)

cyrille-leclerc:
**Description:**

Add Tracer instrumentationVersion (ie `otel.library.version` attribute).

**Existing Issue(s):**

* https://github.com/open-telemetry/opentelemetry-java-contrib/issues/152

**Testing:**

Tested manually.
No unit test because algorithm relies on the jar manifest implementationVersion that is not available during the unit tests.

**Documentation:**

None

**Outstanding items:**

None

### DIFF
--- a/maven-extension/src/main/java/io/opentelemetry/maven/OpenTelemetrySdkService.java
+++ b/maven-extension/src/main/java/io/opentelemetry/maven/OpenTelemetrySdkService.java
@@ -7,7 +7,6 @@ package io.opentelemetry.maven;
 
 import io.opentelemetry.api.OpenTelemetry;
 import io.opentelemetry.api.trace.Tracer;
-import io.opentelemetry.api.trace.TracerBuilder;
 import io.opentelemetry.context.propagation.ContextPropagators;
 import io.opentelemetry.sdk.OpenTelemetrySdk;
 import io.opentelemetry.sdk.autoconfigure.AutoConfiguredOpenTelemetrySdk;
@@ -25,6 +24,9 @@ import org.slf4j.LoggerFactory;
 /** Service to configure the {@link OpenTelemetry} instance. */
 @Component(role = OpenTelemetrySdkService.class, hint = "opentelemetry-service")
 public final class OpenTelemetrySdkService implements Initializable, Disposable {
+
+  public static final String VERSION =
+      OpenTelemetrySdkService.class.getPackage().getImplementationVersion();
 
   private static final Logger logger = LoggerFactory.getLogger(OpenTelemetrySdkService.class);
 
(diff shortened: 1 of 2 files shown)


maven-extension/src/main/java/io/opentelemetry/maven/OpenTelemetrySdkService.java
```diff
@@ -106,7 +107,12 @@ public void initialize() {
             .getBoolean("otel.instrumentation.maven.mojo.enabled");
     this.mojosInstrumentationEnabled = mojoSpansEnabled == null ? true : mojoSpansEnabled;
 
-    this.tracer = this.openTelemetry.getTracer("io.opentelemetry.contrib.maven");
+    TracerBuilder tracerBuilder = openTelemetry.tracerBuilder("io.opentelemetry.contrib.maven");
+    String otelMavenExtensionVersion = this.getClass().getPackage().getImplementationVersion();
+    if (otelMavenExtensionVersion != null) {
+      tracerBuilder.setInstrumentationVersion(otelMavenExtensionVersion);
+    }
+    this.tracer = tracerBuilder.build();
```
kenfinnigan, line 115: Do we need the `null` check? From looking at the implementation it appears we can go with:

```
this.tracer = this.openTelemetry.getTracer("io.opentelemetry.contrib.maven", this.getClass().getPackage().getImplementationVersion());
```

cyrille-leclerc: I initially started like this as `implementationVersion`  will only be null in unit tests.

However, I hardened the code because null is a violation of the contract`OpenTelemetry#getTracer(@Nonnull instrumentationName, @Nonnull instrumentationVersion)`.

It's probably an acceptable violation of the contract as the implementation supports null and we would not be afraid of regressions if the implementation starts rejecting null instrumentationVersion because it would fail in unit tests.

cyrille-leclerc: @kenfinnigan please review the proposed code change. Slightly different from your suggestion in order to also output the otel maven extension version in the logs.

kenfinnigan (approved):
I'm ok with this approach as `VERSION` should only be `null` if there was a problem of some kind

trask (approved):

maven-extension/src/main/java/io/opentelemetry/maven/OpenTelemetrySdkService.java
```diff
@@ -25,6 +25,9 @@
 @Component(role = OpenTelemetrySdkService.class, hint = "opentelemetry-service")
 public final class OpenTelemetrySdkService implements Initializable, Disposable {
 
+  public static final String VERSION =
+      OpenTelemetrySdkService.class.getPackage().getImplementationVersion();
```
anuraaga, line 29: I don't think we want to expose this as a public field

kenfinnigan: Good point @anuraaga, I'd missed that.

cyrille-leclerc: Thanks, PR created https://github.com/open-telemetry/opentelemetry-java-contrib/pull/198

PR #207, opened 2022-01-19
Use type expressions from Java 17

anuraaga:
I've never coded against a Java 17 codebase yet and couldn't resist trying out a new feature

### DIFF
--- a/jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/memory/GCHeapSummaryHandler.java
+++ b/jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/memory/GCHeapSummaryHandler.java
@@ -106,8 +106,7 @@ public final class GCHeapSummaryHandler implements RecordedEventHandler {
     }
     if (after.hasField(HEAP_SPACE)) {
       after.getValue(HEAP_SPACE);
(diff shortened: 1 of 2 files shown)


jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/memory/GCHeapSummaryHandler.java
```diff
@@ -106,8 +106,7 @@ private void recordValues(RecordedEvent before, RecordedEvent after) {
     }
     if (after.hasField(HEAP_SPACE)) {
       after.getValue(HEAP_SPACE);
```
laurit, line 108: can this line be deleted?

laurit (approved):

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/memory/GCHeapSummaryHandler.java
anuraaga: Oops

mateuszrzeszutek (approved):

breedx-splk (approved):
Cool feature.

PR #210, opened 2022-01-19
Add Hadoop Monitoring

jmwilliams89:
**Description:**
- Adds Apache Hadoop monitoring capabilities

**Testing:**
- Added integration tests for Apache Hadoop

**Documentation:**
-  Updated JMX main readme to include Hadoop as a supported integration
- Added Hadoop metric documentation

### DIFF
--- a/jmx-metrics/docs/target-systems/hadoop.md
+++ b/jmx-metrics/docs/target-systems/hadoop.md
@@ -18,42 +18,42 @@ These metrics are sourced from: https://hadoop.apache.org/docs/r2.7.2/hadoop-pro
 
 * Name: `hadoop.hdfs.block.count`
 * Description: The total number of blocks.
-* Unit: `blocks`
+* Unit: `{blocks}`
 * Labels: `node_name`
 * Instrument Type: LongUpDownCounterCallback
 
 * Name: `hadoop.hdfs.block.missing`
 * Description: The number of blocks reported as missing.
-* Unit: `blocks`
+* Unit: `{blocks}`
 * Labels: `node_name`
 * Instrument Type: LongUpDownCounterCallback
 
 * Name: `hadoop.hdfs.block.corrupt`
 * Description: The number of blocks reported as corrupt.
-* Unit: `blocks`
+* Unit: `{blocks}`
 * Labels: `node_name`
 * Instrument Type: LongUpDownCounterCallback
 
 * Name: `hadoop.hdfs.volume.failed`
 * Description: The number of failed volumes.
-* Unit: `volumes`
+* Unit: `{volumes}`
 * Labels: `node_name`
 * Instrument Type: LongUpDownCounterCallback
 
 * Name: `hadoop.hdfs.file.count`
 * Description: The total number of files being tracked by the name node.
-* Unit: `files`
+* Unit: `{files}`
 * Labels: `node_name`
 * Instrument Type: LongUpDownCounterCallback
 
 * Name: `hadoop.hdfs.file.load`
 * Description: The current number of concurrent file accesses.
-* Unit: `operations`
+* Unit: `{operations}`
 * Labels: `node_name`
 * Instrument Type: LongUpDownCounterCallback
 
 * Name: `hadoop.data_node.count`
 * Description: The number of data nodes tracked by the name node.
-* Unit: `nodes`
+* Unit: `{nodes}`
 * Labels: `node_name`, `state`
 * Instrument Type: LongUpDownCounterCallback
--- a/jmx-metrics/docs/target-systems/hadoop.md
+++ b/jmx-metrics/docs/target-systems/hadoop.md
@@ -1,0 +1,59 @@
+# Hadoop Metrics
+
+The JMX Metric Gatherer provides built in Hadoop metric gathering capabilities.
+These metrics are sourced from: https://hadoop.apache.org/docs/r2.7.2/hadoop-project-dist/hadoop-common/Metrics.html
+
+### HDFS Metrics
+* Name: `hdfs.disk.usage`
+* Description: The amount of disk used by data nodes.
+* Unit: `by`
+* Labels: `node_name`
+* Instrument Type: LongUpDownCounterCallback
+
+* Name: `hdfs.disk.limit`
+* Description: The total disk allotted to data nodes.
+* Unit: `by`
+* Labels: `node_name`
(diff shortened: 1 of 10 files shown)


jmx-metrics/docs/target-systems/hadoop.md
```diff
@@ -0,0 +1,59 @@
+# Hadoop Metrics
+
+The JMX Metric Gatherer provides built in Hadoop metric gathering capabilities.
+These metrics are sourced from: https://hadoop.apache.org/docs/r2.7.2/hadoop-project-dist/hadoop-common/Metrics.html
+
+### HDFS Metrics
+* Name: `hdfs.disk.usage`
```
Mrod1598, line 7: all metrics need `hadoop.` prefix added

jmx-metrics/src/main/resources/target-systems/hadoop.groovy
```diff
@@ -0,0 +1,45 @@
...
+otel.instrument(beanHadoopNameNodeFS, "hadoop.hdfs.block.count", "The total number of blocks.", "blocks",
+  ["node_name" : { mbean -> mbean.getProperty("tag.Hostname") }],
+  "BlocksTotal", otel.&longUpDownCounterCallback)
+otel.instrument(beanHadoopNameNodeFS, "hadoop.hdfs.block.missing", "The number of blocks reported as missing.", "blocks",
+  ["node_name" : { mbean -> mbean.getProperty("tag.Hostname") }],
+  "MissingBlocks", otel.&longUpDownCounterCallback)
+otel.instrument(beanHadoopNameNodeFS, "hadoop.hdfs.block.corrupt", "The number of blocks reported as corrupt.", "blocks",
+  ["node_name" : { mbean -> mbean.getProperty("tag.Hostname") }],
+  "CorruptBlocks", otel.&longUpDownCounterCallback)
+otel.instrument(beanHadoopNameNodeFS, "hadoop.hdfs.volume.failed", "The number of failed volumes.", "volumes",
```
Mrod1598, line 33: shouldn't units be wrapped in `{}` i.e. `{volumes}`  if not explicitly called out as a specific type, ex: `By`, `ms`.
https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/metrics/semantic_conventions#instrument-units

Mrod1598 (changes requested):

jmx-metrics/src/main/resources/target-systems/hadoop.groovy
jmwilliams89: Based on the supplied specification, I believe that's correct and this should be updated. However, no other groovy script appears to be following this convention yet, and both activemq and tomcat are breaking it.

jmx-metrics/docs/target-systems/hadoop.md
jmwilliams89: Good catch. This should be addressed now.

Mrod1598 (approved):
LGTM if build passes

jmx-metrics/src/integrationTest/resources/hadoop/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM bmedora/hadoop:2.9-base
+
+COPY yarn-site.xml /hadoop/etc/hadoop/yarn-site.xml
```
anuraaga, line 3: It looks like you should be able to use copyFileToContainer instead of a Dockerfile. We definitely want to avoid building a docker image during test execution if at all possible

jmwilliams89: Thanks for the suggestion. I was able to implement it following the example in Cassandra and it definitely speeds up the testing time.

jmx-metrics/src/main/resources/target-systems/hadoop.groovy
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+def beanHadoopNameNodeFS = otel.mbean("Hadoop:service=NameNode,name=FSNamesystem")
+otel.instrument(beanHadoopNameNodeFS, "hadoop.hdfs.disk.usage", "The amount of disk used by data nodes.", "by",
```
rmfitzpatrick, line 18: Can we track with the descriptions used in the docs? They use `Current used capacity across all DataNodes in bytes` so that "disk" doesn't seem directly applicable. It may also be preferable to identify these metrics as originating from a name node so that the `node_name` attribute is clearer.

jmx-metrics/src/main/resources/target-systems/hadoop.groovy
```diff
@@ -0,0 +1,45 @@
...
+  ["node_name" : { mbean -> mbean.getProperty("tag.Hostname") }],
+  "CapacityUsed", otel.&longUpDownCounterCallback)
+otel.instrument(beanHadoopNameNodeFS, "hadoop.hdfs.disk.limit", "The total disk allotted to data nodes.", "by",
+  ["node_name" : { mbean -> mbean.getProperty("tag.Hostname") }],
+  "CapacityTotal", otel.&longUpDownCounterCallback)
+otel.instrument(beanHadoopNameNodeFS, "hadoop.hdfs.block.count", "The total number of blocks.", "{blocks}",
```
rmfitzpatrick, line 24: I don't think `{}` are common in otel units so would prefer we not use them.

rmfitzpatrick: Oh nice thanks for the head's up. I had missed this update and assumed it was from porting similar scripts.

anuraaga (approved):

jmx-metrics/src/main/resources/target-systems/hadoop.groovy
jmwilliams89: I agree. The origin probably makes more sense than the category for the metric namespace. I've updated the metrics to replace `hdfs` with `name_node`. And I've also replaced `disk` with the `capacity` terminology.

rmfitzpatrick (approved):
Thanks!

PR #223, opened 2022-02-01
Fixes Solr JMX Metrics

StefanKurek:
**Description:**

Fixes Solr metrics from using otel.mbean to otel.mbeans to not ignore additional relevant MBeans.

### DIFF
--- a/jmx-metrics/src/main/resources/target-systems/solr.groovy
+++ b/jmx-metrics/src/main/resources/target-systems/solr.groovy
@@ -14,12 +14,12 @@
  * limitations under the License.
  */
 
-def beanSolrCoreSearcherNumDocs = otel.mbean("solr:dom1=core,dom2=*,category=SEARCHER,scope=searcher,name=numDocs")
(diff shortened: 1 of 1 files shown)


StefanKurek:
@rmfitzpatrick when you have a moment could you please take a look. This is a quick fix regarding me not fully understanding the difference between otel.mbean and otel.mbeans.

jmx-metrics/src/main/resources/target-systems/solr.groovy
```diff
@@ -14,12 +14,12 @@
  * limitations under the License.
  */
 
-def beanSolrCoreSearcherNumDocs = otel.mbean("solr:dom1=core,dom2=*,category=SEARCHER,scope=searcher,name=numDocs")
+def beanSolrCoreSearcherNumDocs = otel.mbeans(["solr:dom1=core,dom2=*,category=SEARCHER,scope=searcher,name=numDocs"])
```
dehaansa, line 17: The function's overloaded, you don't need to call with a bracketed list, you can just call mbeans with a single string and still resolve the issue that you're encountering.

StefanKurek: @dehaansa fixed

rmfitzpatrick (approved):

PR #224, opened 2022-02-02
WildFly Monitoring

jmwilliams89:
**Description:**
- Added WildFly monitoring capabilities.
- Added support for overriding the underlying scraper in an integration test. This makes it possible to test WildFly, because it requires an additional client jar added to the classpath of the scraper.

**Testing:**
- Added integration tests with a custom built scraper.

**Documentation:**
- Added documentation for WildFly metrics.

### DIFF
--- a/jmx-metrics/docs/target-systems/wildfly.md
+++ b/jmx-metrics/docs/target-systems/wildfly.md
@@ -1,0 +1,76 @@
+# WildFly Metrics
+
+The JMX Metric Gatherer provides built in WildFly metric gathering capabilities.
+
+### Deployment Metrics
+* Name: `wildfly.session.count`
+* Description: The number of sessions created.
+* Unit: `{sessions}`
+* Labels: `deployment`
+* Instrument Type: LongCounterCallback
+
+* Name: `wildfly.session.active`
+* Description: The number of currently active sessions.
+* Unit: `{sessions}`
+* Labels: `deployment`
+* Instrument Type: LongUpDownCounterCallback
+
+* Name: `wildfly.session.expired`
+* Description: The number of sessions that have expired.
+* Unit: `{sessions}`
+* Labels: `deployment`
+* Instrument Type: LongCounterCallback
+
+* Name: `wildfly.session.rejected`
+* Description: The number of sessions that have been rejected.
+* Unit: `{sessions}`
+* Labels: `deployment`
+* Instrument Type: LongCounterCallback
+
+### Listener Metrics
+* Name: `wildfly.request.count`
+* Description: The number of requests received.
+* Unit: `{requests}`
+* Labels: `server`, `listener`
+* Instrument Type: LongCounterCallback
+
+* Name: `wildfly.request.time`
+* Description: The total amount of time spent on requests.
+* Unit: `ns`
+* Labels: `server`, `listener`
+* Instrument Type: LongCounterCallback
+
+* Name: `wildfly.request.server_error`
+* Description: The number of requests that have resulted in a 500 response.
+* Unit: `{requests}`
+* Labels: `server`, `listener`
+* Instrument Type: LongCounterCallback
+
+* Name: `wildfly.network.io`
+* Description: The number of bytes transmitted.
+* Unit: `by`
+* Labels: `server`, `listener`, `state`
+* Instrument Type: LongCounterCallback
+
+### Data Source Metrics
+* Name: `wildfly.jdbc.connection.open`
+* Description: The number of open jdbc connections.
+* Unit: `{connections}`
+* Labels: `data_source`, `state`
+* Instrument Type: LongUpDownCounterCallback
+
+* Name: `wildfly.jdbc.request.wait`
+* Description: The number of jdbc connections that had to wait before opening.
+* Unit: `{requests}`
+* Labels: `data_source`
+* Instrument Type: LongCounterCallback
+
+* Name: `wildfly.jdbc.transaction.count`
+* Description: The number of transactions created.
+* Unit: `{transactions}`
+* Instrument Type: LongCounterCallback
+
+* Name: `wildfly.jdbc.rollback.count`
+* Description: The number of transactions rolled back.
+* Unit: `{transactions}`
+* Instrument Type: LongCounterCallback
--- a/jmx-metrics/docs/target-systems/wildfly.md
+++ b/jmx-metrics/docs/target-systems/wildfly.md
@@ -41,7 +41,7 @@ The JMX Metric Gatherer provides built in WildFly metric gathering capabilities.
 * Instrument Type: LongCounterCallback
 
 * Name: `wildfly.request.server_error`
-* Description: The number of requests that have resulted in a 500 response.
+* Description: The number of requests that have resulted in a 5xx response.
 * Unit: `{requests}`
 * Labels: `server`, `listener`
 * Instrument Type: LongCounterCallback
--- a/jmx-metrics/docs/target-systems/wildfly.md
+++ b/jmx-metrics/docs/target-systems/wildfly.md
@@ -1,6 +1,7 @@
 # WildFly Metrics
 
 The JMX Metric Gatherer provides built in WildFly metric gathering capabilities.
+Details about using JMX with WildFly can be found here: https://docs.jboss.org/author/display/WFLY/JMX%20subsystem%20configuration.html
 
 ### Deployment Metrics
 * Name: `wildfly.session.count`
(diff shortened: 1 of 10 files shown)


jmx-metrics/docs/target-systems/wildfly.md
```diff
@@ -0,0 +1,76 @@
+# WildFly Metrics
+
+The JMX Metric Gatherer provides built in WildFly metric gathering capabilities.
```
rmfitzpatrick, line 3: Is there a ~jmx monitoring reference you can link to like is done in the other docs?

jmx-metrics/src/main/resources/target-systems/wildfly.groovy
```diff
@@ -0,0 +1,60 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+def beanWildflyDeployment = otel.mbeans("jboss.as.expr:deployment=*,subsystem=undertow")
+otel.instrument(beanWildflyDeployment, "wildfly.session.count", "The number of sessions created.", "{sessions}",
+  ["deployment": { mbean -> mbean.name().getKeyProperty("deployment")}],
+  "sessionsCreated", otel.&longCounterCallback)
```
rmfitzpatrick, line 20: Would it make sense to use the new attribute map feature and have all of these with the `wildfly.session.count` instrument name and a ~`state` attribute?

```groovy
 otel.instrument(
  beanWildflyDeployment, "wildfly.session.count",
  "The number of sessions in a given state", "{sessions}",
  ["deployment" : { mbean -> mbean.name().getKeyProperty("deployment") }],
  ["sessionsCreated": ["state": {mbean -> "created"}],
   "activeSessions": ["state": {mbean -> "active"}],
   "expiredSessions": ["state": {mbean -> "expired"}],
   "rejectedSessions": ["state": {mbean -> "rejected"}]],
  otel.&longCounterCallback)
```

jmx-metrics/src/integrationTest/resources/wildfly/start.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+/opt/jboss/wildfly/bin/standalone.sh -b 0.0.0.0 -bmanagement 0.0.0.0 & \
+
+/opt/jboss/wildfly/bin/add-user.sh user password --silent & \
+
+java -cp /app/OpenTelemetryJava.jar:/opt/jboss/wildfly/bin/client/jboss-client.jar \
+  -Dotel.jmx.username=user -Dotel.jmx.password=password \
```
rmfitzpatrick, line 8: believe these are unnecessary

jmx-metrics/src/integrationTest/java/io/opentelemetry/contrib/jmxmetrics/target_systems/WildflyIntegrationTest.java
```diff
@@ -0,0 +1,128 @@
...
+  which gives it access to the client jar packaged with WildFly. */
+  @Override
+  protected GenericContainer<?> buildScraper(String otlpEndpoint) {
+    String scraperJarPath = System.getProperty("shadow.jar.path");
+
+    return new GenericContainer<>("jboss/wildfly:23.0.1.Final")
+        .withNetwork(Network.SHARED)
+        .withNetworkAliases("wildfly")
+        .withCopyFileToContainer(
+            MountableFile.forHostPath(scraperJarPath), "/app/OpenTelemetryJava.jar")
```
rmfitzpatrick, line 35: nit 
```suggestion
            MountableFile.forHostPath(scraperJarPath), "/app/OpenTelemetryJMXMetrics.jar")
```

jmx-metrics/docs/target-systems/wildfly.md
```diff
@@ -0,0 +1,76 @@
...
+* Instrument Type: LongCounterCallback
+
+* Name: `wildfly.request.time`
+* Description: The total amount of time spent on requests.
+* Unit: `ns`
+* Labels: `server`, `listener`
+* Instrument Type: LongCounterCallback
+
+* Name: `wildfly.request.server_error`
+* Description: The number of requests that have resulted in a 500 response.
```
rmfitzpatrick, line 44: is this a bit more accurate?

```suggestion
* Description: The number of requests that have resulted in a 5xx response.
```

jmx-metrics/docs/target-systems/wildfly.md
jmwilliams89: Unlike the other JMX integrations, this one has been harder to link to official documentation. The WildFly docs outline how to connect through JMX (https://docs.jboss.org/author/display/WFLY/JMX%20subsystem%20configuration.html), but I've been unable to find a resource listing the available mbeans. The majority of my research was done through manually searching the hierarchy in jconsole and finding the most common metrics.

There are 3rd party resources out there that outline some of the more important mbeans (such as Datadog for example), but I felt it wasn't appropriate to link to another vendor's/product's documentation.

jmx-metrics/src/main/resources/target-systems/wildfly.groovy
jmwilliams89: I originally wanted to data model these after that pattern, but I think the main issue I encountered is that "activeSessions" is a non-monotonic sum. It represents what is currently active. In contrast, created, expired, and rejected are monotonic. If we were to create a single metric and attribute it by those states, the sum of everything would not be meaningful.

jmx-metrics/src/integrationTest/resources/wildfly/start.sh
jmwilliams89: I may be mistaken, but I believe I need those. When the container starts up, I need to start the WildFly process, add a user (since it seems to be a requirement with this JMX implementation), and start the metric scraper. 

I tried to combine these all in a command when creating the generic container, but I discovered it's fairly complicated to execute multiple commands. From what I've seen, it's easier to create a startup script and execute that in your command when creating the generic container. Also, it turns out that the server startup is blocking, so that somewhat complicates matters further, as we need to execute that in the background.

jmx-metrics/src/integrationTest/resources/target-systems/wildfly.properties
```diff
@@ -0,0 +1,9 @@
+otel.jmx.interval.milliseconds = 3000
+otel.metrics.exporter = otlp
+otel.jmx.service.url = service:jmx:remote+http://wildfly:9990
+otel.jmx.target.system = wildfly
+
+# these will be overridden by cmd line
+otel.exporter.otlp.endpoint = http://host.testcontainers.internal
+otel.jmx.username = user
+otel.jmx.password = password
```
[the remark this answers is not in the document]
jmwilliams89: No. My apologies. This was copied from another integration's properties file when first starting WildFly. I don't believe it's accurate. I'll remove it and ensure that tests still pass.

jmx-metrics/docs/target-systems/wildfly.md
jmwilliams89: I agree. That  is more accurate. I've updated the description in the docs, script, and integration test.

jmx-metrics/src/integrationTest/java/io/opentelemetry/contrib/jmxmetrics/target_systems/WildflyIntegrationTest.java
jmwilliams89: I've updated the jar name. However, I pulled that name originally from AbstractIntegrationTest, so we might want to change it there as well at a later point.

jmx-metrics/src/integrationTest/resources/wildfly/start.sh
rmfitzpatrick: They were unnecessary before being removed from the properties file in your later change.

jmx-metrics/src/main/resources/target-systems/wildfly.groovy
rmfitzpatrick: Oh right, though I wonder if the others would still be applicable for this, but the names are pretty clear as is.

jmx-metrics/docs/target-systems/wildfly.md
rmfitzpatrick: What you linked may be better than nothing and is helpful in suggesting this is a supported/provided interface? Unfortunate no better guides are in the rh docs.

jmwilliams89: Agreed. It is unfortunate. I went ahead and added that link, with verbiage similar to what Solr did.

rmfitzpatrick (approved):
Thanks for the addition!

PR #225, opened 2022-02-03
Fixes Solr metric by using alternate attribute

StefanKurek:
**Description:**
solr.cache.size was using the size attribute which actually keeps track of how many "items" are in the cache. Instead we should be using ramBytesUsed to keep track of the size in bytes.


StefanKurek:
@rmfitzpatrick another small Solr metric tweak when you have a chance to look at it. Thanks!

dehaansa (approved):

rmfitzpatrick (approved):

PR #226, opened 2022-02-07
sampler and span processor prototype implementations for consistent sampling

oertl:
**Description:**

This PR adds various Sampler and SpanProcessor prototype implementations for consistent sampling as defined by https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/tracestate-probability-sampling.md and https://github.com/open-telemetry/opentelemetry-specification/pull/2047:
* **ConsistentSampler**: abstract base class of all consistent sampler implementations below
* **ConsistentAlwaysOffSampler**: see https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/tracestate-probability-sampling.md#always-off-sampler 
* **ConsistentAlwaysOnSampler**: see https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/tracestate-probability-sampling.md#always-on-consistent-probability-sampler
* **ConsistentComposedAndSampler**: allows combining two consistent samplers and samples when both samplers would sample
* **ConsistentComposedOrSampler**: allows combining two consistent sampler and samples when at least one of both samplers would sample, see https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/tracestate-probability-sampling.md#requirement-combine-multiple-consistent-probability-samplers-using-the-minimum-p-value
* **ConsistentParentBasedSampler**: see https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/tracestate-probability-sampling.md#parentconsistentprobabilitybased-sampler
* **ConsistentProbabilityBasedSampler**: see https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/tracestate-probability-sampling.md#consistentprobabilitybased-sampler
* **ConsistentRateLimitingSampler**: a rate limiting sampler based on exponential smoothing that dynamically adjusts the sampling probability based on the estimated rate of spans occurring to satisfy a given rate of sampled spans
* **ConsistentReservoirSamplingBatchSpanProcessor**: consistent reservoir sampling of spans, see documentation in the source code

EDIT: ConsistentReservoirSamplingBatchSpanProcessor was removed to make this PR smaller as requested by @anuraaga. It can now be found in https://github.com/dynatrace-oss-contrib/opentelemetry-java-contrib/tree/consistent-reservoir-sampling.


consistent-sampling/build.gradle.kts
```diff
@@ -0,0 +1,13 @@
+plugins {
+  id("otel.java-conventions")
+  id("otel.publish-conventions")
+}
+
+description = "Sampler and exporter implementations for consistent sampling"
+
+dependencies {
+  api("io.opentelemetry:opentelemetry-sdk")
```
anuraaga, line 9:
```suggestion
  api("io.opentelemetry:opentelemetry-sdk-trace")
```

consistent-sampling/build.gradle.kts
```diff
@@ -0,0 +1,13 @@
...
+  testImplementation("com.google.guava:guava:31.0.1-jre")
```
anuraaga, line 10:
```suggestion
  testImplementation("com.google.guava:guava")
```

anuraaga: Though given it'd be very simple to replace the one usage with `HashMap` would recommend doing that instead

consistent-sampling/src/main/java/io/opentelemetry/contrib/export/ConsistentReservoirSamplingBatchSpanProcessor.java
```diff
@@ -0,0 +1,677 @@
...
+/**
+ * Implementation of the {@link SpanProcessor} that batches spans exported by the SDK then pushes
+ * them to the exporter pipeline.
+ *
+ * <p>All spans reported by the SDK implementation are first added to a synchronized queue (with a
+ * {@code maxQueueSize} maximum size, if queue is full spans are dropped). Spans are exported either
+ * when there are {@code maxExportBatchSize} pending spans or {@code scheduleDelayNanos} has passed
+ * since the last export finished.
+ */
+public final class ConsistentReservoirSamplingBatchSpanProcessor implements SpanProcessor {
```
anuraaga, line 49: Is this for tail sampling? It's not intuitive why we'd have a span processor doing sampling otherwise.

If possible I'd recommend holding off on this class until it's in the spec as documentation there would help a lot with understanding it, intuitively it seems unrelated to `ConsistentSampler`

consistent-sampling/src/main/java/io/opentelemetry/contrib/samplers/ConsistentAlwaysOffSampler.java
```diff
@@ -0,0 +1,23 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.contrib.samplers;
+
+import io.opentelemetry.contrib.state.OtelTraceState;
+import javax.annotation.concurrent.Immutable;
+
+@Immutable
+public final class ConsistentAlwaysOffSampler extends ConsistentSampler {
```
anuraaga, line 12: Our general pattern is to not make the implementations public and expose static factories on the `ConsistentSampler` itself

consistent-sampling/src/main/java/io/opentelemetry/contrib/samplers/ConsistentComposedAndSampler.java
```diff
@@ -0,0 +1,57 @@
...
+
+import static java.util.Objects.requireNonNull;
+
+import io.opentelemetry.contrib.state.OtelTraceState;
+import javax.annotation.concurrent.Immutable;
+
+/**
+ * A consistent sampler composes two consistent samplers.
+ *
+ * <p>This sampler samples if both samplers would sample.
```
anuraaga, line 16: Can you describe the difference in behavior vs composing the actual `SamplingResult` instead?

consistent-sampling/src/main/java/io/opentelemetry/contrib/samplers/ConsistentComposedOrSampler.java
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  @Override
+  protected int getP(int parentP, boolean isRoot) {
+    int p1 = sampler1.getP(parentP, isRoot);
+    int p2 = sampler2.getP(parentP, isRoot);
+    if (OtelTraceState.isValidP(p1)) {
+      if (OtelTraceState.isValidP(p2)) {
+        return Math.min(p1, p2);
+      } else {
```
anuraaga, line 49: Prefer to omit else statements when returning

consistent-sampling/src/main/java/io/opentelemetry/contrib/samplers/ConsistentParentBasedSampler.java
```diff
@@ -0,0 +1,62 @@
...
+package io.opentelemetry.contrib.samplers;
+
+import static java.util.Objects.requireNonNull;
+
+import io.opentelemetry.contrib.util.DefaultRandomGenerator;
+import io.opentelemetry.contrib.util.RandomGenerator;
+import javax.annotation.concurrent.Immutable;
+
+/**
+ * A consistent sampler that makes the same sampling decision as the paren and optionally falls back
```
anuraaga, line 15:
```suggestion
 * A consistent sampler that makes the same sampling decision as the parent and optionally falls back
```

consistent-sampling/src/main/java/io/opentelemetry/contrib/samplers/ConsistentProbabilityBasedSampler.java
```diff
@@ -0,0 +1,71 @@
...
+ */
+
+package io.opentelemetry.contrib.samplers;
+
+import io.opentelemetry.contrib.util.DefaultRandomGenerator;
+import io.opentelemetry.contrib.util.RandomGenerator;
+import javax.annotation.concurrent.Immutable;
+
+@Immutable
+public class ConsistentProbabilityBasedSampler extends ConsistentSampler {
```
anuraaga, line 13: Needs javadoc

consistent-sampling/src/main/java/io/opentelemetry/contrib/samplers/ConsistentSampler.java
```diff
@@ -0,0 +1,212 @@
...
+import io.opentelemetry.contrib.util.DefaultRandomGenerator;
+import io.opentelemetry.contrib.util.RandomGenerator;
+import io.opentelemetry.sdk.trace.data.LinkData;
+import io.opentelemetry.sdk.trace.samplers.Sampler;
+import io.opentelemetry.sdk.trace.samplers.SamplingDecision;
+import io.opentelemetry.sdk.trace.samplers.SamplingResult;
+import java.util.List;
+
+/** Abstract base class for consistent samplers. */
+abstract class ConsistentSampler implements Sampler {
```
anuraaga, line 26: We want to reduce public API as much as possible in libraries. We can probably make it so this is the only public class, which has factories returning implementations, and nothing else would need to be public from what I can tell.

anuraaga: Also favor composition over inheritance, so we should have a class that implements `Sampler`, containing a field that's a `ConsistentSampler` rather than talking through hierarchy

anuraaga (commented):
Did a very quick skim - 3700 lines is too huge though, let's start with a PR that only adds `ConsistentProbabilityBasedSampler` please

consistent-sampling/src/main/java/io/opentelemetry/contrib/samplers/ConsistentRateLimitingSampler.java
```diff
@@ -0,0 +1,137 @@
...
+    this.targetSpansPerNanosLimit = 1e-9 * targetSpansPerSecondLimit;
+
+    synchronized (this) {
+      this.effectiveWindowCount = 0;
+      this.effectiveWindowNanos = 0;
+      this.lastNanoTime = nanoTimeSupplier.getAsLong();
+    }
+  }
+
+  private synchronized double updateAndGetSamplingProbability() {
```
PeterF778, line 96: Any chances this algorithm can be adapted to support "worst-case" guarantee, such as a predefined confidence level that the specified rate will not be exceeded?

consistent-sampling/build.gradle.kts
oertl: fixed in 6b4ba10eb16ba8392945f1e59b852e9b4f6be637

consistent-sampling/build.gradle.kts
oertl: fixed in 6b4ba10eb16ba8392945f1e59b852e9b4f6be637

consistent-sampling/src/main/java/io/opentelemetry/contrib/samplers/ConsistentRateLimitingSampler.java
oertl: To calculate such a "confidence level", a model is needed that describes when the spans occur. The simplest model is probably the Poisson process, which assumes that the rate is constant and that all spans are independent. However, both assumptions are rarely true in practice. The rate is usually not constant, and the occurrence of spans is also not independent (consider, for example, a trace that always calls a service twice, so spans are always generated in pairs). 

We also need to specify how the rate limit is defined. Usually this is done in terms of a time interval. E.g., a maximum of 200 spans in each 1-minute interval.  

When everything is well defined, the model, the rate limit definition, and the adaptive sampling algorithm, the expected frequency of rate limit violations can be calculated or simulated. I assume this is what you are ultimately interested in?

PeterF778: Yes, I believe an algorithm that gives some kind of confidence level is a good compromise between a purely probabilistic approach and any algorithm based on "leaky bucket" approach. I believe it can be made almost-unbiased and almost-rate-limiting.
BTW, your algorithm, when faced with spans coming in pairs (at almost the same point of time) can reduce the sampling probability for the second span significantly.

oertl: > BTW, your algorithm, when faced with spans coming in pairs (at almost the same point of time) can reduce the sampling probability for the second span significantly.

Yes, this is true if the parameter for the adaptation time parameter is chosen small (in the order of magnitude of the time span between subsequent pairs or less).

This algorithm was motivated by a scenario where the span rate is suddenly increased. In this case, the algorithm immediately starts to decrease the sampling probability. In contrast, an adaptive sampler that adjusts the sampling rate periodically (e.g., based on the estimated rate of the previous period) would take much longer and the desired sampling rate would be exceeded much further.

consistent-sampling/src/main/java/io/opentelemetry/contrib/samplers/ConsistentComposedOrSampler.java
oertl: addressed in 000ffea29389658bab35f838b7cdd185e74bc97d

consistent-sampling/src/main/java/io/opentelemetry/contrib/samplers/ConsistentProbabilityBasedSampler.java
oertl: addressed in 9258ef4b029ba527179a7c1e6f991873849db5b5

consistent-sampling/src/main/java/io/opentelemetry/contrib/samplers/ConsistentAlwaysOffSampler.java
oertl: addressed in f7911c35c4c502b205493b02e10e5e441d47a00b

consistent-sampling/src/main/java/io/opentelemetry/contrib/samplers/ConsistentComposedAndSampler.java
oertl: please see added javadoc in b3459f9e79d1a69c17e4cdf7aeb811801b692b6b

consistent-sampling/src/main/java/io/opentelemetry/contrib/samplers/ConsistentSampler.java
oertl: > We want to reduce public API as much as possible in libraries. We can probably make it so this is the only public class, which has factories returning implementations, and nothing else would need to be public from what I can tell.

addressed in f7911c35c4c502b205493b02e10e5e441d47a00b

oertl: > Also favor composition over inheritance, so we should have a class that implements `Sampler`, containing a field that's a `ConsistentSampler` rather than talking through hierarchy

I think inheritance is appropriate in this case. First, `ConsistentSampler` is a `Sampler`. And secondly, the construction of a composed consistent sampler that is the AND/OR combination of 2 consistent samplers would be more difficult otherwise.

consistent-sampling/src/main/java/io/opentelemetry/contrib/export/ConsistentReservoirSamplingBatchSpanProcessor.java
oertl: > Is this for tail sampling? It's not intuitive why we'd have a span processor doing sampling otherwise.

No, this is different from tail sampling, where a sampling decision is made for the entire trace after all spans of a trace have been sampled. The ConsistentReservoirSamplingBatchSpanProcessor can be used to implement rate-limited sampling of spans with a hard limit. Since the sampler interface requires an immediate sampling decision, it is not straightforward to ensure that the rate limit is never exceeded, if the number of spans suddenly increases. For example, the ConsistentRateLimitingSampler may exceed the desired span rate in such cases. By delaying the sampling decision, as the ConsistentReservoirSamplingBatchSpanProcessor does, a hard limit can be achieved while giving all spans a fair chance to be sampled.

oertl: > If possible I'd recommend holding off on this class until it's in the spec as documentation there would help a lot with understanding it, intuitively it seems unrelated to ConsistentSampler

I have removed the ConsistentReservoirSamplingBatchSpanProcessor from this PR (see dd3df4ed697e968ee822228a92ad5aba58c32a6a)

consistent-sampling/src/main/java/io/opentelemetry/contrib/samplers/ConsistentRateLimitingSampler.java
```diff
@@ -0,0 +1,146 @@
...
+import java.util.function.LongSupplier;
+import javax.annotation.concurrent.Immutable;
+
+/**
+ * This consistent {@link Sampler} adjust the sampling probability dynamically to limit the rate of
+ * sampled spans.
+ *
+ * <p>This sampler uses exponential smoothing to estimate on irregular data (compare Wright, David
+ * J. "Forecasting data published at irregular time intervals using an extension of Holt's method."
+ * Management science 32.4 (1986): 499-510.) to estimate the current rate of spans.
```
spencerwilson, line 23: Looking at Wright's method (equations 2.5, 2.7 [here](http://quantitative.cz/wp-content/uploads/2018/09/improved-holt-method-for-irregular-time-series-tomas-hanzak-2008.pdf)), there seem to be some differences from what you've implemented. E.g., there are no `1 - decayFactor` terms in this implementation, whereas 1 - alpha appears in Wright. Is this an intentional deviation?

Maybe because the setting in OTel sampling is just different than that of forecasting? My understanding of this implementation is: we're trying to estimate a rate, so you need to maintain both an exponentially smoothed span count and an exponentially smoothed window duration in order for their ratio to be a meaningful: an exponential moving _rate_ (in analogy to exponential moving average). Remarkably, I can't find any literature on "exponential moving rate".

consistent-sampling/src/main/java/io/opentelemetry/contrib/state/OtelTraceState.java
```diff
@@ -0,0 +1,290 @@
...
+   * <p>If the string cannot be successfully parsed. A new OtelTraceState is returned
+   *
+   * @param ts the string
+   * @return the parsed OtelTraceState or a new empty OtelTraceState in case of parsing errors
+   */
+  public static OtelTraceState parse(@Nullable String ts) {
+    List<String> otherKeyValuePairs = null;
+    int p = INVALID_P;
+    int r = INVALID_R;
+    // boolean error = false;
```
jmacd, line 187:
```suggestion
```

consistent-sampling/src/main/java/io/opentelemetry/contrib/state/OtelTraceState.java
```diff
@@ -0,0 +1,290 @@
...
+
+    if (ts == null || ts.isEmpty()) {
+      return new OtelTraceState();
+    }
+
+    if (ts.length() > TRACE_STATE_SIZE_LIMIT) {
+      // error = true;
```
jmacd, line 194:
```suggestion
```

consistent-sampling/src/main/java/io/opentelemetry/contrib/state/OtelTraceState.java
```diff
@@ -0,0 +1,290 @@
...
+    while (true) {
+      int eqPos = tsStartPos;
+      for (; eqPos < len; eqPos++) {
+        char c = ts.charAt(eqPos);
+        if (!isLowerCaseAlpha(c) && (!isLowerCaseNum(c) || eqPos == tsStartPos)) {
+          break;
+        }
+      }
+      if (eqPos == tsStartPos || eqPos == len || ts.charAt(eqPos) != ':') {
+        // error = true;
```
jmacd, line 210:
```suggestion
```

consistent-sampling/src/main/java/io/opentelemetry/contrib/state/OtelTraceState.java
```diff
@@ -0,0 +1,290 @@
...
+        r = parseOneOrTwoDigitNumber(ts, eqPos + 1, sepPos, MAX_R, INVALID_R);
+      } else {
+        if (otherKeyValuePairs == null) {
+          otherKeyValuePairs = new ArrayList<>();
+        }
+        otherKeyValuePairs.add(ts.substring(tsStartPos, sepPos));
+      }
+
+      if (sepPos < len && ts.charAt(sepPos) != ';') {
+        // error = true;
```
jmacd, line 234:
```suggestion
```

jmacd (approved):

trask:
Hi @oertl! As a spec'd component (even if still experimental), I think a better place for this would be in the core java repo, under [sdk-extensions](https://github.com/open-telemetry/opentelemetry-java/tree/main/sdk-extensions) (e.g. alongside the jaeger remote sampler).

We can definitely merge it here first, but if we are planning to move it to the core repo it may be worth moving it sooner than later to avoid user confusion when the groupId changes in the move (from `io.opentelemetry.contrib` to `io.opentelemetry`).

@open-telemetry/java-maintainers what is your preference on how to proceed? I'm happy to support any path forward. It's not a big deal to move it later if you prefer for it to incubate here at least initially. We have renamed many many artifacts already 😭.

oertl:
Hi @trask! I would prefer to merge it here first, since some sampler implementations are not yet described in the spec.

consistent-sampling/src/main/java/io/opentelemetry/contrib/state/OtelTraceState.java
```diff
@@ -0,0 +1,286 @@
...
+  private static final int MAX_P = 63;
+  private static final int MAX_R = 62;
+  private static final int INVALID_P = -1;
+  private static final int INVALID_R = -1;
+  private static final int TRACE_STATE_SIZE_LIMIT = 256;
+
+  private int rval; // valid in the interval [0, MAX_R]
+  private int pval; // valid in the interval [0, MAX_P]
+
+  @Nullable private final List<String> otherKeyValuePairs;
```
trask, line 27: can this use empty list instead of null?

trask (commented):
thx @oertl!

I'm hoping you are able to be a component owner for this? if so, can you update https://github.com/open-telemetry/opentelemetry-java-contrib/blob/main/.github/component_owners.yml?

consistent-sampling/src/main/java/io/opentelemetry/contrib/state/OtelTraceState.java
```diff
@@ -0,0 +1,286 @@
...
+    }
+
+    int tsStartPos = 0;
+    int len = ts.length();
+
+    while (true) {
+      int eqPos = tsStartPos;
+      for (; eqPos < len; eqPos++) {
+        char c = ts.charAt(eqPos);
+        if (!isLowerCaseAlpha(c) && (!isLowerCaseNum(c) || eqPos == tsStartPos)) {
```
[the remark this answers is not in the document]
oertl: The proposed condition is not equivalent. However, I noticed that this condition also passes the unit tests. Therefore, I will add some more test cases, to make clear that the proposed condition is not an option.

consistent-sampling/src/main/java/io/opentelemetry/contrib/util/RandomUtil.java
```diff
@@ -0,0 +1,109 @@
...
+        return true;
+      }
+      boolean b = probability > 0.5;
+      if (randomBooleanSupplier.getAsBoolean()) {
+        return b;
+      }
+      probability += probability;
+      if (b) {
+        probability -= 1;
+      }
```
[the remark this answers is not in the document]
oertl: added some lines to the javadoc to explain this algorithm

consistent-sampling/src/main/java/io/opentelemetry/contrib/util/RandomGenerator.java
```diff
@@ -0,0 +1,89 @@
...
+   * <p>The implementation is based on Daniel Lemire's algorithm as described in "Fast random
+   * integer generation in an interval." ACM Transactions on Modeling and Computer Simulation
+   * (TOMACS) 29.1 (2019): 3.
+   *
+   * @param bound the upper bound (exclusive) for the returned value. Must be positive.
+   * @return a pseudorandomly chosen {@code int} value between zero (inclusive) and the bound
+   *     (exclusive)
+   * @throws IllegalArgumentException if {@code bound} is not positive
+   */
+  default int nextInt(int bound) {
```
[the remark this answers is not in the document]
oertl: addressed in edac8a9a517ed79d772d80b758bc56a6e5d7397a

consistent-sampling/src/main/java/io/opentelemetry/contrib/util/RandomUtil.java
```diff
@@ -0,0 +1,109 @@
...
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.contrib.util;
+
+import java.util.BitSet;
+import java.util.function.BooleanSupplier;
+
+public final class RandomUtil {
```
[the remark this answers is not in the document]
oertl: addressed in edac8a9a517ed79d772d80b758bc56a6e5d7397a

consistent-sampling/src/main/java/io/opentelemetry/contrib/util/RandomGenerator.java
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  /**
+   * Returns a pseudorandomly chosen {@code boolean} value where the probability of returning {@code
+   * true} is predefined.
+   *
+   * @param probability the probability of returning {@code true}
+   * @return a random {@code boolean}
+   */
+  default boolean nextBoolean(double probability) {
```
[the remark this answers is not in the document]
oertl: addressed in edac8a9a517ed79d772d80b758bc56a6e5d7397a

oertl:
@trask, thank you very much for your valuable suggestions! I hope I have not overlooked anything and have addressed all your points.

trask (approved):
@open-telemetry/java-contrib-approvers let me know if anyone is planning to review this further, otherwise I will merge before making the upcoming release, thx!

anuraaga (commented):
It looks like `RandomGenerator` can be moved into the `samplers` package and made package private? We can have package-private factories accepting it for use in unit tests only in this repo, but let's keep the public API using just a default threadlocalrandom based generator without user configuration.

For `OtelTraceState` it also looks like it can be made package private. If you expect other packages needing the utility methods, then it's also ok to move into `io.opentelemetry.contrib.samplers.internal` but suspect we don't need to do more than package private at least at this time.

Note that we cannot have packages present in multiple artifacts in our codebase. So currently the generic sounding packages `io.opentelemetry.contrib.state` and `io.opentelemetry.contrib.util` could cause problems in the future. The above internalizing suggestions will hopefully fix that naturally.

anuraaga (approved):
Thanks!

oertl:
@anuraaga, thanks for your suggestions. I addressed them in da4201893e6ea2674fb4b4f5f7a5337de8d1d23f

trask:
thanks @oertl 🎉

PR #227, opened 2022-02-07
Move to counters

kittylyst:
**Description:**

Change impl to match current version of specification

**Existing Issue(s):**

< Link any applicable issues. >

**Testing:**

Tests continue to pass.

**Outstanding items:**

Initial PR to confirm approach - more to follow


trask:
@roberttoyonaga can you check if there's anything needed from this draft PR, or if it's ok to close? thx!

roberttoyonaga:
@trask There are a few things that might be worth keeping but probably need more discussion before actually including.

1. This PR replaces the jvm cpu usage histogram with a counter that has summary statistics attributes such as max/mean (previously there were no summary attributes). I'm not sure this is really needed because whatever ends up consuming the jvm cpu usage metric can probably generate these stats. Although maybe we want to do this for convenience? 
2. A similar replacement is done for the process.runtime.jvm.memory metrics
3. GC pause duration metric gathering implementation was started. I don't think this was finished though. 

Overall, all the changes here affect JFR metric gatherers that duplicate JMX ones. As per [Jack's comment here](https://github.com/open-telemetry/opentelemetry-specification/issues/3416#issuecomment-1527807713), I think the plan is to shelve replicating the JMX metric gathers with JFR until we actually need to. So,  I think we can worry about discussing these changes later (if we end up deciding to replace the JMX gatherers with JFR) and close this.

trask:
thanks @roberttoyonaga!

PR #232, opened 2022-02-11
Updates jmx-metrics WildFly integration with replacement MBean

StefanKurek:
**Description:**

Updates jmx-metrics WildFly integration to point to integer attributes for some metrics. The session metrics previously pointed at attributes that were Strings and caused errors.

### DIFF
--- a/jmx-metrics/src/main/resources/target-systems/wildfly.groovy
+++ b/jmx-metrics/src/main/resources/target-systems/wildfly.groovy
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-def beanWildflyDeployment = otel.mbeans("jboss.as.expr:deployment=*,subsystem=undertow")
+def beanWildflyDeployment = otel.mbeans("jboss.as:deployment=*,subsystem=undertow")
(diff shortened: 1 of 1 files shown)


StefanKurek:
@rmfitzpatrick FYI this is a small fix which points to an MBean which has integer versions of attributes compared to the old MBean (the String attributes were causing issues).

dehaansa (approved):

rmfitzpatrick:
@StefanKurek have you considered a straightforward update to the integration test that would have detected this?

StefanKurek:
> @StefanKurek have you considered a straightforward update to the integration test that would have detected this?

@rmfitzpatrick it's a good question. The initial implementation for this left out tests for these specific metrics because I believe it involved a more complicated setup of WildFly in order for them to come through (possibly installing an application jar/war which I don't think WildFly ships with any test examples). I can take a look and see if I can come up with something though.

rmfitzpatrick (approved):

StefanKurek:
@rmfitzpatrick not sure how to contact you other than through this. would you be willing to sponsor my otel membership?

Issue #156, opened 2021-12-21
Maven plugin for static instrumentation using Java agent


kubawach:
Dear @open-telemetry/java-contrib-maintainers I'd like to contribute a Maven plugin, capable of ahead-of-time Java application instrumentation using Java OTel agent (aka "static instrumentation").

The goal of the project is:
- implement proof of concept how existing agent can be integrated into the Maven build cycle to instrument applications during build time
- provide a solution for all use cases where dynamic instrumentation is not feasible (primary example - AWS lambda suffering cold start performance issues with dynamic instrumentation)
- measure startup performance dynamic vs static proving the point

Future goal within the community would be to improve the plugin from PoC into real product shape.

The code has been developed by a group of AGH (university in Krakow, Poland) students for their BSc thesis project (3rd year of studies), with active help from Splunk side.

The code can be found here: https://github.com/agh-static-x/Maven-Plugin

kenfinnigan:
Could you provide details on how it differs from https://github.com/open-telemetry/opentelemetry-java-contrib/tree/main/maven-extension ?

FYI @cyrille-leclerc

anuraaga:
I have a feeling this is applying the instrumentation to the actual app at compile time, not measuring the time of the build. Yes we're very interested in this :-) Will try to take a look through the linked repo but can't guarantee for this year

cyrille-leclerc:
I have the same feeling as @anuraaga . This seems to be a proposal to introduce a capability/plugin in Maven to inject Otel instrumentation in applications produced by Maven builds when the Maven Extension is instrumenting Maven itself to observe & monitor the builds but not is not modifying the produced artifacts.

kubawach:
@anuraaga you are right - it's a M2 build plugin to enhance the app with OpenTelemetry. 
@kenfinnigan updated the description a bit to make it a bit clearer :)

kenfinnigan:
Thanks for updating the description @kubawach!

As I understand it then, it's adding the instrumentation to the application at build to remove the need for the agent to be attached to the VM at runtime? That's very interesting.

I'd imagine there are benefits to this approach for constructing native executables with GraalVM, as the agent doesn't need to be explicitly added to the build as the code is already present.

anuraaga:
@kubawach Took a scan of the code - it looks like the major interface into instrumentation is this class which doesn't depend on Maven

[...]

PR #236, opened 2022-02-28
Static instrumenter - core

kubawach:
**Description:**
Core part of the static instrumenter - command line tool enabling modified jar copy.

**Existing Issue(s):**

Part of #156

**Testing:**

- minimal unit tests (more to be added with coming PRs)

**Documentation:**

README updated

**Outstanding items:**

- more test
- agent instrumentation


kubawach:
@java-contrib-approvers I'd love to get some feedback on this PR

static-instrumenter/agent-instrumenter/build.gradle.kts
```diff
@@ -3,3 +3,14 @@ plugins {
 }
 
 description = "OpenTelemetry Java Static Instrumentation Agent"
+
+repositories {
+  mavenLocal()
+  mavenCentral()
+}
+
```
anuraaga, line 11:
```suggestion
```

static-instrumenter/agent-instrumenter/build.gradle.kts
```diff
@@ -3,3 +3,14 @@ plugins {
...
+dependencies {
+  implementation("org.slf4j:slf4j-api")
+  implementation("org.slf4j:slf4j-simple")
+  testImplementation("org.mockito:mockito-core:3.+")
```
anuraaga, line 15:
```suggestion
```

static-instrumenter/agent-instrumenter/build.gradle.kts
```diff
@@ -3,3 +3,14 @@ plugins {
...
 description = "OpenTelemetry Java Static Instrumentation Agent"
+
+repositories {
+  mavenLocal()
+  mavenCentral()
+}
+
+dependencies {
+  implementation("org.slf4j:slf4j-api")
+  implementation("org.slf4j:slf4j-simple")
```
anuraaga, line 14:
```suggestion
  runtimeOnly("org.slf4j:slf4j-simple")
```

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/statical/instrumenter/ArchiveEntry.java
```diff
@@ -0,0 +1,60 @@
...
+  private final boolean shouldBeSkipped;
+
+  private ArchiveEntry(String className, boolean isClass, boolean shouldBeSkipped) {
+    this.className = className;
+    this.classPath = className.replace(".", "/");
+    this.isClass = isClass;
+    this.shouldBeSkipped = shouldBeSkipped;
+  }
+
+  public static ArchiveEntry ofFilePath(String path) {
```
anuraaga, line 25: Let's remove `public` for methods on non-public classes

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/statical/instrumenter/ClassArchive.java
```diff
@@ -0,0 +1,99 @@
...
+  private InputStream getInputStreamForEntry(JarEntry inEntry, ZipEntry outEntry)
+      throws IOException {
+
+    InputStream entryIn = null;
+    ArchiveEntry entry = ArchiveEntry.ofFilePath(inEntry.getName());
+    if (entry.isClass() && !entry.shouldBeSkipped()) {
+      String className = entry.getClassName();
+      try {
+        // sanity check
+        Class.forName(className, false, ClassLoader.getSystemClassLoader());
```
anuraaga, line 80: I wouldn't expect an archive being processed to have its classes on the system classpath. We should load classes into a separate classloader doing the transformation, or some dependencies of this tool could conflict with the transformed archive. We're probably not that far in this PR, but I would remove this sanity check in anticipation of that

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/statical/instrumenter/Main.java
```diff
@@ -0,0 +1,157 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.jar.JarFile;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipOutputStream;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class Main {
+
+  private static final String PATH_SEPARATROR = System.getProperty("path.separator");
```
anuraaga, line 24: Replace with `File.pathSeparator`

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/statical/instrumenter/Main.java
```diff
@@ -0,0 +1,157 @@
...
+  private static final Logger logger = LoggerFactory.getLogger(Main.class);
+
+  private static final Main INSTANCE = new Main(new ClassArchive.Factory() {});
```
anuraaga, line 27: Don't think we need this singleton, pass in dependencies to constructors appropriately

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/statical/instrumenter/StreamUtils.java
```diff
@@ -0,0 +1,24 @@
...
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+
+final class StreamUtils {
+
+  private StreamUtils() {}
+
+  public static void copy(InputStream in, OutputStream out) throws IOException {
```
anuraaga, line 16: Unless anyone knows people with very legacy build systems will want to use this tool, I would target Java 11 for this tool at least. I'd be in favor of Java 17 to "move the ecosystem forward" too if other maintainers are comfortable with it, but Java 8 seems a bit much.

i.e., we would be able to use `transferTo`

static-instrumenter/agent-instrumenter/src/test/java/io/opentelemetry/contrib/statical/instrumenter/ClassArchiveTest.java
```diff
@@ -0,0 +1,46 @@
...
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.util.Collections;
+import java.util.Map;
+import java.util.jar.JarFile;
+import java.util.zip.ZipOutputStream;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+public class ClassArchiveTest {
```
anuraaga, line 18: Since `jupiter`, conventional to avoid `public` in tests

static-instrumenter/agent-instrumenter/src/test/java/io/opentelemetry/contrib/statical/instrumenter/JarTestUtil.java
```diff
@@ -0,0 +1,65 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.contrib.statical.instrumenter;
+
+import static org.junit.jupiter.api.Assertions.assertNotNull;
```
anuraaga, line 8: Use assertj

anuraaga (commented):
Did a first pass

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/statical/instrumenter/ArchiveEntry.java
kubawach: Done!

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/statical/instrumenter/Main.java
kubawach: Ahh.. thanks!

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/statical/instrumenter/StreamUtils.java
kubawach: Yeah, I sticked to Java8 as it's general OTEL compat... I'd be in favour of going to at least 11 if possible. @open-telemetry/java-contrib-maintainers what do you say here?

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/statical/instrumenter/Main.java
kubawach: I think we need it - the other part of the instrumentor runs in the java agent space and needs to store processed bytecode in the Main. So it's either full static or singleton and I opted for the latter. Makes sense for you? :)

static-instrumenter/agent-instrumenter/src/test/java/io/opentelemetry/contrib/statical/instrumenter/JarTestUtil.java
kubawach: Frankly I don't see a benefit of assertJ here as assertions are trivial.

kubawach:
> Did a first pass

Thanks a lot! Looking forward to merging this and following up with the rest of the code :)

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/statical/instrumenter/ClassArchive.java
kubawach: The basic idea was to run the tool in both java agent space (incoming part) and main space (this part) with all of the archives we need to instrument on a classpath. I can remove the sanity although initially it was helpful in debugging the M2 plugin (checking what exactly didn't end up on the instruemntors class path).

anuraaga: I think it would still be better to have the archives needed to instrument on its own classloader, not the system one, to ensure no conflicts with the tool. I don't think that would affect the operation of teh javaagent

static-instrumenter/agent-instrumenter/src/test/java/io/opentelemetry/contrib/statical/instrumenter/JarTestUtil.java
anuraaga: It's to be consistent in our codebase, we always use assertj regardless of the complexity of the assertions

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/statical/instrumenter/Main.java
anuraaga: Hmm - I can imagine the opposite should also be possible, for the javaagent space to store the bytecode and Main reads them out when writing. Perhaps we can fix that when adding th elogic though.

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/statical/instrumenter/StreamUtils.java
trask: I don't mind moving to Java 11 for this component. We can always change course in the future if we see significant user pushback (we have no users of it so far, so it's a bit hard to judge).

I think Java 17 is definitely too soon, given how much work it is for some projects to update their builds to Java 17 😉

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/statical/instrumenter/Main.java
kubawach: Agreed :)

static-instrumenter/agent-instrumenter/src/test/java/io/opentelemetry/contrib/statical/instrumenter/JarTestUtil.java
kubawach: OK, done.

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/statical/instrumenter/ClassArchive.java
kubawach: Sanity removed.

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/statical/instrumenter/Main.java
```diff
@@ -0,0 +1,157 @@
...
+  private static ZipOutputStream outJarFor(File outDir, File inFile) throws FileNotFoundException {
+    File outFile = new File(outDir, inFile.getName());
+    return new ZipOutputStream(new FileOutputStream(outFile));
+  }
+
+  // FIXME: only relevant additional classes should be injected
+  private void injectAdditionalClassesTo(ZipOutputStream outJar) throws IOException {
+    for (String className : additionalClasses.keySet()) {
+      byte[] classData = additionalClasses.get(className);
+
```
mateuszrzeszutek, line 129: Nit: prefer using `entrySet()` or `Map#forEach()` when iterating over a map

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/statical/instrumenter/Main.java
```diff
@@ -0,0 +1,157 @@
...
+      ZipEntry newEntry = new ZipEntry(className);
+      outJar.putNextEntry(newEntry);
+      if (classData != null) {
+        newEntry.setSize(classData.length);
+        StreamUtils.copy(new ByteArrayInputStream(classData), outJar);
+      }
+      outJar.closeEntry();
+
+      logger.debug("Additional class added: " + className);
```
mateuszrzeszutek, line 138:
```suggestion
      logger.debug("Additional class added: {}", className);
```

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/statical/instrumenter/ClassArchive.java
```diff
@@ -0,0 +1,99 @@
...
+        inEntryName.endsWith(".jar") ? new ZipEntry(inEntry) : new ZipEntry(inEntryName);
+
+    try (InputStream entryInputStream = getInputStreamForEntry(inEntry, outEntry)) {
+      try {
+        outJar.putNextEntry(outEntry);
+        StreamUtils.copy(entryInputStream, outJar);
+        outJar.closeEntry();
+      } catch (ZipException e) {
+        if (!isEntryDuplicate(e)) {
+          logger.error("Error while creating entry: " + outEntry.getName(), e);
```
mateuszrzeszutek, line 60:
```suggestion
          logger.error("Error while creating entry: {}", outEntry.getName(), e);
```

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/statical/instrumenter/ClassArchive.java
```diff
@@ -0,0 +1,99 @@
...
+      String className = entry.getClassName();
+      try {
+        // sanity check
+        Class.forName(className, false, ClassLoader.getSystemClassLoader());
+        byte[] modified = instrumentedClasses.get(entry.getClassPath());
+
+        if (modified != null) {
+          logger.debug("Found instrumented class: " + className);
+          entryIn = new ByteArrayInputStream(modified);
+          outEntry.setSize(modified.length);
```
mateuszrzeszutek, line 86: The size here is only set when instrumented class is encountered - what about other classes, or non-class resources? Is this line needed at all?

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/statical/instrumenter/StreamUtils.java
kubawach: Done!

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/statical/instrumenter/Main.java
kubawach: +1

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/statical/instrumenter/ClassArchive.java
kubawach: It will be default (-1 = unknown). I guess it's better to set it if easily available as in this case.

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/statical/instrumenter/ClassArchive.java
```diff
@@ -0,0 +1,87 @@
...
+    if (entry.isClass() && !entry.shouldBeSkipped()) {
+      String className = entry.getClassName();
+      byte[] modified = instrumentedClasses.get(entry.getClassPath());
+      if (modified != null) {
+        logger.debug("Found instrumented class: " + className);
+        entryIn = new ByteArrayInputStream(modified);
+        outEntry.setSize(modified.length);
+      }
+    }
+    return (entryIn == null ? source.getInputStream(inEntry) : entryIn);
```
trask, line 85: undocumented style convention in the otel java repos

```suggestion
    return entryIn == null ? source.getInputStream(inEntry) : entryIn;
```

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/statical/instrumenter/ClassArchive.java
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  private InputStream getInputStreamForEntry(JarEntry inEntry, ZipEntry outEntry)
+      throws IOException {
+
+    InputStream entryIn = null;
+    ArchiveEntry entry = ArchiveEntry.ofFilePath(inEntry.getName());
+    if (entry.isClass() && !entry.shouldBeSkipped()) {
+      String className = entry.getClassName();
+      byte[] modified = instrumentedClasses.get(entry.getClassPath());
```
trask, line 78: unless there are future plans for `ArchiveEntry`, I think it would be clearer to delete it and instead add a few static methods to this class

```suggestion
    String path = inEntry.getName();
    if (isClass(path) && shouldProcess(path)) {
      String className = getClassName(path);
      byte[] modified = instrumentedClasses.get(path);
```

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/statical/instrumenter/ArchiveEntry.java
```diff
@@ -0,0 +1,60 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.contrib.statical.instrumenter;
```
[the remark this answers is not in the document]
kubawach: Makes sense. Initially I wanted to keep this separated for readability but then "statical" package will be always empty (apart from subpackages that is).

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/statical/instrumenter/Main.java
```diff
@@ -0,0 +1,157 @@
...
+    for (String pathItem : jarsList) {
+      logger.info("Classpath item processed: " + pathItem);
+      if (shouldProcessItem(pathItem)) {
+        saveJarTo(new File(pathItem), outDir);
+      }
+    }
+  }
+
+  private static boolean shouldProcessItem(String pathItem) {
+    return (pathItem.endsWith(".jar") || pathItem.endsWith(".war"));
```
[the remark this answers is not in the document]
Enkelian: We haven't taken this format into consideration during plugin development. I think it's best to focus on (fat) jars and war (since all our tests so far were performed for these formats) and leave ears for the future

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/statical/instrumenter/ClassArchive.java
```diff
@@ -0,0 +1,87 @@
...
+        if (!isEntryDuplicate(e)) {
+          logger.error("Error while creating entry: {}", outEntry.getName(), e);
+          throw e;
+        }
+      }
+    }
+  }
+
+  private static boolean isEntryDuplicate(ZipException ze) {
+    return (ze.getMessage() != null && ze.getMessage().contains("duplicate"));
```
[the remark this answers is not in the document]
kubawach: I'd prefer to keep it with master parentheses (personal style so to speak).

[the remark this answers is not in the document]
kubawach: Ok, just read another comment that this is THE style used in other Java repos ;) Will change that then.

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/statical/instrumenter/ClassArchive.java
```diff
@@ -0,0 +1,87 @@
...
+    Enumeration<JarEntry> inEntries = source.entries();
+    while (inEntries.hasMoreElements()) {
+      copyEntry(inEntries.nextElement(), outJar);
+    }
+  }
+
+  private void copyEntry(JarEntry inEntry, ZipOutputStream outJar) throws IOException {
+    String inEntryName = inEntry.getName();
+    ZipEntry outEntry =
+        inEntryName.endsWith(".jar") ? new ZipEntry(inEntry) : new ZipEntry(inEntryName);
```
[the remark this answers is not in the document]
kubawach: To copy JAR's attributes (ZIP ones - CRC, method, size, ...).

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/statical/instrumenter/ClassArchive.java
kubawach: The idea is to keep here additional logic, especially pertaining to skipping. Currently it's quite basic - skipping only opentelemetry entries.

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/statical/instrumenter/Main.java
```diff
@@ -0,0 +1,157 @@
...
+   *
+   * @param outDir directory where jars will be written
+   * @throws IOException in case of file operation problem
+   */
+  public void saveTransformedJarsTo(String[] jarsList, File outDir) throws IOException {
+
+    for (String pathItem : jarsList) {
+      logger.info("Classpath item processed: " + pathItem);
+      if (shouldProcessItem(pathItem)) {
+        saveJarTo(new File(pathItem), outDir);
```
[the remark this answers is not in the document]
kubawach: Sure thing!

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/statical/instrumenter/Main.java
```diff
@@ -0,0 +1,157 @@
...
+  private void injectAdditionalClassesTo(ZipOutputStream outJar) throws IOException {
+    for (Map.Entry<String, byte[]> clazz : additionalClasses.entrySet()) {
+      String className = clazz.getKey();
+      byte[] classData = clazz.getValue();
+
+      ZipEntry newEntry = new ZipEntry(className);
+      outJar.putNextEntry(newEntry);
+      if (classData != null) {
+        newEntry.setSize(classData.length);
+        new ByteArrayInputStream(classData).transferTo(outJar);
```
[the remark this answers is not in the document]
kubawach: +1

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/statical/instrumenter/ClassArchive.java
Enkelian: I'd also vote for keeping `ArchiveEntry` as it is

kubawach:
Another round done :)

static-instrumenter/README.md
```diff
@@ -6,7 +6,7 @@
 
 Module enhancing OpenTelemetry Java Agent for static instrumentation. The modified agent is capable of instrumenting and saving a new JAR with all relevant instrumentations applied and necessary helper class-code included.
 
-In order to statically instrument a JAR, modified agent needs to be both attached (`-javaagent:`) and run as the main method (`io.opentelemetry.javaagent.StaticInstrumenter` class).
+In order to statically instrument a JAR, modified agent needs to be both attached (`-javaagent:`) and run as the main method (`io.opentelemetry.contrib.statical.instrumenter.Main` class).
```
trask, line 9:
```suggestion
In order to statically instrument a JAR, modified agent needs to be both attached (`-javaagent:`) and run as the main method (`io.opentelemetry.contrib.staticinstrumenter.Main` class).
```

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/ArchiveEntry.java
```diff
@@ -0,0 +1,60 @@
...
+  private final boolean shouldBeSkipped;
+
+  private ArchiveEntry(String className, boolean isClass, boolean shouldBeSkipped) {
+    this.className = className;
+    this.classPath = className.replace(".", "/");
+    this.isClass = isClass;
+    this.shouldBeSkipped = shouldBeSkipped;
+  }
+
+  static ArchiveEntry ofFilePath(String path) {
```
trask, line 25:
```suggestion
  static ArchiveEntry fromPath(String path) {
```

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/ArchiveEntry.java
```diff
@@ -0,0 +1,60 @@
...
+
+package io.opentelemetry.contrib.staticinstrumenter;
+
+class ArchiveEntry {
+
+  private static final ArchiveEntry NOT_CLASS =
+      new ArchiveEntry("", /* isClass= */ false, /* shouldBeSkipped= */ true);
+
+  private final String className;
+  private final String classPath;
```
trask, line 14: `classPath` makes me think "classpath" 😄
```suggestion
  private final String path;
```

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/ClassArchive.java
```diff
@@ -0,0 +1,87 @@
...
+  private static boolean isEntryDuplicate(ZipException ze) {
+    return ze.getMessage() != null && ze.getMessage().contains("duplicate");
+  }
+
+  private InputStream getInputStreamForEntry(JarEntry inEntry, ZipEntry outEntry)
+      throws IOException {
+
+    InputStream entryIn = null;
+    ArchiveEntry entry = ArchiveEntry.ofFilePath(inEntry.getName());
+    if (entry.isClass() && !entry.shouldBeSkipped()) {
```
trask, line 76: then I think can simplify ArchiveEntry a bit (don't need the `isClass` field anymore)?

```suggestion
    if (entry.shouldInstrument()) {
```

trask (commented):
a couple suggestions around `ArchiveFile` (since keeping it 😄)

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/ArchiveEntry.java
anuraaga: Also consider using `Path` instead of `String`

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/ArchiveEntry.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+    return new ArchiveEntry(
+        className(path), /* isClass= */ true, /* shouldBeSkipped= */ shouldBeSkipped(path));
+  }
+
+  String getClassName() {
+    return className;
+  }
+
+  private static String className(String path) {
```
anuraaga, line 37: Let's order all the private at the bottom to make it easier to skim the API of the class

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/ArchiveEntry.java
```diff
@@ -0,0 +1,60 @@
...
+package io.opentelemetry.contrib.staticinstrumenter;
+
+class ArchiveEntry {
+
+  private static final ArchiveEntry NOT_CLASS =
+      new ArchiveEntry("", /* isClass= */ false, /* shouldBeSkipped= */ true);
+
+  private final String className;
+  private final String classPath;
+  private final boolean isClass;
```
anuraaga, line 15: Storing this as a field seems like overkill (if `isClass()` is called 10 times per entry, maybe, if 1-3 times per entry, not such a big deal)

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/ClassArchive.java
```diff
@@ -0,0 +1,87 @@
...
+
+  private final JarFile source;
+  private final Map<String, byte[]> instrumentedClasses;
+
+  ClassArchive(JarFile source, Map<String, byte[]> instrumentedClasses) {
+    this.source = source;
+    this.instrumentedClasses = instrumentedClasses;
+  }
+
+  void copyAllClassesTo(ZipOutputStream outJar) throws IOException {
```
anuraaga, line 40: Not sure but since the variable name is `outJar`, should this be `JarOutputStream`? Applies throughout PR

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/ClassArchive.java
```diff
@@ -0,0 +1,87 @@
...
+    String inEntryName = inEntry.getName();
+    ZipEntry outEntry =
+        inEntryName.endsWith(".jar") ? new ZipEntry(inEntry) : new ZipEntry(inEntryName);
+
+    try (InputStream entryInputStream = getInputStreamForEntry(inEntry, outEntry)) {
+      try {
+        outJar.putNextEntry(outEntry);
+        entryInputStream.transferTo(outJar);
+        outJar.closeEntry();
+      } catch (ZipException e) {
```
anuraaga, line 58: It looks like this catch can be applied to the outer try instead of a second try

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/Main.java
```diff
@@ -0,0 +1,156 @@
...
+  private static final Logger logger = LoggerFactory.getLogger(Main.class);
+
+  private static final Main INSTANCE = new Main(new ClassArchive.Factory() {});
+
+  private final ClassArchive.Factory classArchiveFactory;
+
+  private final Map<String, byte[]> instrumentedClasses = new ConcurrentHashMap<>();
+  private final Map<String, byte[]> additionalClasses = new ConcurrentHashMap<>();
+
+  @SuppressWarnings("ThreadLocalUsage")
```
anuraaga, line 32: Think we may as well just make it static

static-instrumenter/agent-instrumenter/src/test/java/io/opentelemetry/contrib/staticinstrumenter/ClassArchiveTest.java
```diff
@@ -0,0 +1,46 @@
...
+
+  @TempDir private File tempDir;
+
+  @Test
+  void shouldCopyAllClasses(@TempDir File destination) throws IOException {
+
+    // given
+    byte[] newTransformed = new byte[] {(byte) 0xCA, (byte) 0xFE, (byte) 0xBA, (byte) 0xBE};
+    Map<String, byte[]> transformed = Collections.singletonMap("test/TestClass", newTransformed);
+    String jar = JarTestUtil.createJar(tempDir, "one.jar", "test/TestClass.class", "second.class");
```
anuraaga, line 28: It probably makes sense to commit JAR files into `resources` for these tests instead of slowing down the tests by creating the same JAR every time

static-instrumenter/agent-instrumenter/src/test/java/io/opentelemetry/contrib/staticinstrumenter/ClassArchiveTest.java
```diff
@@ -0,0 +1,46 @@
...
+    ClassArchive underTest = new ClassArchive(new JarFile(jar), transformed);
+
+    // when
+    File jarOut = new File(destination, "one.jar");
```
anuraaga, line 32: Let's use a different filename for the input and output jar to reduce confusion

static-instrumenter/agent-instrumenter/src/test/java/io/opentelemetry/contrib/staticinstrumenter/MainTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+    // given
+    ClassArchive.Factory factory = mock(ClassArchive.Factory.class);
+    ClassArchive mockArchive = mock(ClassArchive.class);
+    given(factory.createFor(any(), anyMap())).willReturn(mockArchive);
+    Main underTest = new Main(factory);
+    underTest.getAdditionalClasses().put("additionalOne", new byte[0]);
+    underTest.getAdditionalClasses().put("additionalTwo", new byte[0]);
+    String[] jarsList = {
+      JarTestUtil.createJar(tempDir, "one.jar", "first", "second"),
```
anuraaga, line 35: Ditto

static-instrumenter/README.md
kubawach: +1

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/PostTransformer.java
```diff
@@ -0,0 +1,32 @@
...
+  @Override
+  @Nullable
+  public byte[] transform(
+      ClassLoader loader,
+      String className,
+      Class<?> classBeingRedefined,
+      ProtectionDomain protectionDomain,
+      byte[] classfileBuffer) {
+
+    TransformedClass pre = Main.getInstance().getCurrentClass();
```
mateuszrzeszutek, line 23: I guess we don't know yet how these transformers will be used, but maybe in one of the future PRs it'll be possible to pass the state in the constructor instead of a static singleton?

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/ClassArchive.java
kubawach: Done.

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/ArchiveEntry.java
kubawach: Ordered now in order of apperance for better readability.

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/ArchiveEntry.java
kubawach: Done. Not using path as later on the path is used as instrumented classes key (and does not contain file extension).

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/ArchiveEntry.java
kubawach: Changed to "fromZipEntryName" to reflect what's expected as an input.

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/Main.java
kubawach: It's tied to singleton, so implicitely static. I added the suppression as the tool is not smart enough to catch that.

static-instrumenter/agent-instrumenter/src/test/java/io/opentelemetry/contrib/staticinstrumenter/ClassArchiveTest.java
kubawach: The idea here was that under normal conditions out jars will have the same name as input jars (just other folder). But I understand your point.

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/ClassArchive.java
kubawach: True :)

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/ClassArchive.java
kubawach: Interesting point! I read thru std JarOutputStream implementation and it should not make any difference, as we're copying JAR entries so magic number will be in the first entry.
There is also slight inconsistency - the tool is supposed to support both JAR and WAR format (hence: class archive) BUT there is no java model for writing WARs (only JARs). I believe then in next PRs we should switch to using ZIP (file, os) directly.

kubawach:
Another round done.

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/Main.java
anuraaga: Yeah since it's effectively static by being a singleton we may as well declare it static and pass the tool right? There's a reasonable chance of changing the dependency injection strategy in the future while no chance of needing a non-static threadlocal regardless - the tool wins.

kubawach: You lost me here. If we make the currentClass static we need to resign from singleton entirely and make whole code static, right?

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/ClassArchive.java
trask: war files are "jar" files, so shouldn't be a problem to read them with JarOutputStream

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/Main.java
trask: it's rare that ThreadLocals need to be instance fields (thus the errorprone warning) since they are really tied to the thread

the only reason you would need it to be instance field is if you needed to track multiple "current" classes at the same time across different instances of the containing class

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/PostTransformer.java
trask, line 23: can this be `getAndRemoveCurrentClass()`? it's nice to not leak the thread local beyond its scope

trask (approved):
thx @kubawach, looking forward to the next PR!

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/PostTransformer.java
kubawach: 1. Switched do JarOS everywhere.
2. I believe that we can remove after get so implemented as suggested.
3. The TL instance is virtually static (singleton). The original code implemented all methods (and fields ofc) static, but I hated that and for better testability moved to singleton (which isn't my fav design pattern either ;) ).

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/Main.java
kubawach: Singleton was used only to avoid declaring all methods static and using Main in a purely procedural style. From runtime point of view I believe these are equal.

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/ClassArchive.java
kubawach: Done :)

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/Main.java
anuraaga: They should be basically same at runtime so we'd like to use static here to reduce exceptions from the tooling

static-instrumenter/agent-instrumenter/src/test/java/io/opentelemetry/contrib/staticinstrumenter/ClassArchiveTest.java
anuraaga: Reminder of this comment

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/Main.java
kubawach: Right, can change that, although from usage point of view (see PostTransformer) it will look strange:
- get class using static call to Main
- add it to the list kept by Main singleton (instance call)
In my opinion as from runtime point of view there are no differences (singleton vs static) readability wins, so it's better to keep all Main methods instance. Yet, if you feel strong about making field and access methods to currentClass static, will do that (just say that you do ;) ).

static-instrumenter/agent-instrumenter/src/test/java/io/opentelemetry/contrib/staticinstrumenter/ClassArchiveTest.java
kubawach: Yep, another point where I would prefer to leave things as they're now. My reasons:
1. Execution time for tests in contrib is not that important (especially diff made by creating several files on disk)
2. If the JARs are committed, list of classes (that needs to be verified after copy) will need to be hardcoded - tests will be harder to follow (understand)
3. I'm used to creating temp files in tests as a good testing practice

anuraaga: FWIW, I found the current pattern to be hard to understand. Since the code's main job is to create a JAR, it was surprising to see the test also creating a jar. After a while I realized that the created JARs are input for processing - it was not obvious these are test "input constants". I think checked in JARs will make this more obvious and make the tests much easier to follow. A rule of thumb is the less preparation logic in a test, the simpler and easier to read the test is.

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/Main.java
kubawach: Done then :)

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/Main.java
```diff
@@ -0,0 +1,159 @@
...
+import java.util.jar.JarOutputStream;
+import java.util.zip.ZipEntry;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class Main {
+
+  private static final Logger logger = LoggerFactory.getLogger(Main.class);
+
+  private static final Main INSTANCE = new Main(new ClassArchive.Factory() {});
```
mateuszrzeszutek, line 24: Nit: you could make `ClassArchive.Factory` a functional interface (or, alternatively, get rid of it and use `BiFunction`) and use `ClassArchive::New` here instead.

static-instrumenter/agent-instrumenter/src/test/java/io/opentelemetry/contrib/staticinstrumenter/ClassArchiveTest.java
kubawach: As U wish then ;)

anuraaga (approved):
Thanks!

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/Main.java
kubawach: That's a nice trick :)

mateuszrzeszutek (approved):

Enkelian (approved):

trask:
thx @kubawach!

PR #243, opened 2022-03-05
Sync CodeQL workflow with instrumentation repo

trask:
Copy-paste from instrumentation repo. It looks like reusable workflows work across repos too now, will try that in the future, with the primary workflows living in the core java repo.


anuraaga:
The workflow hasn't been working in core for a few months since it doesn't seem to support java 17 well. I don't think codeql does much for us and while working on it now, removing is probably a better option

trask:
> The workflow hasn't been working in core for a few months since it doesn't seem to support java 17 well. I don't think codeql does much for us and while working on it now, removing is probably a better option

I'm kind of reluctant to remove it given it's mentioned in https://github.com/open-telemetry/community/blob/main/docs/how-to-configure-new-repository.md#best-practices

anuraaga:
Hmm but @laurit is also actively working on getting this repo to java 17 I think from the other PRs. I found that codeql is actually officially unsupported :/

https://codeql.github.com/docs/codeql-overview/supported-languages-and-frameworks/#id13

In practical terms I just don't think we can keep it for Java.

anuraaga:
@trask Thanks for looking into fixing codeql. Unfortunately, looks like it may still not be working, what do you think?

image

.github/workflows/nightly-codeql-analysis.yml
```diff
@@ -0,0 +1,50 @@
...
+          java-version: 17
+
+      - name: Initialize CodeQL
+        uses: github/codeql-action/init@v1
+        with:
+          languages: java
+
+      - uses: gradle/gradle-build-action@v2
+        with:
+          arguments: assemble
```
anuraaga, line 28:
```suggestion
          arguments: assemble --no-build-cache
```

Should we try this in SDK repo to see if it fixes the analyzed issue?

trask: oh, good idea, I will send PR

anuraaga:
Cool looks fine now - let's also set the cache to read-only

image

trask:
> let's also set the cache to read-only

just curious why, it seems cache is still potentially helpful for reducing sporadic maven central download failures

anuraaga:
> just curious why, it seems cache is still potentially helpful for reducing sporadic maven central download failures

Yeah that's why we set it to read-only instead of not using cache at all. I don't think there's any specific artifacts in the assemble task we'd need to write in, but it might cause a job-specific cache copy being sent up (not sure if it does in practice but read-only seems fair for it).

trask:
It may not make much difference in this repo, but I would want a CodeQL-specific cache in instrumentation repo, otherwise it might pick up a smoke-test cache, which isn't going to be much help for cached dependencies

anuraaga:
Aren't dependencies cached independently of the job, as something like `dependencies-${hashOfDependencies}` so it should always pick up the correct one as long as it's not been expired? In the instrumentation repo FWIU, we set many jobs to read-only to make sure we don't go over the cache limit, so thought it's a good idea for this one too

anuraaga (approved):

Issue #222, opened 2022-01-31
Sample Groovy yields "Found duplicate metric definition"


rkogelhe:
**Description**
jmxmetrics using the sample groovy works correctly on the first collection.
On the second collection, we get:
`WARNING: Found duplicate metric definition: my.admin.response.count
	at unknown source
		To enable better debugging, run your JVM with -Dotel.experimental.sdk.metrics.debug=true
Causes
- InstrumentType [OBSERVABLE_GAUGE] is async and already registered
Original instrument registered with same name but is incompatible.
	at unknown source
		To enable better debugging, run your JVM with -Dotel.experimental.sdk.metrics.debug=true

io.opentelemetry.sdk.metrics.internal.state.DuplicateMetricStorageException: Async metric with same name has already been created. Found previous metric: MetricDescriptor{name=my.admin.response.count, description=Count of the number of responses from the admin console, unit=By, sourceView=Optional[View{name=null, description=null, aggregation=DefaultAggregation, attributesProcessor=io.opentelemetry.sdk.metrics.internal.view.AttributesProcessor$1@249c65fd, sourceInfo=INSTANCE}], sourceInstrument=InstrumentDescriptor{name=my.admin.response.count, description=Count of the number of responses from the admin console, unit=By, type=OBSERVABLE_GAUGE, valueType=LONG, sourceInfo=INSTANCE}}, cannot register metric: MetricDescriptor{name=my.admin.response.count, description=Count of the number of responses from the admin console, unit=By, sourceView=Optional[View{name=null, description=null, aggregation=DefaultAggregation, attributesProcessor=io.opentelemetry.sdk.metrics.internal.view.AttributesProcessor$1@249c65fd, sourceInfo=INSTANCE}], sourceInstrument=InstrumentDescriptor{name=my.admin.response.count, description=Count of the number of responses from the admin console, unit=By, type=OBSERVABLE_GAUGE, valueType=LONG, sourceInfo=INSTANCE}}
	at io.opentelemetry.sdk.metrics.internal.state.MetricStorageRegistry.register(MetricStorageRegistry.java:75)
	at io.opentelemetry.sdk.metrics.internal.state.MeterSharedState.register(MeterSharedState.java:139)
...`

**Steps to reproduce**
Clone opentelemetry-java-contrib and only build jmx-metrics (full build isn't working) 
Run with 
[...]

PR #253, opened 2022-03-11
jmx-metrics: register all GroovyMetricEnvironment instruments

rmfitzpatrick:
**Description:**
Bug fix  - The `otel.instrument()` helpers do not retrieve the previously built instrument instances on subsequent groovy script runs and this leads to logged duplicate instrument recording warnings. Because the repeated script running pattern is [not aligned with the metric api](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#instrument-type-conflict-detection), instrument creation should be memoized in the metric environment by descriptor to avoid the unsupported usage.

**Existing Issue(s):**
https://github.com/open-telemetry/opentelemetry-java-contrib/issues/222

**Testing:**
Added instance checks for synchronous instrument tests. I briefly investigated testing w/ captured loggers but a straightforward way to access the internal sdk instrument's loggers wasn't apparent so I'm not sure of a clean way to detect warnings in async storage. Since these* unavailable statements are prone to drift it's not a solid practice to begin with.

### DIFF
--- a/jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/GroovyMetricEnvironment.java
+++ b/jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/GroovyMetricEnvironment.java
@@ -14,8 +14,14 @@ import io.opentelemetry.api.metrics.LongCounter;
 import io.opentelemetry.api.metrics.LongHistogram;
 import io.opentelemetry.api.metrics.LongUpDownCounter;
 import io.opentelemetry.api.metrics.Meter;
+import io.opentelemetry.api.metrics.ObservableDoubleCounter;
+import io.opentelemetry.api.metrics.ObservableDoubleGauge;
 import io.opentelemetry.api.metrics.ObservableDoubleMeasurement;
+import io.opentelemetry.api.metrics.ObservableDoubleUpDownCounter;
+import io.opentelemetry.api.metrics.ObservableLongCounter;
+import io.opentelemetry.api.metrics.ObservableLongGauge;
 import io.opentelemetry.api.metrics.ObservableLongMeasurement;
+import io.opentelemetry.api.metrics.ObservableLongUpDownCounter;
 import io.opentelemetry.sdk.autoconfigure.AutoConfiguredOpenTelemetrySdk;
 import io.opentelemetry.sdk.metrics.InstrumentType;
 import io.opentelemetry.sdk.metrics.InstrumentValueType;
@@ -26,6 +32,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
+import java.util.function.Function;
 import javax.annotation.Nullable;
 
 public class GroovyMetricEnvironment {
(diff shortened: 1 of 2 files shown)


jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/GroovyMetricEnvironment.java
```diff
@@ -247,13 +348,25 @@ public void registerLongValueCallback(
...
       final String unit,
       final Consumer<ObservableLongMeasurement> updater) {
-    meter
-        .gaugeBuilder(name)
-        .ofLongs()
-        .setDescription(description)
-        .setUnit(unit)
-        .buildWithCallback(
-            proxiedLongObserver(name, description, unit, InstrumentType.OBSERVABLE_GAUGE, updater));
+    AtomicReference<ObservableLongGauge> existingCounter =
```
dehaansa, line 351: It feels like many of these functions could be refactored to DRY it up.

jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/GroovyMetricEnvironment.java
```diff
@@ -33,15 +39,48 @@
...
+
+    private AtomicReference<T> getExisting(
+        String name,
+        String description,
+        String unit,
+        InstrumentType instrumentType,
+        InstrumentValueType instrumentValueType) {
+      InstrumentDescriptor descriptor =
+          InstrumentDescriptor.create(name, description, unit, instrumentType, instrumentValueType);
+      registry.putIfAbsent(descriptor.hashCode(), new AtomicReference<>());
```
dehaansa, line 53: Should be able to return [`computeIfAbsent`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html#computeIfAbsent-K-java.util.function.Function-) rather than `putIfAbsent` & then return `get`.

carlosalberto:
Thanks for working on this, I was recently wondering about such messages ;)

jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/GroovyMetricEnvironment.java
```diff
@@ -139,7 +178,16 @@ protected static Attributes mapToAttributes(@Nullable final Map<String, String>
...
+    AtomicReference<DoubleCounter> existingCounter =
+        doubleCounterRegistry.getExisting(
+            name, description, unit, InstrumentType.COUNTER, InstrumentValueType.DOUBLE);
+
+    DoubleCounter dc = existingCounter.get();
+    if (dc == null) {
+      dc = meter.counterBuilder(name).setDescription(description).setUnit(unit).ofDoubles().build();
+      existingCounter.set(dc);
+    }
+    return dc;
```
jack-berg, line 190: I don't know the jmx-metrics code well, but these registries should not be necessary: if you ask a meter for an instrument multiple times with the same name, description, unit, type, and value type, there is no conflict and you won't get any warning log. 

Warnings are only produced if you request an instrument with the same name as one previous, but with a different description, unit, type, or value type. 

This code only seems to ensure that the same instrument is returned if it is requested with the same name, description, unit, type, and value type.

jamesmoessis:
I noticed this problem too, is this going to get merged?

trask:
> I noticed this problem too, is this going to get merged?

ping @rmfitzpatrick

jack-berg:
> I noticed this problem too, is this going to get merged?

Do you have a repo or some logs you can share? As mentioned [here](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/253#discussion_r853429001), if I understand correctly, this issue shouldn't happen anymore as the SDK behavior changed a while back so that no warning log is produced when multiple instruments are initialized with the same name, description, unit, type, and value type.

rmfitzpatrick:
@jack-berg when I run the updated unit tests without the registry they fail for identity equivalence. Would this be expected with the current sdk?

```
OtelHelperSynchronousMetricTest > longUpDownCounter() FAILED
    java.lang.AssertionError:
    Expecting actual:
      io.opentelemetry.sdk.metrics.SdkLongUpDownCounter@4d968142
    and actual:
      io.opentelemetry.sdk.metrics.SdkLongUpDownCounter@4d968142
    to refer to the same object
        at io.opentelemetry.contrib.jmxmetrics.OtelHelperSynchronousMetricTest.longUpDownCounter(OtelHelperSynchronousMetricTest.java:289)
```

jack-berg:
They aren't guaranteed to be identical objects (although those logs appear to indicate they are identical 🤔), but they are guaranteed to aggregate and export to the same metric.

trask:
> if I understand correctly, this issue shouldn't happen anymore as the SDK behavior changed a while back so that no warning log is produced when multiple instruments are initialized with the same name, description, unit, type, and value type.

@rmfitzpatrick @jamesmoessis is this still an issue?

PR #267, opened 2022-03-20
Switch from unmaintained actions

trask:
And a few other things:

* Now that the release notes are generated from the change log, remove a couple of more manual steps
* Cleaned up some of the patch release docs since the patch release workflow doesn't support cherry picks anymore (I like the manual cherry-pick proposal/review process)
* Added a step to always create release branch after minor release, to make it easier for contributors to propose patch candidates without having to get a maintainer to create the release branch first

### DIFF
--- a/.github/workflows/release-build.yml
+++ b/.github/workflows/release-build.yml
@@ -124,3 +124,9 @@ jobs:
                             --discussion-category announcements \
                             v${{ github.event.inputs.version }} \
                             opentelemetry-javaagent.jar
+
+        # proactively create release branch so that contributors can suggest patches more easily
+      - name: Create release branch
+        run: |
+          git checkout -b v${{ github.event.inputs.version }}
+          git push origin v${{ github.event.inputs.version }}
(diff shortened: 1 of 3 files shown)


.github/workflows/release-build.yml
```diff
@@ -113,25 +113,20 @@ jobs:
...
+                            --notes-file release-notes.txt \
+                            --discussion-category announcements \
+                            v${{ github.event.inputs.version }} \
+                            opentelemetry-javaagent.jar
+
+        # proactively create release branch so that contributors can suggest patches more easily
+      - name: Create release branch
+        run: |
+          git checkout -b v${{ github.event.inputs.version }}
+          git push origin v${{ github.event.inputs.version }}
```
mateuszrzeszutek, line 132: Wouldn't that create a branch with the same name as the tag? E.g. `v1.2.3` instead of `v.1.2.x`

trask: 🦅👁️❤️

anuraaga (approved):

PR #278, opened 2022-03-30
Static instrumenter - logic

Enkelian:
**Description:**

Another part of `agent-instrumenter`. Now, it provides a tool for modifying the opentelemetry-javaagent.jar so that it is capable of performing static instrumentation.

Modification of the agent classes can be divided into two aspects:
* Adding new classes - all classes located in `io.opentelemetry.contrib.staticinstrumenter.internal` are injected into the agent jar. Most importantly, we add the class `Main`, which provides a main method of the process of static instrumentation, responsible for saving transformed classes to the user's jar.
* Rebasing two classes already present in the agent :
  * We modify `OpenTelemetryAgent` so that it adds two new transformers to the `premain`/`agentmain`’s Instrumentation instance. Those transformers pass classes instrumented by the agent to the Main singleton, which then saves those classes in the user's jar. 
  * We rebase `HelperInjector` to be able to save helper classes created by the agent in a similar manner (via passing them to `Main`).

It's also worth noting that we create two modified agent jars. 
* First one (saved as `opentelemetry-javaagent.jar`) contains only the changes mentroned above and is used only for the process of static instrumentation.
* The second one (saved as `classpath-agent.jar`) has an additional change to the structure of the agent -  the `inst/` prefix is removed and `.classdata` format is changed back to `.class`. When a user wants to start their statically instrumented application, they should add this jar to the classpath. 

While working on our Maven plugin [in this repo](https://github.com/agh-static-x/Maven-Plugin), we were injecting necessary agent classes into the user's jar, so the `classspath-agent.jar` was not necessary. I think it's best to discuss which approach we want to use in the future.

**Existing Issue(s):**

Part of #156

**Testing:**

A small unit test. More to come in following PRs. 

**Documentation:**

README updated

**Outstanding items:**

* more tests, especially a dockerized test involving the whole process of static instrumentation
* separate classloaders for user classes ([this](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/236#discussion_r820366365) had to be added back to the core code to be able to instrument user classes)

### DIFF
--- a/static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/AgentInstrumenter.java
+++ b/static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/AgentInstrumenter.java
@@ -8,7 +8,7 @@ package io.opentelemetry.contrib.staticinstrumenter;
 import static net.bytebuddy.matcher.ElementMatchers.isMethod;
 import static net.bytebuddy.matcher.ElementMatchers.named;
 
-import io.opentelemetry.contrib.staticinstrumenter.advices.AgentAdvices;
+import io.opentelemetry.contrib.staticinstrumenter.advices.AgentAdviceClasses;
 import io.opentelemetry.contrib.staticinstrumenter.advices.HelperInjectorAdvice;
 import io.opentelemetry.contrib.staticinstrumenter.internals.ArchiveEntry;
 import io.opentelemetry.contrib.staticinstrumenter.internals.ClassArchive;
@@ -22,7 +22,6 @@ import io.opentelemetry.contrib.staticinstrumenter.util.AgentExtractor;
 import io.opentelemetry.contrib.staticinstrumenter.util.TmpDirManager;
 import io.opentelemetry.contrib.staticinstrumenter.util.path.IdentityPathGetter;
 import io.opentelemetry.contrib.staticinstrumenter.util.path.SimplePathGetter;
-import java.io.File;
 import java.io.IOException;
 import java.net.URL;
 import java.net.URLClassLoader;
@@ -43,12 +42,12 @@ import org.slf4j.LoggerFactory;
 final class AgentInstrumenter {
   private static final Logger logger = LoggerFactory.getLogger(AgentInstrumenter.class);
   private final Map<String, byte[]> classesToInject;
-  private final File otelJar;
+  private final Path otelJar;
   private final Path finalJarDir;
 
   AgentInstrumenter(String otelJarPath, String finalJarDir) {
     this.classesToInject = new HashMap<>();
-    this.otelJar = new File(otelJarPath);
+    this.otelJar = Path.of(otelJarPath);
     this.finalJarDir = Paths.get(finalJarDir);
   }
 
@@ -72,7 +71,7 @@ final class AgentInstrumenter {
       rebaseOpenTelemetryAgent(openTelemetryAgentClass);
       rebaseHelperInjector(helperInjectorClass);
 
-      prepareAdditionalClasses();
+      prepareStaticInstrumenterClasses();
 
       new AgentCreator(tmpDirManager, new IdentityPathGetter())
           .create(finalJarDir.resolve("opentelemetry-javaagent.jar"), otelJar, classesToInject);
@@ -87,16 +86,24 @@ final class AgentInstrumenter {
     }
   }
 
+  /**
+   * Modifies the <code>io.opentelemetry.javaagent.OpenTelemetryAgent</code> class so that it adds
+   * {@link PreTransformer} and {@link PostTransformer} instances to the {@link
+   * java.lang.instrument.Instrumentation} object used in the agent's premain method. Adding a
+   * PreTransformer instance must be performed after <code>installBootstrapJar</code> method and
+   * adding a PostTransformer must be performed at the end of <code>startAgent</code> method (after
+   * initializing the agent).
+   */
   private void rebaseOpenTelemetryAgent(Class<?> openTelemetryAgentClass) {
 
     DynamicType.Unloaded<?> openTelemetryAgentType =
         new ByteBuddy()
             .rebase(openTelemetryAgentClass)
             .visit(
-                Advice.to(AgentAdvices.AgentMainAdvice.class)
+                Advice.to(AgentAdviceClasses.AgentMainAdvice.class)
                     .on(isMethod().and(named("startAgent"))))
             .visit(
-                Advice.to(AgentAdvices.InstallBootstrapJarAdvice.class)
+                Advice.to(AgentAdviceClasses.InstallBootstrapJarAdvice.class)
                     .on(isMethod().and(named("installBootstrapJar"))))
             .make();
 
@@ -105,6 +112,12 @@ final class AgentInstrumenter {
         openTelemetryAgentType.getBytes());
   }
 
+  /**
+   * Modifies the <code>io.opentelemetry.javaagent.tooling.HelperInjector</code> class so that it
+   * saves additional classes created by the agent in the {@link Main#getAdditionalClasses()} map.
+   * At the end of the static instrumentation process, the Main class saves all additionalClasses to
+   * the result jar.
+   */
   private void rebaseHelperInjector(Class<?> helperInjectorClass) {
 
     DynamicType.Unloaded<?> helperInjectorType =
@@ -120,7 +133,7 @@ final class AgentInstrumenter {
         helperInjectorType.getBytes());
   }
 
-  private void prepareAdditionalClasses() {
+  private void prepareStaticInstrumenterClasses() {
     List<Class<?>> additionalClasses =
         new ArrayList<>(
             Arrays.asList(
--- a/static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/AgentInstrumenter.java
+++ b/static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/AgentInstrumenter.java
@@ -1,0 +1,148 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.contrib.staticinstrumenter;
+
+import static net.bytebuddy.matcher.ElementMatchers.isMethod;
+import static net.bytebuddy.matcher.ElementMatchers.named;
+
+import io.opentelemetry.contrib.staticinstrumenter.advices.AgentAdvices;
+import io.opentelemetry.contrib.staticinstrumenter.advices.HelperInjectorAdvice;
+import io.opentelemetry.contrib.staticinstrumenter.internals.ArchiveEntry;
+import io.opentelemetry.contrib.staticinstrumenter.internals.ClassArchive;
+import io.opentelemetry.contrib.staticinstrumenter.internals.CurrentClass;
+import io.opentelemetry.contrib.staticinstrumenter.internals.Main;
+import io.opentelemetry.contrib.staticinstrumenter.internals.PostTransformer;
+import io.opentelemetry.contrib.staticinstrumenter.internals.PreTransformer;
+import io.opentelemetry.contrib.staticinstrumenter.internals.TransformedClass;
+import io.opentelemetry.contrib.staticinstrumenter.util.AgentCreator;
+import io.opentelemetry.contrib.staticinstrumenter.util.AgentExtractor;
+import io.opentelemetry.contrib.staticinstrumenter.util.TmpDirManager;
+import io.opentelemetry.contrib.staticinstrumenter.util.path.IdentityPathGetter;
+import io.opentelemetry.contrib.staticinstrumenter.util.path.SimplePathGetter;
+import java.io.File;
+import java.io.IOException;
+import java.net.URL;
+import java.net.URLClassLoader;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import net.bytebuddy.ByteBuddy;
+import net.bytebuddy.asm.Advice;
+import net.bytebuddy.dynamic.DynamicType;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Responsible for loading and instrumenting the OpenTelemetry javaagent JAR file */
+final class AgentInstrumenter {
+  private static final Logger logger = LoggerFactory.getLogger(AgentInstrumenter.class);
+  private final Map<String, byte[]> classesToInject;
+  private final File otelJar;
+  private final Path finalJarDir;
+
+  AgentInstrumenter(String otelJarPath, String finalJarDir) {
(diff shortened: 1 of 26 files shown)


static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/util/AgentExtractor.java
```diff
@@ -0,0 +1,79 @@
...
+   * @param otelJarFile JarFile object representing the OpenTelemetry javaagent file
+   * @param entryToSave JAR entry that is extracted
+   * @param sanitized Sanitized JAR entry name
+   * @throws IOException If process of writing to file encounters problems
+   */
+  private void extractEntry(JarFile otelJarFile, JarEntry entryToSave, String sanitized)
+      throws IOException {
+    int lastSlashIdx = sanitized.lastIndexOf("/");
+    String path =
+        sanitized.substring(0, lastSlashIdx).replace("/", System.getProperty("file.separator"));
```
mateuszrzeszutek, line 70: Nit
```suggestion
        sanitized.substring(0, lastSlashIdx).replace("/", File.separator);
```

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/util/AgentExtractor.java
```diff
@@ -0,0 +1,79 @@
...
+    String className = sanitized.substring(lastSlashIdx + 1);
+
+    Path classPackage = Files.createDirectories(extractedAgent.resolve(path));
+
+    Files.write(
+        classPackage.resolve(className),
+        IOUtils.toByteArray(otelJarFile.getInputStream(entryToSave)));
```
mateuszrzeszutek, line 77: Nit: no need to use commons-io, you can achieve the same thing by using plain Java (and this will also close the input stream, unlike `IOUtils`):
```suggestion
    try (InputStream in = otelJarFile.getInputStream(entryToSave);
        OutputStream out = Files.newOutputStream(classPackage.resolve(className))) {
      in.transferTo(out);
    }
```

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/util/AgentExtractor.java
```diff
@@ -0,0 +1,79 @@
...
+    return extractedAgent;
+  }
+
+  /**
+   * Extracts the agent JAR entry to temporary directory under new, sanitized name.
+   *
+   * @param otelJarFile JarFile object representing the OpenTelemetry javaagent file
+   * @param entryToSave JAR entry that is extracted
+   * @param sanitized Sanitized JAR entry name
+   * @throws IOException If process of writing to file encounters problems
```
mateuszrzeszutek, line 64: Nit: there's no need to document everything; especially on private methods - you could rename the method arguments a bit and achieve similar degree of readability, e.g. `otelJarFile` -> `otelJavaagent`, `sanitized` -> `sanitizedEntryName`

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/util/AgentExtractor.java
```diff
@@ -0,0 +1,79 @@
...
+
+  public AgentExtractor(TmpDirManager tmpDirManager) throws IOException {
+    this.extractedAgent = tmpDirManager.createDir("opentelemetry-javaagent");
+  }
+
+  /**
+   * Extracts the agent JAR entry to a temporary directory.
+   *
+   * @param agentFile File object representing the OpenTelemetry javaagent file
+   * @throws IOException If process of writing to file encounters problems
```
mateuszrzeszutek, line 35: Nit: instead of describing the method param/thrown exception (which are kinda obvious), WDYT about mentioning that the `inst/` & `.classdata` convention is dropped, so that agent classes can be "normally" loaded?

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/advices/AgentAdvices.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+package io.opentelemetry.contrib.staticinstrumenter.advices;
+
+import io.opentelemetry.contrib.staticinstrumenter.internals.Main;
+import java.lang.instrument.Instrumentation;
+import net.bytebuddy.asm.Advice;
+
+@SuppressWarnings({"SystemOut", "CatchAndPrintStackTrace"})
+public final class AgentAdvices {
```
mateuszrzeszutek, line 13: "Advice" is an uncountable noun -- WDYT about naming it `AgentAdviceClasses` instead?

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/advices/AgentAdvices.java
```diff
@@ -0,0 +1,38 @@
...
+  @SuppressWarnings("unused")
+  public static final class InstallBootstrapJarAdvice {
+
+    @Advice.OnMethodExit(suppress = Throwable.class, onThrowable = Throwable.class)
+    static void enterInstallBoostrapJar(
+        @Advice.Argument(value = 0, readOnly = false) Instrumentation inst) {
```
mateuszrzeszutek, line 19: This argument can be read-only, you're not overwriting it (same in the other advice class)

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/internals/ClassArchive.java
```diff
@@ -71,11 +71,21 @@ private InputStream getInputStreamForEntry(JarEntry inEntry, ZipEntry outEntry)
...
     if (entry.shouldInstrument()) {
       String className = entry.getName();
-      byte[] modified = instrumentedClasses.get(entry.getPath());
-      if (modified != null) {
-        logger.debug("Found instrumented class: " + className);
-        entryIn = new ByteArrayInputStream(modified);
-        outEntry.setSize(modified.length);
+      try {
+        // TODO: load classes on archive's separate classloader
+        Class.forName(className, false, ClassLoader.getSystemClassLoader());
```
mateuszrzeszutek, line 76: Is this `Class.forName` call triggering the otel javaagent transformer? Can you add an explanation somewhere in the code, how does the transformation happen?

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/AgentInstrumenter.java
```diff
@@ -0,0 +1,148 @@
...
+      URL url = tmpDir.toUri().toURL();
+      URLClassLoader otelClassLoader = new URLClassLoader(new URL[] {url});
+
+      Class<?> openTelemetryAgentClass =
+          otelClassLoader.loadClass("io.opentelemetry.javaagent.OpenTelemetryAgent");
+      Class<?> helperInjectorClass =
+          otelClassLoader.loadClass("io.opentelemetry.javaagent.tooling.HelperInjector");
+
+      rebaseOpenTelemetryAgent(openTelemetryAgentClass);
+      rebaseHelperInjector(helperInjectorClass);
```
mateuszrzeszutek, line 73: Now these methods require some additional explanation - can you add some comments/javadocs that explain what code modification are applied and why?

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/AgentInstrumenter.java
```diff
@@ -0,0 +1,148 @@
...
+
+      prepareAdditionalClasses();
```
mateuszrzeszutek, line 75: How about `prepareStaticInstrumenterClasses()`?

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/util/AgentCreator.java
```diff
@@ -0,0 +1,92 @@
...
+
+          String newName = pathGetter.getPath(entry);
+
+          try {
+            zout.putNextEntry(new JarEntry(newName));
+            if (replacement == null) {
+              in.transferTo(zout);
+            } else {
+              zout.write(replacement);
+              logger.debug("Instrumented " + entry.getName());
```
mateuszrzeszutek, line 70:
```suggestion
              logger.debug("Instrumented {}", entry.getName());
```

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/util/AgentCreator.java
```diff
@@ -0,0 +1,92 @@
...
+          zout.closeEntry();
+        }
+
+        // add extra classes
+        for (Map.Entry<String, byte[]> mapEntry : classesToInstrumentCopy.entrySet()) {
+          zout.putNextEntry(new JarEntry(mapEntry.getKey()));
+          zout.write(mapEntry.getValue());
+          zout.closeEntry();
+
+          logger.debug("Instrumented " + mapEntry.getKey());
```
mateuszrzeszutek, line 86:
```suggestion
          logger.debug("Instrumented {}", mapEntry.getKey());
```

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/util/AgentCreator.java
```diff
@@ -0,0 +1,92 @@
...
+        }
+      }
+    }
+    Files.copy(Paths.get(tmpPath), finalJarPath, REPLACE_EXISTING);
```
mateuszrzeszutek, line 90: How about 
```suggestion
    Files.move(Paths.get(tmpPath), finalJarPath, REPLACE_EXISTING);
```

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/util/AgentCreator.java
```diff
@@ -0,0 +1,92 @@
...
+    // following code mimics ByteBuddy's DynamicType.inject()
+    // we don't use it directly, since it would mean copying jar every time we add one class
+    try (JarInputStream in = new JarInputStream(new FileInputStream(otelJar))) {
+
+      Manifest manifest = in.getManifest();
+
+      try (JarOutputStream zout =
+          manifest == null
+              ? new JarOutputStream(new FileOutputStream(tmpPath))
+              : new JarOutputStream(new FileOutputStream(tmpPath), manifest)) {
```
mateuszrzeszutek, line 55: Instead of using `FileInputStream`/`FileOutputStream` and strings, you can use `Files.newInputStream()`/`Files.newOutputStream()` and `Path`s

mateuszrzeszutek:
Also, it'd be great to see some unit tests of the new code - some of these new classes have lots of complex logic.

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/util/TmpDirManager.java
```diff
@@ -0,0 +1,42 @@
...
+  public TmpDirManager(String rootDirPrefix) throws IOException {
+    this.tmpDir = Files.createTempDirectory(rootDirPrefix);
+    Runtime.getRuntime()
+        .addShutdownHook(
+            new Thread(
+                () -> {
+                  try {
+                    FileUtils.deleteDirectory(tmpDir.toFile());
+                  } catch (IOException e) {
+                    logger.debug("Could not create shutdown hook.");
```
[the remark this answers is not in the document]
Enkelian: Sure!

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/util/AgentExtractor.java
Enkelian: Thx, I'll get rid of the dependency then

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/advices/AgentAdvices.java
Enkelian: I tend to forget about this one, thanks for catching it!

anuraaga:
Can you add to the PR description a high level description of the changes to the javaagent? While there's been some discussion of static instrumentation being appropriate for the -instrumentation repo, this change in particular feels a bit weird - we build and publish the javaagent, but then in our own codebase have to hackily reinstrument the javaagent itself. I think we should provide knobs in the agent itself to support static instrumentation without bytebuddying it after the fact.

static-instrumenter/agent-instrumenter/src/main/java/io/opentelemetry/contrib/staticinstrumenter/internals/ClassArchive.java
mateuszrzeszutek: Thanks!

Enkelian:
@anuraaga I updated the description. Let me know if any further explanation is necessary.
I know that the current approach is far from ideal, though we decided on that during development at the AGH University since it needed no changes in other projects. It would be great if we could replace it with a simpler solution in the instrumentation repo.

anuraaga:
I think the interim approach before possibly integrating directly into the agent is to make this an "agent distro" using the extension mechanism

https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/javaagent-extension-api/src/main/java/io/opentelemetry/javaagent/extension

I think much of what's described in the PR title is already available and others we could probably add to support this.

My opinion is to not merge with the rebuild agent approach as here and go for that. Curious what other @open-telemetry/java-instrumentation-maintainers think

mateuszrzeszutek:
> I think the interim approach before possibly integrating directly into the agent is to make this an "agent distro" using the extension mechanism

I think a distro (not just an extension) could work here: a custom agent class with `premain`/`agentmain` methods that install the transformers, and an `AgentExtension` that installs a custom hook in the `HelperInjector` (we'd have to add that in the main agent codebase). You could even throw a `ConfigPropertySource` that enabled the noop mode and skips teh SDK initialization (you don't need that during static instrumentation). 
The whole instrumentation process would be greatly simplified too: you'd just take the modified distro, run the static instrumentation on your app, and that's it.
I'm in favor of this idea, it simplifies things a lot.

Enkelian:
I'll try to take the approach described by @mateuszrzeszutek - migrating to `AgentExtension` in this repo and creating the necessary hook in `HelperInjector` in the instrumentation repo. I'm closing this PR and I'll follow up with another one when the extension is ready.

PR #285, opened 2022-04-08
Update errorprone

anuraaga:
I'm appreciating errorprone finding when we are recording longs into double histograms (well, assuming we're not supposed to be doing that :P)

### DIFF
--- a/jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/container/ContainerConfigurationHandler.java
+++ b/jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/container/ContainerConfigurationHandler.java
@@ -27,7 +27,6 @@ public final class ContainerConfigurationHandler implements RecordedEventHandler
   public void initializeMeter(Meter meter) {
     meter
         .upDownCounterBuilder(METRIC_NAME)
-        .ofDoubles()
         .setUnit(ONE)
         .buildWithCallback(codm -> codm.record(value));
   }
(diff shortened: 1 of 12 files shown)


jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/container/ContainerConfigurationHandler.java
```diff
@@ -27,7 +27,6 @@ public ContainerConfigurationHandler() {
   public void initializeMeter(Meter meter) {
     meter
         .upDownCounterBuilder(METRIC_NAME)
-        .ofDoubles()
```
anuraaga, line 30: @kittylyst Can you check out this PR, especially this line? There was intention of using doubles even though the value is a long

mateuszrzeszutek (approved):

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/container/ContainerConfigurationHandler.java
kittylyst: @anuraaga Should this value be a long? Is it not popssible to have fractional CPU counts?

kittylyst (approved):
Apart from that one comment about CPU counts, this looks good to me.

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/container/ContainerConfigurationHandler.java
anuraaga: We use `getLong` below to retrieve so it seems no. Unless we should get double?

trask (approved):
> I'm appreciating errorprone finding when we are recording longs into double histograms

ya those are some really good finds 🎉

jack-berg (approved):

PR #293, opened 2022-04-13
Static instr maven plugin core

kubawach:
**Description:**

Maven plugin using static instrumenter.

**Existing Issue(s):**

#156 

**Testing:**

Unit tests and a single integration test

**Outstanding items:**

Once the instrumenter is ready (current @Enkelian work):
1. TODOs in the code resolved
2. Adding the agent to the plugin (currently - binary in the resources folder)


kubawach:
Worked a bit on the maven plugin build. It fails however as created Maven POM contains following reference:
```
<dependencyManagement>
    <dependencies>
      <dependency>
        <groupId>opentelemetry-java-contrib</groupId>
        <artifactId>dependencyManagement</artifactId>
        <version>1.13.0-SNAPSHOT</version>
        <type>pom</type>
        <scope>import</scope>
      </dependency>
    </dependencies>
  </dependencyManagement>
```
Need some gradle expert help with this - @anuraaga ? :)

static-instrumenter/maven-plugin/src/test/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/UnpackerTest.java
```diff
@@ -0,0 +1,60 @@
...
+    Unpacker unpacker = new Unpacker(targetFolder.toPath());
+    PackagingSupport support = PackagingSupport.EMPTY;
+    Path jar = Paths.get(getResourcePath("test.jar"));
+
+    // when
+    List<Path> copied = unpacker.copyAndUnpack(jar, support);
+    // then
+    assertThat(copied).hasSize(3);
+    // copied the right file?
+    assertThat(copied.get(0).getFileName().toString()).isEqualTo("test.jar");
```
mateuszrzeszutek, line 50: Can you verify all the output file names?

static-instrumenter/maven-plugin/src/main/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/PackagingSupportFactory.java
```diff
@@ -0,0 +1,36 @@
...
+
+  private static final Map<String, Supplier<PackagingSupport>> SUPPORTED_FRAMEWORKS =
+      Map.of(
+          "BOOT-INF/lib/", () -> new PackagingSupport("BOOT-INF/classes/"),
+          "WEB-INF/lib/", () -> new PackagingSupport("WEB-INF/classes/"));
+
+  private PackagingSupportFactory() {}
+
+  public static PackagingSupport packagingSupportFor(Path mainArtifact) throws IOException {
+    JarFile jarFile = new JarFile(mainArtifact.toFile());
```
mateuszrzeszutek, line 26: Use try-with-resources for that

static-instrumenter/maven-plugin/src/main/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/PackagingSupportFactory.java
```diff
@@ -0,0 +1,36 @@
...
+    Enumeration<JarEntry> enums = jarFile.entries();
```
mateuszrzeszutek, line 27: How about calling it `entries` instead?

static-instrumenter/maven-plugin/src/main/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/PackagingSupportFactory.java
```diff
@@ -0,0 +1,36 @@
...
+    while (enums.hasMoreElements()) {
```
mateuszrzeszutek, line 28: Wouldn't it be cheaper for iterate over keyset of `SUPPORTED_FRAMEWORKS` and use `getEntry()` to check if a directory entry with the expected name exists?

static-instrumenter/maven-plugin/src/main/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/PackagingSupportFactory.java
```diff
@@ -0,0 +1,36 @@
...
+import java.util.Map;
+import java.util.function.Supplier;
+import java.util.jar.JarEntry;
+import java.util.jar.JarFile;
+
+public final class PackagingSupportFactory {
+
+  private static final Map<String, Supplier<PackagingSupport>> SUPPORTED_FRAMEWORKS =
+      Map.of(
+          "BOOT-INF/lib/", () -> new PackagingSupport("BOOT-INF/classes/"),
```
mateuszrzeszutek, line 20: Just wondering: why are we checking for the existence of `BOOT-INF/lib`, if we want to move files from `BOOT-INF/classes`?

static-instrumenter/maven-plugin/src/main/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/Packer.java
```diff
@@ -0,0 +1,79 @@
...
+    return targetFile;
+  }
+
+  private String createFinalName(String instrumentedFileName) {
+    int lastDotIndex = instrumentedFileName.lastIndexOf('.');
+    return String.format(
+        "%s%s%s",
+        instrumentedFileName.substring(0, lastDotIndex),
+        finalNameSuffix,
+        instrumentedFileName.substring(lastDotIndex));
```
mateuszrzeszutek, line 77: You can just use string concatenation instead

static-instrumenter/maven-plugin/src/main/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/Instrumenter.java
```diff
@@ -0,0 +1,65 @@
...
+        StringBuilder builder =
+            new StringBuilder(
+                    "The instrumentation process for JAR dependencies finished with exit value: ")
+                .append(ret)
+                .append("\nSystem out:\n")
+                .append(output.toString(Charset.defaultCharset()));
+        throw new IOException(builder.toString());
+      }
+    } catch (InterruptedException ie) {
+      throw new IOException(ie);
```
mateuszrzeszutek, line 58:
```suggestion
    } catch (InterruptedException ie) {
      Thread.currentThread().interrupt();
      throw new IOException(ie);
```

static-instrumenter/maven-plugin/src/main/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/PackagingSupportFactory.java
```diff
@@ -0,0 +1,36 @@
...
+
+import java.io.IOException;
+import java.nio.file.Path;
+import java.util.Enumeration;
+import java.util.Map;
+import java.util.function.Supplier;
+import java.util.jar.JarEntry;
+import java.util.jar.JarFile;
+
+public final class PackagingSupportFactory {
```
mateuszrzeszutek, line 16: What's the reason for adding the spring boot/war handling logic here? Can't it be added to the main static instrumenter tool? If we add it to the maven plugin, we'll have to duplicate it in the gradle plugin too. And the standalone static instrumenter will be sort of incomplete, should anyone want to use it by itself (or implement their own tool, e.g. in SBT or Leiningen)

static-instrumenter/maven-plugin/src/main/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/ProjectModel.java
```diff
@@ -0,0 +1,67 @@
...
+        String names =
+            Arrays.toString(allArtifacts.stream().map(Path::toFile).map(File::getName).toArray());
+        logger.debug("Artifact name not provided. Defaults to instrument all artifacts: {}", names);
+      }
+      return allArtifacts;
+    }
+
+    for (Path artifactFile : allArtifacts) {
+      if (artifactName.equals(artifactFile.toFile().getName())) {
+        return new ArrayList<>(Collections.singletonList(artifactFile));
```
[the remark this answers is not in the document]
kubawach: Unfortunately it is - otherwise errorprone complains that some paths return mutable collection and some return immutable. So all returned list instances have to be mutable.

static-instrumenter/maven-plugin/src/main/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/OpentelemetryInstrumenterMojo.java
```diff
@@ -0,0 +1,104 @@
...
+  private String outputFolder;
+
+  @Parameter(readonly = true, defaultValue = "-instrumented")
+  @Nullable
+  private String suffix;
+
+  private final Logger logger = LoggerFactory.getLogger(OpentelemetryInstrumenterMojo.class);
+
+  /**
+   * Conducts the process of application target file instrumentation. Adds the shutdown hook to
```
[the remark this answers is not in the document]
kubawach: Rephrased.

static-instrumenter/maven-plugin/src/test/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/OpentelemetryInstrumenterMojoTest.java
```diff
@@ -0,0 +1,52 @@
...
+
+class OpentelemetryInstrumenterMojoTest {
+
+  @Test
+  void shouldInstrumentSampleApplication(@TempDir File tempdir) throws Exception {
+    // given
+    OpentelemetryInstrumenterMojo mojo = new OpentelemetryInstrumenterMojo();
+    Path testApp = Paths.get(JarTestUtil.getResourcePath("test-http-app.jar"));
+    // when
+    mojo.executeInternal(tempdir.getPath(), "-instrumented", Arrays.asList(testApp));
```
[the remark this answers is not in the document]
kubawach: +1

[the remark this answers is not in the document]
kubawach: +1

static-instrumenter/maven-plugin/src/main/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/OpentelemetryInstrumenterMojo.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+  }
+
+  @VisibleForTesting
+  final void executeInternal(
+      String finalFolder, String finalNameSuffix, List<Path> artifactsToInstrument)
+      throws IOException {
+
+    try {
+      WorkingFolders.create(finalFolder);
```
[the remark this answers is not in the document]
kubawach: Great catch, after last round of refactor it actually ended up being used only here :)

static-instrumenter/maven-plugin/src/main/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/ArtifactProcessor.java
```diff
@@ -0,0 +1,69 @@
...
+    PackagingSupport packagingSupport = packagingSupportFor(artifact);
+    List<Path> artifactsToInstrument = unpacker.copyAndUnpack(artifact, packagingSupport);
+    log.info("Unpacked artifacts: {}", artifactsToInstrument);
+    Path instrumentedArtifact =
+        instrumenter.instrument(artifact.getFileName(), artifactsToInstrument);
+    log.info("Main artifact after instrumentation: {}", instrumentedArtifact);
+    agent.copyAgentClassesTo(instrumentedArtifact, packagingSupport);
+    Path finalArtifact = packer.packAndCopy(instrumentedArtifact, packagingSupport);
+    log.info("Final artifact: {}", finalArtifact);
+    return finalArtifact;
```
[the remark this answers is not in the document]
kubawach: Yep, it ends the processing - user should get the final file where configured.

static-instrumenter/maven-plugin/src/main/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/PackagingSupportFactory.java
kubawach: Makes sense.

static-instrumenter/maven-plugin/src/main/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/PackagingSupportFactory.java
kubawach: Historical reasons I believe. Will change that.

static-instrumenter/maven-plugin/src/main/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/Packer.java
kubawach: Isn't format() prettier? :)

static-instrumenter/maven-plugin/src/main/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/Instrumenter.java
kubawach: What's the reason behind this change?

static-instrumenter/maven-plugin/src/main/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/PackagingSupportFactory.java
kubawach: The idea is that once we start implementing something besides M3 plugin, we should factor out the file handling logic into a plugin-core module or so, as most ops will be quite similar. Then we can decide on the final logic split between the instrumenter and plugin (cores).

static-instrumenter/maven-plugin/src/main/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/Packer.java
trask: it depends on who you ask 😄

we have preferred string concat in the otel java repos

static-instrumenter/maven-plugin/src/main/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/Instrumenter.java
trask: this is a generally good practice, in order to propagate the interrupt

trask:
@kubawach it looks like you may have missed "8 hidden conversations" above, it's my least favorite github feature as they're often missed

static-instrumenter/maven-plugin/src/main/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/Packer.java
kubawach: That's fair ;)

static-instrumenter/maven-plugin/src/main/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/InstrumentationAgent.java
```diff
@@ -0,0 +1,125 @@
...
+  /**
+   * Creates new instance using {@value #JAR_FILE_NAME} classpath agent. Agent is copied to the
+   * target folder.
+   */
+  public static InstrumentationAgent createFromClasspathAgent(Path targetFolder)
+      throws IOException {
+
+    Path targetPath = targetFolder.resolve(JAR_FILE_NAME);
+    InputStream agentSource =
+        InstrumentationAgent.class.getClassLoader().getResourceAsStream(JAR_FILE_NAME);
```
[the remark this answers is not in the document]
kubawach: Currently it is - see `src/main/resources`. In the final version - see "outstanding items" in PR desc.

static-instrumenter/maven-plugin/src/main/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/InstrumentationAgent.java
```diff
@@ -0,0 +1,125 @@
...
+
+  /**
+   * Adds classes from OpenTelemetry javaagent JAR to target file. Removes the shading and replaces
+   * classdata file extension with class file extension.
+   */
+  public void copyAgentClassesTo(Path targetFile, PackagingSupport packagingSupport)
+      throws IOException {
+
+    Set<String> existing = allNames(targetFile);
+    try (FileSystem targetFs = FileSystems.newFileSystem(targetFile, null)) {
```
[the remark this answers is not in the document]
kubawach: The ClassLoader.

static-instrumenter/maven-plugin/src/main/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/InstrumentationAgent.java
```diff
@@ -0,0 +1,125 @@
...
+   */
+  public ProcessBuilder getInstrumentationProcess(String classpath, Path outputFolder) {
+    return new ProcessBuilder(
+        "java",
+        "-Dotel.instrumentation.internal-class-loader.enabled=false",
+        String.format("-javaagent:%s", agentPath),
+        "-cp",
+        classpath,
+        MAIN_CLASS,
+        outputFolder.toString());
```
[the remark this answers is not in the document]
kubawach: Yep, agent with static instrumenter needs to be used here.

static-instrumenter/maven-plugin/src/main/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/InstrumentationAgent.java
```diff
@@ -0,0 +1,125 @@
...
+
+    Set<String> existing = allNames(targetFile);
+    try (FileSystem targetFs = FileSystems.newFileSystem(targetFile, null)) {
+      consumeEntries(
+          agentJar,
+          (entry) -> {
+            String entryName = entry.getName();
+            if (isInstrumentationClass(entry)) {
+              String modifiedName = modifyAgentEntryName(entryName, packagingSupport);
+              if (!existing.contains(modifiedName)) {
```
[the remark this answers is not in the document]
kubawach: U sure about that? 
Current algo is to avoid moving entries from agent jar to target jar if they already exist. Hence:
1. We put all existing target jar entries in `existing` set
2. For each agent jar entry we check if it already exists in the target jar, if not - entry content is copied to target jar

If we switch contains() to add() we won't protect target jar entries from overwrite with agent entries.

static-instrumenter/maven-plugin/src/main/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/Unpacker.java
```diff
@@ -0,0 +1,74 @@
...
+  private Path copy(Path artifact, PackagingSupport packagingSupport) throws IOException {
+    Path targetArtifact = Files.createFile(targetFolder.resolve(artifact.getFileName()));
+    try (ZipOutputStream targetOut = new ZipOutputStream(Files.newOutputStream(targetArtifact));
+        JarFile artifactJar = new JarFile(artifact.toFile())) {
+      packagingSupport.copyRemovingPrefix(artifactJar, targetOut);
+    }
+    return targetArtifact;
+  }
+
+  private void unpack(Path artifact, List<Path> unpacked) throws IOException {
```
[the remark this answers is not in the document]
kubawach: Sure.

static-instrumenter/maven-plugin/src/test/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/UnpackerTest.java
kubawach: Do you mean extracted archives? That's why there is a separate`shouldUnpackNestedJars` test.

static-instrumenter/maven-plugin/src/main/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/PackagingSupportFactory.java
kubawach: Done.

static-instrumenter/maven-plugin/src/main/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/PackagingSupportFactory.java
kubawach: Var removed.

static-instrumenter/maven-plugin/src/main/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/Instrumenter.java
kubawach: Sure.

kubawach:
> @kubawach it looks like you may have missed "8 hidden conversations" above, it's my least favorite github feature as they're often missed

Ahhh yes, best feature ever! ;)

static-instrumenter/maven-plugin/src/main/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/InstrumentationAgent.java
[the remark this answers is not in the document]
mateuszrzeszutek: Hmm, true, looks like I got confused and was wrong. Nevermind then, thanks for being vigil 😛

static-instrumenter/maven-plugin/src/test/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/UnpackerTest.java
mateuszrzeszutek: Strictly speaking, it does not verify that correct paths were returned, it verifies that files were correctly unpacked.

static-instrumenter/maven-plugin/src/main/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/PackagingSupportFactory.java
mateuszrzeszutek: I'd argue that the static agent is that "something" -- it should be possible to use it without any build system plugin, e.g. in a makefile or something (I know, kind of an extreme example, but hey, you can use `javac` without maven or gradle). It is my opinion that the "standalone command line tool" should be truly standalone, and various build plugins should be just thin wrappers over that.
Still, I agree with you that it can be factored out separately in another PR, once #319 is merged -- let's not make this one any larger than it already is.

static-instrumenter/maven-plugin/src/main/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/PackagingSupportFactory.java
```diff
@@ -0,0 +1,36 @@
...
+
+  private static final Map<String, Supplier<PackagingSupport>> SUPPORTED_FRAMEWORKS =
+      Map.of(
+          "BOOT-INF/classes/", () -> new PackagingSupport("BOOT-INF/classes/"),
+          "WEB-INF/classes/", () -> new PackagingSupport("WEB-INF/classes/"));
```
mateuszrzeszutek, line 20: Nit: the value is a simple transformation of the key, this could be just a set of framework classes directory paths and you could just call `new PackagingSupport(...)` in `packagingSupportFor()`

static-instrumenter/maven-plugin/src/test/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/JarTestUtil.java
```diff
@@ -0,0 +1,45 @@
...
+import java.nio.file.Path;
+import java.util.jar.JarFile;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipOutputStream;
+
+final class JarTestUtil {
+
+  private JarTestUtil() {}
+
+  static void assertJar(Path jarFile, String... files) throws IOException {
```
mateuszrzeszutek, line 21: Nit
```suggestion
  static void assertJarContainsFiles(Path jarFile, String... files) throws IOException {
```

static-instrumenter/maven-plugin/src/main/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/ArtifactProcessor.java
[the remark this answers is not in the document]
kubawach: Sure!

static-instrumenter/maven-plugin/src/test/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/UnpackerTest.java
kubawach: True! Will add the latter as well :)

static-instrumenter/maven-plugin/src/main/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/PackagingSupportFactory.java
kubawach: Yep, let's not complicate this further ;)

static-instrumenter/maven-plugin/build.gradle.kts
```diff
@@ -1,5 +1,82 @@
...
+  publications {
+    create<MavenPublication>("mavenPlugin") {
+
+      from(components["java"])
+
+      pom {
+        name.set("OpenTelemetry Java Contrib")
+        url.set("https://github.com/open-telemetry/opentelemetry-java-contrib")
+        artifactId = base.archivesName.get()
+        description.set(project.description)
```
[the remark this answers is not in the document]
kubawach: Sure, no problem.

static-instrumenter/maven-plugin/src/main/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/InstrumentationAgent.java
```diff
@@ -0,0 +1,125 @@
...
+   *
+   * @param classpath classpath for instrumentation process (list of JARs to instrument, separated
+   *     with path separator)
+   * @param outputFolder output folder
+   * @return ProcessBuilder for instrumentation process
+   */
+  public ProcessBuilder getInstrumentationProcess(String classpath, Path outputFolder) {
+    return new ProcessBuilder(
+        "java",
+        "-Dotel.instrumentation.internal-class-loader.enabled=false",
```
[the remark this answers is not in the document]
kubawach: added TODO

static-instrumenter/maven-plugin/src/main/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/PackagingSupportFactory.java
kubawach: True that.

static-instrumenter/maven-plugin/src/test/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/JarTestUtil.java
kubawach: Added.

kubawach:
All resolved now.

static-instrumenter/maven-plugin/src/test/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/JarTestUtil.java
```diff
@@ -0,0 +1,45 @@
...
+
+final class JarTestUtil {
+
+  private JarTestUtil() {}
+
+  static void assertJarContainsFiles(Path jarFile, String... files) throws IOException {
+    assertThat(Files.exists(jarFile)).isTrue();
+    try (JarFile jar = new JarFile(jarFile.toFile())) {
+      for (int i = 0; i < files.length; i++) {
+        String file = files[i];
```
mateuszrzeszutek, line 25: Nit: could be replaced by a simple foreach-style loop

mateuszrzeszutek (approved):
LGTM 👍

kubawach:
@trask @anuraaga please have a look :)

trask (commented):
@Enkelian it would be great to get your approval as a component owner on this

mateuszrzeszutek:
Looks like the build is failing with 
```
No matching variant of project :static-instrumenter:maven-plugin was found. The consumer was configured to find an API of a library compatible with Java 8, preferably in the form of class files, preferably optimized for standard JVMs, and its dependencies declared externally
```
I'm not sure what could be the reason for that; maybe that's because of the publishing configuration? Have you tried running the build without publishing config?

kubawach:
> Looks like the build is failing with
> 
> ```
> No matching variant of project :static-instrumenter:maven-plugin was found. The consumer was configured to find an API of a library compatible with Java 8, preferably in the form of class files, preferably optimized for standard JVMs, and its dependencies declared externally
> ```
> 
> I'm not sure what could be the reason for that; maybe that's because of the publishing configuration? Have you tried running the build without publishing config?

It's a javadoc plugin error. The only way I could work around it was setting all contrib to Java 11 (instead of 8 by default as it's now). We seriously need someone with a decent gradle knowledge here (@anuraaga ?)

static-instrumenter/maven-plugin/build.gradle.kts
```diff
@@ -1,5 +1,82 @@
...
+          connection.set("scm:git:git@github.com:open-telemetry/opentelemetry-java-contrib.git")
+          developerConnection.set("scm:git:git@github.com:open-telemetry/opentelemetry-java-contrib.git")
+          url.set("git@github.com:open-telemetry/opentelemetry-java-contrib.git")
+        }
+
+        // remove dependencyManagement tag causing issues
+        withXml {
+          val dependencyManagement = asNode().get("dependencyManagement") as groovy.util.NodeList
+          asNode().remove(dependencyManagement[0] as Node?)
+        }
```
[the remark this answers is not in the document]
kubawach: Sure! 
So in order to publish a Mojo (M3 plugin) I'm using `de.benediktritter.maven-plugin-development`. It's responsible for creating a plugin descriptor and packaging everything nicely according to M3 plugin spec. Therefore a separate (`macenPlugin`) publication was needed and the convention could not be used. 
Unfortunately the plugin has issues handling "our" way the build is structured - namely the dependencyManagement module is turned into a Maven dependencyManagement tag referring to a dependency that does not exist. To make it work I had to add a piece of code removing the tag and thus generating a valid POM for the plugin. 
I experimented a lot with different options how to create M3 plugins from gradle - this was the only way that worked.

Enkelian (approved):

trask:
> We seriously need someone with a decent gradle knowledge here (@anuraaga ?)

@iNikem?

anuraaga:
I think the easiest way to unblock is to check in the descriptor file and avoid the plugin at least for now

anuraaga:
Ah I think the doc issue will be fixed with #329, which brings a snippet that we have in the SDK repo here.

But we still can't merge with the current duplication of publication config - notably, publish will fail because signing isn't configured. For now let's just check in the plugin descriptor while using the standard publish-conventions instead of using `maven-plugin-development`, we may figure out how to use both together but we can't skip the publish conventions

kubawach:
The build is passing now.

anuraaga:
@kubawach The PR build unfortunately doesn't verify publishing. Please check out my comment

> But we still can't merge with the current duplication of publication config - notably, publish will fail because signing isn't configured (among some other gaps). For now let's just check in the plugin descriptor while using the standard publish-conventions instead of using maven-plugin-development, we may figure out how to use both together but we can't skip the publish conventions

kubawach:
@anuraaga moved to static descriptor and otel-publishing conventions

static-instrumenter/maven-plugin/src/main/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/InstrumentationAgent.java
```diff
@@ -0,0 +1,126 @@
...
+import java.util.Set;
+import java.util.jar.JarEntry;
+import java.util.jar.JarFile;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class InstrumentationAgent {
+
+  // TODO: change after the new static-instr capable agent is released
+  public static final String JAR_FILE_NAME = "opentelemetry-javaagent.jar";
```
trask, line 28: is storing `opentelemetry-javaagent.jar` under source control needed because it's a SNAPSHOT build of the javaagent?

trask (commented):
is there an explanation somewhere of the binary jar and war files that have been added to source control? in particular, how to rebuild them from sources?

static-instrumenter/maven-plugin/src/test/resources/lib/firstNested.jar
```diff
@@ -0,0 +1 @@
+COPIED
```
trask, line 1: maybe?

```suggestion
not a real jar file
```

static-instrumenter/maven-plugin/build.gradle.kts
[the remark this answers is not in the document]
kubawach: Not applicable anymore.

static-instrumenter/maven-plugin/src/main/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/InstrumentationAgent.java
kubawach: It's only temporary - in the final version the agent-with-instrumentation-capabilities (currently - agent with agent-extension) will be fetched during build.

static-instrumenter/maven-plugin/build.gradle.kts
```diff
@@ -1,5 +1,32 @@
 plugins {
   id("otel.java-conventions")
+  id("otel.publish-conventions")
 }
 
 description = "Maven3 plugin for static instrumentation of projects code and dependencies"
+extra["mavenArtifactId"] = "static-instrumentation-maven-plugin"
```
anuraaga, line 7: Is there a reason we need this instead of setting `baseArchives.name`?

kubawach:
> is there an explanation somewhere of the binary jar and war files that have been added to source control? in particular, how to rebuild them from sources?

1. The app required for e2e test (test-http-app) is another module of this project
2. Rest of the JARs are examples to test various JAR and file system operations. Typically I'd create these in-place (in the test code) but in the previous PR @anuraaga requested to have all of such binaries added to test resources. So here they are :)

static-instrumenter/maven-plugin/build.gradle.kts
kubawach: The original ask from @mateuszrzeszutek was to publish the plugin under a specific artifactId - "static-instrumentation-maven-plugin". So this is the way I came up with, mirroring the solution for groupId we had.
As my knowledge of gradle is quite limited, it's probably a question to you - is it sufficient to change the baseArchives.name in the project? 😄

trask:
> > is there an explanation somewhere of the binary jar and war files that have been added to source control? in particular, how to rebuild them from sources?
> 
> 1. The app required for e2e test (test-http-app) is another module of this project
> 2. Rest of the JARs are examples to test various JAR and file system operations. Typically I'd create these in-place (in the test code) but in the previous PR @anuraaga requested to have all of such binaries added to test resources. So here they are :)

Can you add an instructions somewhere of how to (re-)build the binary jar/war files? It seems like that could be needed by someone maintaining the tests?

kubawach:
> > > is there an explanation somewhere of the binary jar and war files that have been added to source control? in particular, how to rebuild them from sources?
> > 
> > 
> > 
> > 1. The app required for e2e test (test-http-app) is another module of this project
> > 2. Rest of the JARs are examples to test various JAR and file system operations. Typically I'd create these in-place (in the test code) but in the previous PR @anuraaga requested to have all of such binaries added to test resources. So here they are :)
> 
> Can you add an instructions somewhere of how to (re-)build the binary jar/war files? It seems like that could be needed by someone maintaining the tests?

Added `CONTRIBUTING.md` with some description of the test data.

static-instrumenter/maven-plugin/build.gradle.kts
kubawach: Done, looks like it's working.

kubawach:
Is there anything left to be done with this PR?

trask:
> Added `CONTRIBUTING.md` with some description of the test data.

thx!

> Is there anything left to be done with this PR?

just to merge it, thx for your patience 👍

kubawach:
TY @trask it's been a long ride :)

PR #313, opened 2022-04-22
small test app for verifying instrumentation

kubawach:
**Description:**

Simple test app for verifying that an instrumentation has been applied. Prints out "SUCCESS" if a http request contains traceparent header, FAILURE otherwise.
The app is used for static instrumentation plugin integration test.


Enkelian (approved):
Thanks!

static-instrumenter/test-app/build.gradle.kts
```diff
@@ -0,0 +1,30 @@
...
+tasks {
+  withType<JavaCompile>().configureEach {
+    with(options) {
+      release.set(11)
+    }
+  }
+
+  jar {
+    manifest {
+      attributes(mapOf("Main-Class" to "io.opentelemetry.contrib.staticinstrumenter.test.HttpClientTest"))
```
trask, line 23:
```suggestion
      attributes("Main-Class" to "io.opentelemetry.contrib.staticinstrumenter.test.HttpClientTest")
```

trask (approved):

static-instrumenter/test-app/src/main/java/io/opentelemetry/contrib/staticinstrumenter/test/HttpClientTest.java
```diff
@@ -0,0 +1,45 @@
...
+
+  private HttpClientTest() {}
+
+  public static void main(String[] args) throws Exception {
+    System.setProperty("otel.traces.exporter", "logging");
+    makeCall();
+  }
+
+  private static void makeCall() throws IOException {
+    try (CloseableHttpClient httpClient = HttpClients.createDefault()) {
```
anuraaga, line 27: Not a big deal but would recommend using OkHttp instead if possible, it's our general "go-to" HTTP client, hence already being in dependency management

static-instrumenter/test-app/src/main/java/io/opentelemetry/contrib/staticinstrumenter/test/HttpClientTest.java
```diff
@@ -0,0 +1,45 @@
...
+      HttpGet request = new HttpGet("https://httpbin.org/get");
+      try (CloseableHttpResponse response = httpClient.execute(request)) {
+        HttpEntity entity = response.getEntity();
+        System.out.println();
```
anuraaga, line 31: Needed?

kubawach: Not at all - thanks for pointing this out.

static-instrumenter/test-app/src/main/java/io/opentelemetry/contrib/staticinstrumenter/test/HttpClientTest.java
kubawach: Oh, next time I'll default to OkHttp then ;)

mateuszrzeszutek (approved):

PR #321, opened 2022-05-03
Add tests on runtime attachment

jeanbisutti:
Hello,

This PR contains an automatic test for the runtime attachment feature.

Hope it will help.

### DIFF
--- a/runtime-attach/build.gradle.kts
+++ b/runtime-attach/build.gradle.kts
@@ -11,4 +11,10 @@ dependencies {
 
   // Used by byte-buddy but not brought in as a transitive dependency.
   compileOnly("com.google.code.findbugs:annotations")
+
+  testImplementation("io.opentelemetry.javaagent:opentelemetry-javaagent:1.6.0")
+  testImplementation("org.junit.jupiter:junit-jupiter-api")
+  testImplementation("org.junit-pioneer:junit-pioneer")
+  testImplementation("org.assertj:assertj-core")
+  testImplementation("org.apache.httpcomponents:httpclient:4.5.13")
 }
--- a/runtime-attach/build.gradle.kts
+++ b/runtime-attach/build.gradle.kts
@@ -12,7 +12,7 @@ dependencies {
   // Used by byte-buddy but not brought in as a transitive dependency.
   compileOnly("com.google.code.findbugs:annotations")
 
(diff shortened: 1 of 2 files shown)


runtime-attach/build.gradle.kts
```diff
@@ -11,4 +11,10 @@ dependencies {
 
   // Used by byte-buddy but not brought in as a transitive dependency.
   compileOnly("com.google.code.findbugs:annotations")
+
+  testImplementation("io.opentelemetry.javaagent:opentelemetry-javaagent:1.13.1")
```
mateuszrzeszutek, line 15: Can you add `io.opentelemetry.instrumentation:opentelemetry-instrumentation-bom-alpha` to `dependencyManagement`? This way you won't have to pass the version here.

runtime-attach/src/test/java/AgentAttachmentTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+    List<String> httpHeaders = makeHttpCallAndReturnHeaders();
+
+    String w3cTraceParentHeader = "traceparent";
+
+    assertThat(httpHeaders).contains(w3cTraceParentHeader);
+  }
+
+  private static List<String> makeHttpCallAndReturnHeaders() throws IOException {
+    try (CloseableHttpClient httpClient = HttpClients.createDefault()) {
```
mateuszrzeszutek, line 51: Nit: as mentioned [elsewhere](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/313#discussion_r857067251), we usually use OkHttp for these kind of things.

runtime-attach/src/test/java/AgentAttachmentTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+    List<String> httpHeaders = makeHttpCallAndReturnHeaders();
+
+    String w3cTraceParentHeader = "traceparent";
+
+    assertThat(httpHeaders).contains(w3cTraceParentHeader);
+  }
+
+  @Test
+  @SetEnvironmentVariable(key = "OTEL_TRACES_EXPORTER", value = "logging")
```
mateuszrzeszutek, line 40: This doesn't really change anything at this point; the agent is already installed (in the `@BeforeAll` method), changing the env var at this point will not affect it.
The 2 cases in this test class look exactly the same, perhaps they can be merged into one?

Issue #322, opened 2022-05-04
jmx-metrics: properties file should be able to set common sdk properties & ssl properties


dehaansa:
**Is your feature request related to a problem? Please describe.**
As the contrib-collector moves towards communicating with the JMX metrics gatherer over primarily a properties file (https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/9686), there are several properties that could be set this way that are not currently supported. Also unit tests are missing for properties file usage and the interaction between properties files & command line properties.

**Describe the solution you'd like**
Implement more support for more fields in properties file. Java trust & key store properties should be supported. My initial plan is just to support a list of these attributes that can be provided in the properties file that will  be set as System properties when loading into JmxConfig.

PR #323, opened 2022-05-04
Support setting more properties in the JMX Metrics properties file

dehaansa:
**Description:**
Enhance properties file support in JMX Metrics Gatherer as collector-contrib is moving towards using the properties file for communication to JMX MG.

**Existing Issue(s):** https://github.com/open-telemetry/opentelemetry-java-contrib/issues/322

**Testing:**
Added unit tests to validate properties file support & expected interaction between properties file values & command line arguments.


jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/JmxConfig.java
```diff
@@ -32,6 +32,16 @@ class JmxConfig {
   static final String JMX_REMOTE_PROFILE = PREFIX + "jmx.remote.profile";
   static final String JMX_REALM = PREFIX + "jmx.realm";
 
+  static final List<String> JAVA_SYSTEM_PROPERTIES =
+      Arrays.asList(
+          "otel.resource.attributes",
```
mateuszrzeszutek, line 37: About `otel.resource.attributes`: take a look at https://github.com/open-telemetry/opentelemetry-java-contrib/blob/main/jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/GroovyMetricEnvironment.java#L63-L85 -- according to that code, the resource property is already being propagated to the OTel SDK autoconfiguration. Does this not work?

jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/JmxConfig.java
```diff
@@ -32,6 +32,16 @@ class JmxConfig {
...
+          "javax.net.ssl.keyStore",
+          "javax.net.ssl.keyStorePassword",
+          "javax.net.ssl.keyStoreType",
+          "javax.net.ssl.trustStore",
+          "javax.net.ssl.trustStorePassword",
+          "javax.net.ssl.trustStoreType");
```
mateuszrzeszutek, line 43: What is consuming these properties? If it's the metric exporter, should these be copied in the `GroovyMetricEnvironment` together with all `otel.*` props?

mateuszrzeszutek, line 43: What is consuming these properties? If it's the metric exporter, should these be copied in the `GroovyMetricEnvironment` together with all `otel.*` props?

mateuszrzeszutek:
Pinging @rmfitzpatrick @Mrod1598

jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/JmxConfig.java
dehaansa: I had missed this component of the groovy metric environment startup! Thanks for pointing it out, I'll have to remove the otel.resource.attributes from the list that gets relocated into system.properties.

jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/JmxConfig.java
dehaansa: These properties are intended to be used internally to the JMXClient, not the metrics exporter.

mateuszrzeszutek (approved):

PR #325, opened 2022-05-05
Initial update of JXM's JVM metrics.

carlosalberto:
Initial pass for a few semantic convention updates:

* Loaded classes.
* Unloaded classes.
* Active threads.

Will update the rest in a second pass:

* Still to be decided whether the GC are Java specific or not.
* Memory ones need to be updated and refactored to be reported differently (not simply a rename).

### DIFF
--- a/jmx-metrics/src/integrationTest/java/io/opentelemetry/contrib/jmxmetrics/target_systems/JvmTargetSystemIntegrationTest.java
+++ b/jmx-metrics/src/integrationTest/java/io/opentelemetry/contrib/jmxmetrics/target_systems/JvmTargetSystemIntegrationTest.java
@@ -31,7 +31,20 @@ class JvmTargetSystemIntegrationTest extends AbstractIntegrationTest {
             "Metaspace",
             "Par Survivor Space");
     waitAndAssertMetrics(
-        metric -> assertGauge(metric, "jvm.classes.loaded", "number of loaded classes", "1"),
+        metric ->
+            assertSum(
+                metric,
+                "process.runtime.jvm.classes.loaded",
+                "Number of classes currently loaded",
+                "{classes}",
+                false),
+        metric ->
+            assertSum(
+                metric,
+                "process.runtime.jvm.classes.unloaded",
+                "Number of classes unloaded since JVM start",
+                "{classes}",
+                false),
         metric ->
             assertTypedSum(
                 metric,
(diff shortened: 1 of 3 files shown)


jmx-metrics/src/integrationTest/java/io/opentelemetry/contrib/jmxmetrics/target_systems/JvmTargetSystemIntegrationTest.java
```diff
@@ -67,6 +80,12 @@ void endToEnd() {
         metric ->
             assertTypedGauge(
                 metric, "jvm.memory.pool.used", "current memory pool usage", "by", gcLabels),
-        metric -> assertGauge(metric, "jvm.threads.count", "number of threads", "1"));
+        metric ->
+            assertSum(
+                metric,
+                "process.runtime.jvm.threads.count",
+                "number of threads",
```
trask, line 87: I'm probably missing something, do u know why this isn't:

```suggestion
                "Number of executing threads",
```

jmx-metrics/src/integrationTest/java/io/opentelemetry/contrib/jmxmetrics/target_systems/KafkaIntegrationTest.java
```diff
@@ -371,7 +371,20 @@ void endToEnd() {
...
+                      "Number of classes currently loaded",
+                      "{classes}",
+                      false),
+              metric ->
+                  assertSum(
+                      metric,
+                      "process.runtime.jvm.classes.unloaded",
+                      "Number of classes unloaded since JVM start",
+                      "{classes}",
+                      false),
```
trask, line 387: @Mrod1598 @rmfitzpatrick @dehaansa do you think there's value in re-asserting the JVM metrics in the Kafka test, or does it make sense to remove these assertions?

rmfitzpatrick: This is intended to test the multiple target system support so requires some redundancy by design: https://github.com/open-telemetry/opentelemetry-java-contrib/blob/main/jmx-metrics/src/integrationTest/resources/target-systems/jvm-and-kafka.properties#L4

jmx-metrics/src/main/resources/target-systems/jvm.groovy
```diff
@@ -15,8 +15,10 @@
  */
 
 def classLoading = otel.mbean("java.lang:type=ClassLoading")
-otel.instrument(classLoading, "jvm.classes.loaded", "number of loaded classes",
-        "1", "LoadedClassCount", otel.&longValueCallback)
+otel.instrument(classLoading, "process.runtime.jvm.classes.loaded", "Number of classes currently loaded",
+        "{classes}", "LoadedClassCount", otel.&longUpDownCounterCallback)
+otel.instrument(classLoading, "process.runtime.jvm.classes.unloaded", "Number of classes unloaded since JVM start",
+        "{classes}", "UnloadedClassCount", otel.&longUpDownCounterCallback)
```
dehaansa, line 21: It looks like this should *not* be an UpDownCounter based on the specification, and rather should be a monotonic counter

trask:
@carlosalberto are you still working on this? thx

PR #328, opened 2022-05-06
MeterProvider that delegates to Micrometer MeterRegistry

HaloFour:
**Description:**

Feature addition - Adds an implementation of OpenTelemetry Metrics `MeterProvider` which wraps and delegates metrics to Micrometer 1.x `MeterRegistry`.  This is to allow an existing project with an investment in Micrometer to iteratively migrate to OpenTelemetry without having to replace existing infrastructure code.

**Existing Issue(s):**

N/A

**Outstanding items:**

- [x] Tests
- [x] Documentation


trask:
Thanks @HaloFour!

Are you able to be a component owner for this? If so, can you update https://github.com/open-telemetry/opentelemetry-java-contrib/blob/main/.github/component_owners.yml in this PR?

I'd suggest renaming the top-level directory from `micrometer-metrics` to `micrometer-meter-provider` just to help avoid confusion from the "recommended" interop solution which is the micrometer shim in the core repo.

HaloFour:
Hey @trask!

I wouldn't mind being a component owner for this at all.  I've updated the component_owners file and renamed the module as suggested.

micrometer-meter-provider/src/main/java/io/opentelemetry/contrib/metrics/micrometer/ScheduledCallbackRegistrar.java
```diff
@@ -0,0 +1,50 @@
...
+    Objects.requireNonNull(scheduledExecutorService, "scheduledExecutorService");
+    return new ScheduledCallbackRegistrarBuilder(scheduledExecutorService);
+  }
+
+  @Override
+  public CallbackRegistration registerCallback(Runnable callback) {
+    if (callback != null) {
+      ScheduledFuture<?> future =
+          scheduledExecutorService.scheduleAtFixedRate(callback, period, period, timeUnit);
+      return () -> future.cancel(false);
```
mateuszrzeszutek, line 40: I don't know if this is an issue, but if it's done this way then all the callbacks will probably be called at slightly different times; meaning some when metrics collection happens some meters might have a somewhat up-to-date value, while some others might be a little bit stale. Is that okay?

micrometer-meter-provider/src/main/java/io/opentelemetry/contrib/metrics/micrometer/internal/instruments/MicrometerDoubleCounter.java
```diff
@@ -0,0 +1,87 @@
...
+    @Override
+    public ObservableDoubleCounter buildWithCallback(
+        Consumer<ObservableDoubleMeasurement> callback) {
+      MicrometerDoubleCounter instrument = build();
+      return instrument.registerDoubleCallback(
+          callback,
+          new ObservableDoubleMeasurement() {
+            @Override
+            public void record(double value) {
+              instrument.counter(Attributes.empty()).increment(value);
```
mateuszrzeszutek, line 77: All async OTel instruments report absolute values, not deltas; see [the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#asynchronous-counter-creation)

> Note: Unlike [Counter.Add()](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#add) which takes the increment/delta value, the callback function reports the absolute value of the counter. To determine the reported rate the counter is changing, the difference between successive measurements is used.

Maybe instead of using a micrometer `Counter`, a `FunctionCounter` would be better suited here?

(this comment applies to all async instruments here, not just the DoubleCounter)

micrometer-meter-provider/src/main/java/io/opentelemetry/contrib/metrics/micrometer/internal/instruments/AbstractUpDownCounter.java
```diff
@@ -0,0 +1,54 @@
...
+        .description(description())
+        .baseUnit(unit())
+        .register(meterRegistry());
+    return counter;
+  }
+
+  private static final class Counter extends AtomicLong {
+    private static final long serialVersionUID = 1927816293512124184L;
+
+    public void incrementDouble(double value) {
```
mateuszrzeszutek, line 39: WDYT about using `DoubleAdder` instead?

micrometer-meter-provider/src/main/java/io/opentelemetry/contrib/metrics/micrometer/internal/instruments/MicrometerDoubleGauge.java
```diff
@@ -0,0 +1,121 @@
...
+import io.opentelemetry.api.metrics.ObservableLongMeasurement;
+import io.opentelemetry.contrib.metrics.micrometer.internal.state.InstrumentState;
+import io.opentelemetry.contrib.metrics.micrometer.internal.state.MeterSharedState;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Consumer;
+import javax.annotation.Nullable;
+
+public final class MicrometerDoubleGauge extends AbstractInstrument {
+  private final MultiGauge multiGauge;
```
mateuszrzeszutek, line 25: Just wondering: wouldn't implementing the gauges in the same way as all other instruments be simpler? Why introduce a `MultiGauge` here?

micrometer-meter-provider/src/main/java/io/opentelemetry/contrib/metrics/micrometer/internal/state/MeterSharedState.java
```diff
@@ -0,0 +1,56 @@
...
+
+  @Nullable
+  public String instrumentationScopeVersion() {
+    return instrumentationScopeVersion;
+  }
+
+  @Nullable
+  public String schemaUrl() {
+    return schemaUrl;
+  }
```
mateuszrzeszutek, line 51: In OTel metrics API, `Meter`s with different instrumentation scopes are not supposed to influence each other; meaning that they can contain instruments with same names/attributes and they won't be merged together.
WDYT about adding the scope info as tags to all Micrometer instruments created within an OTel `Meter`?

micrometer-meter-provider/src/integrationTest/java/io/opentelemetry/contrib/metrics/micrometer/PrometheusIntegrationTest.java
```diff
@@ -0,0 +1,406 @@
...
+    if (!output.contains(value)) {
+      output = prometheusMeterRegistry.scrape();
+    }
+    assertThat(output).contains(value);
+    return output;
+  }
+
+  private <T> Consumer<T> onlyOnce(Consumer<T> consumer) {
+    return new Consumer<T>() {
+      boolean first = true;
```
mateuszrzeszutek, line 386: It'd probably be safer to use `AtomicBoolean`, just in case (to prevent test flakiness)

micrometer-meter-provider/src/test/java/io/opentelemetry/contrib/metrics/micrometer/TestCallbackRegistrar.java
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.contrib.metrics.micrometer;
+
+import java.util.ArrayList;
+
+public class TestCallbackRegistrar extends ArrayList<Runnable>
```
mateuszrzeszutek, line 10: Instead of inheritance, how about making the list a field of this class?

micrometer-meter-provider/src/test/java/io/opentelemetry/contrib/metrics/micrometer/internal/PollingMeterCallbackRegistrarTest.java
```diff
@@ -0,0 +1,72 @@
...
+    Meter pollingMeter;
+    try (PollingMeterCallbackRegistrar underTest =
+        new PollingMeterCallbackRegistrar(() -> meterRegistry)) {
+
+      pollingMeter = meterRegistry.find("otel_polling_meter").meter();
+      assertThat(pollingMeter).isNull();
+
+      underTest.registerCallback(() -> {});
+
+      pollingMeter = meterRegistry.get("otel_polling_meter").meter();
```
mateuszrzeszutek, line 40: `MeterRegistry#get()` will never return a null meter, it will throw instead. For consistency, consider using `find()` everywhere.

mateuszrzeszutek (commented):
Sorry for a late review, it took me a while to find some time to sit down and read through this PR. Left a couple of comments; but overall it looks like a very nice addition to the contrib repo 👍

micrometer-meter-provider/src/main/java/io/opentelemetry/contrib/metrics/micrometer/MicrometerMeterProvider.java
```diff
@@ -0,0 +1,65 @@
...
+  public static MicrometerMeterProviderBuilder builder(MeterRegistry meterRegistry) {
+    Objects.requireNonNull(meterRegistry, "meterRegistry");
+    return new MicrometerMeterProviderBuilder(() -> meterRegistry);
+  }
+
+  /**
+   * Returns a new builder instance for this provider with a {@link Supplier} for a {@link
+   * MeterRegistry}.
+   */
+  public static MicrometerMeterProviderBuilder builder(Supplier<MeterRegistry> meterRegistry) {
```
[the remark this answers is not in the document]
HaloFour: I did this mostly for Spring as it has unfortunate behavior if you materialize the MeterRegistry bean too early during initialization, it results in the BeanPostProcessors which handle the MeterRegistryCustomizers getting skipped.  But you'd likely want to setup the MeterProvider during Spring bootstrap as well.

micrometer-meter-provider/src/main/java/io/opentelemetry/contrib/metrics/micrometer/internal/instruments/AbstractInstrumentBuilder.java
```diff
@@ -0,0 +1,49 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.contrib.metrics.micrometer.internal.instruments;
+
+import io.opentelemetry.contrib.metrics.micrometer.internal.state.InstrumentState;
+import io.opentelemetry.contrib.metrics.micrometer.internal.state.MeterSharedState;
+import javax.annotation.Nullable;
+
+abstract class AbstractInstrumentBuilder<BuilderT extends AbstractInstrumentBuilder<BuilderT>> {
```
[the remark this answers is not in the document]
HaloFour: That's my preferred format as well, either spotless or checkstyle disagreed and required this format.

micrometer-meter-provider/src/main/java/io/opentelemetry/contrib/metrics/micrometer/internal/instruments/MicrometerDoubleCounter.java
HaloFour: That is interesting, and makes a lot more sense.  I'll revisit these instruments.

micrometer-meter-provider/src/main/java/io/opentelemetry/contrib/metrics/micrometer/internal/instruments/MicrometerDoubleGauge.java
HaloFour: MultiGauge was suggested by Micrometer, I don't mind revisiting this though for the sake of simplicity.

micrometer-meter-provider/src/main/java/io/opentelemetry/contrib/metrics/micrometer/internal/instruments/AbstractInstrumentBuilder.java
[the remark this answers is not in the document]
mateuszrzeszutek: Interesting... looks like we might have inconsistent errorprone configs between this repo and instrumentation.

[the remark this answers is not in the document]
HaloFour: Here's the warning:

```
warning: [TypeParameterNaming] Type Parameter BUILDER must be a single letter with an optional numeric suffix, or an UpperCamelCase name followed by the letter 'T'.
abstract class AbstractInstrumentBuilder<BUILDER extends AbstractInstrumentBuilder<BUILDER>> {

                                         ^
    (see https://google.github.io/styleguide/javaguide.html#s5.2.8-type-variable-names)
  Did you mean 'abstract class AbstractInstrumentBuilder<BuilderT extends AbstractInstrumentBuilder<BuilderT>> {' or 'abstract class AbstractInstrumentBuilder<B extends AbstractInstrumentBuilder<B>> {'?
error: warnings found and -Werror specified

```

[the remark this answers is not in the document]
mateuszrzeszutek: Fixed in https://github.com/open-telemetry/opentelemetry-java-contrib/pull/348

micrometer-meter-provider/src/main/java/io/opentelemetry/contrib/metrics/micrometer/internal/Constants.java
```diff
@@ -0,0 +1,23 @@
...
+
+import io.micrometer.core.instrument.Tag;
+
+/**
+ * Constants for common Micrometer {@link Tag} names for the OpenTelemetry instrumentation scope.
+ *
+ * <p>This class is internal and is hence not for public use. Its APIs are unstable and can change
+ * at any time.
+ */
+public class Constants {
```
mateuszrzeszutek, line 16: Nit
```suggestion
public final class Constants {
```

micrometer-meter-provider/src/main/java/io/opentelemetry/contrib/metrics/micrometer/internal/instruments/AtomicDoubleCounter.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.concurrent.atomic.AtomicLongFieldUpdater;
+
+final class AtomicDoubleCounter {
+  private static final AtomicLongFieldUpdater<AtomicDoubleCounter> BITS_UPDATER =
+      AtomicLongFieldUpdater.newUpdater(AtomicDoubleCounter.class, "doubleBits");
+
+  @SuppressWarnings("UnusedVariable")
+  private volatile long doubleBits;
+
+  public double current() {
```
mateuszrzeszutek, line 17: Nit: this class is package private, none of its methods need to be public.

micrometer-meter-provider/src/main/java/io/opentelemetry/contrib/metrics/micrometer/internal/instruments/AbstractUpDownCounter.java
```diff
@@ -0,0 +1,38 @@
...
+    counterMap.computeIfAbsent(attributesOrEmpty(attributes), this::createCounter).increment(value);
+  }
+
+  protected final void record(double value, Attributes attributes) {
+    counterMap.computeIfAbsent(attributesOrEmpty(attributes), this::createCounter).set(value);
+  }
+
+  private AtomicDoubleCounter createCounter(Attributes attributes) {
+    AtomicDoubleCounter counter = new AtomicDoubleCounter();
+    FunctionCounter.builder(name(), counter, AtomicDoubleCounter::current)
```
mateuszrzeszutek, line 31: `FunctionCounter` is supposed to be a monotonically increasing counter; for the `UpDownCounter` it's probably better to use a Micrometer `Gauge`.

micrometer-meter-provider/src/main/java/io/opentelemetry/contrib/metrics/micrometer/MicrometerMeterProvider.java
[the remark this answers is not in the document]
mateuszrzeszutek: I see. I'm a bit worried that someone will pass something like `MeterRegistry::new` that'll create a new instance every time it's called. WDYT about clearly documenting that it's supposed to be used as a late/lazy initialization mechanism, and wrapping the supplier in an equivalent of guava's `MemoizingSupplier` so that the original one is never called more than once?

micrometer-meter-provider/src/test/java/io/opentelemetry/contrib/metrics/micrometer/internal/instruments/MicrometerDoubleCounterTest.java
```diff
@@ -0,0 +1,202 @@
...
+        .containsExactlyInAnyOrder(
+            Tag.of(Constants.OTEL_INSTRUMENTATION_NAME, "meter"),
+            Tag.of(Constants.OTEL_INSTRUMENTATION_VERSION, "1.0"));
+    assertThat(id.getDescription()).isEqualTo("description");
+    assertThat(id.getBaseUnit()).isEqualTo("unit");
+    assertThat(counter.count()).isEqualTo(10.0);
+
+    measurable.setValue(20.0);
+    callbackRegistrar.run();
+    assertThat(counter.count()).isEqualTo(20.0);
```
mateuszrzeszutek, line 158: 👍

micrometer-meter-provider/src/main/java/io/opentelemetry/contrib/metrics/micrometer/internal/instruments/AbstractUpDownCounter.java
HaloFour: You're right, it looked like Micrometer didn't enforce that FunctionCounter was monotonic but the underlying MeterRegistry implementation might enforce it or make assumptions.  I've switched back to Gauge for correctness.

trask:
oh, that was unexpected, re-opening

mateuszrzeszutek (approved):
Looks great, thanks! 👍

trask:
thx @HaloFour!

PR #330, opened 2022-05-10
Add tests on runtime attachment

jeanbisutti:
Hello,

This PR contains 2 automatic tests for the runtime attachment feature.

Hope it will help.

### DIFF
--- a/runtime-attach/src/test/java/AgentAttachmentTest.java
+++ b/runtime-attach/src/test/java/AgentAttachmentTest.java
@@ -1,0 +1,60 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import io.opentelemetry.contrib.attach.RuntimeAttach;
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+import org.apache.http.Header;
+import org.apache.http.client.methods.CloseableHttpResponse;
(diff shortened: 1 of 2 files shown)


runtime-attach/src/test/java/AgentAttachmentTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+    List<String> httpHeaders = makeHttpCallAndReturnHeaders();
+
+    String w3cTraceParentHeader = "traceparent";
+
+    assertThat(httpHeaders).contains(w3cTraceParentHeader);
+  }
+
+  private static List<String> makeHttpCallAndReturnHeaders() throws IOException {
+    try (CloseableHttpClient httpClient = HttpClients.createDefault()) {
```
trask, line 51: can you switch to okhttp? see https://github.com/open-telemetry/opentelemetry-java-contrib/pull/313#discussion_r857067251

runtime-attach/src/test/java/AgentAttachmentTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+    List<String> httpHeaders = makeHttpCallAndReturnHeaders();
+
+    String w3cTraceParentHeader = "traceparent";
+
+    assertThat(httpHeaders).contains(w3cTraceParentHeader);
+  }
+
+  @Test
+  @SetEnvironmentVariable(key = "OTEL_TRACES_EXPORTER", value = "logging")
```
trask, line 40: The Javaagent only configures the exporters during startup, so I suspect it won't pick up this env var?

trask (commented):
thx @jeanbisutti!

PR #331, opened 2022-05-11
Update to OTel 1.14

anuraaga:


static-instrumenter/agent-extension/src/main/java/io/opentelemetry/contrib/staticinstrumenter/extension/AdditionalClassesInjectorListenerInstaller.java
```diff
@@ -20,7 +20,7 @@
 public class AdditionalClassesInjectorListenerInstaller implements AgentListener {
 
   @Override
-  public void beforeAgent(
+  public void afterAgent(
```
anuraaga, line 23: Tests still pass but have a feeling this may not be correct. @mateuszrzeszutek can you check?

Also this started failing automatically - given we hope to release instrumentation within a week or so, I may not have gone with merging a change to use SNAPSHOTs since it always can make the build less reliable.

mateuszrzeszutek: Yes, we should keep using the `beforeAgent` method in this case - the static instrumentation agent is a custom distro anyways, so this should be allowed.

PR #333, opened 2022-05-13
Speed up static instrumentation no-inst jar creation

laurit:
I noticed that creating of the static instrumentation no-inst jar takes a super long time. Took me a while to realise but an alternative way to fix this would be to move `rename` out of `eachFile`.


Enkelian (approved):
Thanks, I see that it also gets rid of the leftover `inst/` dir, so that helps a lot!

mateuszrzeszutek (approved):

PR #319, opened 2022-04-26
Static instrumenter logic - a new agent distro

Enkelian:
**Description:**

An alternative to #278.
As discussed, instead of modifying the agent with ByteBuddy, we create a new distro of the agent. The output of `./gradlew :static-instrumenter:agent-instrumenter:assemble` consists of two jars:
* `opentelemetry-static-agent.jar` - used for static instrumentation
* `opentelemetry-no-inst-agent.jar` - needed on application classpath after instrumentation, contains agent's classes with `/inst' and '.classdata` renamed back to plain `.class` files

The new approach also uses the new listener in `HelperInjector`, added here https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/5910
Until it is merged and released, we need to have local dependencies on `opentelemetry-javaagent` and `opentelemetry-muzzle`.

Modules used in the distro:
* `agent-instrumenter` - contains main static agent classes as well  `OpenTelemetryStaticAgent` - a wrapper for `OpenTelemetryAgent` that surrounds `premain`/`agentmain` calls with installing transformers for static instrumentation (`PreTransformer` and `PostTransformer`)
* `agent-extension` - an extension to the agent *that installs a listener in `HelperInjector`) and some additional configuration (that registers a boostrap package)
* `common` - utility classes that need to be accessible from both the agent classes and the static instrumenter main classes

**Existing Issue(s):**

Part of #156

**Testing:**

* unit test of logger
* integration test of the static agent on a sample application

**Documentation:**

* updated readme

**Outstanding items:**

* using test app from  #313 in the integration test
* better way to create `opentelemetry-no-inst-agent.jar` or implementing a solution that makes it obsolete
* removing local dependencies once new agent is released
* configurable logger levels

PR #334, opened 2022-05-18
Migrate `agent-instrumenter` to jvm test suite

Enkelian:
**Description:**
 As discussed in #319, the integration tests from `agent-instrumenter` is migrated to the new JVM Test Suite Plugin

### DIFF
--- a/static-instrumenter/agent-instrumenter/build.gradle.kts
+++ b/static-instrumenter/agent-instrumenter/build.gradle.kts
@@ -3,7 +3,6 @@ import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
 plugins {
   id("otel.java-conventions")
   id("com.github.johnrengelman.shadow") version "7.1.2"
-  id("jvm-test-suite")
 }
 
 description = "OpenTelemetry Java Static Instrumentation Agent"
--- a/static-instrumenter/agent-instrumenter/build.gradle.kts
+++ b/static-instrumenter/agent-instrumenter/build.gradle.kts
@@ -3,7 +3,7 @@ import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
 plugins {
   id("otel.java-conventions")
   id("com.github.johnrengelman.shadow") version "7.1.2"
-  id("org.unbroken-dome.test-sets") version "4.0.0"
+  id("jvm-test-suite")
 }
 
 description = "OpenTelemetry Java Static Instrumentation Agent"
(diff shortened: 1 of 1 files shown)


static-instrumenter/agent-instrumenter/build.gradle.kts
```diff
@@ -103,10 +98,22 @@ tasks {
   assemble {
     dependsOn(shadowJar, createNoInstAgent)
   }
+}
```
mateuszrzeszutek, line 101: Let's make sure that suites are always run as part of the PR build:
```suggestion

  check {
    dependsOn(testing.suites)
  }
}
```

mateuszrzeszutek (approved):
Thanks!

static-instrumenter/agent-instrumenter/build.gradle.kts
Enkelian: Sure!

static-instrumenter/agent-instrumenter/build.gradle.kts
```diff
@@ -3,7 +3,7 @@ import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
 plugins {
   id("otel.java-conventions")
   id("com.github.johnrengelman.shadow") version "7.1.2"
-  id("org.unbroken-dome.test-sets") version "4.0.0"
+  id("jvm-test-suite")
```
trask, line 6: I don't think this is needed, at least we don't have it in instrumentation repo (must be automatically part of gradle?)
```suggestion
```

anuraaga (approved):

static-instrumenter/agent-instrumenter/build.gradle.kts
```diff
@@ -3,7 +3,6 @@ import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
 plugins {
   id("otel.java-conventions")
   id("com.github.johnrengelman.shadow") version "7.1.2"
```
anuraaga, line 5: By the way, we should never have version numbers in plugins within artifacts as they all need to be the same in the build (this could potentially catch accidentally introducing an unused plugin like test-sets). Shadow in particular is already there

https://github.com/open-telemetry/opentelemetry-java-contrib/blob/main/settings.gradle.kts#L4

anuraaga: So basically this is a point we need to add to CONTRIBUTING.md

https://github.com/open-telemetry/opentelemetry-java-contrib/blob/main/CONTRIBUTING.md

trask: I'll send PR 👍

Issue #174, opened 2022-01-07
Request for a few additional component owners


trask:
We would like to have (at least) 2 owners for each component in this repository. The component owners will be automatically assigned to review PRs that touch files in those components.

`aws-xray`, `jfr-streaming`, `maven-extension`, and `runtime-attach` have (at least) 2 owners already.

`jmx-metrics` needs one additional owner - @Mrod1598 would you be interested? @rmfitzpatrick any other recommendations for people we could ask?

`contrib-samplers` also needs one additional owner - any volunteers?

`disruptor-processor` currently has no owners - @bogdandrutu are you still interested in this component?

Mrod1598:
Yes, I'd be interested!

trask:
> Yes, I'd be interested!

@Mrod1598 awesome, can you request membership into the OpenTelemetry github org? here's the instructions: https://github.com/open-telemetry/community/blob/main/community-membership.md#member. you can put myself and @anuraag as the two sponsors. being a member of the OTel org will allow us to assign you as a reviewer for the jmx-metrics component. thx!

dehaansa:
I'd be happy to join the `jmx-metrics` owners as well if you'd like another. I've contributed a bit there, and helped @Mrod1598 work through some issues on some of his contributions as well. Also haven't applied for membership yet but I've got one sponsor ready, so if anyone here would like to sponsor me as well I'd be happy to take that assistance.

trask:
> I'd be happy to join the `jmx-metrics` owners as well if you'd like another. I've contributed a bit there, and helped @Mrod1598 work through some issues on some of his contributions as well. Also haven't applied for membership yet but I've got one sponsor ready, so if anyone here would like to sponsor me as well I'd be happy to take that assistance.

Great, please add me as a sponsor!

trask:
Closing, all components except `disruptor-processor` have (at least) two owners, and I have opened separate issue to track that.

Thanks everyone!

PR #352, opened 2022-05-27
Consistent reservoir sampling span processor

oertl:
**Description:**

Implementation of a consistent reservoir sampling span processor


consistent-sampling/src/main/java/io/opentelemetry/contrib/samplers/ConsistentReservoirSamplingSpanProcessor.java
```diff
@@ -0,0 +1,564 @@
...
+    private final long priority;
+
+    public static ReadableSpanWithPriority create(
+        ReadableSpan readableSpan, RandomGenerator randomGenerator) {
+      String otelTraceStateString =
+          readableSpan.getSpanContext().getTraceState().get(OtelTraceState.TRACE_STATE_KEY);
+      OtelTraceState otelTraceState = OtelTraceState.parse(otelTraceStateString);
+      int pval;
+      int rval;
+      long priority = randomGenerator.nextLong();
```
PeterF778, line 64: I understand that among spans with the same r-value, the **priority** decides which spans will be removed from the reservoir. Giving each span a new random value may break trace completeness to a larger degree than it is necessary.
To alleviate this problem, the priority might be made dependent on trace-id, for example by combining a hash of trace-id and a (less significant) random number. Since there are no guarantees about quality of trace-id in general, I do not advocate to do it by default, but it would be nice to refactor the code so it would be easy to do for the users who really want it.

oertl: Thanks, this improvement could be added in a later PR without breaking anything. As it would require choosing a hash function, which is a difficult decision by itself, much more discussion would be needed. (Ideally, it should be a fast, high-quality function for which there are compatible implementations in all relevant programming languages.) I think as a first step, this PR should be merged without this improvement.

PeterF778: What I meant by refactoring was something like calculating the priority using an overridable method with trace-id as a parameter. It would remain unused in the default implementation.
But I agree with you that it can be done later.

trask:
@PeterF778 thanks for reviewing! would you be interested in being a "component owner" for the consistent sampling component? (see #174) this is a tricky component and we only have a single component owner currently which slows down reviews and merges for it.

PeterF778:
Yes, I'd be interested! Thanks for the offer!
Peter

________________________________

trask:
great! I'll send a PR to add you

trask:
Closing and re-opening to see if reviewers get updated

trask:
> Closing and re-opening to see if reviewers get updated

It didn't work, I think because the branch is not up-to-date with main (which has the new component owners updates).

This is not a problem though.

@PeterF778 if you can finish reviewing and/or approve then I will do a quick check and merge. thx!

PeterF778 (approved):

consistent-sampling/src/main/java/io/opentelemetry/contrib/samplers/ConsistentReservoirSamplingSpanProcessor.java
```diff
@@ -0,0 +1,564 @@
...
+ */
+public final class ConsistentReservoirSamplingSpanProcessor implements SpanProcessor {
+
+  private static final String WORKER_THREAD_NAME =
+      ConsistentReservoirSamplingSpanProcessor.class.getSimpleName() + "_WorkerThread";
+
+  private final Worker worker;
+  private final AtomicBoolean isShutdown = new AtomicBoolean(false);
+
+  public static final long DEFAULT_EXPORT_TIMEOUT_NANOS = TimeUnit.SECONDS.toNanos(30);
```
trask, line 48: looks like this can be package-private

trask (approved):
thx @oertl!

consistent-sampling/src/main/java/io/opentelemetry/contrib/samplers/ConsistentReservoirSamplingSpanProcessor.java
oertl: thanks, addressed in e6d726c19e88243998b8396c7357ebf5614766b9

PeterF778 (approved):

jmacd (commented):
Thanks, sorry for the late review. I believe I understand what is being done.
LGTM

PR #354, opened 2022-06-13
Runtime attachment additions

jeanbisutti:
**Description:**

Runtime attachment additions

* Add tests
* Add log when agent is disabled with otel.javaagent.enabled property
* Add log when agent is disabled with OTEL_JAVAAGENT_ENABLED environment variable
* Do not try to attach if agent is already attached
* Do not try to attach if attachment is requested from the _main_ thread
* Do not try to attach if attachment is requested from the _main_ method
*  Add log if an unexpected issue has happened during attachment
* Add javadoc
*  Add information to the Readme

### DIFF
--- a/runtime-attach/README.md
+++ b/runtime-attach/README.md
@@ -18,7 +18,7 @@ public class SpringBootApp {
 
     public static void main(String[] args) {
         RuntimeAttach.attachJavaagentToCurrentJVM();
-        SpringApplication.run();
+        SpringApplication.run(SpringBootApp.class, args);
     }
 
 }
--- a/runtime-attach/README.md
+++ b/runtime-attach/README.md
@@ -1,6 +1,6 @@
 # Runtime attachment
 
-If you can't update the JVM arguments to attach the [OpenTelemetry Javaagent](https://github.com/open-telemetry/opentelemetry-java-instrumentation) (_-javaagent:path/to/opentelemetry-javaagent.jar_), this project allows you to do attachment programmatically.
+If you can't update the JVM arguments to attach the [OpenTelemetry Java agent](https://github.com/open-telemetry/opentelemetry-java-instrumentation) (_-javaagent:path/to/opentelemetry-javaagent.jar_), this project allows you to do attachment programmatically.
 
 The `io.opentelemetry.contrib.attach.RuntimeAttach` class has an `attachJavaagentToCurrentJVM` method allowing to trigger the attachment of the OTel agent for Java.
 
--- a/runtime-attach/README.md
+++ b/runtime-attach/README.md
@@ -1,3 +1,29 @@
+# Runtime attachment
+
+If you can't update the JVM arguments to attach [OTel agent for Java](https://github.com/open-telemetry/opentelemetry-java-instrumentation) (_-javaagent:path/to/opentelemetry-javaagent.jar_), this project allows you to do attachment programmatically.
+
+The ```io.opentelemetry.contrib.attach.RuntimeAttach``` class has an ```attachJavaagentToCurrentJVM``` method allowing to trigger the attachment of the OTel agent for Java.
+
+The attachment will not be initiated in the following cases:
+* The  ```otel.javaagent.enabled ``` property is set to ```false```
+* The  ```OTEL_JAVAAGENT_ENABLED ``` environment variable is set to ```false```
+* The attachment is not requested from the _main_ thread
+* The agent is already attached
+
+_The attachment must be requested at the beginning of the ```main``` method._ We give below an example for Spring Boot applications:
+
+```java
+@SpringBootApplication
+public class SpringBootApp {
+
+    public static void main(String[] args) {
+        RuntimeAttach.attachJavaagentToCurrentJVM();
+        SpringApplication.run();
+    }
+
+}
+```
+
 ## Component owners
 
 - [Nikita Salnikov-Tarnovski](https://github.com/iNikem), Splunk
--- a/runtime-attach/README.md
+++ b/runtime-attach/README.md
@@ -1,6 +1,6 @@
 # Runtime attachment
 
-If you can't update the JVM arguments to attach [OTel agent for Java](https://github.com/open-telemetry/opentelemetry-java-instrumentation) (_-javaagent:path/to/opentelemetry-javaagent.jar_), this project allows you to do attachment programmatically.
+If you can't update the JVM arguments to attach the [OpenTelemetry Javaagent](https://github.com/open-telemetry/opentelemetry-java-instrumentation) (_-javaagent:path/to/opentelemetry-javaagent.jar_), this project allows you to do attachment programmatically.
 
 The ```io.opentelemetry.contrib.attach.RuntimeAttach``` class has an ```attachJavaagentToCurrentJVM``` method allowing to trigger the attachment of the OTel agent for Java.
 
--- a/runtime-attach/README.md
+++ b/runtime-attach/README.md
@@ -8,9 +8,10 @@ The attachment will not be initiated in the following cases:
 * The `otel.javaagent.enabled` property is set to `false`
 * The `OTEL_JAVAAGENT_ENABLED` environment variable is set to `false`
 * The attachment is not requested from the _main_ thread
+* The attachment is not requested from the `public static void main(String[] args)` method
 * The agent is already attached
 
-_The attachment must be requested at the beginning of the `main` method._ We give below an example for Spring Boot applications:
+_The attachment must be requested at the beginning of the `public static void main(String[] args)` method._ We give below an example for Spring Boot applications:
 
 ```java
 @SpringBootApplication
--- a/runtime-attach/README.md
+++ b/runtime-attach/README.md
@@ -2,15 +2,15 @@
 
 If you can't update the JVM arguments to attach the [OpenTelemetry Javaagent](https://github.com/open-telemetry/opentelemetry-java-instrumentation) (_-javaagent:path/to/opentelemetry-javaagent.jar_), this project allows you to do attachment programmatically.
 
-The ```io.opentelemetry.contrib.attach.RuntimeAttach``` class has an ```attachJavaagentToCurrentJVM``` method allowing to trigger the attachment of the OTel agent for Java.
+The `io.opentelemetry.contrib.attach.RuntimeAttach` class has an `attachJavaagentToCurrentJVM` method allowing to trigger the attachment of the OTel agent for Java.
 
 The attachment will not be initiated in the following cases:
-* The  ```otel.javaagent.enabled ``` property is set to ```false```
-* The  ```OTEL_JAVAAGENT_ENABLED ``` environment variable is set to ```false```
+* The `otel.javaagent.enabled` property is set to `false`
+* The `OTEL_JAVAAGENT_ENABLED` environment variable is set to `false`
 * The attachment is not requested from the _main_ thread
 * The agent is already attached
 
-_The attachment must be requested at the beginning of the ```main``` method._ We give below an example for Spring Boot applications:
+_The attachment must be requested at the beginning of the `main` method._ We give below an example for Spring Boot applications:
 
 ```java
 @SpringBootApplication
--- a/runtime-attach/build.gradle.kts
+++ b/runtime-attach/build.gradle.kts
@@ -12,7 +12,7 @@ dependencies {
   // Used by byte-buddy but not brought in as a transitive dependency.
   compileOnly("com.google.code.findbugs:annotations")
 
-  testImplementation("io.opentelemetry.javaagent:opentelemetry-javaagent:1.13.1")
+  testImplementation("io.opentelemetry.javaagent:opentelemetry-javaagent:1.14.0")
   testImplementation("io.opentelemetry:opentelemetry-extension-annotations")
   testImplementation("org.junit.jupiter:junit-jupiter-api")
   testImplementation("org.junit-pioneer:junit-pioneer")
--- a/runtime-attach/build.gradle.kts
+++ b/runtime-attach/build.gradle.kts
@@ -11,4 +11,15 @@ dependencies {
 
   // Used by byte-buddy but not brought in as a transitive dependency.
   compileOnly("com.google.code.findbugs:annotations")
+
+  testImplementation("io.opentelemetry.javaagent:opentelemetry-javaagent:1.13.1")
+  testImplementation("io.opentelemetry:opentelemetry-extension-annotations:1.14.0")
+  testImplementation("org.junit.jupiter:junit-jupiter-api")
+  testImplementation("org.junit-pioneer:junit-pioneer")
+  testImplementation("org.assertj:assertj-core")
+}
+
+tasks.test {
+  useJUnitPlatform()
+  setForkEvery(1) // One JVM by test class to avoid a test class launching a runtime attachment influences the behavior of another test class
 }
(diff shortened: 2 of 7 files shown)


runtime-attach/build.gradle.kts
```diff
@@ -11,4 +11,15 @@ dependencies {
 
   // Used by byte-buddy but not brought in as a transitive dependency.
   compileOnly("com.google.code.findbugs:annotations")
+
+  testImplementation("io.opentelemetry.javaagent:opentelemetry-javaagent:1.13.1")
+  testImplementation("io.opentelemetry:opentelemetry-extension-annotations:1.14.0")
```
trask, line 16: I don't think the version is needed here, it should get picked up from `io.opentelemetry:opentelemetry-bom`
```suggestion
  testImplementation("io.opentelemetry:opentelemetry-extension-annotations")
```

runtime-attach/build.gradle.kts
```diff
@@ -11,4 +11,15 @@ dependencies {
 
   // Used by byte-buddy but not brought in as a transitive dependency.
   compileOnly("com.google.code.findbugs:annotations")
+
+  testImplementation("io.opentelemetry.javaagent:opentelemetry-javaagent:1.13.1")
```
trask, line 15: probably we should add to `dependencyManagement/build.gradle.kts` to resolve version there

```suggestion
  testImplementation("io.opentelemetry.javaagent:opentelemetry-javaagent:1.14.0")
```

runtime-attach/README.md
```diff
@@ -1,3 +1,29 @@
+# Runtime attachment
+
+If you can't update the JVM arguments to attach [OTel agent for Java](https://github.com/open-telemetry/opentelemetry-java-instrumentation) (_-javaagent:path/to/opentelemetry-javaagent.jar_), this project allows you to do attachment programmatically.
```
trask, line 3:
```suggestion
If you can't update the JVM arguments to attach the [OpenTelemetry Javaagent](https://github.com/open-telemetry/opentelemetry-java-instrumentation) (_-javaagent:path/to/opentelemetry-javaagent.jar_), this project allows you to do attachment programmatically.
```

runtime-attach/README.md
```diff
@@ -1,3 +1,29 @@
...
+
+The ```io.opentelemetry.contrib.attach.RuntimeAttach``` class has an ```attachJavaagentToCurrentJVM``` method allowing to trigger the attachment of the OTel agent for Java.
```
trask, line 5: I believe triple ``` can be single ` everywhere

runtime-attach/src/main/java/io/opentelemetry/contrib/attach/RuntimeAttach.java
```diff
@@ -7,22 +7,82 @@
 
 import java.io.File;
 import java.lang.management.ManagementFactory;
+import java.util.logging.Logger;
 import net.bytebuddy.agent.ByteBuddyAgent;
 
+/** This class allows you to attach the OTel agent for Java at runtime. */
```
trask, line 13:
```suggestion
/** This class allows you to attach the OpenTelemetry Javaagent at runtime. */
```

runtime-attach/src/main/java/io/opentelemetry/contrib/attach/RuntimeAttach.java
```diff
@@ -7,22 +7,82 @@
...
 public final class RuntimeAttach {
 
+  private static final Logger LOGGER = Logger.getLogger(RuntimeAttach.class.getName());
```
trask, line 16: to match convention in the OTel Java repos
```suggestion
  private static final Logger logger = Logger.getLogger(RuntimeAttach.class.getName());
```

runtime-attach/src/main/java/io/opentelemetry/contrib/attach/RuntimeAttach.java
```diff
@@ -7,22 +7,82 @@
...
+  private static final String AGENT_ENABLED_PROPERTY = "otel.javaagent.enabled";
+  private static final String AGENT_ENABLED_ENV_VAR = "OTEL_JAVAAGENT_ENABLED";
+  static final String MAIN_THREAD_CHECK_PROP = "otel.javaagent.runtimeattach.mainthreadcheck";
```
trask, line 19: I think adding `.testing` somewhere helps to make it clear this property is only for running tests (not for end users), we have some precedent for this in instrumentation repo, and for system property names we typically snake-case them, e.g.

```suggestion
  static final String MAIN_THREAD_CHECK_PROP = "otel.javaagent.testing.runtime-attach.main-thread-check";
```

runtime-attach/src/main/java/io/opentelemetry/contrib/attach/RuntimeAttach.java
```diff
@@ -7,22 +7,82 @@
...
     ByteBuddyAgent.attach(javaagentFile, getPid());
+
+    if (!agentIsAttached()) {
+      LOGGER.warning("Agent was not attached. An unexpected issue has happened.");
+    }
+  }
+
+  private static boolean shouldAttach() {
+    if (agentIsDisabledWithProp()) {
+      LOGGER.warning("Agent was disabled with " + AGENT_ENABLED_PROPERTY + " property.");
```
trask, line 40: I think the first 3 warnings in this method could probably be dropped to debug, e.g. similar message in the agent: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/cc30b5218bf931a842649dc5161fe1253d7db5b5/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java#L93

runtime-attach/src/main/java/io/opentelemetry/contrib/attach/RuntimeAttach.java
```diff
@@ -7,22 +7,82 @@
...
+      LOGGER.warning("Agent was disabled with " + AGENT_ENABLED_ENV_VAR + " environment variable.");
+      return false;
+    }
+    if (agentIsAttached()) {
+      LOGGER.warning("Agent is already attached. It is not attached a second time.");
+      return false;
+    }
+    if (mainThreadCheckIsEnabled() && !isMainThread()) {
+      LOGGER.warning(
+          "Agent is not attached because runtime attachment was requested from main thread.");
```
trask, line 53:
```suggestion
      LOGGER.warning(
          "Agent is not attached because runtime attachment was not requested from main thread.");
```

runtime-attach/src/main/java/io/opentelemetry/contrib/attach/RuntimeAttach.java
```diff
@@ -7,22 +7,82 @@
...
+
+  private static boolean shouldAttach() {
+    if (agentIsDisabledWithProp()) {
+      LOGGER.warning("Agent was disabled with " + AGENT_ENABLED_PROPERTY + " property.");
+      return false;
+    }
+    if (agentIsDisabledWithEnvVar()) {
+      LOGGER.warning("Agent was disabled with " + AGENT_ENABLED_ENV_VAR + " environment variable.");
+      return false;
+    }
```
trask, line 46: do we need to check the env var and system property? it may be ok for the runtime attach component to try to attach and the javaagent will disable itself if these are set

runtime-attach/src/main/java/io/opentelemetry/contrib/attach/RuntimeAttach.java
```diff
@@ -7,22 +7,82 @@
...
+      Class.forName("io.opentelemetry.javaagent.OpenTelemetryAgent", false, null);
+      return true;
+    } catch (ClassNotFoundException e) {
+      return false;
+    }
+  }
+
+  private static boolean mainThreadCheckIsEnabled() {
+    String mainThreadCheck = System.getProperty(MAIN_THREAD_CHECK_PROP);
+    return !"false".equals(mainThreadCheck);
```
trask, line 80: since this is an internal test-only flag, it may be worth inverting it (so the default is `false`, e.g. `...-check-disabled`), then can use `return Boolean.getBoolean(MAIN_THREAD_CHECK_PROP)` here

trask (commented):
thx!

runtime-attach/README.md
jeanbisutti: Done

runtime-attach/src/main/java/io/opentelemetry/contrib/attach/RuntimeAttach.java
jeanbisutti: Done

runtime-attach/src/main/java/io/opentelemetry/contrib/attach/RuntimeAttach.java
jeanbisutti: I understand your point of view @trask . It could make sense.  I would prefer not to try to attach, in such a case, to stop the things sooner. It's the [current behavior](https://github.com/open-telemetry/opentelemetry-java-contrib/commit/858dbae93b80081ebd47261f5dcb1b4abe387a04). What do you think @iNikem?

runtime-attach/src/main/java/io/opentelemetry/contrib/attach/RuntimeAttach.java
jeanbisutti: For me the code would be a little less readable

runtime-attach/src/main/java/io/opentelemetry/contrib/attach/RuntimeAttach.java
trask: I think this comment got collapsed so was missed (this collapsing is my least favorit github feature)

runtime-attach/README.md
```diff
@@ -1,3 +1,29 @@
+# Runtime attachment
+
+If you can't update the JVM arguments to attach the [OpenTelemetry Java agent](https://github.com/open-telemetry/opentelemetry-java-instrumentation) (_-javaagent:path/to/opentelemetry-javaagent.jar_), this project allows you to do attachment programmatically.
+
+The `io.opentelemetry.contrib.attach.RuntimeAttach` class has an `attachJavaagentToCurrentJVM` method allowing to trigger the attachment of the OTel agent for Java.
+
+The attachment will not be initiated in the following cases:
+* The `otel.javaagent.enabled` property is set to `false`
+* The `OTEL_JAVAAGENT_ENABLED` environment variable is set to `false`
+* The attachment is not requested from the _main_ thread
```
trask, line 10: what do you think about adding one more check, that the call stack only has `main` in it, i.e. that it was really called directly from the initial `main` method?

runtime-attach/build.gradle.kts
```diff
@@ -11,4 +11,15 @@ dependencies {
 
   // Used by byte-buddy but not brought in as a transitive dependency.
   compileOnly("com.google.code.findbugs:annotations")
+
+  testImplementation("io.opentelemetry.javaagent:opentelemetry-javaagent:1.14.0")
```
trask, line 15: can you move this version to dependency management section so we don't miss updating it?

runtime-attach/src/main/java/io/opentelemetry/contrib/attach/RuntimeAttach.java
```diff
@@ -7,22 +7,83 @@
...
+      logger.warning(
+          "Agent is not attached because runtime attachment was not requested from main thread.");
+      return false;
+    }
+    return true;
+  }
+
+  private static boolean agentIsDisabledWithProp() {
+    String agentEnabledPropValue = System.getProperty(AGENT_ENABLED_PROPERTY);
+    return "false".equals(agentEnabledPropValue);
```
trask, line 62: I think we use equalsIgnoreCase for boolean properties in the Javaagent, so may as well stay consistent
```suggestion
    return "false".equalsIgnoreCase(agentEnabledPropValue);
```

trask (approved):

runtime-attach/src/main/java/io/opentelemetry/contrib/attach/RuntimeAttach.java
trask: ah, I hadn't noticed this was the current behavior, let's keep it then, thx

runtime-attach/README.md
jeanbisutti: Implemented. The new code be tested with the following class:
``` java
public class MainClass {

  public static void main(String[] args) {
    System.out.println("isMainMethod = " + isMainMethod());
  }

  static boolean isMainMethod() {
    StackTraceElement bottomOfStack = findBottomOfStack(Thread.currentThread());
    String methodName = bottomOfStack.getMethodName();
    return "main".equals(methodName);
  }

  private static StackTraceElement findBottomOfStack(Thread thread) {
    StackTraceElement[] stackTrace = thread.getStackTrace();
    return  stackTrace[stackTrace.length - 1];
  }
}
```

runtime-attach/build.gradle.kts
jeanbisutti: Done

runtime-attach/README.md
trask: I think typically the `main` thread will always be rooted at the `main()` method(?), I was thinking of a more restrictive chck that `attach()` was called directly from `main()`

trask: I'm going to merge this for the 1.15.0 release, and we can discuss / follow-up on this point later

trask:
thx @jeanbisutti!

runtime-attach/src/main/java/io/opentelemetry/contrib/attach/RuntimeAttach.java
```diff
@@ -7,22 +7,99 @@
 
 import java.io.File;
 import java.lang.management.ManagementFactory;
+import java.util.logging.Logger;
 import net.bytebuddy.agent.ByteBuddyAgent;
 
+/** This class allows you to attach the OpenTelemetry Java agent at runtime. */
 public final class RuntimeAttach {
 
+  private static final Logger logger = Logger.getLogger(RuntimeAttach.class.getName());
```
mateuszrzeszutek, line 16: I might be a little bit late here, but... 😄 

Should we use JUL here? This class is supposed to be the very first thing called in `main`, right? JUL won't be set up this early, and these calls certainly won't be redirected to agent's `PatchLogger`. Wouldn't it be safer to use stderr here?

jeanbisutti: > Should we use JUL here? This class is supposed to be the very first thing called in main, right? JUL won't be set up this early, and these calls certainly won't be redirected to agent's PatchLogger. Wouldn't it be safer to use stderr here?

Probably! 

What about the logging added by a user? For example, if a user has a  Spring Boot application and adds an import related to logging (jul, Log4j, logback, ...) in the class annotated with `@SpringBootApplication`, it would also  have a problem? Right?

```java
...
import java.util.logging.Logger;
...
@SpringBootApplication
public class SpringBootApp {
    public static void main(String[] args) {
        RuntimeAttach.attachJavaagentToCurrentJVM();
        SpringApplication.run(SpringBootApp.class, args);
    }
}
```
@mateuszrzeszutek @trask What do you think?

trask: thx, I agree with changing the logging in the runtime attach code itself to use stderr so that we don't trigger loading of j.u.l too early, e.g. in case the user is using jboss-logging (maybe https://www.wildfly.org/news/2020/06/18/Bootable-jar-Wildfly-20/?)

I don't think we need to add any restrictions on the user's logging setup. If whatever they were doing before they added the "runtime attach snippet" worked, it should still work after they add the snippet.

PR #356, opened 2022-06-14
Release workflow simplifications

trask:
Same as https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/6169


.github/workflows/release.yml
```diff
@@ -116,6 +129,8 @@ jobs:
           cat > /tmp/release-notes.txt << EOF
           This release targets the OpenTelemetry SDK $VERSION.
 
+          Note that all artifacts other than \`io.opentelemetry.javaagent:opentelemetry-javaagent\` have the \`-alpha\` suffix attached to their version number, reflecting that they are still alpha quality and will continue to have breaking changes. Please see the [VERSIONING.md](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/VERSIONING.md#opentelemetry-java-instrumentation-versioning) for more details.
```
jack-berg, line 132: This doesn't apply here does it?

.github/workflows/release.yml
```diff
@@ -146,35 +165,125 @@ jobs:
...
         run: |
-          cp jmx-metrics/build/libs/opentelemetry-jmx-metrics-$VERSION-alpha.jar opentelemetry-jmx-metrics.jar
+          cp javaagent/build/libs/opentelemetry-javaagent-${VERSION}.jar opentelemetry-javaagent.jar
           gh release create --target $GITHUB_REF_NAME \
                             --title "Version $VERSION" \
                             --notes-file /tmp/release-notes.txt \
                             --discussion-category announcements \
                             v$VERSION \
-                            opentelemetry-jmx-metrics.jar
+                            opentelemetry-javaagent.jar
```
jack-berg, line 174: This doesn't seem right.

jack-berg (commented):
Might have noticed a couple of copy / paste errors

.github/workflows/release.yml
trask: oops!

.github/workflows/release.yml
trask: indeed it is not!

jack-berg (approved):

PR #358, opened 2022-06-16
Update to 1.15.0

jack-berg:


trask (approved):

jack-berg:
Not sure what's going on wit the failed integration-test build check.

trask:
> Not sure what's going on wit the failed integration-test build check.

ya, the error message seems pretty generic

I re-ran the build both on this PR and on `main` and it does looks like the issue is coming from the updates in this PR.

@Mrod1598 @rmfitzpatrick @dehaansa since the failure is coming from the jmx-metrics component, could one of you take a look at it (and why the underlying error doesn't appear to get logged)?

rmfitzpatrick:
Unfortunately the container logs aren't included in the gradle failure output but are in [the results](https://github.com/open-telemetry/opentelemetry-java-contrib/suites/6968958518/artifacts/272335058). I'll be sure to look into how this can be improved.

```
Exception in thread "main" io.opentelemetry.sdk.autoconfigure.spi.ConfigurationException: OTLP gRPC Metrics Exporter enabled but opentelemetry-exporter-otlp not found on classpath. Make sure to add it as a dependency to enable this feature.
````

Believe we need

```diff
diff --git a/jmx-metrics/build.gradle.kts b/jmx-metrics/build.gradle.kts
index 13769de..74e7a1c 100644
--- a/jmx-metrics/build.gradle.kts
+++ b/jmx-metrics/build.gradle.kts
@@ -40,6 +40,7 @@ dependencies {
   implementation("io.opentelemetry:opentelemetry-sdk-extension-autoconfigure")
   implementation("io.opentelemetry:opentelemetry-sdk-testing")
   implementation("io.opentelemetry:opentelemetry-exporter-logging")
+  implementation("io.opentelemetry:opentelemetry-exporter-otlp")
   implementation("io.opentelemetry:opentelemetry-exporter-prometheus")
   implementation("org.slf4j:slf4j-api")
   implementation("org.slf4j:slf4j-simple")
```

rmfitzpatrick:
I'm now running into failures that appear related to https://github.com/open-telemetry/opentelemetry-java/pull/4376


```
WARNING: An exception occurred invoking callback for Instrument activemq.memory.usage.
java.lang.UnsupportedOperationException: This aggregator does not support recording long values.
	at io.opentelemetry.sdk.metrics.internal.aggregator.AggregatorHandle.doRecordLong(AggregatorHandle.java:126)
	at io.opentelemetry.sdk.metrics.internal.aggregator.AggregatorHandle.recordLong(AggregatorHandle.java:117)
	at io.opentelemetry.sdk.metrics.internal.aggregator.AggregatorHandle.recordLong(AggregatorHandle.java:106)
	at io.opentelemetry.sdk.metrics.internal.aggregator.Aggregator.accumulateLongMeasurement(Aggregator.java:60)
	at io.opentelemetry.sdk.metrics.internal.state.AsynchronousMetricStorage.recordLong(AsynchronousMetricStorage.java:94)
	at io.opentelemetry.sdk.metrics.internal.state.SdkObservableMeasurement.record(SdkObservableMeasurement.java:96)
```

This is unexpected since `activemq.memory.usage` is using a double value callback: https://github.com/open-telemetry/opentelemetry-java-contrib/blob/89aa698ac5a466a906e4df28e495663eef7d5a56/jmx-metrics/src/main/resources/target-systems/activemq.groovy#L51. I can look at this tomorrow but if you're aware of obsoleted api usage that may be to blame it could be helpful.

trask:
thx @rmfitzpatrick!

I sent a fix to the PR for the otlp artifact issue you found.

my guess about the other issue is that it's caused by groovy runtime-binding here:

https://github.com/open-telemetry/opentelemetry-java-contrib/blob/main/jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/InstrumentHelper.groovy#L154

and because `SdkObservableMeasurement` implements both record(long) and record(double).

@jack-berg would it maybe work to make `SdkObservableMeasurement` an abstract class which implements `ObservableMeasurement` and have concrete subclasses `SdkObservableLongMeasurement`/`SdkObservableDoubleMeasurement` which individually implement the respective interfaces?

anuraaga:
`SdkObservableMeasurement` is an internal class, I'd be surprised if we would need to modify it for groovy magic. Presumably adding static API types to the Groovy code will help?

trask:
@jack-berg @anuraaga do you think it's worth tightening up the definition of backwards-incompatible change, since currently https://github.com/open-telemetry/opentelemetry-java/blob/main/VERSIONING.md says incompatible changes include:

> Behavior changes that can require code using the artifact to be changed

jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/InstrumentHelper.groovy
```diff
@@ -118,7 +118,7 @@ class InstrumentHelper {
             def description = tuple.getAt(2)
             def unit = tuple.getAt(3)
 
-            if (instrumentIsObserver(instrument)) {
+            if (instrumentIsDoubleObserver(instrument) || instrumentIsLongObserver(instrument)) {
```
rmfitzpatrick, line 121: Thanks for taking care of this. We have unit tests for the helpers that would need to be updated: https://github.com/open-telemetry/opentelemetry-java-contrib/blob/main/jmx-metrics/src/test/java/io/opentelemetry/contrib/jmxmetrics/InstrumenterHelperTest.java#L628

jack-berg: Yup! Just saw those and pushed a change.

rmfitzpatrick (approved):

jack-berg:
@trask not exactly sure how to proceed. The behavior this code is relying on is outside the normal bounds of how we expect the API to be used, and is only possible by ignoring or casting types. Groovy takes away the guardrails and weird stuff happens 🤷

trask:
ya, I think it's ok to do nothing (and if asked, say that our compatibility guarantee doesn't apply to dynamic JVM languages)

I do think it's an interesting case since it's just "standard" Groovy code using our "standard" public APIs

rmfitzpatrick:
To be fair the registered metric helpers are written in java and provide the expected observable measurement consumer but the groovy context wasn't expecting the overloaded unsupported methods. I have no problem with the required workaround (we probably should have used coercion to begin with) but the change may be a form of footgun.

PR #359, opened 2022-06-17
Fix "URI is not hierarchical" during runtime attachment

jeanbisutti:
When the runtime attachment feature was used with a Spring Boot application (jar), the runtime attachement dit not work and  the following stack trace appeared:
```
    Caused by: java.lang.IllegalArgumentException: URI is not hierarchical
    	at java.base/java.io.File.<init>(File.java:420)
    	at io.opentelemetry.contrib.attach.AgentFileLocator.locateAgentFile(AgentFileLocator.java:25)
    	at io.opentelemetry.contrib.attach.RuntimeAttach.attachJavaagentToCurrentJVM(RuntimeAttach.java:18)
```

This PR fixes the issue.


runtime-attach/src/main/java/io/opentelemetry/contrib/attach/RuntimeAttach.java
```diff
@@ -15,8 +15,15 @@ public static void attachJavaagentToCurrentJVM() {
...
-    File javaagentFile = AgentFileLocator.locateAgentFile();
-    ByteBuddyAgent.attach(javaagentFile, getPid());
+
+    AgentFileProvider agentFileProvider = new AgentFileProvider();
+    File javaagentFile = agentFileProvider.getAgentFile();
+
+    try {
+      ByteBuddyAgent.attach(javaagentFile, getPid());
+    } finally {
+      agentFileProvider.deleteTempDir();
```
trask, line 25: I would think that the JVM would still need access to the jar file for lazy loading additional classes? (in which case we could delete on exit instead)

trask (commented):
is there a way to know whether copying is needed, and only copy in that case?

runtime-attach/src/main/java/io/opentelemetry/contrib/attach/RuntimeAttach.java
jeanbisutti: Not sure because the Javadoc of Byte Buddy's `attach(File agentJar, String processId)` method says _"This operation blocks until the attachment is complete"_.  In case of doubt, I will delete in exit.

jeanbisutti:
> is there a way to know whether copying is needed, and only copy in that case?

The problem has appeared when the application was packaged as a jar file (I expect the same problem with a war file). The only solution I have found is to use a temp jar file (I have tried to write code with real path but without success). The same thing is done in [Elastic agent](https://github.com/elastic/apm-agent-java/blob/9db44a05b135420a7e7333f03389b6881cadcb07/apm-agent-attach/src/main/java/co/elastic/apm/attach/ElasticApmAttacher.java#L208). We could use a ` try/catch `with `new File(URI)`  to only copy when it is needed. But I see several  drawbacks to this approach:
- It would only be related to the case where the code is executed in an IDE
- It would add more code to maintain
- The fast tests of this [PR](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/354) would not use the same execution path as in production

trask:
makes sense, we may want to revisit later due to users on read-only file systems, but I agree with keeping it simple for now

jeanbisutti:
> makes sense, we may want to revisit later due to users on read-only file systems, but I agree with keeping it simple for now

We may use the ` install()` method of Byte Buddy

trask (approved):

runtime-attach/src/main/java/io/opentelemetry/contrib/attach/AgentFileProvider.java
```diff
@@ -0,0 +1,75 @@
...
+
+    Path tempAgentJarPath = createTempAgentJarFile(tempDirPath);
+
+    deleteTempDirOnJvmExit(tempDirPath, tempAgentJarPath);
+
+    return tempAgentJarPath.toFile();
+  }
+
+  private static void deleteTempDirOnJvmExit(Path tempDirPath, Path tempAgentJarPath) {
+    if (tempDirPath != null && tempAgentJarPath != null) {
```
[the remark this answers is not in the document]
jeanbisutti: This can't be null but i was afraid by build checks 😅

trask:
@jeanbisutti looks like this needs a `spotlessApply` then ready for merge!

jeanbisutti:
> I pushed spotlessApply 👍

Thank you! Maybe it would be possible to automatically add a _spotless:apply_ commit in the CI

PR #371, opened 2022-06-21
Remove logger from runtime attachment

jeanbisutti:
See [this](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/354#discussion_r902163975)

### DIFF
--- a/runtime-attach/src/main/java/io/opentelemetry/contrib/attach/RuntimeAttach.java
+++ b/runtime-attach/src/main/java/io/opentelemetry/contrib/attach/RuntimeAttach.java
@@ -7,13 +7,11 @@ package io.opentelemetry.contrib.attach;
 
 import java.io.File;
 import java.lang.management.ManagementFactory;
-import java.util.logging.Logger;
 import net.bytebuddy.agent.ByteBuddyAgent;
 
 /** This class allows you to attach the OpenTelemetry Java agent at runtime. */
 public final class RuntimeAttach {
 
-  private static final Logger logger = Logger.getLogger(RuntimeAttach.class.getName());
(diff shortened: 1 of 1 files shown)


mateuszrzeszutek (approved):
Thanks!

trask (approved):

runtime-attach/src/main/java/io/opentelemetry/contrib/attach/RuntimeAttach.java
```diff
@@ -32,30 +30,32 @@ public static void attachJavaagentToCurrentJVM() {
...
 
     if (!agentIsAttached()) {
-      logger.warning("Agent was not attached. An unexpected issue has happened.");
+      printError("Agent was not attached. An unexpected issue has happened.");
     }
   }
 
+  @SuppressWarnings("SystemOut")
+  private static void printError(String message) {
+    System.err.println(message);
```
trask, line 39:
```suggestion
    // not using java.util.logging in order to avoid initializing the global LogManager
    // too early (and incompatibly with the user's app),
    // and because this is too early to use the Javaagent's PatchLogger
    System.err.println(message);
```

laurit: I'm not sure whether mentioning PatchLogger here makes sense as this code is independent from the java agent.

trask: can you clarify what you mean by independent from the java agent? from my perspective I think of it as very dependent since it is attaching the java agent at runtime

PR #377, opened 2022-06-28
Fix #376: `AwsXrayRemoteSampler` doesn’t poll for update

### LINKED ISSUE #376: AWS X-Ray remote sampler does not poll for new rules
**Description**

AWS X-Ray remote sampler does not poll for updated sampling configuration in the default configuration.

Any configuration which has a polling interval smaller than roughly 3.3 minutes is expected to work currently.

I already addeed a comment in #133 but decided to open a separate issue instead.

**Steps to reproduce**

Run in default configuration.

**Expectation**

I expect the sampling configuration to be fetched every 5 minutes.

**What applicable config did you use?**

AWS X-Ray OpenTelemetry distro Java agent with java option `-Dotel.traces.sampler=xray`

**Relevant Environment Information**

Version: 1.14.0 but main should be affecteed as well.
OS: macOS and AWS Fargate

**Additional context**

I configured the `-Dotel.traces.sampler=xray` on my development machine where it is expected to fail as I don't have any AWS proxy running locally but should still at least try every 5 minutes.

I set a breakpoint here

https://github.com/open-telemetry/opentelemetry-java-contrib/blob/7c23e02e2b3a6ddca888d71806071dd86a51d5bb/aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsXrayRemoteSampler.java#L151

and noticed that in my case `jitterNanos = -1294967296`. This screams truncation error so I checked where `jitterNanos` is set. It is set here

https://github.com/open-telemetry/opentelemetry-java-contrib/blob/7c23e02e2b3a6ddca888d71806071dd86a51d5bb/aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsXrayRemoteSampler.java#L98

so 1% of the `pollingIntervalNanos` but cast to `int`. 

Default is `pollingIntervalNanos = 300000000000 = 3 * 10^11`, so `pollingIntervalNanos / 100 = 3 * 10^9` which is bigger than `Integer.MAX_VALUE = 2147483647 ~= 2.1 * 10^9`

I see the reason why it is cast to int is because there is a method `Random.nextInt(Int)` but no method `Random.nextLong(Long)`.

I think the easiest fix would be to store the polling interval in milliseconds instead. I don't think nanoseconds makes sense here.

felixscheinost:
**Description:**

In the default configuration `pollingIntervalNanos = 3 * 10^11` so `pollingIntervalMillis / 100 > Integer.MAX_VALUE`.

Switch to storing everything in milliseconds instead. This should be a non-breaking change as all the nanoseconds related logic isn’t part of the public API.

**Existing Issue(s):**

#376

**Testing:**

I don't know how this could be tested as all the logic isn't part of the public API.

~~Best would be a sort of integration test where time could be skipped forward and we could observe a request. Is there some code already set up in this project to do something like this?~~
EDIT: I noticed that there both real integration tests and a test that tests polling with a very short polling interval.

I tested this manually by shadowing the two affected classes in my project and manually testing there.

### DIFF
--- a/aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsXrayRemoteSampler.java
+++ b/aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsXrayRemoteSampler.java
@@ -17,8 +17,10 @@ import io.opentelemetry.sdk.trace.data.LinkData;
 import io.opentelemetry.sdk.trace.samplers.Sampler;
 import io.opentelemetry.sdk.trace.samplers.SamplingResult;
 import java.io.Closeable;
+import java.time.Duration;
 import java.time.Instant;
 import java.util.Date;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Random;
@@ -49,7 +51,7 @@ public final class AwsXrayRemoteSampler implements Sampler, Closeable {
   // Unique per-sampler client ID, generated as a random string.
   private final String clientId;
   private final long pollingIntervalNanos;
-  private final int jitterNanos;
+  private final Iterator<Long> jitterNanos;
 
   @Nullable private volatile ScheduledFuture<?> pollFuture;
   @Nullable private volatile ScheduledFuture<?> fetchTargetsFuture;
@@ -94,8 +96,8 @@ public final class AwsXrayRemoteSampler implements Sampler, Closeable {
     sampler = initialSampler;
 
     this.pollingIntervalNanos = pollingIntervalNanos;
-    // Add ~1% of jitter. Truncating to int is safe for any practical polling interval.
-    jitterNanos = (int) (pollingIntervalNanos / 100);
+    // Add ~1% of jitter
+    jitterNanos = RANDOM.longs(0, pollingIntervalNanos / 100).iterator();
 
     // Execute first update right away on the executor thread.
     executor.execute(this::getAndUpdateSampler);
@@ -148,8 +150,23 @@ public final class AwsXrayRemoteSampler implements Sampler, Closeable {
   }
 
   private void scheduleSamplerUpdate() {
-    long delay = pollingIntervalNanos + RANDOM.nextInt(jitterNanos);
+    long delay = pollingIntervalNanos + jitterNanos.next();
     pollFuture = executor.schedule(this::getAndUpdateSampler, delay, TimeUnit.NANOSECONDS);
+  }
+
+  /**
+   * returns the duration until the next scheduled sampler update or null if no next update is
+   * scheduled yet.
+   *
+   * <p>only used for testing.
+   */
+  @Nullable
+  Duration getNextSamplerUpdateScheduledDuration() {
+    ScheduledFuture<?> pollFuture = this.pollFuture;
+    if (pollFuture == null) {
+      return null;
+    }
+    return Duration.ofNanos(pollFuture.getDelay(TimeUnit.NANOSECONDS));
   }
 
   private void fetchTargets() {
--- a/aws-xray/src/test/java/io/opentelemetry/contrib/awsxray/AwsXrayRemoteSamplerTest.java
+++ b/aws-xray/src/test/java/io/opentelemetry/contrib/awsxray/AwsXrayRemoteSamplerTest.java
@@ -168,6 +168,25 @@ class AwsXrayRemoteSamplerTest {
     }
   }
 
+  // https://github.com/open-telemetry/opentelemetry-java-contrib/issues/376
+  @Test
+  void testJitterTruncation() {
+    try (AwsXrayRemoteSampler samplerWithLongerPollingInterval =
+        AwsXrayRemoteSampler.newBuilder(Resource.empty())
+            .setInitialSampler(Sampler.alwaysOn())
+            .setEndpoint(server.httpUri().toString())
+            .setPollingInterval(Duration.ofMinutes(5))
+            .build()) {
+      assertThat(samplerWithLongerPollingInterval.getNextSamplerUpdateScheduledDuration()).isNull();
+      await()
+          .untilAsserted(
+              () -> {
+                assertThat(samplerWithLongerPollingInterval.getNextSamplerUpdateScheduledDuration())
+                    .isCloseTo(Duration.ofMinutes(5), Duration.ofSeconds(10));
+              });
+    }
+  }
+
   private static SamplingDecision doSample(Sampler sampler, String name) {
     return sampler
         .shouldSample(


willarmiros (commented):
Thanks for catching this & and contributing this fix! Just to confirm, did you run the integration tests in the package as part of this change?

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsXrayRemoteSampler.java
```diff
@@ -148,8 +149,8 @@ private void getAndUpdateSampler() {
   }
 
   private void scheduleSamplerUpdate() {
-    long delay = pollingIntervalNanos + RANDOM.nextInt(jitterNanos);
-    pollFuture = executor.schedule(this::getAndUpdateSampler, delay, TimeUnit.NANOSECONDS);
+    long delay = pollingIntervalMillis + RANDOM.nextInt(jitterMillis);
```
anuraaga, line 152: Can we stick with nanos and switch to `Random.longs(0, jitterNanos)` instead of switching to millis? It's nice to avoid dropping precision mysteriously when possible, even in practice it generally shouldn't matter here.

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsXrayRemoteSampler.java
```diff
@@ -148,10 +149,23 @@ private void getAndUpdateSampler() {
...
+    @SuppressWarnings("OptionalGetWithoutIsPresent") // getAsLong: RANDOM.longs should always provide a value, so we can do an unchecked unwrap here
+    long delay = pollingIntervalNanos + RANDOM.longs(0, jitterNanos).findFirst().getAsLong();
     pollFuture = executor.schedule(this::getAndUpdateSampler, delay, TimeUnit.NANOSECONDS);
   }
 
+  /**
+   * returns the duration until the next scheduled sampler update or null if no next update is scheduled yet
+   */
+  @Nullable
+  public Duration getNextSamplerUpdateScheduledDuration() {
```
felixscheinost, line 161: This is currently only required for the test. Would that be okay?

I don't have an idea how to test this properly because we don't want to wait for 5 minutes in the test, right? As only long polling intervals trigger this bug so we should use a long polling interval.

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsXrayRemoteSampler.java
felixscheinost: Good point! This requires significantly less changes. I pushed an updated commit.

felixscheinost:
@willarmiros Does running the integration tests mean running `./gradlew :aws-xray:awsTest`? 

If yes, running this command returns an error on my machine:

```
> Task :aws-xray:compileAwsTestJava FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':aws-xray:compileAwsTestJava'.
> Could not resolve all files for configuration ':aws-xray:awsTestCompileClasspath'.
   > Could not find io.opentelemetry:opentelemetry-exporter-otlp-trace:.
     Required by:
         project :aws-xray
```

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsXrayRemoteSampler.java
willarmiros: Hmmm I think the maintainers would have final say but I don't think we want to expose additional public methods in the sampler like this. Is it possible to mock the system clock and trick the test into thinking 5 minutes have passed? Or just otherwise verifying the interval isn't negative when set to 5 minutes?

willarmiros (commented):
@felixscheinost for the test failure it looks like you're not resolving the package version for some reason... Maybe you need to do a build first? https://github.com/open-telemetry/opentelemetry-java-contrib#getting-started

Not sure if @anuraaga would have any other ideas why this could happen.

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsXrayRemoteSampler.java
trask: can it be package-private and still accessed from test?

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsXrayRemoteSampler.java
```diff
@@ -148,10 +149,26 @@ private void getAndUpdateSampler() {
   }
 
   private void scheduleSamplerUpdate() {
-    long delay = pollingIntervalNanos + RANDOM.nextInt(jitterNanos);
+    @SuppressWarnings(
+        "OptionalGetWithoutIsPresent") // getAsLong: RANDOM.longs should always provide a value, so
+    // we can do an unchecked unwrap here
+    long delay = pollingIntervalNanos + RANDOM.longs(0, jitterNanos).findFirst().getAsLong();
```
anuraaga, line 155: Let's replace the `jitterNanos` field with `RANDOM.longs(0, jitterNanos)` to not construct the stream every time.

anuraaga:
Sorry for the confusion on `awsTest` - it's not run with the build currently since we don't have AWS credentials for integration tests in this repo. It seems to have drifted and doesn't build anymore... It's also not a true integration test as it just polls and expects manually checking the logs after operating the console. @willarmiros I'd suggest someone at AWS checks out the PR and runs it manually until we could have AWS creds in this repo for running integ tests.

felixscheinost:
I think one way to test that the scheduling works correctly without the `getNextSamplerUpdateScheduledDuration` method I added would be to add (also package-private) a way to override scheduler. The scheduler could also be set in the builder and passeed into the sampler.

This way we could use a mocked scheduler that could verify what's scheduleed.

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsXrayRemoteSampler.java
felixscheinost: Sorry, I am not very used to working with streams. How would I get the next value from a stream? From reading examples and the documentation a little streams aren't supposed to use that way?

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsXrayRemoteSampler.java
felixscheinost: Good point, I made the method package-private.

aws-xray/src/test/java/io/opentelemetry/contrib/awsxray/AwsXrayRemoteSamplerTest.java
```diff
@@ -168,6 +168,27 @@ void defaultInitialSampler() {
     }
   }
 
+  // https://github.com/open-telemetry/opentelemetry-java-contrib/issues/376
+  @Test
+  void testJitterTruncation() {
+    sampler.close();
```
anuraaga, line 174: Let's use the same pattern as `defaultInitialSampler()`, not operating on the test's "base sampler" and defining a local one within try/resources.

anuraaga (commented):
Thanks - just small nit left

anuraaga (approved):

willarmiros (approved):

PR #389, opened 2022-07-11
Remove hardcoded agent version in runtime attachment

jeanbisutti:
Remove hardcoded agent version in runtime attachment

PR #392, opened 2022-07-11
Remove hardcoded agent version

trask:
alternative to #389

now I (partly) understand the point of `isCanBeResolved` and `isCanBeConsumed` in:

```
... by configurations.creating {
  isCanBeResolved = true
  isCanBeConsumed = false
}
```

see https://github.com/open-telemetry/opentelemetry-java-contrib/blob/7d19e674f301838a528266a7be4f3778e77c795b/buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts#L94-L96


mateuszrzeszutek (approved):
Nice! 🚀

PR #409, opened 2022-07-21
Prevent the runtime attachment from launching multiple times

jeanbisutti:
Prevent the runtime attachment from launching multiple times

### DIFF
--- a/runtime-attach/runtime-attach-core/src/main/java/io/opentelemetry/contrib/attach/CoreRuntimeAttach.java
+++ b/runtime-attach/runtime-attach-core/src/main/java/io/opentelemetry/contrib/attach/CoreRuntimeAttach.java
@@ -19,6 +19,8 @@ public final class CoreRuntimeAttach {
 
   private final String agentJarResourceName;
 
+  private boolean runtimeAttachmentRequested;
+
   /**
    * Creates a new {@code DistroRuntimeAttach} from the resource name of the agent jar.
    *
@@ -36,6 +38,11 @@ public final class CoreRuntimeAttach {
     if (!shouldAttach()) {
       return;
     }
+
+    if (runtimeAttachmentRequested) {
+      return;
+    }
+    runtimeAttachmentRequested = true;
 
     AgentFileProvider agentFileProvider = new AgentFileProvider(agentJarResourceName);
 


runtime-attach/runtime-attach-core/src/main/java/io/opentelemetry/contrib/attach/CoreRuntimeAttach.java
```diff
@@ -37,6 +39,11 @@ public void attachJavaagentToCurrentJVM() {
       return;
     }
 
+    if (runtimeAttachmentRequested) {
+      return;
+    }
+    runtimeAttachmentRequested = true;
```
mateuszrzeszutek, line 45: I'm just wondering, what kind of situations are we trying to prevent here? A simple duplicated line of code? Or runtime attach being called from completely different places in the application? Would it make sense to print out a warning with a stack trace to help pinpoint the unnecessary invocations?

mateuszrzeszutek (approved):

runtime-attach/runtime-attach-core/src/main/java/io/opentelemetry/contrib/attach/CoreRuntimeAttach.java
jeanbisutti: This PR aims to prevent the runtime attachment from being invoked several times from the main method. If a runtime attachment is invoked while another runtime attachment is in progress and not finished, I am not sure about the result with JVM attachment mechanism. The instrumentation might not be correctly installed.

>  Would it make sense to print out a warning with a stack trace to help pinpoint the unnecessary invocations?

Nothing is displayed to be consistent with the current display policy. [First, this type of message was reported with a debug log](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/354#discussion_r896359682). [After, the warning log messages were replaced with `System.err.println` and the debug messages removed](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/371). Perhaps it may be worth rediscussing this display policy.

mateuszrzeszutek: > If a runtime attachment is invoked while another runtime attachment is in progress and not finished

Again, is that a valid scenario? Don't we already have a check that verifies that only the `main` thread can call this method?
Otherwise, if a concurrent scenario is possible, then this code needs to be thread safe (use `AtomicBoolean` instead).

jeanbisutti: > Don't we already have a check that verifies that only the main thread can call this method?
> Otherwise, if a concurrent scenario is possible, then this code needs to be thread safe (use AtomicBoolean instead).
> 

The scenario is about two runtime attachments requested from the main thread since the new check is after the main thread check.


`Again, is that a valid scenario?`

I have not today enough knowledge about the `Attach Listener` JVM thread implementation.  I would prefer to stop the things ealier because it is not the kind of issue I would like to debug. We can close this PR on OTel if you prefer.

mateuszrzeszutek: > I would prefer to stop the things ealier because it is not the kind of issue I would like to debug.

👍 

> We can close this PR on OTel if you prefer.

No, I approve of this PR, I think this kind of check definitely makes sense. I was wondering whether this was motivated by a real use case, and if so, what kind of issue it was.

PR #417, opened 2022-08-12
Support traceId-based r-values

trask:
hey @oertl! I'm interested in using ConsistentProbabilityBasedSampler and ConsistentRateLimitingSampler, but I'm currently limited by backwards compatibility with an existing ecosystem which requires sampling to be based on a specific hash of the traceId (where lower hash values are more likely to be sampled-in).

I realize that I can't get true parity with any traceId-based ratio sampler, since the consistent samplers have some randomization in the decision making when the probability is not a power of 1/2, but I think I can live with that variance.

This PR explores an option to allow this kind of traceId-based decision when constructing the r-value.

I realize it's not an ideal end-state, but I think it could be a good bridge for us (and maybe others?) to the new consistent samplers.


consistent-sampling/src/main/java/io/opentelemetry/contrib/samplers/ConsistentAlwaysOffSampler.java
```diff
@@ -10,7 +10,9 @@
 @Immutable
 final class ConsistentAlwaysOffSampler extends ConsistentSampler {
 
-  private ConsistentAlwaysOffSampler() {}
+  private ConsistentAlwaysOffSampler() {
+    super(s -> RandomGenerator.getDefault().numberOfLeadingZerosOfRandomLong());
```
trask, line 14: @oertl what's the reason for having/using a ConsistentAlwaysOffSampler vs a normal always off sampler?

oertl: The `ConsistentAlwaysOffSampler` sets an r-value when it is not present. This way, the spans of different child branches will all have the same r-value and the corresponding sampling decisions will be consistent. Furthermore, it sets the p-value to `OtelTraceState.getMaxP()` which corresponds to a probability of 0, so that a subsequent `ConsistentParentBasedSampler` can make the sampling decision based on the p-value and not on the sampled flag.

consistent-sampling/src/main/java/io/opentelemetry/contrib/samplers/ConsistentSampler.java
```diff
@@ -69,20 +90,21 @@ static final ConsistentSampler probabilityBased(
...
    * or falls-back to the given sampler if it is a root span.
    *
    * @param rootSampler the root sampler
-   * @param randomGenerator a random generator
+   * @param rValueGenerator the function to use for generating the r-value
    */
-  static final ConsistentSampler parentBased(
-      ConsistentSampler rootSampler, RandomGenerator randomGenerator) {
-    return new ConsistentParentBasedSampler(rootSampler, randomGenerator);
+  static ConsistentSampler parentBased(
```
oertl, line 105: missing `public`?

consistent-sampling/src/main/java/io/opentelemetry/contrib/samplers/ConsistentSampler.java
```diff
@@ -47,8 +48,25 @@ public static final ConsistentSampler alwaysOff() {
...
+        samplingProbability,
+        s -> randomGenerator.numberOfLeadingZerosOfRandomLong(),
+        randomGenerator);
+  }
+
+  /**
+   * Returns a {@link ConsistentSampler} that samples each span with a fixed probability.
+   *
+   * @param samplingProbability the sampling probability
+   * @param rValueGenerator the function to use for generating the r-value
```
oertl, line 63: Probably needs some more documentation. The `rValueGenerator` must map a trace ID to a value in range [0,62] with probabilities as specified in https://opentelemetry.io/docs/reference/specification/trace/tracestate-probability-sampling/#appendix. I would prefer a new functional interface instead of `ToIntFunction<String>` where all this can be documented.

PeterF778: I agree. A new interface could be also used to define the default behavior "s -> RandomGenerator.getDefault().numberOfLeadingZerosOfRandomLong()" and avoid repeating this code phrase in multiple places.

consistent-sampling/src/main/java/io/opentelemetry/contrib/samplers/ConsistentSampler.java
```diff
@@ -212,8 +240,7 @@ public final SamplingResult shouldSample(
 
     // generate new r-value if not available
     if (!otelTraceState.hasValidR()) {
-      otelTraceState.setR(
-          Math.min(randomGenerator.numberOfLeadingZerosOfRandomLong(), OtelTraceState.getMaxR()));
+      otelTraceState.setR(Math.min(rValueGenerator.applyAsInt(traceId), OtelTraceState.getMaxR()));
```
PeterF778, line 243: I'd love to see this approach made more general with rValueGenerator taking not only traceId, but also the Attributes as arguments.
The use case for this is to assign the same rValue to a group of traces, so they are likely to survive sampling as an intact group - still while ensuring the correct statistical distribution of the rValues over the whole trace population.

consistent-sampling/src/main/java/io/opentelemetry/contrib/samplers/ConsistentSampler.java
trask: Great ideas, implemented. I put the default behavior in a separate class `RValueGenerators` to keep it package-private.

oertl (approved):

PeterF778 (approved):

mateuszrzeszutek (approved):

PR #427, opened 2022-08-23
Add Spring Boot service name guesser / ResourceProvider

breedx-splk:
This may not be the final home for `ResourceProvider` implementations, but it's a start. I know that the core project is likely going to migrate at least some of the existing `ResourceProvider`s out of core....maybe some go to the instrumentation repo and some go to contrib? I'm open to reorganization suggestions. 

This adds a new module that contains a single `ResourceProvider` called `SpringBootServiceNameGuesser` that does what it says. There are some additional, more complicated strategies that have not yet been included (such as the inline JSON doc as environment variable).

This new module will publish a new contrib artifact that distros can include to do spring app name detection.

This class could have been built as a composition of each of the strategies with a `Resource.merge()` step at the end, but that felt a little overengineered.


resource-providers/src/test/java/io/opentelemetry/resource/spring/SpringBootServiceNameGuesserTest.java
```diff
@@ -0,0 +1,111 @@
...
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+@ExtendWith(MockitoExtension.class)
+class SpringBootServiceNameGuesserTest {
+
+  public static final String PROPS = "application.properties";
+  public static final String APPLICATION_YML = "application.yml";
```
mateuszrzeszutek, line 26: Nit
```suggestion
  static final String PROPS = "application.properties";
  static final String APPLICATION_YML = "application.yml";
```

resource-providers/src/main/java/io/opentelemetry/resource/spring/SpringBootServiceNameGuesser.java
```diff
@@ -0,0 +1,255 @@
...
+    @SuppressWarnings("unchecked")
+    String attemptGetCommandLineViaReflection() throws Exception {
+      Class<?> clazz = Class.forName("java.lang.ProcessHandle");
+      Method currentMethod = clazz.getDeclaredMethod("current");
+      Method infoMethod = clazz.getDeclaredMethod("info");
+      Object currentInstance = currentMethod.invoke(null);
+      Object info = infoMethod.invoke(currentInstance);
+      Method commandLineMethod = info.getClass().getInterfaces()[0].getMethod("commandLine");
+      Object optionalCommandLine = commandLineMethod.invoke(info);
+      return ((Optional<String>) optionalCommandLine).get();
```
mateuszrzeszutek, line 252: Nit: avoid creating NPE:
```suggestion
      return ((Optional<String>) optionalCommandLine).orElse(null);
```

resource-providers/src/main/java/io/opentelemetry/resource/spring/SpringBootServiceNameGuesser.java
```diff
@@ -0,0 +1,255 @@
...
+      Map<String, Object> data = yaml.load(in);
+      Map<String, Map<String, Object>> spring =
+          (Map<String, Map<String, Object>>) data.get("spring");
+      if (spring != null) {
+        Map<String, Object> app = spring.get("application");
+        if (app != null) {
+          Object name = app.get("name");
+          if (name != null) {
+            return (String) name;
+          }
```
mateuszrzeszutek, line 158: No need for a null check here:
```suggestion
          return (String) app.get("name");
```

mateuszrzeszutek:
I'm in favor of including this in the base javaagent distro -- it looks like it'll be very useful for spring boot users, and the accuracy of guessing the service name is rather high I think.

resource-providers/src/main/java/io/opentelemetry/resource/spring/SpringBootServiceNameGuesser.java
```diff
@@ -0,0 +1,255 @@
...
+     * succeed on java 9+.
+     */
+    @SuppressWarnings("unchecked")
+    String attemptGetCommandLineViaReflection() throws Exception {
+      Class<?> clazz = Class.forName("java.lang.ProcessHandle");
+      Method currentMethod = clazz.getDeclaredMethod("current");
+      Method infoMethod = clazz.getDeclaredMethod("info");
+      Object currentInstance = currentMethod.invoke(null);
+      Object info = infoMethod.invoke(currentInstance);
+      Method commandLineMethod = info.getClass().getInterfaces()[0].getMethod("commandLine");
```
laurit, line 250: Using `arguments` instead of `commandLine` could allow accurately detecting application name that contains spaces.
I'd probably add `Class<?> infoClazz = Class.forName("java.lang.ProcessHandle$Info");` and use that instead of `info.getClass().getInterfaces()[0]`

resource-providers/src/main/java/io/opentelemetry/resource/spring/SpringBootServiceNameGuesser.java
```diff
@@ -0,0 +1,255 @@
...
+  private String readNameFromAppProperties() {
+    return loadFromClasspath(
+        "application.properties", SpringBootServiceNameGuesser::getAppNamePropertyFromStream);
+  }
+
+  @Nullable
+  private static String getAppNamePropertyFromStream(InputStream in) {
+    Properties properties = new Properties();
+    try {
+      properties.load(in);
```
laurit, line 204: would it make sense to add a comment that `Properties.load` use `ISO-8859-1` encoding which is also what spring uses by default for properties files?

resource-providers/src/main/java/io/opentelemetry/resource/spring/SpringBootServiceNameGuesser.java
```diff
@@ -0,0 +1,255 @@
...
+    logger.debug("Performing Spring Boot service name auto-detection...");
+    Stream<Supplier<String>> finders =
+        Stream.of(
+            this::findByEnvironmentVariable,
+            this::findBySystemProperties,
+            this::findByClasspathApplicationProperties,
+            this::findByCurrentDirectoryApplicationProperties,
+            this::findByClasspathApplicationYaml,
+            this::findByCurrentDirectoryApplicationYaml,
+            this::findByCommandlineArgument);
```
laurit, line 78: I believe you need to adjust the order of these. For example command line arguments should have higher priority than properties file.

resource-providers/src/main/java/io/opentelemetry/resource/spring/SpringBootServiceNameGuesser.java
```diff
@@ -0,0 +1,255 @@
...
+    String getProperty(String key) {
+      return System.getProperty(key);
+    }
+
+    InputStream openClasspathResource(String filename) {
+      return getClass().getClassLoader().getResourceAsStream(filename);
+    }
+
+    InputStream openFile(String filename) throws Exception {
+      return Files.newInputStream(Paths.get(filename));
```
laurit, line 236: I would check that the file exists before attempting to open it

trask:
> I'm in favor of including this in the base javaagent distro -- it looks like it'll be very useful for spring boot users, and the accuracy of guessing the service name is rather high I think.

+1

resource-providers/src/main/java/io/opentelemetry/resource/spring/SpringBootServiceNameGuesser.java
```diff
@@ -0,0 +1,255 @@
...
+ *   <li>Check for application.yml in the current working dir
+ *   <li>Check for --spring.application.name program argument (not jvm arg) via ProcessHandle
+ *   <li>Check for --spring.application.name program argument via sun.java.command system property
+ * </ul>
+ */
+@AutoService(ResourceProvider.class)
+public class SpringBootServiceNameGuesser implements ResourceProvider {
+
+  private static final Pattern COMMANDLINE_PATTERN =
+      Pattern.compile(".*--spring\\.application\\.name=([a-zA-Z.\\-_]+).*");
```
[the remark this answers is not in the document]
breedx-splk: I wanted to include the hyphen `-` and the period `.` because I think people use those pretty frequently in application names and they are not part of `\w`. 😿

resource-providers/src/main/java/io/opentelemetry/resource/spring/SpringBootServiceNameGuesser.java
breedx-splk: Sure, since we chatted about encodings, I think that makes sense. 👍🏻

resource-providers/src/main/java/io/opentelemetry/resource/spring/SpringBootServiceNameGuesser.java
breedx-splk: This is a good point -- I hadn't considered the user/ergonomics of this ordering decision. I had attempted to guess at the most common/frequent things first...but you're totally right, a user that bothers to put stuff on the commandline or environment definitely wants to override what might be on the classpath or filesystem. Will reorder.

resource-providers/src/main/java/io/opentelemetry/resource/spring/SpringBootServiceNameGuesser.java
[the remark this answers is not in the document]
laurit: you could just take everything up to the first space `([^ ]+)`

resource-providers/src/main/java/io/opentelemetry/resource/spring/SpringBootServiceNameGuesser.java
breedx-splk: Why, just to avoid the FileNotFound exception? I'm pretty "meh" on this. 

I mean, if we're going to do that we might as well check to ensure that the file exists AND that it's a regular file AND that I have permission to read it also, right?

laurit: This can be super annoying if you need to debug some exception during agent startup and use an exception breakpoint and you get a million random exceptions before the one where you want to stop. Imo it is best not to cause exceptions if you can easily avoid it. I'd add regular file check, but I wouldn't bother with checking permissions.

breedx-splk: Is it more annoying that code that's just littered with checked exception handlers and guarded code all over the place? 😁 

In this case, if I add check, what should I do in the event that the file doesn't exist? 

Apparently I shouldn't throw a different exception, so I guess I slap `@Nullable` on the method and return a `null` `InputStream`? That will cause an NPE in the caller unless guarded, and the callers are currently using try-with-resources. So I'll have to unwrap those into separate assignments. Yuck. Alternatively, I could return a `ByteArrayInputStream` with nothing in it and hope that the parsers can grok an empty stream without exploding -- but returning an empty stream also suggests to any caller that the file was actually read but contains nothing. That's not great either. I dunno. Situations like this always feel like a language shortcoming to me.

Coincidentally, I was reading the [nodejs docs on `fs.exists()`](https://nodejs.org/api/fs.html#fsexistspath-callback) yesterday and they explicitly discourage checking before attempting because it's not multi-thread/process safe. That doesn't really apply here, but I thought it was interesting nonetheless.

laurit: As far as I know try with resource works nicely with `null`, it won't NPE.

breedx-splk:
Closing and will reopen in the main instrumentation repo.

resource-providers/src/main/java/io/opentelemetry/resource/spring/SpringBootServiceNameGuesser.java
breedx-splk: Yeah this is much nicer. I've aded that in the instrumentation PR. Thank you!

PR #446, opened 2022-09-21
More dependabot

trask:
Similar to work going on in instrumentation repo: https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/6645


trask:
In this repo, we may be able to be more aggressive and search all gradle files, but I think this is a good first start

mateuszrzeszutek (approved):

PR #472, opened 2022-09-22
Target maven 3.5.0

### LINKED ISSUE #467: What is the minimum maven version to support for maven-extension and static-instrumenter?
It looks like `maven-extension` supports 3.5.0+, that seems like a reasonable baseline, and we could update `static-instrumenter` to the same.

trask:
Resolves #467

### DIFF
--- a/micrometer-meter-provider/build.gradle.kts
+++ b/micrometer-meter-provider/build.gradle.kts
@@ -9,7 +9,7 @@ dependencies {
   api("io.opentelemetry:opentelemetry-api")
   api("io.opentelemetry:opentelemetry-sdk-metrics")
 
-  compileOnly("io.micrometer:micrometer-core:1.1.0") // do not update this version
+  compileOnly("io.micrometer:micrometer-core:1.1.0") // do not auto-update this version
   compileOnly("io.opentelemetry:opentelemetry-sdk-extension-autoconfigure")
 
   annotationProcessor("com.google.auto.service:auto-service")
--- a/static-instrumenter/maven-plugin/build.gradle.kts
+++ b/static-instrumenter/maven-plugin/build.gradle.kts
@@ -18,7 +18,7 @@ dependencies {
   compileOnly("org.slf4j:slf4j-api")
 
   testImplementation("org.apache.maven.plugin-tools:maven-plugin-annotations:3.6.4")
-  testImplementation("org.apache.maven:maven-core:3.8.6")
+  testImplementation("org.apache.maven:maven-core:3.5.0")
   testImplementation("org.slf4j:slf4j-simple")
 }
 
--- a/static-instrumenter/maven-plugin/build.gradle.kts
+++ b/static-instrumenter/maven-plugin/build.gradle.kts
@@ -11,14 +11,14 @@ description = "Maven3 plugin for static instrumentation of projects code and dep
 base.archivesName.set("static-instrumentation-maven-plugin")
 
 dependencies {
-  implementation("org.apache.maven:maven-plugin-api:3.8.6")
+  implementation("org.apache.maven:maven-plugin-api:3.5.0") // do not auto-update this version
   implementation("org.apache.maven:maven-project:2.2.1")
   compileOnly("org.apache.maven.plugin-tools:maven-plugin-annotations:3.6.4")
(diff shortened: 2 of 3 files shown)


static-instrumenter/maven-plugin/build.gradle.kts
```diff
@@ -11,14 +11,14 @@ description = "Maven3 plugin for static instrumentation of projects code and dep
...
   implementation("org.apache.maven:maven-project:2.2.1")
   compileOnly("org.apache.maven.plugin-tools:maven-plugin-annotations:3.6.4")
-  compileOnly("org.apache.maven:maven-core:3.5.0")
+  compileOnly("org.apache.maven:maven-core:3.5.0") // do not auto-update this version
   compileOnly("org.slf4j:slf4j-api")
 
-  testImplementation("org.apache.maven.plugin-tools:maven-plugin-annotations:3.6.0")
-  testImplementation("org.apache.maven:maven-core:3.5.0")
+  testImplementation("org.apache.maven.plugin-tools:maven-plugin-annotations:3.6.4")
+  testImplementation("org.apache.maven:maven-core:3.8.6")
```
trask, line 21: @cyrille-leclerc two questions for you here:
* is it ok to use latest `maven-plugin-annotations` artifact, or should we pin that also (I couldn't figure out which, if any, version of `maven-plugin-annotations` corresponds to maven 3.5.0)
* do you have a preference for whether we test against `3.5.0` or test against the latest version? I'm ok with whichever you think is better

micrometer-meter-provider/build.gradle.kts
```diff
@@ -9,7 +9,7 @@ dependencies {
   api("io.opentelemetry:opentelemetry-api")
   api("io.opentelemetry:opentelemetry-sdk-metrics")
 
-  compileOnly("io.micrometer:micrometer-core:1.1.0") // do not update this version
+  compileOnly("io.micrometer:micrometer-core:1.1.0") // do not auto-update this version
```
mateuszrzeszutek, line 12: Is there any reason we're targeting 1.1 here? Micrometer 1.1 is super old, and not compatible with the more recent releases - the API stabilized around 1.5, and since then all the changes have been strictly incremental. Before that there's been a few breaking changes.

static-instrumenter/maven-plugin/build.gradle.kts
cyrille-leclerc: > is it ok to use latest `maven-plugin-annotations` artifact, or should we pin that also (I couldn't figure out which, if any, version of `maven-plugin-annotations` corresponds to maven 3.5.0)

Yes we should bump to the latest version of `maven-plugin-annotations`.

The [Maven Plugin Tool for Annotations](https://maven.apache.org/plugin-tools/maven-plugin-tools-annotations/index.html#) is a Maven tool with a versioning that is mostly decoupled from the version of Maven Core. The latest `maven-plugin-annotations`, the v3.6.4, requires `maven-core` v3.2.5, a very old version (see [here](https://github.com/apache/maven-plugin-tools/blob/maven-plugin-tools-3.6.4/pom.xml#L96))

> do you have a preference for whether we test against 3.5.0 or test against the latest version? I'm ok with whichever you think is better

I'm wondering if it would be beneficial to run the automated tests with Maven 3.5.0 that is less often used as we can assume that the most recent version of Maven Core will be tested by the Static Instrumenter developers are very likely to have this version installed on their workstation.

micrometer-meter-provider/build.gradle.kts
trask: makes sense, I've opened #476 to track and will send follow-up PR to make sure @HaloFour has a chance to review

static-instrumenter/maven-plugin/build.gradle.kts
trask: Thanks for the clarification on the annotations artifact, and testing against maven 3.5.0 sounds like a good plan, I've updated this PR

micrometer-meter-provider/build.gradle.kts
HaloFour: That's fine by me, I was concerned about targeting a version that would support a critical mass of users.

mateuszrzeszutek (approved):

PR #474, opened 2022-09-23
Fix static instrumenter test broken since 1.18.0

### LINKED ISSUE #436: Static instrumenter may be broken since 1.18.0
The static instrumenter artifacts are not yet published, so this is not urgent, but wanted to open tracking issue so not forgotten about. I had to disable its `JarTest` which was failing after updating to 1.18.0.

jeanbisutti:
Fix #436 

It's not a regression coming from 1.18.0.

The failing test has been modified, mainly:
* The jar used was regenerated from [test-app](https://github.com/open-telemetry/opentelemetry-java-contrib/tree/main/static-instrumenter/test-app) project. The reason is that the previous jar contained OTel classes that are now supposed to be provided by the `no-inst-agent` jar. The build/tests may be improved later to avoid having jar files in `src/test/resources`.
* [The failing assertion depended on the error output](https://github.com/open-telemetry/opentelemetry-java-contrib/blob/b931535ae742fec72f328585f56ba2bf7fa56ce8/static-instrumenter/agent-instrumenter/src/integrationTest/java/JarTest.java#L71). This assertion now depends on the standard output.


static-instrumenter/agent-instrumenter/src/integrationTest/java/JarTest.java
```diff
@@ -43,17 +46,25 @@ void testSampleJar() throws Exception {
...
+          .contains(
+              "Sharing is only supported for boot loader classes because bootstrap classpath has been appended");
+    }
+
     assertThat(instrumentationProcess.exitValue()).isEqualTo(0);
-    assertThat(instrumentationLog).isNotEmpty();
 
     Path resultAppPath = outPath.resolve("app.jar");
+    assertThat(resultAppPath).exists();
     assertThat(Files.exists(resultAppPath)).isTrue();
```
mateuszrzeszutek, line 65: These two lines assert the same thing

static-instrumenter/agent-instrumenter/src/integrationTest/java/JarTest.java
```diff
@@ -43,17 +46,25 @@ void testSampleJar() throws Exception {
...
-    String instrumentationLog =
-        new String(instrumentationIn.readAllBytes(), StandardCharsets.UTF_8);
+    boolean isHotJdkJvm = System.getProperty("java.vm.name").contains("OpenJDK");
+    if (isHotJdkJvm) {
+      InputStream errorOutput = instrumentationProcess.getErrorStream();
+      String errorOutputAsString = new String(errorOutput.readAllBytes(), StandardCharsets.UTF_8);
+      assertThat(errorOutputAsString)
+          .isNotEmpty()
+          .contains(
+              "Sharing is only supported for boot loader classes because bootstrap classpath has been appended");
```
mateuszrzeszutek, line 58: Do we actually need to assert on this string? It does not seem to be important to this scenario.

static-instrumenter/agent-instrumenter/src/integrationTest/java/JarTest.java
```diff
@@ -43,17 +46,25 @@ void testSampleJar() throws Exception {
...
 
     Process instrumentationProcess = instrumentationProcessBuilder.start();
-    InputStream instrumentationIn = instrumentationProcess.getErrorStream();
     instrumentationProcess.waitFor(10, TimeUnit.SECONDS);
 
-    String instrumentationLog =
-        new String(instrumentationIn.readAllBytes(), StandardCharsets.UTF_8);
+    boolean isHotJdkJvm = System.getProperty("java.vm.name").contains("OpenJDK");
+    if (isHotJdkJvm) {
+      InputStream errorOutput = instrumentationProcess.getErrorStream();
```
mateuszrzeszutek, line 53: Can you use try-with-resources for IO streams?

static-instrumenter/agent-instrumenter/src/integrationTest/java/JarTest.java
```diff
@@ -64,13 +75,14 @@ void testSampleJar() throws Exception {
...
     Process runtimeProcess = runtimeProcessBuilder.start();
-    InputStream err = runtimeProcess.getErrorStream();
     runtimeProcess.waitFor(10, TimeUnit.SECONDS);
 
     assertThat(runtimeProcess.exitValue()).isEqualTo(0);
-    assertThat(new String(err.readAllBytes(), StandardCharsets.UTF_8))
-        .startsWith(
-            "[main] INFO io.opentelemetry.exporter.logging.LoggingSpanExporter - 'HTTP GET'");
+
+    InputStream standardOutputOfInstrumentedApp = runtimeProcess.getInputStream();
```
mateuszrzeszutek, line 82: Can you use try-with-resources for IO streams?

static-instrumenter/agent-instrumenter/src/integrationTest/java/JarTest.java
```diff
@@ -43,17 +46,25 @@ void testSampleJar() throws Exception {
             outPath.toString());
 
     Process instrumentationProcess = instrumentationProcessBuilder.start();
-    InputStream instrumentationIn = instrumentationProcess.getErrorStream();
     instrumentationProcess.waitFor(10, TimeUnit.SECONDS);
 
-    String instrumentationLog =
-        new String(instrumentationIn.readAllBytes(), StandardCharsets.UTF_8);
+    boolean isHotJdkJvm = System.getProperty("java.vm.name").contains("OpenJDK");
+    if (isHotJdkJvm) {
```
mateuszrzeszutek, line 52: Does the instrumentation process print out output only on OpenJDK? I'm wondering what's the reason behind this condition

jeanbisutti: It's to check that the java agent is started. This message is generated for OpenJDK. I would prefer to have this check to fail fast if something unexpected happens.

static-instrumenter/agent-instrumenter/src/integrationTest/java/JarTest.java
jeanbisutti: See my comment above.

jeanbisutti: See my comment below.

static-instrumenter/agent-instrumenter/src/integrationTest/java/JarTest.java
jeanbisutti: I have forgotten to remove the previous way of doing this assertion. Fixed.

static-instrumenter/agent-instrumenter/src/integrationTest/java/JarTest.java
jeanbisutti: The `close()` of `InputStream` does nothing (see https://docs.oracle.com/javase/8/docs/api/java/io/InputStream.html#close--).

trask: Subclasses of InputStream may though

static-instrumenter/agent-instrumenter/src/integrationTest/java/JarTest.java
trask: could we check for `INFO io.opentelemetry.javaagent.tooling.VersionLogger - opentelemetry-javaagent` instead? I think the purpose of that check would probably be clearer than checking the class sharing message

jeanbisutti: Done. I was not aware of this OTel message. It's better. However, this message and the debug messages are printed into the error output. Do you know why?

static-instrumenter/agent-instrumenter/src/integrationTest/java/JarTest.java
```diff
@@ -43,34 +45,39 @@ void testSampleJar() throws Exception {
...
             APP_MAIN);
 
     Process runtimeProcess = runtimeProcessBuilder.start();
-    InputStream err = runtimeProcess.getErrorStream();
     runtimeProcess.waitFor(10, TimeUnit.SECONDS);
 
     assertThat(runtimeProcess.exitValue()).isEqualTo(0);
-    assertThat(new String(err.readAllBytes(), StandardCharsets.UTF_8))
-        .startsWith(
-            "[main] INFO io.opentelemetry.exporter.logging.LoggingSpanExporter - 'HTTP GET'");
```
trask, line 73: let's try to keep this assertion to verify that http client instrumentation was applied

trask (commented):
it would be great to understand what changed in 1.18.0 javaagent that caused the test to fail

static-instrumenter/agent-instrumenter/src/integrationTest/java/JarTest.java
trask: Looks like the logging isn't working since the Java agent started using SPI to initialize its InternalLogger, because while the Java SDK SPIs are loaded and executed at runtime (when running statically instrumented application), the Java agent SPIs are not. I'll open an issue to track this.

trask:
> it would be great to understand what changed in 1.18.0 javaagent that caused the test to fail

see comment above

trask (approved):

PR #477, opened 2022-09-24
More dependabot

trask:
probably need to revisit later what's a "core dependency" and what's not, but doesn't seem to be any real difference other than organization, see https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/6712


trask:
bump

jack-berg (approved):

PR #479, opened 2022-09-28
Static instrumentation: remove Gradle plugin mentions

jeanbisutti:
The Gradle plugin of the static instrumentation does not seem implemented.

This PR proposes to remove the mentions of this plugin to avoid misunderstandings.


trask (approved):
👍

PR #492, opened 2022-09-30
Disable by default the internal debug logs of static instrumentation

jeanbisutti:
Disable by default the internal debug logs of static instrumentation.

With DEBUG level some tests never seem to end with my computer.


mateuszrzeszutek (approved):

trask (approved):
sounds like maybe this issue:

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/dc7e072a5602ca25b113f45b32b8887c1cc8d4ec/muzzle/src/main/java/io/opentelemetry/javaagent/tooling/TransformSafeLogger.java#L16-L21

PR #497, opened 2022-10-04
Static instrumentation: replace NPE with a custom message

jeanbisutti:
In the Maven plugin for static instrumentation, an NPE was raised if no file was found for the project artifact. Now it is replaced by a custom message.

### DIFF
--- a/static-instrumenter/maven-plugin/src/main/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/ProjectModel.java
+++ b/static-instrumenter/maven-plugin/src/main/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/ProjectModel.java
@@ -57,7 +57,10 @@ class ProjectModel {
             + " not found. The available artifacts are: "
             + toFileNames(allArtifacts)
             + ". Project artifact: "
-            + project.getArtifact();
+            + project.getArtifact()
+            + ". "
+            + "File of project artifact: "
+            + project.getArtifact().getFile();
     throw new IllegalArgumentException(message);
   }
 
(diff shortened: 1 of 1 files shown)


static-instrumenter/maven-plugin/src/main/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/ProjectModel.java
```diff
@@ -50,18 +51,42 @@ List<Path> chooseForInstrumentation(@Nullable String artifactName) {
         return new ArrayList<>(Collections.singletonList(artifactFile));
       }
     }
-    String message = "Artifact with name " + artifactName + " not found.";
-    logger.error(message);
```
mateuszrzeszutek, line 54: Should we keep that log statement?

mateuszrzeszutek (approved):

static-instrumenter/maven-plugin/src/main/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/ProjectModel.java
jeanbisutti: We don't have to keep it because [an exception with the message will be raised](https://github.com/jeanbisutti/opentelemetry-java-contrib/blob/c1ae5bb1f134094f325c8b1fd0e1e2d796fc6a2e/static-instrumenter/maven-plugin/src/main/java/io/opentelemetry/contrib/staticinstrumenter/plugin/maven/ProjectModel.java#L64) and the plugin execution will stop with this exception.

PR #509, opened 2022-10-11
Fix no Class in opentelemetry-runtime-attach jar

### LINKED ISSUE #508: No Class in opentelemetry-runtime-attach jar
**Description**
I wanted to try opentelemetry-runtime-attach module. 
I import it in maven project.
But when I would like to import `RuntimeAttach` : there is no class.
In jar, there is only otel-agent.jar in `io.opentelemetry.contrib.attach`

**Steps to reproduce**
[Just open jar from maven and see the content](https://search.maven.org/remotecontent?filepath=io/opentelemetry/contrib/opentelemetry-runtime-attach/1.18.0-alpha/opentelemetry-runtime-attach-1.18.0-alpha.jar)

**Expectation**
Find RuntimeAttach.class  in jar

**Relevant Environment Information**
Version: 1.18.0-alpha (but doesn't exist since 1.16.0-alpha)
OS: Ubuntu 20.04
Maven: 3.6.3

jeanbisutti:
Fix #508

### DIFF
--- a/runtime-attach/runtime-attach/build.gradle.kts
+++ b/runtime-attach/runtime-attach/build.gradle.kts
@@ -29,7 +29,7 @@ tasks {
     from({
       agent.singleFile
     })
-    rename { "otel-agent.jar" }
+    rename("^(.*)\\.jar\$", "otel-agent.jar")
   }
 
   withType<Test>().configureEach {


trask (approved):
I added topic to Thu SIG meeting to discuss moving runtime-attach artifact over to instrumentation repo so we can use the smoke-test infra over there to catch this kind of issue

PR #533, opened 2022-10-14
Populate log group in EKS Resource Detector

rapphil:
**Description:**

Populate log group in the EKS resource detector.

**Existing Issue(s):**

https://github.com/aws-observability/aws-otel-java-instrumentation/issues/205

**Testing:**

Unit tests were updated.

**Documentation:**

NA

**Outstanding items:**

NA

### DIFF
--- a/aws-resources/src/main/java/io/opentelemetry/contrib/aws/resource/EksResource.java
+++ b/aws-resources/src/main/java/io/opentelemetry/contrib/aws/resource/EksResource.java
@@ -17,6 +17,7 @@ import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Paths;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.logging.Level;
@@ -72,6 +73,9 @@ public final class EksResource {
     String clusterName = getClusterName(httpClient);
     if (clusterName != null && !clusterName.isEmpty()) {
       attrBuilders.put(ResourceAttributes.K8S_CLUSTER_NAME, clusterName);
+
+      String logGroupName = String.format("/aws/containerinsights/%s/application", clusterName);
+      attrBuilders.put(ResourceAttributes.AWS_LOG_GROUP_NAMES, Arrays.asList(logGroupName));
     }
 
     String containerId = dockerHelper.getContainerId();
--- a/aws-resources/src/test/java/io/opentelemetry/contrib/aws/resource/EksResourceTest.java
+++ b/aws-resources/src/test/java/io/opentelemetry/contrib/aws/resource/EksResourceTest.java
@@ -21,6 +21,7 @@ import io.opentelemetry.sdk.resources.Resource;
 import io.opentelemetry.semconv.resource.attributes.ResourceAttributes;
 import java.io.File;
 import java.io.IOException;
+import java.util.Arrays;
 import java.util.ServiceLoader;
 import org.junit.jupiter.api.Test;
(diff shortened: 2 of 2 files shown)


willarmiros (approved):
Thanks for adding!

rapphil:
Marking as draft to check if this is the best approach as this will only support Containers Insight.

aws-resources/src/main/java/io/opentelemetry/contrib/aws/resource/EksResource.java
```diff
@@ -72,6 +73,9 @@ static Resource buildResource(
     String clusterName = getClusterName(httpClient);
     if (clusterName != null && !clusterName.isEmpty()) {
       attrBuilders.put(ResourceAttributes.K8S_CLUSTER_NAME, clusterName);
+
+      String logGroupName = String.format("/aws/containerinsights/%s/application", clusterName);
```
BT-22, line 77: Hello, does this mean there must be a group name literally ends with `application` and the full name must be `"/aws/containerinsights/{clusterName}/application"`? 

Just wondering / confirming because I do not see there is such a log group name in my environment.

If that's true, then we'll need to adjust our log group setup.

willarmiros: This is good to know @BT-22 - this log group name is based on the log groups if you've setup CloudWatch Container Insights in your EKS cluster: https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/Container-Insights-setup-logs-FluentBit.html#Container-Insights-FluentBit-setup

Do you know of anywhere in your EKS environment where you can inspect the name of the CW Log group you're sending log data to? A configuration file for instance?

BT-22: Thanks for the swift reply @willarmiros - while I'm still finding more details in my environment, I learned the following first:

> we are using a gruntworks module that wraps the fluent bit helm chart https://github.com/aws/eks-charts/tree/master/stable/aws-for-fluent-bit
> there's only one log group with that

Where I get the container logs is `/aws/eks/{clusterName}/container-logs`

willarmiros: I see! In the repo you linked, it says you configure the log group destination: https://github.com/aws/eks-charts/tree/master/stable/aws-for-fluent-bit#configuration

So I wonder if we can read that config from the resource detector somehow...

lorelei-rupp-imprivata: Yeah I do not think you can assume everyone uses the same log group name. Depending on how people get their container logs to cloudwatch. We like to have ours formatted a specific way, so I think this needs to be exposed in your collector so you can specify what it is. Ours is "/aws/eks/<eks clustername}/container-logs" as well. As BT-22 says, we are using fluent bit, but we pass the log group to a terraform module(gruntworks) that intern calls the fluent bit helm chart. I think this needs to be exposed in the helm charts for rolling out the collectors that collect this stuff and send it to aws?

lorelei-rupp-imprivata: We actually use the adot operator and the adot collector statefulset to collect our traces and send to xray https://github.com/aws-observability/aws-otel-collector.

rapphil:
Closing this PR and opening https://github.com/open-telemetry/opentelemetry-java-contrib/issues/564. Discussion should be moved there.

PR #561, opened 2022-11-01
Minor build improvements

breedx-splk:
With some other work here I noticed some kotlin related warnings coming from the build. The change from `disabled_rules` to `ktlint_ disabled_rules` fixes that.

The other change relaxes where we are setting `options.release.set(8)` to only the `compileJava` tasks. The previous `  withType<JavaCompile>().configureEach {...}` seemed to be also matching on the test compilation, which meant that Java 11+ syntax could not be used in tests. With that change, it can be allowed.

### DIFF
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -19,7 +19,7 @@ dependencies {
 
 spotless {
   kotlinGradle {
-    ktlint().editorConfigOverride(mapOf("indent_size" to "2", "continuation_indent_size" to "2", "disabled_rules" to "no-wildcard-imports"))
+    ktlint().editorConfigOverride(mapOf("indent_size" to "2", "continuation_indent_size" to "2", "ktlint_disabled_rules" to "no-wildcard-imports"))
     target("**/*.gradle.kts")
   }
 }
--- a/buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
@@ -22,10 +22,12 @@ java {
 }
 
 tasks {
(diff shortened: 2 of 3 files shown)


trask:
> With some other work here I noticed some kotlin related warnings coming from the build. The change from `disabled_rules` to `ktlint_ disabled_rules` fixes that.

nice 👍, can you send PR to fix this in the other 2 repos also?

> The other change relaxes where we are setting `options.release.set(8)` to only the `compileJava` tasks. The previous ` withType<JavaCompile>().configureEach {...}` seemed to be also matching on the test compilation, which meant that Java 11+ syntax could not be used in tests. With that change, it can be allowed.

wouldn't using Java 11+ syntax mean that we can no longer run the tests on Java 8?

breedx-splk:
> nice 👍, can you send PR to fix this in the other 2 repos also?

Sure, can do.
 
> wouldn't using Java 11+ syntax mean that we can no longer run the tests on Java 8?

Yeah, that's true. Do we have that need? 
Maybe some tests are run in a matrix of versions? Straightforward unit tests should be fine to only run on 11 tho right?

breedx-splk:
Actually, I think this doesn't actually work.  Contrib isn't using any wildcard imports in the kotlin buildscripts, so the fact that we're breaking this isn't showing itself.

I think we'll have to live with the warnings until ktlint is updated within spotless.

trask:
> I think we'll have to live with the warnings until ktlint is updated within spotless.

bummer, thx for checking

> Maybe some tests are run in a matrix of versions?

it looks like we aren't running tests against a Java version matrix in this repo. in the other 2 repos, all tests are run against 8, 11, 17, and 19.

PR #563, opened 2022-11-02
Update the SDK and instrumentation versions to 1.19.2

opentelemetrybot:
Update the SDK and instrumentation versions to `1.19.2`.

Note: you will likely need to re-run the checks on this PR in an hour or so, once the
updated dependencies are visible in maven central.

PR #553, opened 2022-10-20
Update the SDK and instrumentation versions to 1.19.1

opentelemetrybot:
Update the SDK and instrumentation versions to `1.19.1`.

Note: you will likely need to re-run the checks on this PR in an hour or so, once the
updated dependencies are visible in maven central.

PR #567, opened 2022-11-03
Split versions

trask:
see #553 and #563


dependencyManagement/build.gradle.kts
```diff
@@ -12,17 +12,19 @@ val dependencyVersions = hashMapOf<String, String>()
 rootProject.extra["versions"] = dependencyVersions
 
 // this line is managed by .github/scripts/update-sdk-version.sh
-val otelVersion = "1.19.0"
+val otelSdkVersion = "1.19.0"
+// this line is managed by .github/scripts/update-instrumentation-version.sh
+val otelInstrumentationVersion = "1.19.0"
```
trask, line 17: intentionally left this at 1.19.0 to see if dependabot picks it up

... and if it doesn't, then I'll use it to test out updated workflow

jack-berg (approved):

.github/scripts/update-instrumentation-version.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash -e
+
+version=$1
+
+sed -Ei "s/val otelInstrumentationVersion = \"[^\"]*\"/val otelInstrumentationVersion = \"$version\"/" dependencyManagement/build.gradle.kts
```
jack-berg, line 5: Didn't check, but is this file executable?

jack-berg: Checked and it is. 👍

trask: thanks for checking, I usually forget until it fails during the release process, but somehow remembered this time 😂

PR #568, opened 2022-11-03
Use dependabot to update OTel SDK and Instrumentation dependencies

trask:
Seems simpler and more scalable to other repos, see https://github.com/open-telemetry/opentelemetry-collector-releases/issues/196#issuecomment-1258428187

### DIFF
--- a/dependencyManagement/build.gradle.kts
+++ b/dependencyManagement/build.gradle.kts
@@ -11,20 +11,14 @@ data class DependencySet(val group: String, val version: String, val modules: Li
 val dependencyVersions = hashMapOf<String, String>()
 rootProject.extra["versions"] = dependencyVersions
 
-// this line is managed by .github/scripts/update-sdk-version.sh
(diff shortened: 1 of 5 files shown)


dependencyManagement/build.gradle.kts
```diff
@@ -11,20 +11,14 @@ data class DependencySet(val group: String, val version: String, val modules: Li
 val dependencyVersions = hashMapOf<String, String>()
 rootProject.extra["versions"] = dependencyVersions
 
-// this line is managed by .github/scripts/update-sdk-version.sh
-val otelSdkVersion = "1.19.0"
-// this line is managed by .github/scripts/update-instrumentation-version.sh
-val otelInstrumentationVersion = "1.19.0"
```
trask, line 17: there seem to be some situations where dependabot understands version variables and some that it doesn't, but I gave up 🤷‍♂️

mateuszrzeszutek (approved):

PR #571, opened 2022-11-07
Thread count and classes loaded handlers

roberttoyonaga:
**Description:**
Add handlers for thread count and loaded classes count metrics.  These are derived from `jdk.ClassLoadingStatistics` and `jdk.JavaThreadStatistics` JFR events

**Testing:**
A new unit test was added to test these handlers in `jfr-streaming/src/test/java/io/opentelemetry/contrib/jfr/metrics`.


jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/classLoading/ClassesLoadedHandler.java
```diff
@@ -0,0 +1,50 @@
...
+
+  @Override
+  public String getEventName() {
+    return EVENT_NAME;
+  }
+
+  @Override
+  public void initializeMeter(Meter meter) {
+    meter
+        .gaugeBuilder(METRIC_NAME)
```
breedx-splk, line 40: This event has `unloadedClassCount` and `loadedClassCount`, both of which seem like they would be better modeled by monotonically increasing counters instead of a gauge. While unloads may be uncommon, they do happen, and if one of the goals of this class is to give an idea of the total classes _currently_ loaded, both values should be tracked and the delta reported in the callback.

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/classLoading/ClassesLoadedHandler.java
```diff
@@ -0,0 +1,50 @@
...
+        .ofLongs()
+        .setDescription(METRIC_DESCRIPTION)
+        .buildWithCallback(measurement -> measurement.record(value));
+  }
+
+  @Override
+  public Optional<Duration> getPollingDuration() {
+    return Optional.of(Duration.ofSeconds(1));
+  }
+}
```
breedx-splk, line 50: Would appreciate a unit test that covers the behavior of this class in isolation.

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/cpu/ThreadCountHandler.java
```diff
@@ -0,0 +1,50 @@
...
+        .ofLongs()
+        .setDescription(METRIC_DESCRIPTION)
+        .buildWithCallback(measurement -> measurement.record(value));
+  }
+
+  @Override
+  public Optional<Duration> getPollingDuration() {
+    return Optional.of(Duration.ofSeconds(1));
+  }
+}
```
breedx-splk, line 50: Would like to see a unit test that covers this class in isolation as well.

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/classLoading/ClassesLoadedHandler.java
breedx-splk: Also, please take a look at the [OpenTelemetry JVM Metrics specification](https://github.com/open-telemetry/opentelemetry-specification/blob/7c51a04570cca391eba8926958e6bfbf5682a955/specification/metrics/semantic_conventions/runtime-environment-metrics.md#jvm-metrics), which currently lists these metrics, which you should conform to:

* `process.runtime.jvm.classes.loaded`
* `process.runtime.jvm.classes.unloaded`
* `process.runtime.jvm.classes.current_loaded`

I think this handler can probably provide all 3, but if you wanted to wait an do `current_loaded` in another implementation/PR that would be fine too.

breedx-splk (changes requested):
Hi @roberttoyonaga. Thanks for this contribution. I made a few suggestions, but it looks pretty good overall. Appreciate it.

jfr-streaming/src/test/java/io/opentelemetry/contrib/jfr/metrics/JfrPeriodicMetricsTest.java
```diff
@@ -0,0 +1,83 @@
...
+    synchronized (obj) {
+      // Spin until everyone is at critical section. All threads must be running now.
+      while (count < THREADS) {
+        Thread.sleep(MILLIS);
+      }
+    }
+  }
+
+  @Test
+  public void shouldHaveJfrPeriodicEvents() throws Exception {
```
breedx-splk, line 47: I think there's bad precedent elsewhere in this repo, but with JUnit5 you shouldn't need the test class nor test method to be public I think.

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/cpu/ThreadCountHandler.java
```diff
@@ -0,0 +1,50 @@
...
+import static io.opentelemetry.contrib.jfr.metrics.internal.RecordedEventHandler.defaultMeter;
+
+import io.opentelemetry.api.metrics.Meter;
+import io.opentelemetry.contrib.jfr.metrics.internal.RecordedEventHandler;
+import java.time.Duration;
+import java.util.Optional;
+import jdk.jfr.consumer.RecordedEvent;
+
+public final class ThreadCountHandler implements RecordedEventHandler {
+  private static final String METRIC_NAME = "process.runtime.jvm.cpu.active_threads";
```
mateuszrzeszutek, line 17: The OpenTelemetry spec defines [`process.runtime.jvm.threads.count`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/runtime-environment-metrics.md#jvm-metrics), can you make sure that this implementation of JVM metrics conforms to it?

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/classLoading/ClassesLoadedHandler.java
roberttoyonaga: Thank you for the review! I've changed it to use monotonically increasing counters and also added an implementation for `process.runtime.jvm.classes.current_loaded` with an `upDownCounter`. I've corrected the instrument, unit, and descriptions so that they now should conform to the specification.

jfr-streaming/src/test/java/io/opentelemetry/contrib/jfr/metrics/JfrPeriodicMetricsTest.java
roberttoyonaga: done!

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/cpu/ThreadCountHandler.java
roberttoyonaga: Thanks for the review! I've changed the thread count handler so that it now properly implements the specification.

breedx-splk (approved):
Thanks for following up with these changes! Looking good to me.

It's a slight bummer that the tests currently have these sleeps in them. That approach tends not to scale well as the size of the codebase grows, but we can circle back on that another time. I still would have liked to see _unit_ tests for the handlers and not just _integration_ style tests, but that too can be added later.

breedx-splk (approved):

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/classes/ClassesLoadedHandler.java
```diff
@@ -0,0 +1,68 @@
...
+import io.opentelemetry.api.metrics.Meter;
+import io.opentelemetry.contrib.jfr.metrics.internal.RecordedEventHandler;
+import java.time.Duration;
+import java.util.Optional;
+import jdk.jfr.consumer.RecordedEvent;
+
+public final class ClassesLoadedHandler implements RecordedEventHandler {
+  private static final String METRIC_NAME_LOADED = "process.runtime.jvm.classes.loaded";
+  private static final String METRIC_NAME_UNLOADED = "process.runtime.jvm.classes.unloaded";
+  private static final String METRIC_NAME_CURRENT = "process.runtime.jvm.classes.current_loaded";
```
kittylyst, line 20: Do we document somewhere what the difference is between classes.loaded and classes.current_loaded? Maybe link to that definition here?

kittylyst (approved):
LGTM. I +1 @breedx-splk comments about not using sleep() in tests, but I don't think that should hold up a merge of this.

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/threads/ThreadCountHandler.java
```diff
@@ -0,0 +1,62 @@
...
+  @Override
+  public void initializeMeter(Meter meter) {
+    meter
+        .upDownCounterBuilder(METRIC_NAME)
+        .setDescription(METRIC_DESCRIPTION)
+        .setUnit(UNIT_THREADS)
+        .buildWithCallback(
+            measurement -> {
+              measurement.record(daemonCount, ATTR_DAEMON_TRUE);
+              measurement.record(activeCount - daemonCount, ATTR_DAEMON_FALSE);
```
mateuszrzeszutek, line 54: Nit: you can read the `daemonCount` volatile field just once, for better observation consistency:
```suggestion
              long d = daemonCount;
              measurement.record(d, ATTR_DAEMON_TRUE);
              measurement.record(activeCount - d, ATTR_DAEMON_FALSE);
```

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/Constants.java
```diff
@@ -43,4 +46,8 @@ private Constants() {}
   public static final AttributeKey<String> ATTR_ARENA_NAME = AttributeKey.stringKey("arena");
   public static final AttributeKey<String> ATTR_NETWORK_MODE = AttributeKey.stringKey("mode");
   public static final AttributeKey<String> ATTR_USAGE = AttributeKey.stringKey("usage.type");
+  public static final AttributeKey<String> ATTR_DAEMON = AttributeKey.stringKey(DAEMON);
```
mateuszrzeszutek, line 49: This should be a boolean attribute (see the [JMX based implementation](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/runtime-metrics/library/src/main/java/io/opentelemetry/instrumentation/runtimemetrics/Threads.java#L36) for comparison)

jfr-streaming/src/test/java/io/opentelemetry/contrib/jfr/metrics/JfrThreadCountTest.java
```diff
@@ -0,0 +1,81 @@
...
+        metric ->
+            metric
+                .hasName("process.runtime.jvm.threads.count")
+                .hasUnit(UNIT_THREADS)
+                .satisfies(
+                    metricData -> {
+                      SumData<?> sumData = metricData.getLongSumData();
+                      boolean foundDaemon = false;
+                      boolean foundNonDaemon = false;
+                      for (PointData point : sumData.getPoints()) {
```
mateuszrzeszutek, line 60: Nit: you could use assertJ's `anySatisfy` or `anyMatch` instead (pseudocode):
```java
assertThat(sumData.getPoints())
    .map(LongPointData.class::cast)
    .anyMatch(p -> p.getValue() > 0 && isDaemon(p))
    .anyMatch(p -> p.getValue() > 0 && !isDaemon(p));
```

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/classes/ClassesLoadedHandler.java
roberttoyonaga: I added some comments for further clarity. The metric descriptions also add some information.

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/Constants.java
roberttoyonaga: changed!

jfr-streaming/src/test/java/io/opentelemetry/contrib/jfr/metrics/JfrThreadCountTest.java
roberttoyonaga: done!

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/threads/ThreadCountHandler.java
roberttoyonaga: that's a good point!

mateuszrzeszutek (approved):

jfr-streaming/src/test/java/io/opentelemetry/contrib/jfr/metrics/JfrThreadCountTest.java
```diff
@@ -0,0 +1,65 @@
...
+
+class JfrThreadCountTest extends AbstractMetricsTest {
+  private static final int SAMPLING_INTERVAL = 1000;
+
+  private static void doWork() throws InterruptedException {
+    Thread.sleep(2 * SAMPLING_INTERVAL);
+  }
+
+  private static boolean isDaemon(LongPointData p) {
+    return p.getAttributes().asMap().get(AttributeKey.booleanKey(DAEMON)).equals(true);
```
mateuszrzeszutek, line 25: There are actually 3 possible states for the `DAEMON` attribute: true, false and null. We're not expecting a null value since it's invalid in this context, so let's add an assertion for that:
```suggestion
    Boolean daemon = p.getAttributes().get(AttributeKey.booleanKey(DAEMON));
    assertThat(daemon).isNotNull();
    return daemon;
```

trask:
thx @roberttoyonaga!

PR #579, opened 2022-11-09
Glassfish service name detector

breedx-splk:
This is a continuation of #562 that introduces the first of several web app servers for which we will detect the service name.

Because this is the first (GlassFish), it includes some of the helper/plumbing utilities. Subsequent additions will mostly be the `AppServer` implementations (and adding them to the delegate list).

### DIFF
--- a/resource-providers/README.md
+++ b/resource-providers/README.md
@@ -10,10 +10,14 @@ to populate the `service.name` resource attribute based on the runtime configura
 of an app server. This is useful when a user has not yet specified the `service.name`
 resource attribute manually.
 
-It is capable of detecting common scenarios among the following popular application servers:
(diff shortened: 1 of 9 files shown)


mateuszrzeszutek (approved):

trask (commented):
can you add to doc, with a brief explanation of where the service name is read from for each application server (Glassfish so far)?

breedx-splk:
> can you add to doc, with a brief explanation of where the service name is read from for each application server (Glassfish so far)?

Sure, I assume you want it in this module README or did you have something else in mind?

trask:
> Sure, I assume you want it in this module README or did you have something else in mind?

👍

trask:
thx @breedx-splk!

PR #644, opened 2022-12-14
More JFR streaming handlers

roberttoyonaga:
**Description:**
These changes add more handlers in an effort to help the JFR streaming implementation meet the semantic conventions defined [here](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/runtime-environment-metrics.md#jvm-metrics).

With the addition of these new changes, the JFR streaming implementation will be as complete as possible with the data emitted by current JFR events. Continuing to make this JFR streaming implementation more complete would require changes to JFR itself in OpenJDK, such as adding new events or adding additional data to existing events. 

**Testing:**

Tests were added in `BufferMetricTest`, `GCDurationMetricTest`, `MemoryCommittedMetricTest`, `MemoryLimitMetricTest`, `MemoryInitMetricTest`, and `MemoryUsageMetricTest`.  Existing tests were also updated where needed. 

**Outstanding items:**
A list of changes to OpenJDK JFR in order to help meet the semantic conventions should be compiled. The next step is to contribute the new events/instrumentation to the JDK, then to utilize those changes by updating the handlers in this repo.

PR #650, opened 2022-12-15
Add buffer handlers and implement buffer metrics

roberttoyonaga:
**Description:**
Add handlers for buffer metrics `process.runtime.jvm.buffer.limit`, `process.runtime.jvm.buffer.count`, and `process.runtime.jvm.buffer.usage`.

This was a part of a larger PR here:  https://github.com/open-telemetry/opentelemetry-java-contrib/pull/644.

`jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/Constants.java` has been kept the same as it was in the larger PR because it's used in all of the smaller constituent PRs. This will hopefully also avoid some conflicts as the PRs are sequentially merged. As a result, there are some constants that are defined, but not used in this PR.

### DIFF
--- a/jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/buffer/DirectBufferStatisticsHandler.java
+++ b/jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/buffer/DirectBufferStatisticsHandler.java
@@ -1,0 +1,93 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.contrib.jfr.metrics.internal.buffer;
+
+import static io.opentelemetry.contrib.jfr.metrics.internal.Constants.ATTR_POOL;
+import static io.opentelemetry.contrib.jfr.metrics.internal.Constants.BYTES;
+import static io.opentelemetry.contrib.jfr.metrics.internal.Constants.UNIT_BUFFERS;
+import static io.opentelemetry.contrib.jfr.metrics.internal.RecordedEventHandler.defaultMeter;
+
+import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.api.metrics.Meter;
+import io.opentelemetry.contrib.jfr.metrics.internal.RecordedEventHandler;
+import java.time.Duration;
+import java.util.Optional;
+import jdk.jfr.consumer.RecordedEvent;
+
+public final class DirectBufferStatisticsHandler implements RecordedEventHandler {
+  private static final String METRIC_NAME_USAGE = "process.runtime.jvm.buffer.usage";
+  private static final String METRIC_NAME_LIMIT = "process.runtime.jvm.buffer.limit";
+  private static final String METRIC_NAME_COUNT = "process.runtime.jvm.buffer.count";
+  private static final String METRIC_DESCRIPTION_USAGE = "Measure of memory used by buffers";
+  private static final String METRIC_DESCRIPTION_LIMIT =
+      "Measure of total memory capacity of buffers";
+  private static final String METRIC_DESCRIPTION_COUNT = "Number of buffers in the pool";
+  private static final String COUNT = "count";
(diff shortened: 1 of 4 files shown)


kittylyst (approved):

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/buffer/DirectBufferStatisticsHandler.java
```diff
@@ -0,0 +1,93 @@
...
+              measurement.record(limit, ATTR);
+            });
+    meter
+        .upDownCounterBuilder(METRIC_NAME_COUNT)
+        .setDescription(METRIC_DESCRIPTION_COUNT)
+        .setUnit(UNIT_BUFFERS)
+        .buildWithCallback(
+            measurement -> {
+              measurement.record(count, ATTR);
+            });
```
jack-berg, line 86: #nit: these curly braces aren't used

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/buffer/DirectBufferStatisticsHandler.java
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  @Override
+  public Optional<Duration> getPollingDuration() {
+    return Optional.of(Duration.ofSeconds(1));
```
jack-berg, line 91: What's the significance of the polling duration? Why 1s vs. 5s, 10s, etc?

jfr-streaming/src/test/java/io/opentelemetry/contrib/jfr/metrics/BufferMetricTest.java
```diff
@@ -0,0 +1,89 @@
...
+   *
+   * <p>This test handles all 3 buffer related metrics defined in the OpenTelemetry Java runtime
+   * Semantic Conventions.
+   *
+   * <p>Currently JFR only has support for the "direct" buffer pool. The "mapped" and "mapped -
+   * 'non-volatile memory'" pools do not have corresponding JFR events. In the future, events should
+   * be added for those missing pools.
+   */
+  @Test
+  void shouldHaveJfrLoadedClassesCountEvents() throws Exception {
```
jack-berg, line 32: #nit: can remove `throws Exception`

jfr-streaming/src/test/java/io/opentelemetry/contrib/jfr/metrics/BufferMetricTest.java
```diff
@@ -0,0 +1,89 @@
...
+                    sum ->
+                        sum.hasPointsSatisfying(
+                            point ->
+                                point.satisfies(
+                                    pointData -> {
+                                      Assertions.assertTrue(pointData.getValue() > 0);
+                                      Assertions.assertTrue(
+                                          pointData
+                                              .getAttributes()
+                                              .equals(Attributes.of(ATTR_POOL, "direct")));
```
jack-berg, line 52: #nit: Minor improvement in readibility:
```suggestion
                                      assertThat(pointData.getValue()).isGreaterThan(0);
                                      assertThat(pointData.getAttributes())
                                          .isEqualTo(Attributes.of(ATTR_POOL, "direct"));
```
Same advice applies to the other two metrics.

jack-berg (approved):
Just a few minor comments. Thanks!

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/buffer/DirectBufferStatisticsHandler.java
roberttoyonaga: The JFR streaming system flushes the in memory JFR buffers holding event data to the disk repo at 1s intervals, at which point they may be consumed by the handlers. So polling at 1s intervals should provide the most up to date data without adding unnecessary overhead.

jack-berg (approved):

PR #651, opened 2022-12-15
cpu handlers

roberttoyonaga:
**Description:**

This PR updates the existing CPU event handler to also implement the metric `process.runtime.jvm.system.cpu.load_1m`. 

This was a part of a [larger PR here](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/644). 

`jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/Constants.java` has been kept the same as it was in the larger PR because it is used in all of the smaller constituent PRs.

### DIFF
--- a/jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/Constants.java
+++ b/jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/Constants.java
@@ -26,11 +26,19 @@ public final class Constants {
   public static final String REGION_COUNT = "region.count";
   public static final String COMMITTED = "committed";
   public static final String RESERVED = "reserved";
+  public static final String INITIAL_SIZE = "initialSize";
   public static final String USED = "used";
   public static final String COMMITTED_SIZE = "committedSize";
+  public static final String RESERVED_SIZE = "reservedSize";
 
   public static final String DAEMON = "daemon";
   public static final String HEAP = "heap";
+  public static final String NON_HEAP = "nonheap";
+  public static final String NETWORK_MODE_READ = "read";
+  public static final String NETWORK_MODE_WRITE = "write";
+  public static final String DURATION = "duration";
+  public static final String END_OF_MINOR_GC = "end of minor GC";
+  public static final String END_OF_MAJOR_GC = "end of major GC";
 
   public static final String METRIC_NAME_NETWORK_BYTES = "process.runtime.jvm.network.io";
   public static final String METRIC_DESCRIPTION_NETWORK_BYTES = "Network read/write bytes";
@@ -38,25 +46,33 @@ public final class Constants {
   public static final String METRIC_DESCRIPTION_NETWORK_DURATION = "Network read/write duration";
   public static final String METRIC_NAME_COMMITTED = "process.runtime.jvm.memory.committed";
   public static final String METRIC_DESCRIPTION_COMMITTED = "Measure of memory committed";
-  public static final String NETWORK_MODE_READ = "read";
-  public static final String NETWORK_MODE_WRITE = "write";
   public static final String METRIC_NAME_MEMORY = "process.runtime.jvm.memory.usage";
+  public static final String METRIC_DESCRIPTION_MEMORY = "Measure of memory used";
   public static final String METRIC_NAME_MEMORY_AFTER =
       "process.runtime.jvm.memory.usage_after_last_gc";
-  public static final String METRIC_DESCRIPTION_MEMORY = "Measure of memory used";
   public static final String METRIC_DESCRIPTION_MEMORY_AFTER =
       "Measure of memory used, as measured after the most recent garbage collection event on this pool";
   public static final String METRIC_NAME_MEMORY_ALLOCATION =
       "process.runtime.jvm.memory.allocation";
   public static final String METRIC_DESCRIPTION_MEMORY_ALLOCATION = "Allocation";
+  public static final String METRIC_NAME_MEMORY_INIT = "process.runtime.jvm.memory.init";
+  public static final String METRIC_DESCRIPTION_MEMORY_INIT = "Measure of initial memory requested";
+  public static final String METRIC_NAME_MEMORY_LIMIT = "process.runtime.jvm.memory.limit";
+  public static final String METRIC_DESCRIPTION_MEMORY_LIMIT = "Measure of max obtainable memory";
+  public static final String METRIC_NAME_GC_DURATION = "process.runtime.jvm.gc.duration";
+  public static final String METRIC_DESCRIPTION_GC_DURATION =
+      "Duration of JVM garbage collection actions";
 
   public static final AttributeKey<String> ATTR_THREAD_NAME = AttributeKey.stringKey("thread.name");
   public static final AttributeKey<String> ATTR_ARENA_NAME = AttributeKey.stringKey("arena");
   public static final AttributeKey<String> ATTR_NETWORK_MODE = AttributeKey.stringKey("mode");
   public static final AttributeKey<String> ATTR_TYPE = AttributeKey.stringKey("type");
   public static final AttributeKey<String> ATTR_POOL = AttributeKey.stringKey("pool");
+  public static final AttributeKey<String> ATTR_GC = AttributeKey.stringKey("pool");
+  public static final AttributeKey<String> ATTR_ACTION = AttributeKey.stringKey("action");
   public static final AttributeKey<Boolean> ATTR_DAEMON = AttributeKey.booleanKey(DAEMON);
   public static final String UNIT_CLASSES = "{classes}";
   public static final String UNIT_THREADS = "{threads}";
+  public static final String UNIT_BUFFERS = "{buffers}";
   public static final String UNIT_UTILIZATION = "1";
 }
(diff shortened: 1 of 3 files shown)


jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/cpu/OverallCPULoadHandler.java
```diff
@@ -17,16 +17,23 @@
...
   private static final String EVENT_NAME = "jdk.CPULoad";
   private static final String JVM_USER = "jvmUser";
   private static final String JVM_SYSTEM = "jvmSystem";
   private static final String MACHINE_TOTAL = "machineTotal";
   private volatile double process = 0;
   private volatile double machine = 0;
+  private volatile double machineMinuteAverage = 0;
+  private static final int POLLING_INTERVAL_S = 1;
+  private int count = 0;
+  private double machineSum = 0;
```
breedx-splk, line 36: Does JFR guarantee that events are always dispatched on the same thread? If not, then this should probably be volatile as well.

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/cpu/OverallCPULoadHandler.java
```diff
@@ -50,16 +57,31 @@ public void initializeMeter(Meter meter) {
...
+        .setUnit(UNIT_UTILIZATION)
+        .buildWithCallback(
+            measurement -> {
+              measurement.record(machineMinuteAverage);
+            });
   }
 
   @Override
   public void accept(RecordedEvent ev) {
+    if (count % 60 == 0) {
```
breedx-splk, line 72: I think this 60 literal also needs to take into consideration the polling interval, right? So it needs to be computed.

Like if the interval changes to 2s, then the 60 would need to be changed to 30.

breedx-splk (commented):
Looks mostly ok to me. Two small comments.

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/cpu/OverallCPULoadHandler.java
kittylyst: Good question. The _assumption_ is that the events are fired onto a SingleThreadExecutor but it is in theory possible that some could use a different model, although I would expect that the extra complexity would very much outweigh any possible throughput gain.

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/cpu/OverallCPULoadHandler.java
kittylyst: We have a hardcoded `METRIC_NAME_MACHINE_MINUTE` and the metric defined in spec is, per, https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/runtime-environment-metrics.md#jvm-metrics specifically a 1-minute metric. So, I think this is OK - it isn't supposed to represent an average over a general interval.

kittylyst (approved):

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/cpu/OverallCPULoadHandler.java
roberttoyonaga: I made it volatile and added synchronization

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/Constants.java
```diff
@@ -26,37 +26,53 @@ private Constants() {}
...
+  public static final String RESERVED_SIZE = "reservedSize";
 
   public static final String DAEMON = "daemon";
   public static final String HEAP = "heap";
+  public static final String NON_HEAP = "nonheap";
+  public static final String NETWORK_MODE_READ = "read";
+  public static final String NETWORK_MODE_WRITE = "write";
+  public static final String DURATION = "duration";
+  public static final String END_OF_MINOR_GC = "end of minor GC";
+  public static final String END_OF_MAJOR_GC = "end of major GC";
```
jack-berg, line 41: Only a few of these are used in the current PR. Can you prune down the unused ones?

jfr-streaming/src/test/java/io/opentelemetry/contrib/jfr/metrics/JfrOverallCPULoadHandlerTest.java
```diff
@@ -25,6 +25,12 @@ public void shouldHaveCPULoadEvents() throws Exception {
                 .hasName("process.runtime.jvm.system.cpu.utilization")
                 .hasUnit(UNIT_UTILIZATION)
                 .hasDescription("Recent CPU utilization for the whole system")
+                .hasDoubleGaugeSatisfying(gauge -> {}),
+        metric ->
+            metric
+                .hasName("process.runtime.jvm.system.cpu.load_1m")
+                .hasUnit(UNIT_UTILIZATION)
+                .hasDescription("Average CPU load of the whole system for the last minute")
                 .hasDoubleGaugeSatisfying(gauge -> {}));
```
jack-berg, line 34: Can you rename this class to `OverallCPULoadHandlerTest`?

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/cpu/OverallCPULoadHandler.java
```diff
@@ -50,16 +56,35 @@ public void initializeMeter(Meter meter) {
...
 
   @Override
   public void accept(RecordedEvent ev) {
+    // Synchronized to avoid races on machineSum
+    synchronized (this) {
+      if (count % 60 == 0) {
+        machineMinuteAverage = machineSum / 60;
+        count = 0;
+        machineSum = 0;
+      }
```
jack-berg, line 77: I think the most correct thing to do here would be to retain a rolling list of the last 60 measurements. Each time a new measurement comes in, drop the oldest. Then when the callback is invoked take the average of all the entries.

roberttoyonaga: Ok it's been changed now. I agree that the suggested approach is better. That way the granularity will be per second instead of per minute.

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/cpu/OverallCPULoadHandler.java
```diff
@@ -50,16 +58,37 @@ public void initializeMeter(Meter meter) {
...
+        machine = ev.getDouble(MACHINE_TOTAL);
+        // Remove oldest data point if queue is at capacity
+        if (machineMinuteQueue.size() == 60) {
+          machineSum -= machineMinuteQueue.poll();
+        }
+        // Add new data point
+        machineSum += machine;
+        machineMinuteQueue.add(machine);
+        // Compute new result
+        machineMinuteAverage = machineSum / machineMinuteQueue.size();
```
jack-berg, line 85: This math isn't quite right. `machineSum` is a running total of all the `machine` measurements, and its being divided by the size of the deque which is always <= 60. 

You need to iterate through `machineMinuteAverage`, sum up the entries, and divide by the size.

Should do this in the callback on line 66 so you're only computing as needed as opposed to for each measurement.  And will need to synchronize so the deque isn't modified while computing the average.

Would be good to somehow breakout the pieces of code involved so it can be unit tested.

jack-berg, line 85: Nice.

Would be good to somehow breakout the pieces of code involved in this so they can be unit tested.

roberttoyonaga: I added a [commit](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/651/commits/283ccc38063a7ab6d2c204c6c8bb3ecc33268ee1) that extracts the moving average code and adds a test for it. But I'm not sure it's an ideal solution because the method is now `public`.

jack-berg (approved):

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/cpu/OverallCPULoadHandler.java
jack-berg: It can be package private.

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/cpu/OverallCPULoadHandler.java
trask: I think `process.runtime.jvm.system.cpu.load_1m` is intended to be a "load average", e.g. https://en.wikipedia.org/wiki/Load_(computing)#Unix-style_load_calculation and https://docs.oracle.com/javase/8/docs/api/java/lang/management/OperatingSystemMXBean.html#getSystemLoadAverage--, instead of an "average cpu utilization"

roberttoyonaga: ohh I see. I think that part of the semantic conventions was written with JMX in mind because JFR does not gather that information. A new event emitting that information will have to be created for this metric. Closing the PR.

roberttoyonaga:
Closed because `process.runtime.jvm.system.cpu.load_1m` is intended to be a "load average". JFR does not yet emit this info.

PR #652, opened 2022-12-15
Implement memory metrics

roberttoyonaga:
**Description:**
This PR updates the implementation for the metrics `process.runtime.jvm.memory.limit` , `process.runtime.jvm.memory.usage_after_last_gc` ,  `process.runtime.jvm.memory.committed`, `process.runtime.jvm.memory.init`, `process.runtime.jvm.memory.usage`

This was a part of a [larger PR here](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/644). 

`jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/Constants.java` has been kept the same as it was in the larger PR because it is used in all of the smaller constituent PRs. This will hopefully also avoid some conflicts as the PRs are sequentially merged. As a result, there are some constants that are defined, but not used in this PR.

Some of the attribute key-value pairs outlined [here](https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation/runtime-metrics/library#garbage-collector-dependent-metrics) are not present with the current implementation. This is because they require additional JFR data, or new events. In the future, when JFR is updated, these handlers should be revisited and updated to cover all the missing attribute key-value pairs.


jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/memory/MetaspaceSummaryHandler.java
```diff
@@ -0,0 +1,142 @@
...
+
+  @Override
+  public String getEventName() {
+    return EVENT_NAME;
+  }
+
+  @Override
+  public void accept(RecordedEvent ev) {
+
+    recordValues(ev);
```
jack-berg, line 91: Feels unnecessary to split out a separate method for this

jfr-streaming/src/test/java/io/opentelemetry/contrib/jfr/metrics/MemoryUsageMetricTest.java
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  /**
+   * This is a basic test for process.runtime.jvm.memory.usage and
+   * process.runtime.jvm.memory.usage_after_last_gc metrics.
+   */
+  @Test
+  void shouldHaveMemoryUsageMetrics() {
+    System.gc();
+    if (garbageCollectors.contains("G1 Young Generation")) {
```
jack-berg, line 53: Shouldn't the garbage collector in use for a given build be deterministic? 

Consider breaking out this test by garbage collector and using [assumeThat](https://github.com/open-telemetry/opentelemetry-java/blob/main/exporters/otlp/testing-internal/src/main/java/io/opentelemetry/exporter/otlp/testing/internal/AbstractGrpcTelemetryExporterTest.java#L241-L250) to only verify the conditions if the assumption is true.

jfr-streaming/src/test/java/io/opentelemetry/contrib/jfr/metrics/MemoryLimitMetricTest.java
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  @Test
+  void shouldHaveMemoryLimitMetrics() {
+
+    System.gc();
+
+    if (garbageCollectors.contains("G1 Young Generation")) {
+      // TODO: needs JFR support
+
```
jack-berg, line 39: There's a handful of errant blank lines such as this, line 34, and others. Can you do a pass to clean them up?

jfr-streaming/src/test/java/io/opentelemetry/contrib/jfr/metrics/MemoryLimitMetricTest.java
```diff
@@ -0,0 +1,61 @@
...
+import static io.opentelemetry.contrib.jfr.metrics.internal.Constants.METRIC_DESCRIPTION_MEMORY_LIMIT;
+import static org.assertj.core.api.Assertions.assertThat;
+
+import io.opentelemetry.sdk.metrics.data.MetricData;
+import io.opentelemetry.sdk.metrics.data.SumData;
+import org.assertj.core.api.ThrowingConsumer;
+import org.junit.jupiter.api.Test;
+
+class MemoryLimitMetricTest extends AbstractMetricsTest {
+  private void check(ThrowingConsumer<MetricData> attributeCheck) {
```
jack-berg, line 22: #nit: Let's move these utility methods to the end of the class

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/memory/GCHeapConfigurationHandler.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.Optional;
+import jdk.jfr.consumer.RecordedEvent;
+
+/** This class handles GCHeapConfiguration JFR events. */
+public final class GCHeapConfigurationHandler implements RecordedEventHandler {
+  private static final String EVENT_NAME = "jdk.GCHeapConfiguration";
+
+  private static final String MAX_SIZE = "maxSize";
+  private static final Attributes ATTR =
+      Attributes.of(ATTR_TYPE, HEAP, ATTR_POOL, "Java heap space");
```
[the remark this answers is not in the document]
roberttoyonaga: removed.

jfr-streaming/src/test/java/io/opentelemetry/contrib/jfr/metrics/MemoryUsageMetricTest.java
roberttoyonaga: I'm a bit confused. Isn't the GC specified at runtime instead of buildtime? I'm looking for a way to specify a different GC for each test so I can ensure that metrics for the various pools are correct. Right now, I haven't figured out how to do that.  Additionally, I'm not sure how to specify a particular GC without passing it as a runtime option (so the JVM chooses a default depending on the system). This is why the tests are set up like this. 

I can break out these tests by GC, but they still won't be that useful until the points above are resolved.

jack-berg: You can create different test suites that specify different JVM args. Check out this [commit](https://github.com/jack-berg/opentelemetry-java-contrib/commit/0c83192f98ef17ffe4a1aeae3d1b191241e11dab) where I introduce a new test suite called "testShenandoah", which runs with JVM args `-XX:+UnlockExperimentalVMOptions XX:+UseShenandoahGC`. Could use the same technique to create a different test suite for each GC that needs to be tested.

roberttoyonaga: That's exactly what I was looking for!

jfr-streaming/src/g1GcTest/java/io/opentelemetry/contrib/jfr/metrics/G1MemoryCommittedMetricTest.java
```diff
@@ -0,0 +1,40 @@
...
+import static io.opentelemetry.contrib.jfr.metrics.internal.Constants.METRIC_DESCRIPTION_COMMITTED;
+import static org.assertj.core.api.Assertions.assertThat;
+
+import io.opentelemetry.sdk.metrics.data.MetricData;
+import io.opentelemetry.sdk.metrics.data.SumData;
+import org.assertj.core.api.ThrowingConsumer;
+import org.junit.jupiter.api.Test;
+
+class G1MemoryCommittedMetricTest extends AbstractMetricsTest {
+  private void check(ThrowingConsumer<MetricData> attributeCheck) {
```
jack-berg, line 19: Why is this a separate method?

jfr-streaming/src/g1GcTest/java/io/opentelemetry/contrib/jfr/metrics/G1MemoryLimitMetricTest.java
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.contrib.jfr.metrics;
+
+import org.junit.jupiter.api.Test;
+
+class G1MemoryLimitMetricTest extends AbstractMetricsTest {
```
jack-berg, line 10: Should we group all the memory related tests together?

jfr-streaming/src/parallelGcTest/java/io/opentelemetry/contrib/jfr/metrics/PsGcDurationMetricTest.java
```diff
@@ -18,35 +18,16 @@
 import org.assertj.core.api.ThrowingConsumer;
 import org.junit.jupiter.api.Test;
 
-class GCDurationMetricTest extends AbstractMetricsTest {
+class PsGcDurationMetricTest extends AbstractMetricsTest {
   static class AttributeCheck implements ThrowingConsumer<PointData> {
```
jack-berg, line 22: I don't think having a separate class makes the test code more readable. Let's inline this.

jfr-streaming/src/parallelGcTest/java/io/opentelemetry/contrib/jfr/metrics/PsGcDurationMetricTest.java
```diff
@@ -18,35 +18,16 @@
...
-              .isEqualTo(Attributes.of(ATTR_GC, "PS MarkSweep", ATTR_ACTION, END_OF_MAJOR_GC));
-        } else if (garbageCollectors.contains("MarkSweepCompact")) {
-          assertThat(pointData.getAttributes())
-              .isEqualTo(Attributes.of(ATTR_GC, "MarkSweepCompact", ATTR_ACTION, END_OF_MAJOR_GC));
-        }
+        assertThat(pointData.getAttributes())
+            .isEqualTo(Attributes.of(ATTR_GC, "PS MarkSweep", ATTR_ACTION, END_OF_MAJOR_GC));
       }
     }
   }
```
jack-berg, line 33: There's a comment below that is no longer true:
```
// TODO: Need a reliable way to test old and young gen GC in isolation.
```

jfr-streaming/src/g1GcTest/java/io/opentelemetry/contrib/jfr/metrics/G1MemoryLimitMetricTest.java
roberttoyonaga: Ok I grouped the memory related tests into classes per their GC

jfr-streaming/src/parallelGcTest/java/io/opentelemetry/contrib/jfr/metrics/PsGcDurationMetricTest.java
roberttoyonaga: I left that comment in because, although we can reliably test different collectors independently, there still isn't any code that enforces young vs old gen collection for GCs that divide up the heap.

jfr-streaming/src/g1GcTest/java/io/opentelemetry/contrib/jfr/metrics/G1MemoryCommittedMetricTest.java
roberttoyonaga: I thought it helped readability. now inlined

jfr-streaming/src/g1GcTest/java/io/opentelemetry/contrib/jfr/metrics/AbstractMetricsTest.java
```diff
@@ -0,0 +1,53 @@
...
+  @BeforeAll
+  static void initializeOpenTelemetry() {
+    if (isInitialized) {
+      return;
+    }
+    isInitialized = true;
+    metricReader = InMemoryMetricReader.create();
+    meterProvider = SdkMeterProvider.builder().registerMetricReader(metricReader).build();
+    GlobalOpenTelemetry.set(OpenTelemetrySdk.builder().setMeterProvider(meterProvider).build());
+    JfrMetrics.enable(meterProvider);
```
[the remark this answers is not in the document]
roberttoyonaga: Yup I agree with all those points. 
>Right now that's required because JfrMetrics can't be cancelled once started
Maybe `@BeforeEach` test I can reinitialize JFR streaming and all openTelemetry variables, and `@AfterEach` test the JFR streaming can be ended.  I think that should be enough for tests to fully clean up after themselves.

 > Should be put in a common place where it can be used by each of the different testing suites

I was trying to figure out a good way of doing this today. Maybe something related to test fixtures. I'll have to investigate more.

jack-berg (approved):

jfr-streaming/src/g1GcTest/java/io/opentelemetry/contrib/jfr/metrics/AbstractMetricsTest.java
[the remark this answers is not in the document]
roberttoyonaga: Update:
I renamed the base test class to `AbstractJfrTest`. I was able to get the `testFixtures` working. So now the class resides in a single place, instead of being duplicated for each GC type.

jfr-streaming/build.gradle.kts
```diff
@@ -1,11 +1,20 @@
 plugins {
   id("otel.java-conventions")
+  id("java-library")
+  id("java-test-fixtures")
```
jack-berg, line 4: TIL about `java-test-fixtures`. Nice!

trask: 💯 cc @mateuszrzeszutek @laurit

PR #653, opened 2022-12-15
Implement GC duration metric

roberttoyonaga:
**Description:**

This PR adds changes that implement the metric `process.runtime.jvm.gc.duration`.

This was a part of a larger PR here:  https://github.com/open-telemetry/opentelemetry-java-contrib/pull/644.

`jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/Constants.java` has been kept the same as it was in the larger PR because it's used in all of the smaller constituent PRs. This will hopefully also avoid some conflicts as the PRs are sequentially merged. As a result, there are some constants that are defined, but not used in this PR.

**Testing:**

Tests were added but they don't have great coverage. In the future we need to devise a reliable way to test each GC. Currently, I'm not sure of a way to change GCs at runtime in order to test them all. I'm also uncertain of how to reliably force young gen GC and old gen GC. As a result, the test added in this PR basically just invokes "some" GC and ensures that the resulting metric has data that is expected.

### DIFF
--- a/jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/HandlerRegistry.java
+++ b/jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/HandlerRegistry.java
@@ -28,8 +28,6 @@ import java.lang.management.ManagementFactory;
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
-import java.util.Map;
-import java.util.function.Supplier;
 
 public final class HandlerRegistry {
   private static final String INSTRUMENTATION_NAME = "io.opentelemetry.contrib.jfr";
@@ -37,34 +35,44 @@ public final class HandlerRegistry {
 
   private final List<RecordedEventHandler> mappers;
 
-  public static HashSet<String> garbageCollectors = new HashSet<String>();
-
-  private static final Map<String, List<Supplier<RecordedEventHandler>>> HANDLERS_PER_GC =
-      Map.of(
-          "G1",
-          List.of(G1HeapSummaryHandler::new, G1GarbageCollectionHandler::new),
-          "Parallel",
-          List.of(ParallelHeapSummaryHandler::new));
-
   private HandlerRegistry(List<? extends RecordedEventHandler> mappers) {
     this.mappers = new ArrayList<>(mappers);
   }
 
   static HandlerRegistry createDefault(MeterProvider meterProvider) {
+    HashSet<String> garbageCollectors = new HashSet<>();
+
     var handlers = new ArrayList<RecordedEventHandler>();
-    var seen = new HashSet<String>();
+    // Must gather all GC names before creating GC handlers that require the list of active GC names
     for (var bean : ManagementFactory.getGarbageCollectorMXBeans()) {
-      var name = bean.getName();
-      garbageCollectors.add(name);
-      for (var gcType : HANDLERS_PER_GC.keySet()) {
-        if (name.contains(gcType)
-            && !seen.contains(gcType)
-            && HANDLERS_PER_GC.get(gcType) != null) {
-          handlers.addAll(HANDLERS_PER_GC.get(gcType).stream().map(s -> s.get()).toList());
-          seen.add(gcType);
-        }
+      garbageCollectors.add(bean.getName());
+    }
+
+    // Configure GC specific handlers
+    for (var name : garbageCollectors) {
+      switch (name) {
+        case "G1 Young Generation":
+          {
+            handlers.add(new G1HeapSummaryHandler());
+            handlers.add(new G1GarbageCollectionHandler());
+            break;
+          }
+        case "Copy":
+          {
+            handlers.add(new YoungGarbageCollectionHandler(garbageCollectors));
+            break;
+          }
+        case "PS Scavenge":
+          {
+            handlers.add(new YoungGarbageCollectionHandler(garbageCollectors));
+            handlers.add(new ParallelHeapSummaryHandler());
+            break;
+          }
+        default:
+          // If none of the above GCs are detected, no action.
       }
     }
+
     var grouper = new ThreadGrouper();
     var basicHandlers =
         List.of(
@@ -79,12 +87,8 @@ public final class HandlerRegistry {
             new LongLockHandler(grouper),
             new ThreadCountHandler(),
             new ClassesLoadedHandler(),
-            new OldGarbageCollectionHandler());
+            new OldGarbageCollectionHandler(garbageCollectors));
     handlers.addAll(basicHandlers);
-
-    if (!garbageCollectors.contains("G1 Young Generation")) {
-      handlers.add(new YoungGarbageCollectionHandler());
-    }
 
     var meter =
         meterProvider
--- a/jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/HandlerRegistry.java
+++ b/jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/HandlerRegistry.java
@@ -26,10 +26,9 @@ import io.opentelemetry.contrib.jfr.metrics.internal.network.NetworkWriteHandler
 import io.opentelemetry.contrib.jfr.metrics.internal.threads.ThreadCountHandler;
 import java.lang.management.ManagementFactory;
 import java.util.ArrayList;
-import java.util.HashSet;
 import java.util.List;
 
-public final class HandlerRegistry {
+final class HandlerRegistry {
   private static final String INSTRUMENTATION_NAME = "io.opentelemetry.contrib.jfr";
   private static final String INSTRUMENTATION_VERSION = "1.7.0-SNAPSHOT";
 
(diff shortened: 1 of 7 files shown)


jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/HandlerRegistry.java
```diff
@@ -28,16 +31,18 @@
...
 import java.util.function.Supplier;
 
-final class HandlerRegistry {
+public final class HandlerRegistry {
   private static final String INSTRUMENTATION_NAME = "io.opentelemetry.contrib.jfr";
   private static final String INSTRUMENTATION_VERSION = "1.7.0-SNAPSHOT";
 
   private final List<RecordedEventHandler> mappers;
 
+  public static HashSet<String> garbageCollectors = new HashSet<String>();
```
jack-berg, line 40: A public static mutable constant is a code smell. Let's get rid of this and instead pass the names of the garbage collectors as an constructor argument to `OldGarbageCollectionHandler`.

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/HandlerRegistry.java
```diff
@@ -50,6 +55,7 @@ static HandlerRegistry createDefault(MeterProvider meterProvider) {
     var seen = new HashSet<String>();
     for (var bean : ManagementFactory.getGarbageCollectorMXBeans()) {
       var name = bean.getName();
+      garbageCollectors.add(name);
```
jack-berg, line 58: It's pretty tricky to tell which handlers get added in response to which garbage collectors. How about just a switch statement with cases for each known garbage collector bean name, where each case explicitly registers its required handlers?

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/HandlerRegistry.java
roberttoyonaga: done!

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/HandlerRegistry.java
roberttoyonaga: I think that's a good idea. Implemented suggestion.

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/GarbageCollection/OldGarbageCollectionHandler.java
```diff
@@ -0,0 +1,74 @@
...
+  @Override
+  public void initializeMeter(Meter meter) {
+    // Set the attribute's GC based on which GC is being used.
+    if (garbageCollectors.contains("PS MarkSweep")) {
+      attributes = Attributes.of(ATTR_GC, "PS MarkSweep", ATTR_ACTION, END_OF_MAJOR_GC);
+    } else if (garbageCollectors.contains("G1 Old Generation")) {
+      attributes = Attributes.of(ATTR_GC, "G1 Old Generation", ATTR_ACTION, END_OF_MAJOR_GC);
+    } else if (garbageCollectors.contains("MarkSweepCompact")) {
+      attributes = Attributes.of(ATTR_GC, "MarkSweepCompact", ATTR_ACTION, END_OF_MAJOR_GC);
+    }
```
jack-berg, line 59: You can resolve these attributes once in the constructor.

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/GarbageCollection/OldGarbageCollectionHandler.java
```diff
@@ -0,0 +1,74 @@
...
+public final class OldGarbageCollectionHandler implements RecordedEventHandler {
+  private final HashSet<String> garbageCollectors;
+  private static final String EVENT_NAME = "jdk.OldGarbageCollection";
+  // This could be changed later
+  private Attributes attributes =
+      Attributes.of(ATTR_GC, "MarkSweepCompact", ATTR_ACTION, END_OF_MAJOR_GC);
+
+  private LongHistogram histogram;
+
+  public OldGarbageCollectionHandler(HashSet<String> garbageCollectors) {
```
jack-berg, line 35: `Set<String>` is preferable to `HashSet<String>` because this code doesn't care which implementation of `Set` is used.

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/HandlerRegistry.java
```diff
@@ -25,40 +28,51 @@
...
+
+    // Configure GC specific handlers
+    for (var name : garbageCollectors) {
+      switch (name) {
+        case "G1 Young Generation":
+          {
+            handlers.add(new G1HeapSummaryHandler());
+            handlers.add(new G1GarbageCollectionHandler());
+            break;
+          }
```
jack-berg, line 59: #nit: probably can drop the curly braces here as they don't improve readability.

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/HandlerRegistry.java
```diff
@@ -25,40 +28,51 @@
...
-            && HANDLERS_PER_GC.get(gcType) != null) {
-          handlers.addAll(HANDLERS_PER_GC.get(gcType).stream().map(s -> s.get()).toList());
-          seen.add(gcType);
-        }
+      garbageCollectors.add(bean.getName());
+    }
+
+    // Configure GC specific handlers
+    for (var name : garbageCollectors) {
+      switch (name) {
```
jack-berg, line 53: What do you think of moving the switch statement into the for loop above so we have one loop instead of two?

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/HandlerRegistry.java
```diff
@@ -25,40 +28,51 @@
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
-import java.util.Map;
-import java.util.function.Supplier;
 
-final class HandlerRegistry {
+public final class HandlerRegistry {
```
jack-berg, line 32: Does this still need to be public?

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/HandlerRegistry.java
```diff
@@ -72,7 +86,8 @@ static HandlerRegistry createDefault(MeterProvider meterProvider) {
             new ContainerConfigurationHandler(),
             new LongLockHandler(grouper),
             new ThreadCountHandler(),
-            new ClassesLoadedHandler());
+            new ClassesLoadedHandler(),
+            new OldGarbageCollectionHandler(garbageCollectors));
```
jack-berg, line 90: These could be added in the switch statement as well. Instead of passing the set of `garbageCollectors` you could pass the name of the garbage collector as a constructor argument.

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/HandlerRegistry.java
roberttoyonaga: Not anymore. updated

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/HandlerRegistry.java
roberttoyonaga: This is a good idea. It also allows us to get rid of the set caching all the GC types. Done!

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/GarbageCollection/OldGarbageCollectionHandler.java
```diff
@@ -0,0 +1,70 @@
...
+
+  public OldGarbageCollectionHandler(String gc) {
+    // Set the attribute's GC based on which GC is being used.
+    if (gc.equals("PS MarkSweep")) {
+      attributes = Attributes.of(ATTR_GC, "PS MarkSweep", ATTR_ACTION, END_OF_MAJOR_GC);
+    } else if (gc.equals("G1 Old Generation")) {
+      attributes = Attributes.of(ATTR_GC, "G1 Old Generation", ATTR_ACTION, END_OF_MAJOR_GC);
+    } else if (gc.equals("MarkSweepCompact")) {
+      attributes = Attributes.of(ATTR_GC, "MarkSweepCompact", ATTR_ACTION, END_OF_MAJOR_GC);
+    }
```
jack-berg, line 41:
```suggestion
    attributes = Attributes.of(ATTR_GC, gc, ATTR_ACTION, END_OF_MAJOR_GC);
```

jfr-streaming/src/main/java/io/opentelemetry/contrib/jfr/metrics/internal/GarbageCollection/YoungGarbageCollectionHandler.java
```diff
@@ -0,0 +1,67 @@
...
+  private LongHistogram histogram;
+
+  public YoungGarbageCollectionHandler(String gc) {
+    // Set the attribute's GC based on which GC is being used.
+    // G1 young collection is already handled by G1GarbageCollectionHandler.
+    if (gc.equals("PS Scavenge")) {
+      attributes = Attributes.of(ATTR_GC, "PS Scavenge", ATTR_ACTION, END_OF_MINOR_GC);
+    } else if (gc.equals("Copy")) {
+      attributes = Attributes.of(ATTR_GC, "Copy", ATTR_ACTION, END_OF_MINOR_GC);
+    }
```
jack-berg, line 38:
```suggestion
    attributes = Attributes.of(ATTR_GC, gc, ATTR_ACTION, END_OF_MINOR_GC);
```

jack-berg (commented):
One last simplification

jack-berg (approved):

PR #672, opened 2023-01-03
Add dependabot ignores for maven artifacts

trask:
Should close #670


trask:
oops! closing in favor of #671

PR #689, opened 2023-01-13
Rename prometheus-collector to prometheus-client-bridge and prepare 1.22.0 change log

trask:


prometheus-client-bridge/src/main/java/io/opentelemetry/contrib/prometheusclientbridge/MetricAdapter.java
```diff
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-package io.opentelemetry.contrib.prometheus;
+package io.opentelemetry.contrib.prometheusclientbridge;
```
jkwatson, line 6: how about `..contrib.prometheus.clientbridge` ?

jack-berg (approved):

prometheus-client-bridge/src/main/java/io/opentelemetry/contrib/clientbridge/Serializer.java
```diff
@@ -19,7 +19,7 @@
  * Ocelli project by Netflix Inc. (https://github.com/Netflix/ocelli/).
  */
 
-package io.opentelemetry.contrib.prometheus;
+package io.opentelemetry.contrib.clientbridge;
```
mateuszrzeszutek, line 22: Should it be `clientbridge` or `prometheus. clientbridge`?

jkwatson (approved):

PR #709, opened 2023-01-19
JFR features can be enabled and disabled

### LINKED ISSUE #673: Clarify intent of JFR module
There's been some recent activity in the JFR module to implement the [JVM runtime semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/runtime-environment-metrics.md#jvm-metrics) by observing JFR events. It's got my wondering what exactly we're trying to achieve. The recent changes attempt to recreate the same data which is already available through JMX, but with more effort and lower quality.

I like the idea of using JFR events to expose data about the JVM which is not otherwise available, but I'm increasingly skeptical about the value of using JFR to duplicate the metrics we easily get from JMX.

@roberttoyonaga and @kittylyst - I'm especially interested in your perspective on this. Are there situations where a JFR source of these metrics would be superior to JMX? 

(Also, apologies for not raising this concern sooner. I appreciate the effort that @roberttoyonaga has recently dedicated to moving this forward.) 

Relavant issues: #653, #652, #651, #650

jack-berg:
Resolves #673.

Groups `RecordedEventHandlers` into "features", which can be toggled on and off, and which are enabled by default if not already implemented in the JMX based [runtime-metrics](https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation/runtime-metrics/library). 

This was discussed in the 1/5 SIG, as mentioned in this [comment](https://github.com/open-telemetry/opentelemetry-java-contrib/issues/673#issuecomment-1374223384).

A side effect of this is that the tests can enable the precise set of features that they test, further increasing isolation.

Also extends `JfrTelemetry#close()` to close all the observable instruments so that they stop reported data. Does this by having `RecordedEventHandler` extend `Closeable`, and having each implementation track its observable instruments.


jfr-streaming/README.md
```diff
@@ -4,7 +4,48 @@
...
+| MEMORY_ALLOCATION_METRICS | true            | `process.runtime.jvm.memory.allocation`                                                                                                                                                                     |
+| MEMORY_POOL_METRICS       | false           | `process.runtime.jvm.memory.init`\n`process.runtime.jvm.memory.committed`\n`process.runtime.jvm.memory.usage`\n`process.runtime.jvm.memory.limit`\n``\n``\n`process.runtime.jvm.memory.usage_after_last_gc` |
+| NETWORK_IO_METRICS        | true            | `process.runtime.jvm.network.io`\n`process.runtime.jvm.network.time`                                                                                                                                        |
+| THREAD_METRICS            | false           | `process.runtime.jvm.threads.count`                                                                                                                                                                         |
```
jack-berg, line 48: I've added descriptive text and this table to the readme to help explain how all the stuff in this PR comes together.

roberttoyonaga (approved):
This looks great to me!

kittylyst (approved):
LGTM. Nice PR!

trask (approved):

jack-berg:
> So we're enabling the additional metrics gathered via JFR streaming even though they are not yet outlined in the semantic conventions?

I think so. Should try to get feedback and have these added to the semantic conventions if they're useful (I think they will be).

PR #711, opened 2023-01-26
Fix and enhance resource detection logging.

breedx-splk:
I learned (the hard way) today that jul doesn't expand numeric placeholders in string templates if they are surrounded by single quotes. Yup.

* `logger.log(INFO, "{0}", "test")` -> `test`
* `logger.log(INFO, "'{0}'", "test")` -> `{0}`

So this results in pretty useless logs here. I also used this as an opportunity to enhance the details about when/why the resource provider doesn't kick in.

### DIFF
--- a/resource-providers/src/main/java/io/opentelemetry/resourceproviders/AppServerServiceNameProvider.java
+++ b/resource-providers/src/main/java/io/opentelemetry/resourceproviders/AppServerServiceNameProvider.java
@@ -7,6 +7,7 @@ package io.opentelemetry.resourceproviders;
 
 import static io.opentelemetry.semconv.resource.attributes.ResourceAttributes.SERVICE_NAME;
 import static java.util.logging.Level.INFO;
+import static java.util.logging.Level.WARNING;
 
 import com.google.auto.service.AutoService;
 import io.opentelemetry.api.common.Attributes;
@@ -37,11 +38,14 @@ public final class AppServerServiceNameProvider implements ConditionalResourcePr
   @Override
   public Resource createResource(ConfigProperties config) {
     String serviceName = detectServiceName();
-    if (serviceName != null) {
-      logger.log(INFO, "Auto-detected service name {0}.", serviceName);
-      return Resource.create(Attributes.of(SERVICE_NAME, serviceName));
+    if (serviceName == null) {
+      logger.log(
+          WARNING,
+          "Service name could not be detected using common application server strategies.");
+      return Resource.empty();
     }
-    return Resource.empty();
+    logger.log(INFO, "Auto-detected service name {0}.", serviceName);
+    return Resource.create(Attributes.of(SERVICE_NAME, serviceName));
   }
 
   @Nullable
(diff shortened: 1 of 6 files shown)


breedx-splk:
❌ is a Sonatype flaky link failure in markdown check. Re-run should clear it up (I verified the link works).

resource-providers/src/main/java/io/opentelemetry/resourceproviders/AppServerServiceNameProvider.java
```diff
@@ -37,11 +38,14 @@ public AppServerServiceNameProvider() {
   @Override
   public Resource createResource(ConfigProperties config) {
     String serviceName = detectServiceName();
-    if (serviceName != null) {
-      logger.log(INFO, "Auto-detected service name '{0}'.", serviceName);
-      return Resource.create(Attributes.of(SERVICE_NAME, serviceName));
+    if (serviceName == null) {
+      logger.log(
+          WARNING,
```
mateuszrzeszutek, line 43: `WARNING` and `INFO` are pretty high logging levels -- these messages will be printed out for everyone who's using this module (so, all Splunk distro users)

breedx-splk: True, but only if it got past `shouldApply()`. In that case, I do think it's helpful to see what is happening...but if you feel strongly about bumping all of the logging down to  `FINE` or something I can do that. 

Personally, I'd rather have more logs for troubleshooting than not enough to speculate....

mateuszrzeszutek: `shouldApply()` also logs a couple `INFO` messages that perhaps should be `FINE`.

I totally get it that it's valuable for troubleshooting, but we've had customers complain about the agent printing out stuff that does not match their logging patterns. E.g. let's assume that somebody runs a simple one jar app and expects the service name to be detected from the file name -- this resource provider would print 4 extra logs that are totally unrelated to their use case (because the jar name detector runs last).

resource-providers/src/main/java/io/opentelemetry/resourceproviders/AppServerServiceNameProvider.java
```diff
@@ -37,11 +38,14 @@ public AppServerServiceNameProvider() {
...
+          "Service name could not be detected using common application server strategies.");
+      return Resource.empty();
     }
-    return Resource.empty();
+    logger.log(INFO, "Auto-detected service name {0}.", serviceName);
```
mateuszrzeszutek, line 47: For the record, I think that `INFO` actually fits this log - you could add something mentioning where the service name was extracted from (e.g. `... in a app server deployment`)

resource-providers/src/main/java/io/opentelemetry/resourceproviders/AppServerServiceNameProvider.java
breedx-splk: `FINE` I'll change it. 🙃

resource-providers/src/main/java/io/opentelemetry/resourceproviders/AppServerServiceNameProvider.java
breedx-splk: Ok, I kept that one at `INFO`.

mateuszrzeszutek (approved):

mateuszrzeszutek:
Thanks!

PR #715, opened 2023-01-28
Generate JFR telemetry table

jack-berg:
Adds a gradle task called `./gradlew generateDocs` which programatically generates the telemetry table in `jfr-streaming/README.md`. 

This will ensure that the docs stay up to date following changes to the instrumentation.


jfr-streaming/src/test/java/io/opentelemetry/contrib/jfr/metrics/GenerateDocs.java
```diff
@@ -0,0 +1,150 @@
...
+                          + System.lineSeparator());
+                } catch (InterruptedException e) {
+                  Thread.currentThread().interrupt();
+                }
+              });
+      threads.add(thread);
+      thread.start();
+    }
+
+    for (Thread thread : threads) {
```
breedx-splk, line 88: I think it's marginally cleaner to use an `ExecutorService` for this stuff, but no biggie.

breedx-splk (approved):
This is cool, I appreciate the automation. I'm curious if we can think of any interesting failure modes that might cause us to miss data and cause the README to be misleading. The only thing I can think of is that sometimes GHA lags and maybe the hard coded sleep value won't be enough. 🤷🏻

jfr-streaming/src/test/java/io/opentelemetry/contrib/jfr/metrics/GenerateDocs.java
jack-berg: Realized I don't need to do these in separate threads at all. Can just start a bunch of JfrTelemetry instances in the current thread. They'll each subscribe to independent event streams and record only metrics for a particular JfrFeature.

jfr-streaming/src/test/java/io/opentelemetry/contrib/jfr/metrics/GenerateDocs.java
```diff
@@ -0,0 +1,147 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import java.util.regex.Pattern;
+
+public class GenerateDocs {
```
trask, line 27: since it's a little confusing to come across this in the tests
```suggestion
// this is called from ./gradlew generateDocs
public class GenerateDocs {
```

trask (approved):
👍

PR #716, opened 2023-01-28
Publish jfr streaming

jack-berg:
- Rename package from `io.opentelemetry.contrib.jfr.metrics` to `io.opentelemetry.contrib.jfr.streaming`
- Enable publishing, disable publishing of test fixtures


kittylyst (approved):
LGTM.

trask (approved):

PR #721, opened 2023-01-31
Jfr streaming version

### LINKED ISSUE #707: JFR HandlerRequest should source version from file
As part of #705, @jack-berg pointed out that the `HandlerRequest` has the version string used for instrumentation scope hard coded. This should really be consistent with the version of the artifacts, which means that we should source it from a file, like `version.properties`.

jack-berg:
Resolves #707. Builds on #716. 

Turns out this wasn't a good first issue because there was some gradle buildSrc work needed to add a file with the version that the source code could read. Lifted the source for this from [opentelemetry-java](https://github.com/open-telemetry/opentelemetry-java/blob/main/buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts#L144-L165).


jfr-streaming/src/test/java/io/opentelemetry/contrib/jfr/streaming/JfrTelemetryTest.java
```diff
@@ -45,7 +44,16 @@ void create_Default() {
       assertThat(logs.getEvents()).hasSize(1);
       logs.assertContains("Starting JfrTelemetry");
 
-      await().untilAsserted(() -> assertThat(reader.collectAllMetrics()).isNotEmpty());
+      assertThat(reader.collectAllMetrics())
+          .isNotEmpty()
+          .allSatisfy(
+              metric -> {
+                assertThat(metric.getInstrumentationScopeInfo().getName())
+                    .isEqualTo("io.opentelemetry.contrib.jfr");
```
trask, line 52: maybe, to match the module name?

```suggestion
                    .isEqualTo("io.opentelemetry.contrib.jfr.streaming");
```

consistent-sampling/build.gradle.kts
```diff
@@ -4,6 +4,7 @@ plugins {
 }
 
 description = "Sampler and exporter implementations for consistent sampling"
+otelJava.moduleName.set("io.opentelemetry.contrib.samplers")
```
trask, line 7: looks like we are using this package in two modules

runtime-attach/runtime-attach-core/build.gradle.kts
```diff
@@ -4,6 +4,7 @@ plugins {
 }
 
 description = "To help in create an OpenTelemetry distro able to runtime attach an OpenTelemetry Java Instrumentation agent"
+otelJava.moduleName.set("io.opentelemetry.contrib.attach")
```
trask, line 7: we could put this under `io.opentelemetry.contrib.attach.core` to not conflict with below

trask (approved):

resource-providers/build.gradle.kts
```diff
@@ -4,6 +4,7 @@ plugins {
 }
 
 description = "Contributed ResourceProviders"
+otelJava.moduleName.set("io.opentelemetry.resourceproviders")
```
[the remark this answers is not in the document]
jack-berg: Seems like just an oversight. @breedx-splk, @mateuszrzeszutek, @laurit do you care if we adjust the root package and module to be `io.opentelemetry.contrib.resourceproviders`?

runtime-attach/runtime-attach-core/build.gradle.kts
jack-berg: 👍

consistent-sampling/build.gradle.kts
jack-berg: I've made the following changes:
- `:samplers` base package changed to `io.opentelemetry.contrib.sampler`
- `:consistent-sampling` base package changed to `io.opentelemetry.contrib.sampler.consistent`

resource-providers/build.gradle.kts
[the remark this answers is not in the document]
breedx-splk: I'm ok with that change, thanks for checking. We'll have to touch our distro, but no biggie.
If anything, I'm sure I just wanted a flatter namespace (always, heh), but `contrib` makes sense here.

trask (approved):
thx 👍

PR #730, opened 2023-02-07
Cleanup jfr-streaming tests

jack-berg:
Makes the jfr-streaming tests a bit idiomatic:
- Eliminates couple usages of junit assertions
- Package private test classes and methods
- Fluent assertions instead of helper functions

### DIFF
--- a/jfr-streaming/src/g1GcTest/java/io/opentelemetry/contrib/jfr/streaming/G1GcMemoryMetricTest.java
+++ b/jfr-streaming/src/g1GcTest/java/io/opentelemetry/contrib/jfr/streaming/G1GcMemoryMetricTest.java
@@ -24,9 +24,7 @@ import static org.assertj.core.api.Assertions.assertThat;
 import io.opentelemetry.api.common.Attributes;
 import io.opentelemetry.sdk.metrics.data.HistogramData;
 import io.opentelemetry.sdk.metrics.data.HistogramPointData;
-import io.opentelemetry.sdk.metrics.data.MetricData;
 import io.opentelemetry.sdk.metrics.data.SumData;
-import org.assertj.core.api.ThrowingConsumer;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.RegisterExtension;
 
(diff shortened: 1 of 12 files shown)


jfr-streaming/src/g1GcTest/java/io/opentelemetry/contrib/jfr/streaming/G1GcMemoryMetricTest.java
```diff
@@ -41,34 +39,32 @@ class G1GcMemoryMetricTest {
...
             metric
                 .hasName(METRIC_NAME_MEMORY)
                 .hasUnit(BYTES)
                 .hasDescription(METRIC_DESCRIPTION_MEMORY)
-                .satisfies(attributeCheck),
+                .satisfies(
+                    data ->
+                        assertThat(data.getLongSumData().getPoints())
+                            .anyMatch(p -> p.getAttributes().equals(ATTR_G1_EDEN_SPACE))
+                            .anyMatch(p -> p.getAttributes().equals(ATTR_G1_SURVIVOR_SPACE))),
```
breedx-splk, line 57: I think it would be nice to have a method encapsulating this because it's duplicated below. It would at least read something more like 
```
.satisfies(this::verifyAttributes)
```

jfr-streaming/src/parallelGcTest/java/io/opentelemetry/contrib/jfr/streaming/PsGcMemoryMetricTest.java
```diff
@@ -43,33 +41,32 @@ class PsGcMemoryMetricTest {
...
                 .hasName(METRIC_NAME_MEMORY)
                 .hasUnit(BYTES)
                 .hasDescription(METRIC_DESCRIPTION_MEMORY)
-                .satisfies(attributeCheck),
+                .satisfies(
+                    data ->
+                        assertThat(data.getLongSumData().getPoints())
+                            .anyMatch(p -> p.getAttributes().equals(ATTR_PS_EDEN_SPACE))
+                            .anyMatch(p -> p.getAttributes().equals(ATTR_PS_SURVIVOR_SPACE))
+                            .anyMatch(p -> p.getAttributes().equals(ATTR_PS_OLD_GEN))),
```
breedx-splk, line 58: Same idea about deduping with a method.

jfr-streaming/src/serialGcTest/java/io/opentelemetry/contrib/jfr/streaming/SerialGcMemoryMetricTest.java
```diff
@@ -48,27 +46,24 @@ void shouldHaveGCDurationMetrics() throws Exception {
...
+                                                        "Copy",
+                                                        ATTR_ACTION,
+                                                        END_OF_MINOR_GC))
+                                            || p.getAttributes()
+                                                .equals(
+                                                    Attributes.of(
+                                                        ATTR_GC,
+                                                        "MarkSweepCompact",
+                                                        ATTR_ACTION,
+                                                        END_OF_MAJOR_GC))))));
```
breedx-splk, line 67: Spotless makes this really hard to read! Maybe if those two attrs were local vars it could clean up a bit?

jfr-streaming/src/test/java/io/opentelemetry/contrib/jfr/streaming/JfrThreadCountTest.java
```diff
@@ -22,26 +22,13 @@ class JfrThreadCountTest {
...
-
   @Test
   void shouldHaveJfrThreadCountEvents() throws Exception {
     // This should generate some events
     Runnable work =
         () -> {
-          // create contention between threads for one lock
           try {
-            doWork();
+            Thread.sleep(2000);
```
breedx-splk, line 31: I like removing `doWork()` but I thought the `2 * SAMPLING_INTERVAL` was slightly helpful.

breedx-splk (approved):
I like it! Had a few smaller ideas along the way, but definitely some improvements.

jfr-streaming/src/g1GcTest/java/io/opentelemetry/contrib/jfr/streaming/G1GcMemoryMetricTest.java
kittylyst: +1

jfr-streaming/src/test/java/io/opentelemetry/contrib/jfr/streaming/JfrThreadCountTest.java
kittylyst: +1

Also - should we consider moving to Awaitility rather than explicit Thread.sleep() calls?

kittylyst (approved):
Couple of minor comments, but generally LGTM.

jfr-streaming/src/test/java/io/opentelemetry/contrib/jfr/streaming/JfrThreadCountTest.java
jack-berg: > I like removing doWork() but I thought the 2 * SAMPLING_INTERVAL was slightly helpful.

Added an explanatory comment instead, which is even more useful since the reader doesn't have to guess what SAMPLING_INTERVAL represents.

> Also - should we consider moving to Awaitility rather than explicit Thread.sleep() calls?

We need to let the threads run long enough to produce `jdk.JavaThreadStatistics` JFR events to be consumed by the `ThreadCountHandler`, which awaitility doesn't help us with.

jfr-streaming/src/serialGcTest/java/io/opentelemetry/contrib/jfr/streaming/SerialGcMemoryMetricTest.java
jack-berg: Local vars improves it a lot 👍

trask (approved):

PR #749, opened 2023-02-20
Rename JFR streaming to JFR metrics

trask:
I think this would good to do before 1.23.0 release, since we only just enabled publishing of this module, and I don't think `jfr-streaming` a great fit.


jack-berg:
I anticipate other telemetry like events and other perhaps profiles (depending on how that area takes shape) being produced by this instrumentation. The metrics suffix seems too restrictive. Is `jfr-telemetry` overly broad?

trask:
> I anticipate other telemetry like events and other perhaps profiles (depending on how that area takes shape) being produced by this instrumentation. The metrics suffix seems too restrictive. Is `jfr-telemetry` overly broad?

is there a benefit to having a single module for all JFR-based telemetry, or do you think we would split things out?

when this goes over to instrumentation repo, one naming I was thinking of is

* `runtime-metrics-jfr`
* `runtime-metrics-jmx`

mateuszrzeszutek:
WDYT about renaming `runtime-metrics` to `runtime-telemetry-jmx` instead?

> is there a benefit to having a single module for all JFR-based telemetry, or do you think we would split things out?

Don't we usually try to have all possible telemetry signals in one package? E.g. HTTP metrics, Kafka metrics

jack-berg:
> is there a benefit to having a single module for all JFR-based telemetry, or do you think we would split things out?

There's definitely some setup / boilerplate code that would need to be shared amongst the different signals. I can't imagine the handlers for the additional pieces being so heavy weight that users would want it split out into multiple modules. 

> WDYT about renaming runtime-metrics to runtime-telemetry-jmx instead?

Sounds good to me. `runtime-telemetry-jmx` and `runtime-telemetry-jfr`

trask:
> Sounds good to me. `runtime-telemetry-jmx` and `runtime-telemetry-jfr`

👍

should we just not publish the contrib module for now?

cc @roberttoyonaga

jack-berg:
@trask if you're willing to sponsor it for inclusion in instrumentation I think we should postpone publishing rather than publishing under different coordinates in subsequent releases.

trask:
Closing, will send PR to disable publishing instead.

PR #759, opened 2023-02-23
add extended tracer module

zeitlinger:
The Opentelemetry APIs are rather difficult to use - therefore I've developed the following utility method, which is part of a company internal library - so it has already been tested by some users.

Based on [this PR](https://github.com/open-telemetry/opentelemetry-java/pull/5197), I'm now creating a PR here.


zeitlinger:
@trask I hope this PR makes more sense now - I've kept the idea of being able to wrap a `Tracer` - but I've also kept the static methods in the `Tracing` class.

extended-tracer/src/main/java/io/opentelemetry/contrib/tracer/Tracing.java
```diff
@@ -0,0 +1,269 @@
...
+                    entry -> entry.getKey().toLowerCase(Locale.ROOT), Map.Entry::getValue));
+    return GlobalOpenTelemetry.get()
+        .getPropagators()
+        .getTextMapPropagator()
+        .extract(current, normalizedTransport, TEXT_MAP_GETTER);
+  }
+
+  /** Sets baggage items which are active in given block. */
+  @SuppressWarnings("NullAway")
+  public static <T> T setBaggage(Map<String, String> baggage, Callable<T> callable) {
```
trask, line 192: maybe `runWithBaggage`

extended-tracer/src/main/java/io/opentelemetry/contrib/tracer/Tracing.java
```diff
@@ -0,0 +1,269 @@
...
+  public static void setSpanError(Span span, String description, Throwable exception) {
+    span.setStatus(StatusCode.ERROR, description);
+    span.recordException(exception);
+  }
+
+  /**
+   * Injects the current context into a string map, which can then be added to HTTP headers or the
+   * metadata of an event.
+   */
+  public static Map<String, String> injectContext() {
```
trask, line 149: `injectContext` sounds like it's doing the injection, instead of just returning the headers that should be injected

extended-tracer/src/main/java/io/opentelemetry/contrib/tracer/Tracing.java
```diff
@@ -0,0 +1,269 @@
...
+    return serviceTracer().call(spanName, callable);
+  }
+
+  public static <T> T call(Span span, Callable<T> callable) {
+    return call(span, callable, Tracing::setSpanError);
+  }
+
+  @SuppressWarnings("NullAway")
+  public static <T> T call(
+      Span span, Callable<T> callable, BiConsumer<Span, Exception> handleException) {
```
trask, line 80: it seems unclear that this method ends the span, I think I prefer the version that takes SpanBuilder since that alleviates the confusion

trask (commented):
I'm a little hesitant to introduce too much syntactic sugar on top of opentelemetry-api, since that creates more ways to do the same thing, which could cause confusion

at the same time, it makes sense to have somewhere to experiment with things that we might want to add into the opentelemetry-api itself

we also have opentelemetry-instrumentation-api, which possibly has a little bit of overlap

I left a couple of initial comments, overall I think it might help to pair the PR down and just focus on the `Tracing` class at least initially

extended-tracer/src/main/java/io/opentelemetry/contrib/tracer/CurrentSpan.java
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  /**
+   * Marks the current span as error.
+   *
+   * @param exception the exception that caused the error
+   */
+  public static void setSpanError(Throwable exception) {
+    Tracing.setSpanError(Span.current(), exception);
+  }
```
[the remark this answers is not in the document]
zeitlinger: Yes, that was the intention.

extended-tracer/src/main/java/io/opentelemetry/contrib/tracer/Tracing.java
zeitlinger: good point - what about `getPropagationHeaders`?

extended-tracer/src/main/java/io/opentelemetry/contrib/tracer/Tracing.java
zeitlinger: ending the span is the essential part of this method, but I'm not sure I get your point, since all methods in this class end the spans.

extended-tracer/src/main/java/io/opentelemetry/contrib/tracer/Tracing.java
zeitlinger: good point - I suggest `callWithBaggage` to make it consistent with the other methods.

zeitlinger:
> we also have opentelemetry-instrumentation-api, which possibly has a little bit of overlap

I checked `Instrumenter` but didn't see anything that could be reused.

> I left a couple of initial comments, overall I think it might help to pair the PR down and just focus on the `Tracing` class at least initially

I like that idea and have adjusted the PR accordingly :smile:

extended-tracer/src/main/java/io/opentelemetry/contrib/tracer/Tracing.java
trask: it seems clear to me that methods which create spans also end them, but it seems unclear to me for methods which get an active span passed into them

zeitlinger: great point - I've renamed to "(call|run)AndEndSpan"

extended-tracer/src/main/java/io/opentelemetry/contrib/tracer/Tracing.java
```diff
@@ -0,0 +1,267 @@
...
+    }
+  }
+
+  /**
+   * Return the tracer to be used in a service
+   *
+   * @return the tracer to be used in a service
+   */
+  public static Tracer serviceTracer() {
+    return GlobalOpenTelemetry.getTracer("service");
```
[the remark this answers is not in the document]
zeitlinger: good idea - I've extracted `GlobalTracing` to make that clear

extended-tracer/src/main/java/io/opentelemetry/contrib/tracer/GlobalTracing.java
```diff
@@ -0,0 +1,119 @@
...
+import io.opentelemetry.api.GlobalOpenTelemetry;
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.SpanBuilder;
+import io.opentelemetry.api.trace.Tracer;
+import io.opentelemetry.context.Context;
+import java.util.Map;
+import java.util.concurrent.Callable;
+import java.util.function.BiConsumer;
+
+public class GlobalTracing {
```
jkwatson, line 17: Recommend adding some class-level Javadoc to explain the use-cases for this class.

extended-tracer/src/main/java/io/opentelemetry/contrib/tracer/Tracing.java
```diff
@@ -0,0 +1,261 @@
...
+import java.util.HashMap;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.Callable;
+import java.util.function.BiConsumer;
+import java.util.stream.Collectors;
+import javax.annotation.Nullable;
+
+public final class Tracing {
```
jkwatson, line 31: Recommend Javadoc here

extended-tracer/src/main/java/io/opentelemetry/contrib/tracer/Tracing.java
```diff
@@ -0,0 +1,261 @@
...
+        @Nullable
+        public String get(@Nullable Map<String, String> carrier, String key) {
+          //noinspection ConstantConditions
+          return carrier == null ? null : carrier.get(key);
+        }
+      };
+
+  private Tracing() {}
+
+  public static void run(Tracer tracer, String spanName, Runnable runnable) {
```
jkwatson, line 50: all the public methods, etc should have javadoc on them.

extended-tracer/src/main/java/io/opentelemetry/contrib/tracer/Tracing.java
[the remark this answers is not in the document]
trask: > Is there a way we can rename this class to make it super clear that it's utilizing the global, and should only be used in conjunction with the agent?

alternatively, what do you think of injecting an instance of `OpenTelemetry` into `Tracing` (and changing all methods from static to instance)?

extended-tracer/src/main/java/io/opentelemetry/contrib/tracer/Tracing.java
trask: maybe worth waiting on javadoc until we agree a bit more on scope / shape

extended-tracer/src/main/java/io/opentelemetry/contrib/tracer/Tracing.java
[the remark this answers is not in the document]
zeitlinger: that is pretty close to what I had with `ExtendedTracer` :+1:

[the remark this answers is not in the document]
zeitlinger: I have both now - injetcted `OpenTelemetry` into `Tracing` and a `GlobalTracing` method (which I still find useful).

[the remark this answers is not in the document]
trask: > I have both now - injetcted `OpenTelemetry` into `Tracing` and a `GlobalTracing` method (which I still find useful).

I agree the global can be convenient, but it also creates problems and so we have been trying to eliminate its use from all of our instrumentation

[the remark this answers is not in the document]
zeitlinger: the `GlobalTracing` is meant to stay in contrib basically forever, giving users the choice to use the global tracer - in the same way as they will always be able to use `GlobalOpenTelemetry`.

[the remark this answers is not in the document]
trask: I think that currently the recommended use of `GlobalOpenTelemetry` is to use it only once, and inject the resulting `OpenTelemetry` instance wherever you need it. This would probably make a good Java SIG discussion.

extended-tracer/src/main/java/io/opentelemetry/contrib/tracer/Tracing.java
```diff
@@ -0,0 +1,252 @@
...
+          return null;
+        });
+  }
+
+  /**
+   * Runs a block of code with a new span - ending the span at the end and recording any exception.
+   *
+   * @param spanName name of the new span
+   */
+  public <T> T call(Tracer tracer, String spanName, Callable<T> callable) {
```
trask, line 71: inject `Tracer` into the constructor?

zeitlinger:
@trask @jkwatson @jack-berg as discussed yesterday, I've removed GlobalTracer, injected Tracer into Tracing, and added before/after examples.

extended-tracer/README.md
```diff
@@ -0,0 +1,149 @@
...
+} finally {
+  span.end();
+}
+```
+
+After:
+
+```java
+Tracing tracing = new Tracing(openTelemetry, "service");
+String transactionId = tracing.call("reset_checkout", () -> resetCheckout(cartId));
```
kenfinnigan, line 31: Firstly, thanks for putting together the examples @zeitlinger.

One question with this example, the checked exception is used to set an error on the span but the application code never sees it. Or are these usages targeted at instrumentation developers and not application developers?

zeitlinger: the exceptions should be re-thrown after being recorded, e.g. [here](https://github.com/zeitlinger/opentelemetry-java-contrib/blob/09c67e117fad6e44e48c10e9862e76473711a6b3/extended-tracer/src/main/java/io/opentelemetry/contrib/tracer/Tracing.java#L90)

trask: Should we change `Exception` to `Throwable` in the example (and in the `Tracing` class)?

kenfinnigan: Could the example be updated to show application code needing to change from catching `Exception` to `Throwable`? It's an important point to note as developers will need to be aware the checked exception their code would normally throw is being hidden in a Throwable.

zeitlinger: good points - both of them :smile: 

I changed to use `Throwable` consistently & added a separate section to explain error handling

extended-tracer/src/main/java/io/opentelemetry/contrib/tracer/Tracing.java
```diff
@@ -0,0 +1,255 @@
...
+
+  public Tracing(OpenTelemetry openTelemetry, String tracerName) {
+    this.openTelemetry = openTelemetry;
+    this.tracer = openTelemetry.getTracer(tracerName);
+  }
+
+  public void run(String spanName, Runnable runnable) {
+    runAndEndSpan(tracer.spanBuilder(spanName).startSpan(), runnable);
+  }
+
```
PeterF778, line 59: I think a useful extension for Runnables would be ability to wrap them in Spans, forming another Runnable which could be suitable for submission to Executor services for asynchronous execution.

breedx-splk: I agree, but would advocate for doing that in a follow-up PR.

breedx-splk: Relates to https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/9325

zeitlinger:
I've added javadocs - so please give it a final review :smile:

breedx-splk (approved):
Looks good to me! Thanks.

trask (commented):
what do you think of changing `runAndEndSpan(Span, Runnable)` to `run(SpanBuilder, Runnable)`? (and same for `callAndEndSpan`)

zeitlinger:
> what do you think of changing `runAndEndSpan(Span, Runnable)` to `run(SpanBuilder, Runnable)`? (and same for `callAndEndSpan`)

great idea - makes the API more intuitive!

zeitlinger:
@trask can you check again?

extended-tracer/src/main/java/io/opentelemetry/contrib/tracer/Tracing.java
```diff
@@ -0,0 +1,386 @@
...
+   *
+   * <p>If an exception is thrown by the {@link Runnable}, the span will be marked as error, and the
+   * exception will be recorded.
+   *
+   * @param spanBuilder the {@link SpanBuilder} to use
+   * @param callable the {@link Callable} to call
+   * @param <T> the type of the result
+   * @return the result of the {@link Callable}
+   */
+  public <T> T call(SpanBuilder spanBuilder, Callable<T> callable) {
```
trask, line 148: I'm not sure about hiding checked exceptions in these methods. I'm interested in hearing from @jack-berg and @jkwatson since the goal is for this module to align with OpenTelemetry API practices.

```suggestion
  public <T> T call(SpanBuilder spanBuilder, Callable<T> callable) throws Exception {
```

extended-tracer/src/main/java/io/opentelemetry/contrib/tracer/Tracing.java
```diff
@@ -0,0 +1,386 @@
...
+   * <p>If an exception is thrown by the {@link Callable}, the span will be marked as error, and the
+   * exception will be recorded.
+   *
+   * @param transport the transport where to extract the span context from
+   * @param spanBuilder the {@link SpanBuilder} to use
+   * @param callable the {@link Callable} to call
+   * @param <T> the type of the result
+   * @return the result of the {@link Callable}
+   */
+  public <T> T traceConsumerSpan(
```
trask, line 341: are separate `traceConsumerSpan` / `traceServerSpan` methods worth it now that they accept spanBuilder and user can setSpanKind before passing that in?

extended-tracer/src/main/java/io/opentelemetry/contrib/tracer/Tracing.java
```diff
@@ -0,0 +1,386 @@
...
+
+  /**
+   * Marks a span as error.
+   *
+   * @param span the span
+   * @param description what went wrong
+   */
+  public void setSpanError(Span span, String description) {
+    span.setStatus(StatusCode.ERROR, description);
+  }
```
trask, line 190: how important is it having the `setSpanError` methods be public? it feels like they are not needed for the primary use cases

extended-tracer/src/main/java/io/opentelemetry/contrib/tracer/Tracing.java
```diff
@@ -0,0 +1,386 @@
...
+            });
+
+    return transport;
+  }
+
+  /**
+   * Extract the context from a string map, which you get from HTTP headers of the metadata of an
+   * event you're processing.
+   */
+  public Context extractContext(Map<String, String> transport) {
```
trask, line 241: is it important for this to be public?

extended-tracer/src/main/java/io/opentelemetry/contrib/tracer/Tracing.java
zeitlinger: changed using ThrowingSupplier as discussed

extended-tracer/src/main/java/io/opentelemetry/contrib/tracer/Tracing.java
zeitlinger: no - fixed

extended-tracer/src/main/java/io/opentelemetry/contrib/tracer/Tracing.java
zeitlinger: no - fixed

extended-tracer/src/main/java/io/opentelemetry/contrib/tracer/Tracing.java
zeitlinger: I would say yes - it's guiding the user to set the span kind.

I also can't find a good name otherwise - for making this method public:

```java
private <T, E extends Throwable> T extractAndRun(
      SpanKind spanKind,
      Map<String, String> transport,
      SpanBuilder spanBuilder,
      ThrowingSupplier<T, E> throwingSupplier,
      BiConsumer<Span, Throwable> handleException)
      throws E {
    try (Scope ignore = extractContext(transport).makeCurrent()) {
      return call(spanBuilder.setSpanKind(spanKind), throwingSupplier, handleException);
    }
  }
```

zeitlinger:
@trask please check again :smile:

extended-tracer/src/main/java/io/opentelemetry/contrib/tracer/ThrowingSupplier.java
```diff
@@ -0,0 +1,18 @@
...
+import java.util.function.Supplier;
+
+/**
+ * A utility interface representing a {@link Supplier} that may throw.
+ *
+ * @param <E> Thrown exception type.
+ */
+@FunctionalInterface
+public interface ThrowingSupplier<T, E extends Throwable> {
+  T get() throws E;
```
trask, line 17: since this is our interface, we could potentially give it more descriptive class/method names

a possible option, borrowing from Spring's [TransactionCallback/doInTransaction](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/transaction/reactive/TransactionCallback.html):

```suggestion
public interface SpanCallback<T, E extends Throwable> {
  T doInSpan() throws E;
```

zeitlinger: great idea :smile:

extended-tracer/src/main/java/io/opentelemetry/contrib/tracer/SpanCallback.java
```diff
@@ -0,0 +1,18 @@
...
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.contrib.tracer;
+
+import java.util.function.Supplier;
+
+/**
+ * A utility interface representing a {@link Supplier} that may throw.
```
trask, line 11: could be more descriptive/specific now after the rename

extended-tracer/src/main/java/io/opentelemetry/contrib/tracer/Tracing.java
```diff
@@ -0,0 +1,374 @@
...
+
+  private final Tracer tracer;
+
+  /**
+   * Creates a new instance of {@link Tracing}.
+   *
+   * @param openTelemetry the {@link OpenTelemetry} instance
+   * @param tracerName the name of the tracer to use
+   */
+  public Tracing(OpenTelemetry openTelemetry, String tracerName) {
```
trask, line 63:
```suggestion
  public Tracing(OpenTelemetry openTelemetry, String instrumentationScopeName) {
```

extended-tracer/src/test/java/io/opentelemetry/contrib/tracer/TracingTest.java
```diff
@@ -0,0 +1,149 @@
...
+        tracing.call(
+            "parent",
+            () ->
+                tracing.callWithBaggage(
+                    Collections.singletonMap("key", "value"),
+                    () -> Baggage.current().getEntryValue("key")));
+
+    assertThat(value).isEqualTo("value");
+  }
+
```
trask, line 72: it could be worth adding tests for other public methods, not only for the coverage but for showing usage examples

extended-tracer/src/test/java/io/opentelemetry/contrib/tracer/TracingTest.java
```diff
@@ -0,0 +1,149 @@
...
+        ThrowingBiConsumer<Tracing, SpanCallback<Void, Throwable>> extractAndRun,
+        SpanKind wantKind,
+        StatusData wantStatus) {
+      this.extractAndRun = extractAndRun;
+      this.wantKind = wantKind;
+      this.wantStatus = wantStatus;
+    }
+  }
+
+  @Keep
```
trask, line 88: what does this annotation do?

extended-tracer/src/main/java/io/opentelemetry/contrib/tracer/Tracing.java
```diff
@@ -0,0 +1,374 @@
...
+    span.setStatus(StatusCode.ERROR);
+    span.recordException(exception);
+  }
+
+  /**
+   * Injects the current context into a string map, which can then be added to HTTP headers or the
+   * metadata of an event.
+   */
+  public Map<String, String> getPropagationHeaders() {
+    Map<String, String> transport = new HashMap<>();
```
jack-berg, line 205: Rename transport -> carrier

extended-tracer/src/main/java/io/opentelemetry/contrib/tracer/Tracing.java
```diff
@@ -0,0 +1,374 @@
...
+   * The span will be ended at the end of the {@link Runnable}.
+   *
+   * <p>If an exception is thrown by the {@link Runnable}, the span will be marked as error, and the
+   * exception will be recorded.
+   *
+   * @param spanBuilder the {@link SpanBuilder} to use
+   * @param runnable the {@link Runnable} to run
+   */
+  @SuppressWarnings("NullAway")
+  public void run(SpanBuilder spanBuilder, Runnable runnable) {
```
jack-berg, line 112: The run variants aren't symmetric to the call variants - there aren't any overloads for handling exceptions. I think you probably did this because `Runnable#run` doesn't throw, but you added a custom functional interface in SpanCallback which combines characteristics of both Supplier and Callable, allowing the implementation's throwable to be a parameterized type. Why not do the same for Runnable for symmetry?

extended-tracer/src/main/java/io/opentelemetry/contrib/tracer/Tracing.java
```diff
@@ -0,0 +1,374 @@
...
+  private static void setSpanError(Span span, Throwable exception) {
+    span.setStatus(StatusCode.ERROR);
+    span.recordException(exception);
+  }
+
+  /**
+   * Injects the current context into a string map, which can then be added to HTTP headers or the
+   * metadata of an event.
+   */
+  public Map<String, String> getPropagationHeaders() {
```
jack-berg, line 204: Calling this headers isn't correct since this method could easily be used to obtain context and propagate it through some different means. Maybe something more generate like `getTextMapPropagationContext()`?

extended-tracer/src/main/java/io/opentelemetry/contrib/tracer/Tracing.java
```diff
@@ -0,0 +1,374 @@
...
+            });
+
+    return transport;
+  }
+
+  /**
+   * Extract the context from a string map, which you get from HTTP headers of the metadata of an
+   * event you're processing.
+   */
+  private Context extractContext(Map<String, String> transport) {
```
jack-berg, line 226: If you go with `getTextMapPropagationContext` as suggested above, maybe this should be `extractTextMapPropagationContext()`?

Its weird that in order to access this method, which is purely related to context propagation, you need to configure an object called `Tracing` and include information to configure a `Tracer`.

What do you think about creating a second object called `Propagation` to hold the methods that are purely propagation related?

jack-berg (commented):
Left a few comments but I don't feel too strongly about any of it. Just thoughts I had reading through it.

extended-tracer/src/test/java/io/opentelemetry/contrib/tracer/TracingTest.java
zeitlinger: it flags the method as being used (by reflection) to avoid warnings about unused methods.

extended-tracer/src/main/java/io/opentelemetry/contrib/tracer/Tracing.java
```diff
@@ -0,0 +1,255 @@
...
+  }
+
+  /**
+   * Marks a span as error.
+   *
+   * @param span the span
+   * @param description what went wrong
+   */
+  public void setSpanError(Span span, String description) {
+    span.setStatus(StatusCode.ERROR, description);
```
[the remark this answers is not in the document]
zeitlinger: it's already private

extended-tracer/src/main/java/io/opentelemetry/contrib/tracer/Tracing.java
```diff
@@ -0,0 +1,374 @@
...
+    this(openTelemetry, openTelemetry.getTracer(tracerName));
+  }
+
+  /**
+   * Creates a new instance of {@link Tracing}.
+   *
+   * @param openTelemetry the {@link OpenTelemetry} instance
+   * @param tracer the {@link Tracer} to use
+   */
+  public Tracing(OpenTelemetry openTelemetry, Tracer tracer) {
```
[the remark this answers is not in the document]
zeitlinger: no, I actually had the case in mind, where you already have a tracer object from somewhere (e.g. a spring bean) - and you just want to keep using that.

extended-tracer/src/main/java/io/opentelemetry/contrib/tracer/Tracing.java
zeitlinger: great idea :+1:

extended-tracer/src/main/java/io/opentelemetry/contrib/tracer/Tracing.java
zeitlinger: it's really just that one method - now called `getTextMapPropagationContext`.

I'm not sure it helps discoverability if we move this method to a different class.

zeitlinger:
@jack-berg @trask thanks for the review - I incorporated most changes.

extended-tracer/src/main/java/io/opentelemetry/contrib/tracer/Tracing.java
jack-berg: > I'm not sure it helps discoverability if we move this method to a different class.

Yeah it definitely hurts. What about making the methods which don't depend on `Tracer` be pure functions? I.e. public static method which accept `OpenTelemetry` as a parameter. That way, users who wish to benefit from their convenience but don't want to use any of the methods that depend on `Tracer` can use them without having to create a dummy tracer.

zeitlinger: I can make this one method a static method in `Propagation` (as you suggested) - and leave the existing method as-is.

zeitlinger: `callWithBaggage` can be made static as well.

zeitlinger: done @jack-berg

breedx-splk:
@zeitlinger check out the few small build failures.

zeitlinger:
@jack-berg @trask thanks for the review - I incorporated most changes.

jack-berg (approved):

extended-tracer/README.md
```diff
@@ -0,0 +1,168 @@
...
+    span.end();
+  }
+}
+```
+
+After:
+
+```java
+Tracing tracing = new Tracing(openTelemetry, "service");
+Map<String, String> propagationHeaders = tracing.getPropagationHeaders();
```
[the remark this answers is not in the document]
zeitlinger: good catch - changed to `Map<String, String> propagationHeaders = tracing.getTextMapPropagationContext();`

extended-tracer/README.md
```diff
@@ -0,0 +1,168 @@
...
+`Tracing` calls and handle them as you would without `Tracing`.
+
+Note that the `Tracing` methods do not declare any checked exceptions
+(the idea is taken from [@SneakyThrows](https://projectlombok.org/features/SneakyThrows)).
+Declaring a checked exception would force callers to handle it, which would create a lot of
+boilerplate code. Instead, `Tracing` re-throws checked exceptions as unchecked exceptions.
+
+## Component owners
+
+- [Gregor Zeitlinger](https://github.com/zeitlinger), Grafana Labs
```
trask, line 166: it would be great to get one more component owner, any takers?

extended-tracer/build.gradle.kts
```diff
@@ -0,0 +1,13 @@
+plugins {
+  id("otel.java-conventions")
+  id("otel.publish-conventions")
+}
+
+description = "Tracing Utilities"
```
trask, line 6:
```suggestion
description = "Extended Tracer"
```

extended-tracer/src/test/java/io/opentelemetry/contrib/tracer/TracingTest.java
trask: oh, nice, we generally use `@SuppressWarnings("unused")` but this is shorter 😄

extended-tracer/build.gradle.kts
```diff
@@ -0,0 +1,13 @@
+plugins {
+  id("otel.java-conventions")
+  id("otel.publish-conventions")
```
trask, line 3: @zeitlinger we discussed in Java SIG today, and would like to merge this without publishing, and then move it into the core repo incubator

```suggestion
```

trask (approved):

extended-tracer/README.md
zeitlinger: is this still applicable if the plan is to move this to core repo incubator?

zeitlinger:
@trask build is passing now :smile:

extended-tracer/README.md
trask: nope

trask:
thanks @zeitlinger for your patience and incorporating all the feedback!

PR #777, opened 2023-03-10
Add AwsAttributePropagatingSpanProcessor component to AWS X-Ray

bjrara:
## Context

`AwsAttributePropagatingSpanProcessor` is designed to handle attributes `aws.remote.application`, `aws.remote.operation` and `aws.local.operation` which will later be used to generate metrics under the background of https://github.com/open-telemetry/opentelemetry-java-contrib/issues/789.

It has a prerequisite PR https://github.com/open-telemetry/opentelemetry-java-contrib/pull/800.

## Description

`AwsAttributePropagatingSpanProcessor` accomplishes two things:

### 1. Propagating remote attributes.
The `aws.remote.application` and `aws.remote.operation` attributes are designed to be set by users into local client spans through manual instrumentation. However, since client spans are mostly generated by auto-instrumentation libraries which make them unaccessible to application developers.

[OTEP-207](https://github.com/open-telemetry/oteps/pull/207) seems to be a general solution that could help address this requirement, but it hasn't been finalized yet. We took the idea from [this comment](https://github.com/open-telemetry/oteps/pull/207#pullrequestreview-1006473604) which suggests to let SpanProcessor to do the propagation work. 

`AwsAttributePropagatingSpanProcessor ` is used as an alternative solution that inherit `aws.remote.application` and `aws.remote.operation` attributes from parent span to the child spans if present.

### 2. Setting and propagating local attributes.
`SpanMetricsProcessor` from https://github.com/open-telemetry/opentelemetry-java-contrib/issues/789 requires `aws.local.operation` (the initial entry point of a execution) to be injected into the local client spans to generate metrics that can later be used for other correlations.

`AwsAttributePropagatingSpanProcessor` sets this attribute with the span name of the local root span if the request is from remote, and passes it to child spans.

## Testing

JUnit test is created for this new component in the package.

## Documentation

It's an independent component that would require manual installation in the TracerProvider. We'll create a doc to describe the usage.

## Outstanding items

NONE.

### DIFF
--- a/aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsAttributeKeys.java
+++ b/aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsAttributeKeys.java
@@ -1,0 +1,23 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.contrib.awsxray;
+
+import io.opentelemetry.api.common.AttributeKey;
+
+/** Utility class holding attribute keys with special meaning to AWS components */
+final class AwsAttributeKeys {
+
+  private AwsAttributeKeys() {}
+
+  static final AttributeKey<String> AWS_LOCAL_OPERATION =
+      AttributeKey.stringKey("aws.local.operation");
+
+  static final AttributeKey<String> AWS_REMOTE_APPLICATION =
+      AttributeKey.stringKey("aws.remote.application");
+
+  static final AttributeKey<String> AWS_REMOTE_OPERATION =
+      AttributeKey.stringKey("aws.remote.operation");
+}
--- a/aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsAttributePropagatingSpanProcessor.java
+++ b/aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsAttributePropagatingSpanProcessor.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.contrib.awsxray;
+
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.SpanContext;
+import io.opentelemetry.api.trace.SpanKind;
+import io.opentelemetry.context.Context;
+import io.opentelemetry.sdk.trace.ReadWriteSpan;
+import io.opentelemetry.sdk.trace.ReadableSpan;
+import io.opentelemetry.sdk.trace.SpanProcessor;
+
+/**
+ * AwsAttributePropagatingSpanProcessor handles the creation of the aws.local.operation attribute,
+ * and the inheritance of the {@link AwsAttributeKeys#AWS_LOCAL_OPERATION}, {@link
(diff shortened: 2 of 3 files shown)


willarmiros (approved):
This change LGTM and I support it being included in the X-Ray contrib repo.

@open-telemetry/java-instrumentation-approvers PTAL and merge when ready :)

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/RemoteAttributesSpanProcessor.java
```diff
@@ -0,0 +1,58 @@
...
+import io.opentelemetry.sdk.trace.ReadWriteSpan;
+import io.opentelemetry.sdk.trace.ReadableSpan;
+import io.opentelemetry.sdk.trace.SpanProcessor;
+
+/**
+ * RemoteAttributesSpanProcessor copies attributes aws.remote.application and aws.remote.operation
+ * from the parent span to children after receiving them. These two attributes will be used as
+ * RemoteApplication and RemoteOperation dimension values in dependency metrics generation.
+ */
+public class RemoteAttributesSpanProcessor implements SpanProcessor {
```
jack-berg, line 20: Reminder that this artifact is stable and we can't make breaking changes to any of the API surface area. Therefore, we should think carefully about this addition.

Comments:
- Make the class final
- Add a private constructor
- Add a `public static RemoteAttributesSpanProcessor create()` method for initializing.

bjrara: Updated.

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/RemoteAttributesSpanProcessor.java
```diff
@@ -0,0 +1,64 @@
...
+  @Override
+  public void onStart(Context parentContext, ReadWriteSpan span) {
+    Span parentSpan = Span.fromContextOrNull(parentContext);
+    if (!(parentSpan instanceof ReadableSpan)) {
+      return;
+    }
+    ReadableSpan parentReadableSpan = (ReadableSpan) parentSpan;
+
+    String remoteApplication = parentReadableSpan.getAttribute(REMOTE_APPLICATION);
+    if (remoteApplication != null) {
```
jack-berg, line 42: Can you add test coverage for the case where `aws.remote.application` / `aws.remote.operation` are null?

jack-berg (approved):
This looks ok to me. Note to other reviews that this artifact is stable so please take a close look at the new API surface area.

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/RemoteAttributesSpanProcessor.java
bjrara: Thanks for the review. Added a new test case for null check.

willarmiros (approved):
Additional changes LGTM! This can be merged whenever and thanks for the review @jack-berg

jack-berg:
@open-telemetry/java-contrib-maintainers PTAL.

bjrara:
Put this PR on hold as the component might need to extend its scope to cooperate with https://github.com/open-telemetry/opentelemetry-java-contrib/issues/789.

thpierce:
Because the author of this PR is not currently available, I have created a new PR here: https://github.com/open-telemetry/opentelemetry-java-contrib/pull/856

Please close this PR if you are able.

Issue #789, opened 2023-03-16
Add new Sampler and SpanProcessor to allow for generating metrics from 100% of spans without impacting sampling


thpierce:
**Is your feature request related to a problem? Please describe.**
Yes. We wish to generate key metrics from OTEL auto-instrumented spans. I had [previously engaged](https://github.com/open-telemetry/opentelemetry-specification/issues/3145) the community to discuss this, and after some conversations, decided that the best approach would be to implement a Sampler and a Span Processor to the `contrib` repositories.

The major requirements for our project are as follows:
* Metrics should be produced from 100% of Client/Server/Producer/Consumer spans produced by OTEL Auto-Instrumentation, regardless of sampling decision
   * Span sampling decisions should still be respected w.r.t. exporting spans to the collector and propagating sampling decisions to child spans
* Metric generation must be done when the span is ended, as metric attributes are derived from some span attributes that are not always available when the span is started (e.g. `http.route`).
* Metric attributes need to be added to the span attributes before export to Collector, to be used to correlate metrics and spans in a downstream process.

I am raising this issue to garner attention for this change before I raise a PR, and perhaps run conversations in parallel with implementation, which is ongoing. Please note there may be some differences in the final implementation, but the broad strokes should remain the same.

**Describe the solution you'd like**

All contributions would go to [opentelemetry-java-contrib/aws-xray/src/main/java/io/opentelemetry/contrib/awsxray](https://github.com/open-telemetry/opentelemetry-java-contrib/tree/main/aws-xray/src/main/java/io/opentelemetry/contrib/awsxray).

* First, a new Sampler, tentatively called `AlwaysRecordSampler`.
   * This is an aggregate Sampler that is initialized with a root Sampler (akin to [ParentBasedSampler](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/ParentBasedSampler.java)).
   * `shouldSample()` will call the root sampler and, if the `SampleDecision` is `RECORD_AND_SAMPLE` or `RECORD_ONLY`, it simply returns the result. However, if the `SampleDecision` is `DROP`, it will instead create a new result with `SampleDecision` set to `RECORD_ONLY` (all other result behaviour unchanged).
   * This Sampler will ensure that 100% of spans get send to the SpanProcessor, without changing the actual Sampling rate/behaviour.
[...]

PR #801, opened 2023-03-30
Breakout ResourceHolder from AwsXrayRemoteSamplerProvider

thpierce:
**Description:**

ResourceHolder provides generic utility for other components in the awsxray package, and could be more broadly used. In this commit, we are breaking out the ResourceHolder inner class to it's own class, and exposing the functionality of getResource, so that other classes can rely on it.

Specifically, we are working on new components to meet the needs of https://github.com/open-telemetry/opentelemetry-java-contrib/issues/789, which will require the use of this ResourceHolder. Opening up ResourceHolder seems to us to be the best path forward.

**Existing Issue(s):**

Tangentially related: https://github.com/open-telemetry/opentelemetry-java-contrib/issues/789

**Testing:**

* Added unit tests, validated they all pass
* `./gradlew clean assemble` succeeds

**Documentation:**

Unclear if any documentation is required here, can help contribute this as needed. ResourceHolder is a new independent component that can be used to retrieve the Resource from autoconfiguration.

**Outstanding items:**

None

### DIFF
--- a/aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsXrayRemoteSamplerProvider.java
+++ b/aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsXrayRemoteSamplerProvider.java
@@ -13,18 +13,13 @@ import io.opentelemetry.sdk.autoconfigure.spi.traces.ConfigurableSamplerProvider
 import io.opentelemetry.sdk.resources.Resource;
 import io.opentelemetry.sdk.trace.samplers.Sampler;
 import java.util.Map;
-import javax.annotation.Nullable;
 
 @AutoService(ConfigurableSamplerProvider.class)
 public class AwsXrayRemoteSamplerProvider implements ConfigurableSamplerProvider {
 
   @Override
   public Sampler createSampler(ConfigProperties config) {
-    Resource resource = ResourceHolder.resource;
-    if (resource == null) {
-      // Should never be the case in practice.
-      resource = Resource.getDefault();
-    }
+    Resource resource = io.opentelemetry.contrib.awsxray.ResourceHolder.getResource();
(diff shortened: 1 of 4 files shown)


thpierce:
CLA signed, but can't get the workflow to update. Will just close and re-open.

willarmiros (approved):
lgtm!

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsXrayRemoteSamplerProvider.java
```diff
@@ -41,22 +34,4 @@ public Sampler createSampler(ConfigProperties config) {
   public String getName() {
     return "xray";
   }
-
-  // Currently the only way to read the Resource from autoconfiguration. Best would be if the SPI
-  // could return a Function<SamplerFactoryArgs, Sampler> where SamplerFactoryArgs has
-  // SDK-constructed components like Resource and Clock.
-  @AutoService(AutoConfigurationCustomizerProvider.class)
-  public static final class ResourceHolder implements AutoConfigurationCustomizerProvider {
```
trask, line 49: is it ok to delete this public class given `aws-xray` module is stable?

trask: thanks, how about mark it `@Deprecated` and make it a no-op, since that will still conform to our versioning guarantees for stable artifacts: https://github.com/open-telemetry/opentelemetry-java/blob/main/VERSIONING.md#compatibility-requirements

thpierce: Makes sense, I'll make the changes today.

trask (approved):

willarmiros (approved):
Should be good to merge @trask !

PR #802, opened 2023-03-30
Add new components to allow for generating metrics from 100% of spans without impacting sampling

thpierce:
**Description:**

We wish to generate key metrics from OTEL auto-instrumented spans. I had [previously engaged](https://github.com/open-telemetry/opentelemetry-specification/issues/3145) the community to discuss this, and after some conversations, decided that the best approach would be to implement a Sampler and a Span Processor to the `contrib` repositories.

The major requirements for our project are as follows:
* Metrics should be produced from 100% of Client/Server/Producer/Consumer spans produced by OTEL Auto-Instrumentation, regardless of sampling decision
   * Span sampling decisions should still be respected w.r.t. exporting spans to the collector and propagating sampling decisions to child spans
* Metric generation must be done when the span is ended, as metric attributes are derived from some span attributes that are not always available when the span is started (e.g. `http.route`).
* Metric attributes need to be added to the span attributes before export to Collector, to be used to correlate metrics and spans in a downstream process.

To meet these goals, we have developed components which we wish to contribute to the OTEL community:
* `AlwaysRecordSampler`.
   * This is an aggregate Sampler that is initialized with a root Sampler (akin to [ParentBasedSampler](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/ParentBasedSampler.java)).
   * `shouldSample()` will call the root sampler and, if the `SampleDecision` is `RECORD_AND_SAMPLE` or `RECORD_ONLY`, it simply returns the result. However, if the `SampleDecision` is `DROP`, it will instead create a new result with `SampleDecision` set to `RECORD_ONLY` (all other result behaviour unchanged).
   * This Sampler will ensure that 100% of spans get send to the SpanProcessor, without changing the actual Sampling rate/behaviour.
* `MetricAttributeGenerator`
   * A generic interface for components that consume spans and resources and export attributes for metrics generated from these spans.
   * Implemented by `AwsMetricAttributeGenerator` to produce metric attributes specific to our usecase.
* `AwsSpanMetricsProcessor`.
   * This SpanProcessor will create desired metrics and embed attributes into processed spans before sending them to the downstream SpanProcessor (for additional processing/export).
   * The AwsSpanMetricsProcessor is initialized (via `AwsSpanMetricsProcessorBuilder`) with metering instruments, a MetricAttributeGenerator and a resource.
   * All APIs are uninteresting except `onEnd()`, which will:
      * Determine attributes for metrics (using the provided MetricAttributeGenerator)
      * Emit metrics with attributes (with the provided metering instruments)
* `AwsMetricAttributesSpanExporter`.
   * This SpanExporter will embed desired metric attributes into processed spans before sending them to the downstream SpanExporter for export.
   * The AwsMetricAttributesSpanExporter is initialized (via `AwsMetricAttributesSpanExporterBuilder`) with a MetricAttributeGenerator, a Resource, and a delegate SpanExporter.
   * All APIs are uninteresting except `export()`, which will:
      * Determine attributes to embed(using the provided MetricAttributeGenerator)
      * Replace the span with a new `SpanData` identical to the original span, but with metric attributes embedded into span attributes.
      * Pass the spans to the delegate SpanExporter's `export` method.

**Existing Issue(s):**

https://github.com/open-telemetry/opentelemetry-java-contrib/issues/789

**Testing:**

* Unit tests added and run successfully
* `./gradlew assemble` Passes
* `./gradlew spotlessCheck` Passes
* E2E testing done to verify emitted metrics align with expectations.

**Documentation:**

These are independent components that will require configuration in the TracerProvider. We will create a document to describe the usage once it is ready for distribution.

**Outstanding items:**

* We may change in the details of the metric attributes created by `AwsMetricAttributeGenerator` in future, but for now these are sufficient.


aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsMetricAttributeGenerator.java
```diff
@@ -0,0 +1,229 @@
...
+import io.opentelemetry.semconv.trace.attributes.SemanticAttributes;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * AwsMetricAttributeGenerator generates very specific metric attributes based on low-cardinality
+ * span and resource attributes. If such attributes are not present, we fallback to default values.
+ *
+ * <p>The goal of these particular metric attributes is to get metrics for incoming and outgoing
+ * traffic for an application. Namely, {@link SpanKind#SERVER} amd {@link SpanKind#CONSUMER} spans
```
srprash, line 39: nit: typo in "and".

thpierce: ty, resolved.

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsMetricAttributeGenerator.java
```diff
@@ -0,0 +1,222 @@
...
+  private static final String GRAPHQL = "graphql";
+
+  // Default attribute values if no valid span attribute value is identified
+  private static final String UNKNOWN_SERVICE = "UnknownService";
+  private static final String UNKNOWN_OPERATION = "UnknownOperation";
+  private static final String UNKNOWN_REMOTE_SERVICE = "UnknownRemoteService";
+  private static final String UNKNOWN_REMOTE_OPERATION = "UnknownRemoteOperation";
+
+  @Override
+  public Attributes generateMetricAttributesFromSpan(SpanData span, Resource resource) {
```
srprash, line 60: The `Resource` looks like a dependency of this class. Probably better to supply it via a parametrized constructor?

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsSpanMetricsProcessor.java
```diff
@@ -0,0 +1,124 @@
...
+ * latency is measured with a histogram. Metrics are emitted with attributes derived from span
+ * attributes.
+ *
+ * <p>For highest fidelity metrics, this processor should be coupled with the {@link
+ * AlwaysRecordSampler}, which will result in 100% of spans being sent to the processor.
+ */
+@Immutable
+public final class AwsSpanMetricsProcessor implements SpanProcessor {
+
+  private static final double NANOS_TO_MILLIS = 1_000_000.0;
```
srprash, line 40: nit: Shouldn't it be named `MILLIS_TO_NANOS` since 1 ms = 1000000 ns?

srprash (approved):
A couple of small comments. Otherwise LGTM.

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsMetricAttributeGenerator.java
thpierce: Maybe - I actually prefer feeding it into the exporter/processor so it's more clear where it is coming from. I don't think this is a blocking comment, so going to leave this as it is.

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsSpanMetricsProcessor.java
thpierce: It's used to convert nanos to millis, so prefer to leave it as it is. Is a multiplication/division factor, so either name is probably reasonable.

thpierce:
@trask - PR is ready for your review.

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsMetricAttributeGenerator.java
```diff
@@ -0,0 +1,222 @@
...
+    } else {
+      logUnknownAttribute(AWS_REMOTE_SERVICE, span);
+      builder.put(AWS_REMOTE_SERVICE, UNKNOWN_REMOTE_SERVICE);
+      logUnknownAttribute(AWS_REMOTE_OPERATION, span);
+      builder.put(AWS_REMOTE_OPERATION, UNKNOWN_REMOTE_OPERATION);
+    }
+
+    // Peer service takes priority as RemoteService over everything but AWS Remote.
+    if (isKeyPresent(span, PEER_SERVICE) && !isKeyPresent(span, AWS_REMOTE_SERVICE)) {
+      setRemoteService(span, builder, PEER_SERVICE);
```
mxiamxia, line 182: is it expected that we have added the log entry by `logUnknownAttribute(AWS_REMOTE_SERVICE, span);` but  the `RemoteService` is actually set properly here?

mxiamxia (approved):
LGTM, ty!

mxiamxia (approved):
LGTM, ty!

mxiamxia (approved):
LGTM, ty!

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsMetricAttributeGenerator.java
thpierce: Yea I see what you mean, it's a little bit odd. I think for now it's ok. We plan to revise a bit of this logic in future, and it's not totally unreasonable for the time being. We don't expect `PEER_SERVICE` to be frequently applied. Will make a note for the future.

mxiamxia:
Hi @trask , just kindly check if you can help to take a look this PR.

trask (approved):

PR #813, opened 2023-04-06
Links based sampler

atshaw43:
**Description:**
Feature addition

Problem
What if we only want to sample if and only if one of the span links is sampled.

Solution
Add a LinksBasedSampler sampler that samples if and only if a span link is sampled.

**Testing:**
Unit Tests

### DIFF
--- a/samplers/src/main/java/io/opentelemetry/contrib/sampler/LinksBasedSampler.java
+++ b/samplers/src/main/java/io/opentelemetry/contrib/sampler/LinksBasedSampler.java
@@ -17,8 +17,8 @@ import javax.annotation.Nullable;
 import javax.annotation.concurrent.Immutable;
 
 /**
- * A Sampler that uses the sampled flag of the span links, if present. If at least one span link is
- * sampled then, this span will be sampled. Otherwise, it is not sampled.
+ * A Sampler that uses the sampled flag of the span links if present. If at least one span link is
+ * sampled, then this span will be sampled. Otherwise, it is not sampled.
  * If the span has no span links, this Sampler will use the "root" sampler that it is built with.
  */
 @Immutable
@@ -32,7 +32,7 @@ final class LinksBasedSampler implements Sampler {
   LinksBasedSampler(Sampler root) {
     this.root = root;
   }
-  
+
   @Override
   public SamplingResult shouldSample(
       Context parentContext,
--- a/samplers/src/main/java/io/opentelemetry/contrib/sampler/LinksBasedSampler.java
+++ b/samplers/src/main/java/io/opentelemetry/contrib/sampler/LinksBasedSampler.java
@@ -28,7 +28,7 @@ final class LinksBasedSampler implements Sampler {
   private static final SamplingResult POSITIVE_SAMPLING_RESULT = SamplingResult.recordAndSample();
   private static final SamplingResult NEGATIVE_SAMPLING_RESULT = SamplingResult.drop();
 
-  LinksBasedSampler(Sampler root) {
+  public LinksBasedSampler(Sampler root) {
     this.root = root;
   }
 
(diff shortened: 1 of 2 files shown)


samplers/src/main/java/io/opentelemetry/contrib/sampler/LinksBasedSampler.java
```diff
@@ -0,0 +1,83 @@
...
+ */
+@Immutable
+final class LinksBasedSampler implements Sampler {
+
+  private final Sampler root;
+
+  private static final SamplingResult POSITIVE_SAMPLING_RESULT = SamplingResult.recordAndSample();
+  private static final SamplingResult NEGATIVE_SAMPLING_RESULT = SamplingResult.drop();
+
+  LinksBasedSampler(Sampler root) {
```
jack-berg, line 31: The class and constructor are package private so it doesn't appear this can actually be used.

samplers/src/main/java/io/opentelemetry/contrib/sampler/LinksBasedSampler.java
```diff
@@ -0,0 +1,83 @@
...
+      return true;
+    }
+    if (!(o instanceof LinksBasedSampler)) {
+      return false;
+    }
+
+    LinksBasedSampler that = (LinksBasedSampler) o;
+
+    return root.equals(that.root);
+  }
```
jack-berg, line 77: Let's omit the `equals` / `hashcode` implementations - I don't imagine this being added to maps or being compared to other samplers for equality very often.

samplers/src/main/java/io/opentelemetry/contrib/sampler/LinksBasedSampler.java
atshaw43: Good point! Updated.

samplers/src/main/java/io/opentelemetry/contrib/sampler/LinksBasedSampler.java
atshaw43: Removed.

atshaw43: @jack-berg 
Bump

atshaw43: @jack-berg 
Bump

samplers/src/main/java/io/opentelemetry/contrib/sampler/LinksBasedSampler.java
```diff
@@ -0,0 +1,63 @@
...
+import java.util.List;
+import javax.annotation.concurrent.Immutable;
+
+/**
+ * A Sampler that uses the sampled flag of the span links if present. If at least one span link is
+ * sampled, then this span will be sampled. Otherwise, it is not sampled. If the span has no span
+ * links, this Sampler will use the "root" sampler that it is built with.
+ */
+@Immutable
+final class LinksBasedSampler implements Sampler {
```
jack-berg, line 23: This class needs to be public. 

Also, do we need the `@Immutable` annotation?

jack-berg (commented):
Sorry for the delay. Couple minor comments but looks good.

jack-berg (approved):

samplers/src/main/java/io/opentelemetry/contrib/sampler/LinksBasedSampler.java
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * A Sampler that uses the sampled flag of the span links if present. If at least one span link is
+ * sampled, then this span will be sampled. Otherwise, it is not sampled. If the span has no span
+ * links, this Sampler will use the "root" sampler that it is built with.
+ */
+public final class LinksBasedSampler implements Sampler {
+  private final Sampler root;
+
+  private static final SamplingResult POSITIVE_SAMPLING_RESULT = SamplingResult.recordAndSample();
+  private static final SamplingResult NEGATIVE_SAMPLING_RESULT = SamplingResult.drop();
```
trask, line 25: I don't think constants help improve readability or performance here, so I'd suggest inlining

trask (approved):

samplers/src/main/java/io/opentelemetry/contrib/sampler/LinksBasedSampler.java
trask: @atshaw43 just checking in to make sure you saw this comment, thx

PR #819, opened 2023-04-11
Update to open-telemetry/assign-reviewers-action

trask:


.github/workflows/assign-reviewers.yml
```diff
@@ -12,4 +12,6 @@ jobs:
   assign-reviewers:
     runs-on: ubuntu-latest
     steps:
-      - uses: dyladan/component-owners@main
+      - uses: open-telemetry/assign-reviewers-action@main
+        with:
+          config-file: .github/component_owners.yml
```
trask, line 17: @dyladan I like the language of "component owners" so kept that part intact, e.g. https://github.com/open-telemetry/opentelemetry-java-contrib/tree/main/maven-extension#component-owners

dyladan: I also like owners better. I changed the name in order to better describe the actual behavior before, but I think we're going to extend it to do additional things aside from reviewer assignment. Maybe we should consider changing it back before too many people switch to the new repo?

dyladan: The file name and action name both use "assign-reviewers". The language should at least be consistent within the repo?

dyladan (approved):
LGTM. Renaming the workflow is a non-blocking nit

mateuszrzeszutek (approved):

Issue #834, opened 2023-04-19
jmx-metrics unable to connect to server with SSL enabled on the RMI registry


dmitchnewrelic:
**Description**
The JMX Metric Gatherer is not working properly with servers that have SSL protection enabled on the RMI registry.  [Oracle's Java documentation](https://docs.oracle.com/javadb/10.10.1.2/adminguide/radminjmxenablepwdssl.html) describes this setting as follows: 

_The system property `com.sun.management.jmxremote.registry.ssl=true` was new in JDK 6 and aims at resolving security issues with the RMI registry used in relation with JMX. This property must be used in conjunction with `com.sun.management.jmxremote.ssl.need.client.auth=true` in order to fully secure the RMI registry._

**Steps to reproduce**
1. Configure a java server with the following settings: 

```
 -Dcom.sun.management.jmxremote.local.only=false 
 -Dcom.sun.management.jmxremote.password.file=jmxremote.password 
 -Dcom.sun.management.jmxremote.access.file=jmxremote.access 
 -Dcom.sun.management.jmxremote 
 -Dcom.sun.management.jmxremote.authenticate=true 
 -Dcom.sun.management.jmxremote.ssl.need.client.auth=true 
 -Dcom.sun.management.jmxremote.registry.ssl=true 
 -Dcom.sun.management.jmxremote.ssl=true 
 -Dcom.sun.management.jmxremote.port=1098 
 -Dcom.sun.management.jmxremote.rmi.port=1099 
 -Djava.rmi.server.hostname=localhost 
 -Djavax.net.ssl.keyStore=serverkeystore 
 -Djavax.net.ssl.keyStorePassword=<password> 
 -Djavax.net.ssl.trustStore=servertruststore 
 -Djavax.net.ssl.trustStorePassword=<password>
```

2. Configure the JMX Metric Gatherer to connect to the server using the following session.properties file: 

```
otel.jmx.service.url=service:jmx:rmi:///jndi/rmi://localhost:1098/jmxrmi
otel.jmx.target.system = kafka
otel.jmx.interval.milliseconds = 60000
otel.metrics.exporter = otlp
otel.exporter.otlp.timeout = 30000
otel.exporter.otlp.endpoint = http://localhost:4317
otel.jmx.username=<username>
otel.jmx.password=<password>
javax.net.ssl.keyStoreType=JKS
javax.net.ssl.keyStore=clientkeystore
javax.net.ssl.keyStorePassword=<password>
[...]

PR #835, opened 2023-04-20
Adding support for scenarios where the RMI registry has SSL enabled

dmitchnewrelic:
**Description:**

The JMX Metric Gatherer is not working properly with servers that have SSL protection enabled on the RMI registry.  [Oracle's Java documentation](https://docs.oracle.com/javadb/10.10.1.2/adminguide/radminjmxenablepwdssl.html) describes this setting as follows: 

_The system property `com.sun.management.jmxremote.registry.ssl=true` was new in JDK 6 and aims at resolving security issues with the RMI registry used in relation with JMX. This property must be used in conjunction with `com.sun.management.jmxremote.ssl.need.client.auth=true` in order to fully secure the RMI registry._

Changes were made to the JMX Metrics Gatherer to support this scenario, by adding a new helper class called JmxConnectorHelper. This class implements different connection logic when SSL is enabled on the RMI registry. 

Changes were also made to JmxConfig, as a new parameter is required to handle this scenario.  Specifically, the following parameter was added: 

```
otel.jmx.remote.registry.ssl=true
```
The readme file was updated to explain when to use this new parameter. 

**Existing Issue(s):**

Issue #834 

**Testing:**

The following scenarios were tested: 

- SSL not enabled, and no credentials required
- SSL not enabled, but with credentials
- SSL enabled for JMX, but not on the RMI registry 
- SSL enabled for JMX including the RMI registry  

**Documentation:**

README.md was updated to reflect the new configuration parameters. 

**Outstanding items:**

Nothing at this time.

### DIFF
--- a/jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/JmxConnectorHelper.java
+++ b/jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/JmxConnectorHelper.java
@@ -66,10 +66,10 @@ public class JmxConnectorHelper {
     return jmxConn;
   }
 
-  private static void getStub(String hostName, int port) throws IOException {
+  private static RMIServer getStub(String hostName, int port) throws IOException {
     try {
       Registry registry = LocateRegistry.getRegistry(hostName, port, sslRMIClientSocketFactory);
-      stub = (RMIServer) registry.lookup("jmxrmi");
+      return (RMIServer) registry.lookup("jmxrmi");
     } catch (NotBoundException nbe) {
       throw new IOException(nbe);
     }
--- a/jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/JmxConnectorHelper.java
+++ b/jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/JmxConnectorHelper.java
@@ -59,7 +59,7 @@ public class JmxConnectorHelper {
     }
 
     if (stub == null) {
-      getStub(hostName, port);
+      stub = getStub(hostName, port);
     }
     JMXConnector jmxConn = new RMIConnector(stub, null);
     jmxConn.connect(env);
--- a/jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/JmxConnectorHelper.java
+++ b/jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/JmxConnectorHelper.java
@@ -1,0 +1,77 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.contrib.jmxmetrics;
+
+import java.io.IOException;
+import java.net.URI;
+import java.rmi.NotBoundException;
+import java.rmi.registry.LocateRegistry;
+import java.rmi.registry.Registry;
+import java.util.Map;
+import java.util.logging.Level;
+import java.util.logging.Logger;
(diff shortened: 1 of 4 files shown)


jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/JmxConnectorHelper.java
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  private static void getStub(String hostName, int port) throws IOException {
+    try {
+      Registry registry = LocateRegistry.getRegistry(hostName, port, sslRMIClientSocketFactory);
+      stub = (RMIServer) registry.lookup("jmxrmi");
+    } catch (NotBoundException nbe) {
+      throw new IOException(nbe);
+    }
+  }
```
breedx-splk, line 76: nit: it's a little clunky to have this method just do side effect. Prefer to return the thing and assign in the caller.

```suggestion
  private static RMIServer getStub(String hostName, int port) throws IOException {
    try {
      Registry registry = LocateRegistry.getRegistry(hostName, port, sslRMIClientSocketFactory);
      return (RMIServer) registry.lookup("jmxrmi");
    } catch (NotBoundException nbe) {
      throw new IOException(nbe);
    }
  }
```

breedx-splk (approved):
This seems good to me. Had one rather small suggestion. 

I'd really like to see an integration test that covers this. Would you mind opening an issue to do that work in a follow-up PR?

jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/JmxConnectorHelper.java
jeffalder: Yo @breedx-splk! Good idea, me gusta

jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/JmxConnectorHelper.java
```diff
@@ -0,0 +1,77 @@
...
+      final URI uri = URI.create(components[2]);
+      hostName = uri.getHost();
+      port = uri.getPort();
+    } else {
+      hostName = serviceURL.getHost();
+      port = serviceURL.getPort();
+    }
+
+    if (stub == null) {
+      getStub(hostName, port);
```
jeffalder, line 62: @dmitchnewrelic You'll need to assign the variable here, now.

dmitchnewrelic:
Thanks @breedx-splk.  I've made the suggested change and will create a separate issue for the integration test.

dmitchnewrelic:
I've created issue #838 for this.

jeffalder:
@breedx-splk per https://github.com/open-telemetry/opentelemetry-java-contrib/blob/main/RELEASING.md#release-cadence does this mean the next normal release would be May 12th?

trask:
just a heads up, you can ignore the EasyCLA failure, it is having an issue currently (https://github.com/open-telemetry/community/issues/1445)

> @breedx-splk per https://github.com/open-telemetry/opentelemetry-java-contrib/blob/main/RELEASING.md#release-cadence does this mean the next normal release would be May 12th?

👍

mateuszrzeszutek (approved):

breedx-splk:
> @breedx-splk per https://github.com/open-telemetry/opentelemetry-java-contrib/blob/main/RELEASING.md#release-cadence does this mean the next normal release would be May 12th?

I don't think we'd want to promise a date, but yeah, basically monthly and more strongly coordinated with the core and instrumentation releases. I defer to the maintainers on anything more specific.

PR #862, opened 2023-05-08
Fix native mode error cause by static init of random

brunobat:
**Description:**

Native build fails because AwsXrayRemoteSampler.java statically inits a Random()
This is not allowed, as explained here:
https://foivos.zakkak.net/tutorials/working-with-randoms-native-images/

See discussion here: https://github.com/quarkusio/quarkus/discussions/33154


brunobat:
> is it possible to add a native test (e.g. similar to [open-telemetry/opentelemetry-java-instrumentation#8163](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/8163)) to prevent regressions on this?

This would require a build step that builds a native app with graalvm and then run tests against it. With Quarkus this is easy, but for a standalone lib like this, it would require research.

trask:
@srprash @wangzlei can you review? thanks!

wangzlei (approved):

trask (approved):

trask:
closing and re-opening to try to jump start CI

PR #885, opened 2023-05-22
Test against Java 8, 11, 17, and 20

### LINKED ISSUE #586: Add CI test matrix to run tests against Java 8 and 11 (in addition to 17)
Currently tests are only run against Java 17.

It would be good to add a test matrix to run them against 8 and 11 as well (where the module under test supports those versions), similar to what is done in core and instrumentation repos.

trask:
Resolves #586

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -85,6 +85,13 @@ jobs:
           arguments: integrationTest
           cache-read-only: ${{ github.event_name == 'pull_request' }}
 
+      - name: Save integration test results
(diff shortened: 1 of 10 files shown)


.github/workflows/build.yml
```diff
@@ -54,13 +85,6 @@ jobs:
           arguments: integrationTest
           cache-read-only: ${{ github.event_name == 'pull_request' }}
 
-      - name: Save integration test results
```
laurit, line 57: aren't the test results useful?

laurit (approved):

PR #888, opened 2023-05-23
Stop publishing static instrumenter maven plugin

trask:
Since we don't publish any other parts of the static-instrumenter project


mateuszrzeszutek (approved):
😢

PR #906, opened 2023-06-06
Enhance AWS APM metrics mapping implementation

mxiamxia:
**Description:**
Enhance AWS APM metrics mapping implementation to resolve a few of the following "Unknown" metric attributes use cases.

- if span name used in ingress operation is not valid with values of 'null', 'UnknownOperation' and 'HttpMethod value` for HTTP request use cases. We will try to extract the ingress operation from `http.method` and `http.target` attributes.
- if the remote service attribute can not be directly derived from the attributes such `rpc.name`, `db.system`. We will try to fill it with the `net.peer.name` and `net.sock.peer.addr` attribute values and have Collector to do the further resolution on the populated values. 
- if the remote operation attribute is `Unknown`, We will try to extract the remote operation from  `http.method` and `http.url` attributes.
- Add a new `aws.remote.target` attribute for supporting the following 4 AWS resources when invoking it with Client Spans  
   - AWS S3 Bucket Name
   - AWS Queue Name
   - AWS Stream Name
   - AWS DynamoDB table name

**Testing:**
unit test and adhoc

**Documentation:**
N/A

**Outstanding items:**
N/A

### DIFF
--- a/aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsAttributeKeys.java
+++ b/aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsAttributeKeys.java
@@ -24,4 +24,12 @@ final class AwsAttributeKeys {
 
   static final AttributeKey<String> AWS_REMOTE_OPERATION =
       AttributeKey.stringKey("aws.remote.operation");
+
+  static final AttributeKey<String> AWS_REMOTE_TARGET = AttributeKey.stringKey("aws.remote.target");
+
+  // use the same AWS Resource attribute name defined by OTel java auto-instr for aws_sdk_v_1_1
+  static final AttributeKey<String> AWS_BUCKET_NAME = AttributeKey.stringKey("aws.bucket.name");
+  static final AttributeKey<String> AWS_QUEUE_NAME = AttributeKey.stringKey("aws.queue.name");
+  static final AttributeKey<String> AWS_STREAM_NAME = AttributeKey.stringKey("aws.stream.name");
+  static final AttributeKey<String> AWS_TABLE_NAME = AttributeKey.stringKey("aws.table.name");
 }
--- a/aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsAttributeKeys.java
+++ b/aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsAttributeKeys.java
@@ -28,6 +28,10 @@ final class AwsAttributeKeys {
   static final AttributeKey<String> AWS_REMOTE_TARGET = AttributeKey.stringKey("aws.remote.target");
 
   // use the same AWS Resource attribute name defined by OTel java auto-instr for aws_sdk_v_1_1
+  // TODO: all AWS specific attributes should be defined in semconv package and reused cross all
+  // otel packages. Related sim -
+  // https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/8710
+
   static final AttributeKey<String> AWS_BUCKET_NAME = AttributeKey.stringKey("aws.bucket.name");
   static final AttributeKey<String> AWS_QUEUE_NAME = AttributeKey.stringKey("aws.queue.name");
   static final AttributeKey<String> AWS_STREAM_NAME = AttributeKey.stringKey("aws.stream.name");
--- a/aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsMetricAttributeGenerator.java
+++ b/aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsMetricAttributeGenerator.java
@@ -5,19 +5,31 @@
 
 package io.opentelemetry.contrib.awsxray;
 
+import static io.opentelemetry.contrib.awsxray.AwsAttributeKeys.AWS_BUCKET_NAME;
 import static io.opentelemetry.contrib.awsxray.AwsAttributeKeys.AWS_LOCAL_OPERATION;
 import static io.opentelemetry.contrib.awsxray.AwsAttributeKeys.AWS_LOCAL_SERVICE;
+import static io.opentelemetry.contrib.awsxray.AwsAttributeKeys.AWS_QUEUE_NAME;
 import static io.opentelemetry.contrib.awsxray.AwsAttributeKeys.AWS_REMOTE_OPERATION;
 import static io.opentelemetry.contrib.awsxray.AwsAttributeKeys.AWS_REMOTE_SERVICE;
+import static io.opentelemetry.contrib.awsxray.AwsAttributeKeys.AWS_REMOTE_TARGET;
 import static io.opentelemetry.contrib.awsxray.AwsAttributeKeys.AWS_SPAN_KIND;
+import static io.opentelemetry.contrib.awsxray.AwsAttributeKeys.AWS_STREAM_NAME;
+import static io.opentelemetry.contrib.awsxray.AwsAttributeKeys.AWS_TABLE_NAME;
 import static io.opentelemetry.semconv.resource.attributes.ResourceAttributes.SERVICE_NAME;
 import static io.opentelemetry.semconv.trace.attributes.SemanticAttributes.DB_OPERATION;
 import static io.opentelemetry.semconv.trace.attributes.SemanticAttributes.DB_SYSTEM;
 import static io.opentelemetry.semconv.trace.attributes.SemanticAttributes.FAAS_INVOKED_NAME;
-import static io.opentelemetry.semconv.trace.attributes.SemanticAttributes.FAAS_INVOKED_PROVIDER;
+import static io.opentelemetry.semconv.trace.attributes.SemanticAttributes.FAAS_TRIGGER;
 import static io.opentelemetry.semconv.trace.attributes.SemanticAttributes.GRAPHQL_OPERATION_TYPE;
+import static io.opentelemetry.semconv.trace.attributes.SemanticAttributes.HTTP_METHOD;
+import static io.opentelemetry.semconv.trace.attributes.SemanticAttributes.HTTP_TARGET;
+import static io.opentelemetry.semconv.trace.attributes.SemanticAttributes.HTTP_URL;
 import static io.opentelemetry.semconv.trace.attributes.SemanticAttributes.MESSAGING_OPERATION;
 import static io.opentelemetry.semconv.trace.attributes.SemanticAttributes.MESSAGING_SYSTEM;
+import static io.opentelemetry.semconv.trace.attributes.SemanticAttributes.NET_PEER_NAME;
+import static io.opentelemetry.semconv.trace.attributes.SemanticAttributes.NET_PEER_PORT;
+import static io.opentelemetry.semconv.trace.attributes.SemanticAttributes.NET_SOCK_PEER_ADDR;
+import static io.opentelemetry.semconv.trace.attributes.SemanticAttributes.NET_SOCK_PEER_PORT;
 import static io.opentelemetry.semconv.trace.attributes.SemanticAttributes.PEER_SERVICE;
 import static io.opentelemetry.semconv.trace.attributes.SemanticAttributes.RPC_METHOD;
 import static io.opentelemetry.semconv.trace.attributes.SemanticAttributes.RPC_SERVICE;
@@ -30,6 +42,9 @@ import io.opentelemetry.sdk.resources.Resource;
 import io.opentelemetry.sdk.trace.data.SpanData;
 import io.opentelemetry.semconv.resource.attributes.ResourceAttributes;
 import io.opentelemetry.semconv.trace.attributes.SemanticAttributes;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.util.Optional;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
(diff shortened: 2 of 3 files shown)


aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsMetricAttributeGenerator.java
```diff
@@ -152,35 +212,130 @@ private static void setEgressOperation(SpanData span, AttributesBuilder builder)
...
-      setRemoteOperation(span, builder, GRAPHQL_OPERATION_TYPE);
+      remoteService = GRAPHQL;
+      remoteOperation = getRemoteOperation(span, GRAPHQL_OPERATION_TYPE);
     } else {
-      logUnknownAttribute(AWS_REMOTE_SERVICE, span);
-      builder.put(AWS_REMOTE_SERVICE, UNKNOWN_REMOTE_SERVICE);
-      logUnknownAttribute(AWS_REMOTE_OPERATION, span);
-      builder.put(AWS_REMOTE_OPERATION, UNKNOWN_REMOTE_OPERATION);
+      // try to derive RemoteService and RemoteOperation from the other un-directive attributes
+      remoteService = generateRemoteService(span);
```
bjrara, line 237: It's hard to tell the difference between `generateRemoteService/Operation` and `getRemoteService/Operation`, only that `generate*` does some specific to HTTP/NET spans which makes me feel the method name is ambiguous.

I'm thinking if it will be better to assign `UNKNOWN*` as the initial values to `remoteService` and `remoteOperation`, and only do HTTP/NET handling using function like `getRemote*FromNETSpan` in the `else` block.

bjrara: Thanks! It's much clear now.

bjrara (approved):
LGTM

mxiamxia:
Hi @trask , could you please take a look this PR as code approver when you get the chance? Thanks!

bjrara (approved):

wangzlei:
LGTM!

wangzlei (approved):

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsMetricAttributeGenerator.java
```diff
@@ -152,35 +212,132 @@ private static void setEgressOperation(SpanData span, AttributesBuilder builder)
...
+      remoteService = getRemoteService(span, DB_SYSTEM);
+      remoteOperation = getRemoteOperation(span, DB_OPERATION);
+    } else if (isKeyPresent(span, FAAS_INVOKED_NAME) || isKeyPresent(span, FAAS_TRIGGER)) {
+      remoteService = getRemoteService(span, FAAS_INVOKED_NAME);
+      remoteOperation = getRemoteOperation(span, FAAS_TRIGGER);
     } else if (isKeyPresent(span, MESSAGING_SYSTEM) || isKeyPresent(span, MESSAGING_OPERATION)) {
-      setRemoteService(span, builder, MESSAGING_SYSTEM);
-      setRemoteOperation(span, builder, MESSAGING_OPERATION);
+      remoteService = getRemoteService(span, MESSAGING_SYSTEM);
+      remoteOperation = getRemoteOperation(span, MESSAGING_OPERATION);
```
thpierce, line 231: I thought this would be: `{messaging.destination.name} {messaging.operation}`

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsMetricAttributeGenerator.java
```diff
@@ -152,35 +212,132 @@ private static void setEgressOperation(SpanData span, AttributesBuilder builder)
...
+      remoteOperation = getRemoteOperation(span, GRAPHQL_OPERATION_TYPE);
     }
 
     // Peer service takes priority as RemoteService over everything but AWS Remote.
     if (isKeyPresent(span, PEER_SERVICE) && !isKeyPresent(span, AWS_REMOTE_SERVICE)) {
-      setRemoteService(span, builder, PEER_SERVICE);
+      remoteService = getRemoteService(span, PEER_SERVICE);
     }
+
+    // try to derive RemoteService and RemoteOperation from the other un-directive attributes
```
thpierce, line 242: What is un-directive?

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsMetricAttributeGenerator.java
```diff
@@ -94,14 +134,34 @@ private static void setService(Resource resource, SpanData span, AttributesBuild
    * name.
    */
```
thpierce, line 135: Please update comment, behaviour has substantially changed. Please review other comments in this code and update accordingly.

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsMetricAttributeGenerator.java
```diff
@@ -152,35 +212,132 @@ private static void setEgressOperation(SpanData span, AttributesBuilder builder)
...
+    String operation = UNKNOWN_OPERATION;
+    if (isKeyPresent(span, HTTP_TARGET)) {
+      String httpTarget = span.getAttributes().get(HTTP_TARGET);
+      // get the first part from API path string as operation value
+      // the more levels/parts we get from API path the higher chance for getting high cardinality
+      // data
+      if (httpTarget != null) {
+        operation = extractAPIPathValue(httpTarget);
+      }
+      if (isKeyPresent(span, HTTP_METHOD)) {
```
thpierce, line 268: if `HTTP_METHOD` is not found, are we sure we want to keep just the first part of the target?

Also this code is confusing because in `isValidOperation`, we are saying that HTTP_METHOD alone is invalid, but here we consider it valid if target is null. A cleaner solution would be to update the if block on line 158 to attempt to append target if only HTTP_METHOD is being used as span name.

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsMetricAttributeGenerator.java
```diff
@@ -152,35 +212,132 @@ private static void setEgressOperation(SpanData span, AttributesBuilder builder)
...
+        logger.log(Level.FINEST, "invalid http.url attribute: ", httpUrl);
+      }
+    }
+    if (isKeyPresent(span, HTTP_METHOD)) {
+      String httpMethod = span.getAttributes().get(HTTP_METHOD);
+      remoteOperation = httpMethod + " " + remoteOperation;
+    }
+    if (remoteOperation.equals(UNKNOWN_REMOTE_OPERATION)) {
+      logUnknownAttribute(AWS_REMOTE_OPERATION, span);
+    }
```
thpierce, line 302: Not blocking: This doesn't seem appropriate for this method - I would probably keep it in setRemoteServiceAndOperation. Ditto to generateRemoteService

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsMetricAttributeGenerator.java
```diff
@@ -152,35 +212,132 @@ private static void setEgressOperation(SpanData span, AttributesBuilder builder)
...
+  }
+
+  /**
+   * Extract the first part from API http target if it exists
+   *
+   * @param httpTarget http request target string value. Eg, /payment/1234
+   * @return the first part from the http target. Eg, /payment
+   */
+  private static String extractAPIPathValue(String httpTarget) {
+    if (httpTarget.isEmpty()) {
```
thpierce, line 313: Should we do null as well?

aws-xray/src/test/java/io/opentelemetry/contrib/awsxray/AwsMetricAttributeGeneratorTest.java
```diff
@@ -5,19 +5,32 @@
 
 package io.opentelemetry.contrib.awsxray;
```
thpierce, line 6: Did not have time to review tests, will rely on reviews from others to cover these.

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsMetricAttributeGenerator.java
```diff
@@ -152,35 +212,132 @@ private static void setEgressOperation(SpanData span, AttributesBuilder builder)
    * and RPC attributes to use here, but this is a sufficient starting point.
    */
   private static void setRemoteServiceAndOperation(SpanData span, AttributesBuilder builder) {
```
thpierce, line 214: I thought we said that remote service/remote operation would be done independently (e.g. could have rpc service and db operation)

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsAttributeKeys.java
```diff
@@ -24,4 +24,12 @@ private AwsAttributeKeys() {}
 
   static final AttributeKey<String> AWS_REMOTE_OPERATION =
       AttributeKey.stringKey("aws.remote.operation");
+
+  static final AttributeKey<String> AWS_REMOTE_TARGET = AttributeKey.stringKey("aws.remote.target");
+
+  // use the same AWS Resource attribute name defined by OTel java auto-instr for aws_sdk_v_1_1
```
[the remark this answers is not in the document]
carolabadeer: +1. I was going to recommend adding a TODO note here to update these attributes to match the semantic attributes, but I prefer @thpierce suggestion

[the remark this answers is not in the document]
mxiamxia: The listed AWS resource attributes are defined in [opentelemetry-java-instrumentation#AwsExperimentalAttributes ](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/0070148b3424db5b11be257b1e086cf8ba59ed0b/instrumentation/aws-sdk/aws-sdk-1.11/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v1_11/AwsExperimentalAttributes.java#L15) which is not visible to SDK. I think the long term solution is to move these resource name definitions into [semconv#SemanticAttributes](https://github.com/open-telemetry/opentelemetry-java/blob/main/semconv/src/main/java/io/opentelemetry/semconv/trace/attributes/SemanticAttributes.java). 

I think we have to define it again here before we can add it into semconv package.

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsMetricAttributeGenerator.java
mxiamxia: thanks for calling it out. I am thinking moving `{messaging.destination.name}` to `aws.remote.target` attribute later and keep `remote.operation` with `messaging.operation`, what do you think?

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsMetricAttributeGenerator.java
mxiamxia: removed it.

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsMetricAttributeGenerator.java
mxiamxia: changed it to: only add `http.method` as prefix for RemoteOperation when `http.target` is not null.

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsMetricAttributeGenerator.java
mxiamxia: it could happen in the following blocks which is not restricted by pair of values.

```
// try to derive RemoteService and RemoteOperation from the other related attributes
    if (remoteService.equals(UNKNOWN_REMOTE_SERVICE)) {
      remoteService = generateRemoteService(span);
    }
    if (remoteOperation.equals(UNKNOWN_REMOTE_OPERATION)) {
      remoteOperation = generateRemoteOperation(span);
    }
```

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsAttributeKeys.java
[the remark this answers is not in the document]
mxiamxia: Create an issue for better way to manage AWS specific attributes https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/8710

carolabadeer (approved):

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsMetricAttributeGenerator.java
thpierce: Discussed offline - not exactly what I was talking about, but we decided this is desirable behaviour.

thpierce (approved):

mxiamxia:
Hi @trask, kindly follow up if you can help to review this PR and it has been reviewed by AWS reviewers. Thanks!

PR #913, opened 2023-06-09
Signal disk buffering

LikeTheSalad:
**Description:**

Feature addition - Allows to cache signals in the disk and later send them on-demand.
This feature is pretty similar to the one already existing in the [OpenTelemetry Swift SDK](https://github.com/open-telemetry/opentelemetry-swift/tree/main/Sources/Exporters/Persistence).

**Requirements**
- Java 8 compatible (avoiding to use non-Android-friendly tools).
- Configurable disk space limit.
- Configurable disk cache dir.
- Allows executing write and read operations in parallel.
- Ensures FIFO order when reading from disk.
- Prioritizes data writing, deleting the oldest data available in case the max disk space limit has been reached.
- Lightweight.

**Existing Issue(s):**

**Testing:**

* Unit testing.
* Manual testing.

**Documentation:**

Javadoc and README.

**How it works**

You can take a look at the [CONTRIBUTING](https://github.com/open-telemetry/opentelemetry-java-contrib/blob/e89728569bae06d5f04e4a3a74229b8da71bb7cd/disk-buffering/CONTRIBUTING.md) file to get a more detailed overview of how it all works.

**Outstanding items:**

* A serialization/deserialization mechanism that relies on the [Java proto lib](https://github.com/open-telemetry/opentelemetry-proto-java) is being used despite its large size. A future PR might be needed to replace the serialization tool with a smaller alternative in order to better comply with the lightweight requirement.


bherbst:
Just chiming in to say that this functionality is highly desirable for mobile clients, where we don't have gaurantees of a persistent network connection and the OS can kill the process at any time. The ability to buffer signals to disk for later upload helps prevent lost data in those cases.

Splunk's Android instrumentation [supports buffering to disk](https://github.com/signalfx/splunk-otel-android/blob/main/splunk-otel-android/src/main/java/com/splunk/rum/ZipkinToDiskSender.java) but only in Zipkin format, and that instrumentation is [on track to be donated to OTel](https://github.com/open-telemetry/community/issues/1400), so this work pairs nicely with better supporting that functionality in the Android instrumentation.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/storage/StorageConfiguration.java
```diff
@@ -0,0 +1,78 @@
...
+  /** A creator of temporary files needed to do the disk reading process. */
+  public abstract TemporaryFileProvider getTemporaryFileProvider();
+
+  public static StorageConfiguration getDefault() {
+    return builder().build();
+  }
+
+  public static Builder builder() {
+    return new AutoValue_StorageConfiguration.Builder()
+        .setMaxFileSize(1024 * 1024) // 1MB
```
zeitlinger, line 54: @jack-berg I remember we discussed making `ConfigProperties` available to other components as well.

Here it would be nice if we could configure all settings using those config properties. 
Is that possible / can we do something to enable it?

zeitlinger (commented):
I'll do another round once I've understood the high level approach better.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/exporters/AbstractDiskExporter.java
```diff
@@ -0,0 +1,91 @@
...
+import io.opentelemetry.sdk.common.CompletableResultCode;
+import java.io.File;
+import java.io.IOException;
+import java.util.Collection;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+public abstract class AbstractDiskExporter<EXPORT_DATA> {
```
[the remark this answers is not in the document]
LikeTheSalad: Sounds good, I'll take a look at that approach, I really don't have strong opinions on this so if it works then I'm fine with this change.

disk-buffering/README.md
```diff
@@ -0,0 +1,88 @@
...
+```
+
+### Reading data
+
+Each of the exporter wrappers can read from the disk and send the retrieved data over to their
+wrapped exporter by calling this method from them:
+
+```java
+try {
+    if(diskExporter.exportStoredBatch(1, TimeUnit.SECONDS)) {
```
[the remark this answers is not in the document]
LikeTheSalad: Not sure if I fully understood the question, though essentially this method is meant to be called on the consumer side whenever they want to start exporting the stored data. At the moment I didn't create any mechanism to automatically do this work, mostly because I'm thinking that, in Android, there are tools considered to be best practices for this kind of repetitive, background work such as [WorkManager](https://developer.android.com/topic/libraries/architecture/workmanager), that are preferable over going with a more generic solution of maybe spinning up a Thread with a loop inside that calls this method. So I left this uncalled on purpose so that each platform can use the mechanism that suits it best, and also to avoid making this PR bigger 😅 though I'm open to discussing an automatic read/export mechanism by default that could be optional for implementations that aren't willing or don't need to use specific optimized solutions.

> Where is the missing piece that detects that the regular (e.g. grpc) exporter fails - and only sends this part later?

Regarding this, the way it currently works is that, whenever the consumer calls the [exportStoredBatch](https://github.com/LikeTheSalad/opentelemetry-java-contrib/blob/disk-buffer/disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/exporters/AbstractDiskExporter.java#L39) method, 2 tasks are attempted inside it. The first one is to try to read a batch of signals (a JSON line) and then if it finds one, it proceeds to delegate said batch to the wrapped exporter and wait for it to finish. This is what could happen:

- If there's no data in disk, then `exportStoredBatch` will return false.
- If there's data but the wrapped exporter (for example, the grpc one) doesn't succeed when trying to export the batch, then the data is not removed from the disk and `exportStoredBatch` would return false.
- If there's data and also the wrapped exporter succeeds in exporting the batch, then the exported data is removed from the disk, and `exportStoredBatch` would return true.

So based on that information, the retry strategy would have to be decided by the consumer of this library, since they would be the ones calling this method on-demand. In the meantime, before they call this method, all it would happen is that the intercepted batches would keep on getting stored in disk, and if the configured maximum space (or [maxFolderSize](src/main/java/io/opentelemetry/contrib/disk/buffering/storage/StorageConfiguration.java#L43)) is reached, then the oldest file will be removed to make space for new data to be stored.

disk-buffering/README.md
```diff
@@ -0,0 +1,88 @@
+# Disk buffering
```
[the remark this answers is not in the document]
LikeTheSalad: It should be done now, please have a look in case the information I added about you is not accurate. Cheers.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/exporters/AbstractDiskExporter.java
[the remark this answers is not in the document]
LikeTheSalad: Btw I'm planning to do the serialization clean-up first to get it out of the way and avoid possible issues with doing so later after other code changes.

disk-buffering/README.md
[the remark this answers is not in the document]
zeitlinger: I just revisited the readme and realized that I misunderstood the high-level approach.

I think I get it now - maybe you want to add paragraph - provided it's correct :smile: 

- the disk exporter is meant to write data to disk **instead** of sending it directly (i.e. it's not meant to store only failed batches to disk)
- the disk exporter doesn't include an export loop (at the moment), because that is framework specific (e.g. [android background work](https://developer.android.com/guide/background))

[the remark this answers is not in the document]
LikeTheSalad: I agree, I'm glad someone else takes a look at the docs because is usually difficult for me to describe these things properly when I wrote the code 😅 - I'll add that information to it, cheers!

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/exporters/MetricDiskExporter.java
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  @Override
+  public CompletableResultCode export(Collection<MetricData> metrics) {
+    return onExport(metrics);
+  }
+
+  @Override
+  public CompletableResultCode flush() {
+    return wrapped.flush();
```
zeitlinger, line 66: why not just return success?
It's weird that flush should be propagated, but write gets buffered.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/exporters/AbstractDiskExporter.java
```diff
@@ -0,0 +1,91 @@
...
+  public boolean exportStoredBatch(long timeout, TimeUnit unit) throws IOException {
+    logger.log(Level.INFO, "Attempting to export batch from disk.");
+    AtomicBoolean exportSucceeded = new AtomicBoolean(false);
+    boolean foundDataToExport =
+        storage.read(
+            bytes -> {
+              logger.log(Level.INFO, "About to export stored batch.");
+              CompletableResultCode join =
+                  doExport(getSerializer().deserialize(bytes)).join(timeout, unit);
+              exportSucceeded.set(join.isSuccess());
```
zeitlinger, line 48: what's the reason for the atomic boolean?

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/exporters/AbstractDiskExporter.java
```diff
@@ -0,0 +1,91 @@
...
+      logger.log(Level.INFO, "Could not store batch in disk. Exporting it right away.");
+      return doExport(data);
+    }
+  }
+
+  private File getSignalFolder(File rootDir) {
+    File folder = new File(rootDir, getStorageFolderName());
+    if (!folder.exists()) {
+      if (!folder.mkdirs()) {
+        throw new IllegalStateException("Could not create the signal folder");
```
zeitlinger, line 79: is throwing an exception (and thus crashing the app) appropriate? can we just return an optional instead?

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/serialization/converters/AttributesJsonConverter.java
```diff
@@ -0,0 +1,220 @@
...
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.function.Function;
+
+@SuppressWarnings({"rawtypes", "unchecked"})
+@JsonConverter(target = Attributes.class)
+public final class AttributesJsonConverter {
```
zeitlinger, line 32: looks like  generated class - but doesn't have a "generated by .." comment

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/exceptions/MaxAttemptsReachedException.java
```diff
@@ -0,0 +1,11 @@
...
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.contrib.disk.buffering.internal.storage.exceptions;
+
+import java.io.IOException;
+
+@SuppressWarnings("serial")
+public class MaxAttemptsReachedException extends IOException {}
```
zeitlinger, line 11: using exception for control flow is discouraged in Java. Can we use some kind of result enum or object instead?

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/utils/TimeProvider.java
```diff
@@ -0,0 +1,19 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.contrib.disk.buffering.internal.storage.utils;
+
+public class TimeProvider {
+  @SuppressWarnings({"FieldCanBeFinal", "FieldMayBeFinal"}) // Needed for tests
+  private static TimeProvider instance = new TimeProvider();
```
zeitlinger, line 10: can we pass the timeprovider in tests that need this instead of the static field?

disk-buffering/src/test/java/io/opentelemetry/contrib/disk/buffering/exporters/MetricDiskExporterTest.java
```diff
@@ -0,0 +1,77 @@
...
+
+  @Test
+  public void onFlush_flushWrappedExporter() {
+    exporter.flush();
+
+    verify(wrapped).flush();
+  }
+
+  @Test
+  public void provideWrappedAggregationTemporality() {
```
zeitlinger, line 68: a little bit too fine granular test - but it's a matter of preference
In other words, tests that only check that a particular method is forwarded are not very valuable IMO.

disk-buffering/src/test/java/io/opentelemetry/contrib/disk/buffering/internal/storage/files/WritableFileTest.java
```diff
@@ -0,0 +1,122 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.util.List;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+class WritableFileTest {
```
zeitlinger, line 32: not sure if I missed it, but I couldn't find any integration test that tries to test all components together.
such a test could also reduce the number of required unit tests without hurting coverage.

zeitlinger (commented):
I reviewed everything except for what looks like generated classes.
Can you explain how those classes are generated - and which ones?

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/FolderManager.java
```diff
@@ -0,0 +1,145 @@
...
+            oldestFileAvailable = existingFile;
+          }
+        }
+      }
+    }
+    // Checking if the oldest available file is currently the writable file.
+    if (oldestFileAvailable != null
+        && currentWritableFile != null
+        && oldestFileAvailable.equals(currentWritableFile.file)) {
+      currentWritableFile.close();
```
zeitlinger, line 84: how can this condition be true? I thought the reader and writer live in different processes?

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/exporters/AbstractDiskExporter.java
```diff
@@ -0,0 +1,91 @@
...
+   * @param unit The unit of the time provided.
+   * @return TRUE if there was data available and it was successfully exported within the timeout
+   *     provided. FALSE if either of those conditions didn't meet.
+   * @throws IOException If an unexpected error happens.
+   */
+  public boolean exportStoredBatch(long timeout, TimeUnit unit) throws IOException {
+    logger.log(Level.INFO, "Attempting to export batch from disk.");
+    AtomicBoolean exportSucceeded = new AtomicBoolean(false);
+    boolean foundDataToExport =
+        storage.read(
```
zeitlinger, line 43: I don't understand what happens if the export thread gets killed after `storage.read`? Do we assume that the exporter thread lives forever?

I see that `ReadableFile` has an index for the current reading position, but I can't see an index of the current file that is being read - `findReadableFile` seems to pick the first stored file.

Maybe add a CONTRIBUTING.md file with a simple data flow diagram (e.g. with https://asciiflow.com) would help.

disk-buffering/README.md
```diff
@@ -0,0 +1,95 @@
...
+* The exporter to be wrapped.
+* A File instance of the root directory where all the data is going to be written. The same root dir
+  can be used for all the wrappers, since each will create their own folder inside it.
+* An instance
+  of [StorageConfiguration](src/main/java/io/opentelemetry/contrib/disk/buffering/storage/StorageConfiguration.java)
+  with the desired parameters. You can create one with default values by
+  calling `StorageConfiguration.getDefault()`.
+
+After wrapping your exporters, you must register the wrapper as the exporter you'll use. It will
+take care of always storing the data it receives.
```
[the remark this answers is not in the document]
LikeTheSalad: You're right. I'm not fully sure but I think flushing the data might only be needed when the output stream is buffered, which was the case at first, but then I replaced it with FileOutputStream directly when I ran some tests and found out that some data was lost when the app might be terminated without closing/flushing the buffered stream before. So it seemed to me that without the buffer, each time the `write` was called, the data was written right away, which seems more reliable at least for Android.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/exporters/MetricDiskExporter.java
LikeTheSalad: Do you mean returning success without calling `wrapped.flush()`? Or just returning success afterward?

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/exporters/AbstractDiskExporter.java
LikeTheSalad: Since I needed to know the outcome of the delegated exporter by line 51, the IDE suggested either using AtomicBoolean or an array of booleans with a single item to be able to change its value in a callback and read it outside of the callback later.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/serialization/converters/AttributesJsonConverter.java
LikeTheSalad: I see, though it's not generated 😅 I had to create 2 custom Json converters, one for Attributes and one for generic MetricData, both in the same folder. However, these will be removed from this PR to make it smaller.

disk-buffering/README.md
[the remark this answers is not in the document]
zeitlinger: great - this would be a valuable piece to add to the readme (maybe a design section)

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/exporters/MetricDiskExporter.java
zeitlinger: without calling wrapped - the point of the disk exporter is to do all interactions with the wrapped exporter later.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/exporters/AbstractDiskExporter.java
zeitlinger: can we simply to?

we'd have to pass the result of the consumer back to the caller in https://github.com/LikeTheSalad/opentelemetry-java-contrib/blob/9ab9aa91e7a92e0e4e208c794da16bc9e4cdf1e5/disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/files/ReadableFile.java#L81 - but I think that would be easier to understand

```java
    return storage.read(
        bytes -> {
          logger.log(Level.INFO, "About to export stored batch.");
          CompletableResultCode join =
              doExport(getSerializer().deserialize(bytes)).join(timeout, unit);
          return join.isSuccess();
        });
```

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/exceptions/MaxAttemptsReachedException.java
LikeTheSalad: I wonder if this applies for any kind of flow? I'm not aware of the details of the discouragement for this cases, but the reason I decided to use exceptions for non-happy path cases is because it allows for a more straightforward API usage. For example, this exception is thrown [in here](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/913/files/549ea97609ea7ee71ae2e069fd1e8cfb4ea2c9da#diff-4ab92b18edc8e550a570fd5e6f577e0d5d08df1f1aed47fd700ffe3c79e0694bR45) in case that some unexpected amount of recursive `write` calls happen (which in theory shouldn't, but it's there to make it more robust). Though essentially, `write` should not have to return anything because its happy path would just complete without any exceptions. If we make it to return an enum for example, some manual verification has to be done on the caller side in order to avoid ignoring an unexpected scenario, and the compiler won't be able to help by forcing the caller to handle that scenario unlike what happens with this exception. So generally speaking I'm always in favor to raise issues when something unexpected happens to make sure they are easily visible and taken into account at the time of writing the code, seems to make for a safer runtime.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/utils/TimeProvider.java
LikeTheSalad: I think we can yeah, I remember I had to change this several times so maybe in the end it's not needed to have it as a static field. I'll double check.

disk-buffering/src/test/java/io/opentelemetry/contrib/disk/buffering/exporters/MetricDiskExporterTest.java
LikeTheSalad: Trust me I want to spend time writing more tests than needed 😄 so I understand your point, though since the main focus of this implementation is to be a wrapper, I wanted to make it bulletproof in case somebody might remove this method call forwarding by mistake in the future (even myself, I tend to forget about the details of my code after like 2 months 😅)

disk-buffering/src/test/java/io/opentelemetry/contrib/disk/buffering/internal/storage/files/WritableFileTest.java
LikeTheSalad: You're right, for now I only focused on unit tests. Though I think we can add an integration one before handing it over for review later. If you have some ideas of best practices for these kinds of tests I'd be glad to get some insights from your side since I'm currently too into Android and Gradle plugins integration test suites, so I'm not aware of the latest tools for other environments that might make these easier to create.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/FolderManager.java
LikeTheSalad: It's probably a bit of an edge-case, though since the reading and writing processes can be done in parallel, there's a case in which a file is open for writing new data for too long (and stopped receiving new data), and at the same time a reader in a separate thread would want to start reading from the oldest available file, and there might only be a single file in the folder which happens to be the one open for writing because it's been idle for a while so the checking for closing it has not run yet. So in that case this validation makes sure to close said file before opening it up for reading to avoid IO errors. This can happen because both processes (write/read) can be done in parallel, now, in most cases they will be done in separate files, but it's not impossible that they might clash into the same one if there are no other files available to be read and the writing process is idle to the point that its writing timeout expires.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/exporters/AbstractDiskExporter.java
LikeTheSalad: I'm not sure I'm following the question. Though it seems like the concern is related to how to continue reading from the same file after the app gets destroyed and reopened?

`findReadableFile` will always get the oldest (unexpired) available file, so if the reading process gets killed after starting to read from a file, what would happen when a new read process is started is:

- findReadableFile will search for the oldest available file, which should be the same file it was being read before, so it will pick up where it left off.
- If the oldest file, which is the same as it was being read before, has expired due to the configured [maxFileAgeForRead](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/913/files/549ea97609ea7ee71ae2e069fd1e8cfb4ea2c9da#diff-f2b3e0dc8ee13418cbbdc79a8d73e8c201fae10caea943142626dd8c448c74a2R30) parameter, then the leftover data in that file will be discarded because it's stale, and the next oldest file will be picked up.

I hope it helps clarify it. And yeah I agree a data flow diagram would be really helpful. I'll find the time to create after the code changes are done.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/exporters/MetricDiskExporter.java
LikeTheSalad: I see, my only concern is that some exporters tend to batch signals and send them later, so the same might happen even when the disk exporter delegates data to them, so in those cases this might cause that data to never get exported. However, I believe for that case it might be better to call `wrapped.flush()` right after the disk exporter delegates the data (during the reading process) to avoid these kinds of issues, and after that, we can simply remove the flush call from here and just return success right away as you suggest 👍  I'm glad I got this insight because without calling `flush` during the reading/delegating process, the disk exporter might be getting unnecessary delays during the delegation calls right now, so it's a great catch!

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/exceptions/MaxAttemptsReachedException.java
zeitlinger: here's an article: https://softwareengineering.stackexchange.com/questions/189222/are-exceptions-as-control-flow-considered-a-serious-antipattern-if-so-why

In my experience, it's much harder to understand control flow with exceptions, because an exception can terminate multiple layers in the call stack.
On the other hand, it's very easy to check that a returned result code is not ignored.

disk-buffering/src/test/java/io/opentelemetry/contrib/disk/buffering/internal/storage/files/WritableFileTest.java
zeitlinger: such an "integration test" can be called like a unit test - no special gradle plugin or similar is needed.

The integration test should cover most classes that make the core logic are tested (and not mocked) - in our case that would be the entire storage layer. 
The integration test would first create files (using the exporters) and then read (and possibly delete) the same files using the export loop.
Technically, you can create the required object graph with constructor injection - like you're already doing in some tests.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/exporters/AbstractDiskExporter.java
LikeTheSalad: I see, that looks way more readable for sure. I'll apply the changes, though maybe I'll also rename the `read` method to something like `readAndProcess` in order to express the full intention and probably give a hint on what to expect as return from it.

disk-buffering/src/test/java/io/opentelemetry/contrib/disk/buffering/internal/storage/files/WritableFileTest.java
LikeTheSalad: Sounds simple enough.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/FolderManager.java
zeitlinger: > and at the same time a reader in a separate thread ...

This is the critical part I'm trying to understand. Is the assumption that reader and writer
- always use the same JVM?
- always use the same instance of the storage object?

LikeTheSalad: That's correct. The storage object is not exposed from the signal exporter, and there should be a single exporter per signal which is used for both read/write operations. The writing process happens automatically as new signals are provided from the signal processor, whereas the reading process is triggered (using the same signal disk exporter) elsewhere in the consumer of the lib.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/exporters/AbstractDiskExporter.java
LikeTheSalad: That's correct, it doesn't matter what was sent before when it comes to picking what file to read from, as long as the file has not expired. Now, if the same file is chosen more than once, the second time it is read it won't provide the data that was previously sent though, this is because the data is removed from the file as soon as it gets successfully exported.

zeitlinger: > because the data is removed from the file as soon as it gets successfully exported.

where is that logic?

LikeTheSalad: It's [in here](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/913/files#diff-fb0920b798b73be9fdbda266cabf3752079478b8d8ac91983c3ee11b754da5fcR86).

zeitlinger:
[Discussion doc](https://docs.google.com/document/d/1HxujHEFXIv2YT1Cw7t-SXs8aR_ivT4uEbEYmapvepCQ/edit)

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/exporters/AbstractDiskExporter.java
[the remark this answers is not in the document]
LikeTheSalad: Thank you, I've made the changes.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/exporters/MetricDiskExporter.java
LikeTheSalad: Thank you, I've made the changes. Also, after double-checking, it seems like no exporters are batching anything (I must have confused them with some processors) and the ones available right now are always returning success here, so there's no need to do the flushing after the reading/delegating process either.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/exceptions/MaxAttemptsReachedException.java
LikeTheSalad: Thank you, I've made the changes.

disk-buffering/src/test/java/io/opentelemetry/contrib/disk/buffering/internal/storage/files/WritableFileTest.java
LikeTheSalad: Thank you, I've made the changes.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/utils/TimeProvider.java
LikeTheSalad: Thank you, I've made the changes.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/exporters/AbstractDiskExporter.java
LikeTheSalad: Thank you, I've made the changes.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/exporters/AbstractDiskExporter.java
LikeTheSalad: I missed this comment for the previous changes I made. So from a consumer point of view, this code will be executed when creating a new instance of either of the signal disk exporters. So maybe, in order to return some value instead of throwing an exception, we might have to make the constructors private and use some static function for creating the instances instead. Apart from that, I'm just realizing that this exception isn't present in the constructor's signature, so there's no compilation warning that people can see to become aware of this possible scenario, so that's an issue.

Now, I think there has to be some sort of specific feedback for this because, if the folder cannot be created, it's most likely due to some issue with the `rootDir` File provided by the consumer, so my concern with using an optional, or even just returning null when creating an instance through a static function, would be that the reason why the class couldn't be instantiated wouldn't be clear, so the consumer wouldn't easily know what to fix, which they have to I believe since they should be aware of the disk space and permissions available for the folder they provide. Though I'm open to suggestions to achieve the same with another mechanism. Now, if the exception is what we end up choosing though, I think I'll have to convert it into an IOException one to provide the compile-time checks.

zeitlinger: sounds good :smile:

jack-berg:
A few weeks back we discussed using lower code alternatives to serialization / deserialization to make reviewing more manageable. I've pushed a commit which demonstrates using the hand rolled serialization from `opentelemetry-java/exporters/otlp/common`, and using `opentelemetry-proto-java` deserialization: https://github.com/jack-berg/opentelemetry-java-contrib/commit/5d2e8455201b2727969cad937c55602cba04742e

I think this type of approach is a good first step to landing this work.

disk-buffering/README.md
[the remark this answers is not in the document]
LikeTheSalad: I've just added some changes to the README, I think it should be better now.

disk-buffering/README.md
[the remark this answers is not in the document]
LikeTheSalad: Thanks! I've just mentioned this in the CONTRIBUTING.md file.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/exporters/AbstractDiskExporter.java
LikeTheSalad: I've decided to create a static `create` function for each disk exporter since it seems to be a common pattern in other classes of the Java SDK.

LikeTheSalad:
Thanks for taking the time @jack-berg! I've made some updates to replace the custom Json serialization to use `opentelemetry-proto-java` instead, but even after that it seems like the PR might still be quite large, so from a first glance at your code I think I got an idea that could help further reduce serialization part a lot since I see it's taking advantage of the existing process. I'll take a deeper look to see how to integrate it. Cheers!

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/exporters/DiskExporter.java
```diff
@@ -0,0 +1,102 @@
...
+        new Storage(
+            new FolderManager(getSignalFolder(rootDir, folderName), configuration, timeProvider));
+    this.serializer = serializer;
+    this.exportFunction = exportFunction;
+  }
+
+  @Override
+  public boolean exportStoredBatch(long timeout, TimeUnit unit) throws IOException {
+    logger.log(Level.INFO, "Attempting to export batch from disk.");
+    AtomicBoolean exportSucceeded = new AtomicBoolean(false);
```
zeitlinger, line 50: This boolean is redundant now.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/responses/ReadableResult.java
```diff
@@ -0,0 +1,14 @@
...
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.contrib.disk.buffering.internal.storage.responses;
+
+public enum ReadableResult {
+  SUCCEEDED,
+  CLOSED,
+  FILE_HAS_EXPIRED,
```
zeitlinger, line 11: this enum value is returned, but no caller reacts on it - so maybe it's effectively the same as processing failed

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/responses/ReadableResult.java
```diff
@@ -0,0 +1,14 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.contrib.disk.buffering.internal.storage.responses;
+
+public enum ReadableResult {
+  SUCCEEDED,
+  CLOSED,
```
zeitlinger, line 10: this enum value is returned, but no caller reacts on it - so maybe it's effectively the same as processing failed

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/responses/ReadableResult.java
```diff
@@ -0,0 +1,14 @@
...
+  FILE_HAS_EXPIRED,
+  PROCESSING_FAILED,
+  NO_CONTENT_AVAILABLE
```
zeitlinger, line 13: this enum value is returned, but no caller reacts on it - so maybe it's effectively the same as processing failed

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/responses/WritableResult.java
```diff
@@ -0,0 +1,13 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.contrib.disk.buffering.internal.storage.responses;
+
+public enum WritableResult {
+  SUCCEEDED,
+  CLOSED,
```
zeitlinger, line 10: this enum value is returned, but no caller reacts on it

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/responses/WritableResult.java
```diff
@@ -0,0 +1,13 @@
...
+  FILE_EXPIRED,
```
zeitlinger, line 11: this enum value is returned, but no caller reacts on it

zeitlinger (commented):
great job :smile: 

Only some small things left.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/exporters/DiskExporter.java
LikeTheSalad: Thanks, it's updated now.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/responses/ReadableResult.java
LikeTheSalad: Thanks, it's updated now.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/responses/ReadableResult.java
LikeTheSalad: Thanks, it's updated now.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/responses/ReadableResult.java
LikeTheSalad: Thanks, it's updated now.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/responses/WritableResult.java
LikeTheSalad: Thanks, it's updated now.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/responses/WritableResult.java
LikeTheSalad: Thanks, it's updated now.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/responses/WritableResult.java
```diff
@@ -0,0 +1,13 @@
...
+  FILE_IS_FULL
```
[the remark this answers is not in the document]
LikeTheSalad: Thanks, it's updated now.

zeitlinger:
looks good from my side - let's just see the outcome of the serializer discussion with @jack-berg

LikeTheSalad:
Sounds good, I'm planning to discuss it in the SIG, though basically, I tried to reuse the existing serialization process with the one I created using the proto lib, and there's a missing piece, which is that the serialization currently doesn't create the proto object that contains an array of resource-gathered signals, which is handy to store full batches as one item that can be easily retrieved independently during the reading process. @jack-berg provided a solution for that when using Json, though after the change, I was using proto serialization to avoid adding more weight to the lib 😅 so the rest of the code changes are expecting the proto format at the moment. Which I can change to use Json in case there might be an issue with using the proto format for now. However, after checking Jack's code, it occurred to me that another way to further reduce this PR was by using directly the internal SDK implementations of the signals data interfaces instead of creating my own, which actually worked and helped to reduce it a lot! So I'm hoping it's much better now to review it with the amount of files left.

LikeTheSalad:
Hi @jack-berg, we couldn't discuss the serialization during yesterday's call so maybe we can discuss it over here? I'm just curious about what you think of the current approach, which is just using the proto format directly from the proto java source, in relation to making this PR better for reviewing. Or if you think it's still worth making the switch to proto+json?

zeitlinger (approved):

trask:
will leave this open for a few days in case anyone else would like to review, will plan to merge it before the 1.28.0 release

disk-buffering/README.md
```diff
@@ -0,0 +1,114 @@
...
+```
+
+Both, the writing and reading processes can run in parallel and the way they don't overlap is
+because each is supposed to happen in different files, and the way it is ensured that they don't
+accidentally meet in the same file is by using the configurable parameters, which ensure to set time
+frames for each action to be done on a single file at a time. On top of that, there's a mechanism in
+place to avoid overlapping on edge cases where the time frames ended but the resources haven't been
+released, for that mechanism to work properly, this tool assumes that both, the reading and the
+writing actions, are executed within the same application process.
+
```
breedx-splk, line 108:
```suggestion
Both the writing and reading processes can run in parallel and they don't overlap
because each is supposed to happen in different files. We ensure that reader and writer don't
accidentally meet in the same file by using the configurable parameters. These parameters set non-overlapping time frames for each action to be done on a single file at a time. On top of that, there's a mechanism in
place to avoid overlapping on edge cases where the time frames ended but the resources haven't been
released. For that mechanism to work properly, this tool assumes that both the reading and the
writing actions are executed within the same application process.

```

breedx-splk: Oh and this obviously isn't a show-stopper.

breedx-splk (approved):
Disclaimer: I have not looked at every line of code, but I have looked at the overall design and a bunch of the code and think this is a great start/addition. I do wish it would have been more incremental, but alas, he were are.

I have a number of things I am thinking about iterating on with this, but since this is new and experimental contrib code, at this point I would prefer to merge the big module and do piecemeal issues/changes after. 

Thanks @LikeTheSalad !

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/exporters/LogRecordDiskExporter.java
```diff
@@ -0,0 +1,100 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.contrib.disk.buffering.exporters;
```
jack-berg, line 6: Let's put all the public classes in `io.opentelemetry.contrib.disk.buffering`, and put the remaining stuff in `io.opentelemetry.contrib.disk.buffering.internal.*`.

In other words, I don't think there's a need for the breaking out `i.o.c.d.b.exporters` and `i.o.c.d.b.storage`.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/storage/files/DefaultTemporaryFileProvider.java
```diff
@@ -0,0 +1,20 @@
...
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.contrib.disk.buffering.storage.files;
+
+import java.io.File;
+import java.io.IOException;
+
+public class DefaultTemporaryFileProvider implements TemporaryFileProvider {
```
jack-berg, line 11:
```suggestion
public final class DefaultTemporaryFileProvider implements TemporaryFileProvider {
```

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/storage/files/DefaultTemporaryFileProvider.java
```diff
@@ -0,0 +1,20 @@
...
+  public static final TemporaryFileProvider INSTANCE = new DefaultTemporaryFileProvider();
```
jack-berg, line 12: We prefer exposing singletons via public static methods rather than public static fields:
```
private static final TemporaryFileProvider INSTANCE = new DefaultTemporaryFileProvider();

public static TemporaryFileProvider getInstance() {
  return INSTANCE;
}
```

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/exporters/StoredBatchExporter.java
```diff
@@ -0,0 +1,23 @@
...
+  /**
+   * Reads data from the disk and attempts to export it.
+   *
+   * @param timeout The amount of time to wait for the wrapped exporter to finish.
+   * @param unit The unit of the time provided.
+   * @return TRUE if there was data available and it was successfully exported within the timeout
+   *     provided. FALSE if either of those conditions didn't meet.
+   * @throws IOException If an unexpected error happens.
+   */
+  boolean exportStoredBatch(long timeout, TimeUnit unit) throws IOException;
```
jack-berg, line 22: Seems like this method has roughly the same semantics as `CompletableResultCode {Signal}Exporter#flush()`.

Can we get rid of this method and instead have users use `flush()` to force an read from disk and export?

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/utils/TimeProvider.java
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.contrib.disk.buffering.internal.storage.utils;
+
+public class TimeProvider {
```
jack-berg, line 8: Consider using [io.opentelemetry.sdk.common.Clock](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/common/src/main/java/io/opentelemetry/sdk/common/Clock.java) instead of introducing a new abstraction.

jack-berg (commented):
Some minor comments. Definitely easier to grok with the simplified serialization logic.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/utils/TimeProvider.java
breedx-splk: Oof, I forgot to make this same observation! Good call jack.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/exporters/LogRecordDiskExporter.java
LikeTheSalad: Sounds good to me, I've applied the changes.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/storage/files/DefaultTemporaryFileProvider.java
LikeTheSalad: Thanks, I've applied this change to all the cases where an INSTANCE field was directly accessed.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/utils/TimeProvider.java
LikeTheSalad: I just made the change, although I'm not sure if the use case here is properly addressed by a Clock implementation because the time format that the storage classes use is milliseconds, which is also the format used in the configuration parameters, making it handy to do the math when checking for changes in the state of the cache files based on the user-provided parameters. So if we provide nanoseconds as local time, we'd have to make conversions all the time, which would turn in overall to more than one conversion needed per action, considering that the way a simple clock like the one used here would provide the current time in nanoseconds would be by converting the system current milliseconds to nanos.

I think it's also worth noting that this is meant to be an internal utility class which makes it possible to change the time for testing purposes, whereas I believe the usage of Clock is also handy for when we need to provide the consumers the ability to use their own implementation, though at the moment that's not needed in here.

So based on the above, the changes I made are essentially an incorrect implementation of Clock, which returns the current time in milliseconds to avoid unnecessary conversion issues. It's a bit ugly to me, though it's a simple way to reuse the Clock interface, though it's ugly nevertheless, so I'd like to know your opinion on it.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/exporters/StoredBatchExporter.java
LikeTheSalad: I think it can work yeah. My only concern with doing so would be that the expectations of someone using this method won't probably match the reality of what actually happens, in the sense that, I believe the `flush()`method suggests that all the unexported signals will be exported right away. Though for the case of the disk exporter, it would mean that the next available batch of signals in the queue will be sent, but not all the batches at once. So I'm wondering if that could cause misunderstandings and probably invalid issues to be created in the repo because of it.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/files/utils/Constants.java
```diff
@@ -0,0 +1,17 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.contrib.disk.buffering.internal.storage.files.utils;
+
+import java.nio.charset.StandardCharsets;
+
+public final class Constants {
+
+  public static final byte[] NEW_LINE_BYTES =
```
laurit, line 12: This is only used in a test. Is this a remnant from the json based serialzier?

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/Storage.java
```diff
@@ -0,0 +1,105 @@
...
+  @Nullable private ReadableFile readableFile;
+  private static final int MAX_ATTEMPTS = 3;
+  private final AtomicBoolean isClosed = new AtomicBoolean(false);
+
+  public Storage(FolderManager folderManager) {
+    this.folderManager = folderManager;
+  }
+
+  /**
+   * Attempts to write a line into a writable file.
```
laurit, line 30: `line` feels a bit misleading as far as I can tell it's just binary data

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/files/ReadableFile.java
```diff
@@ -0,0 +1,153 @@
...
+      StorageClock clock,
+      StorageConfiguration configuration,
+      StreamReader.Factory readerFactory)
+      throws IOException {
+    super(file);
+    this.clock = clock;
+    expireTimeMillis = createdTimeMillis + configuration.getMaxFileAgeForReadMillis();
+    originalFileSize = (int) file.length();
+    temporaryFile = configuration.getTemporaryFileProvider().createTemporaryFile(file.getName());
+    copyFile(file, temporaryFile);
```
laurit, line 60: I found the copy source file to temp file, read temp file and overwrite the source after each batch logic hard to follow. Comments might help.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/files/ReadableFile.java
```diff
@@ -0,0 +1,153 @@
...
+   *     If the processing function returns TRUE, then the provided line will be deleted from the
+   *     source file. If the function returns FALSE, no changes will be applied to the source file.
+   */
+  public synchronized ReadableResult readAndProcess(Function<byte[], Boolean> processing)
+      throws IOException {
+    if (isClosed.get()) {
+      return ReadableResult.FAILED;
+    }
+    if (hasExpired()) {
+      close();
```
laurit, line 80: something else is supposed to delete the expired file?

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/files/ReadableFile.java
```diff
@@ -0,0 +1,153 @@
...
+  public synchronized void close() throws IOException {
+    if (isClosed.compareAndSet(false, true)) {
+      unconsumedResult = null;
+      fileTransferUtil.close();
+      reader.close();
+      temporaryFile.delete();
+    }
+  }
+
+  private static void copyFile(File from, File to) throws IOException {
```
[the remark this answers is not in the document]
LikeTheSalad: [It is](https://developer.android.com/reference/java/nio/file/Files#copy(java.nio.file.Path,%20java.nio.file.Path,%20java.nio.file.CopyOption[])), it's just that it was added in API level 26 which is quite high for apps that need to support older devices, we've got a couple of customers which are on level 24. Though now that I think about it better, if the core OTel SDK requires library desugaring, then probably there's no need to make this lib safe for older versions. I'll take a deeper look.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/files/utils/Constants.java
LikeTheSalad: Nice catch, it is from when we were using the custom json serialization.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/files/ReadableFile.java
laurit: Instead of copying the data around it might be more efficient to alter the file format so that it would be possible to tell which chunks have already been handled. Having some kind of header in the data file could also be useful. If it is possible that files written by older version of library are read by newer version then having a header could allow skipping data in case file format has changed.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/files/ReadableFile.java
LikeTheSalad: A clean up is done when creating a new file [here](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/913/files#diff-5132fe2f62753f8cdbc84a13763f484ddbeaea87f8af84fcc8519bdbbe213f5eR84).

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/files/ReadableFile.java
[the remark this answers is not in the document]
laurit: Thanks, that makes sense. I had a quick peek at desugared api list and couldn't find `Files.copy`. I'd just add a comment there explaining why you are not using `Files.copy`. An alternative would be to implement copy using `FileChannel.transferTo`.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/files/ReadableFile.java
LikeTheSalad: @zeitlinger and I also discussed some more optimized approaches as well, such as having a separate file to keep track of the bytes already read and updating that number instead. The header approach I think it's a good idea as well, especially because it'd keep everything in the same file, however, I'd have to double-check how could we update the header values at the same time the file is being read to make sure we don't lose the current position if the app gets terminated without a chance to properly close the reader. Though ultimately, since there's a lot to this PR already, and, based on a benchmark added to this specific functionality, I found out that the simple approach is quite decent in terms of performance, I was thinking it would be ideal to add these optimizations on a further PR if needed.

laurit: Exploring these ideas in the future is fine.

disk-buffering/build.gradle.kts
```diff
@@ -0,0 +1,32 @@
+plugins {
+  id("otel.java-conventions")
+  id("otel.publish-conventions")
+  id("me.champeau.jmh") version "0.7.1"
```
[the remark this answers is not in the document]
LikeTheSalad: Good idea, I've just added it to verify API 24 supported code

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/serialization/serializers/LogRecordDataSerializer.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.List;
+import javax.annotation.Nullable;
+
+public final class LogRecordDataSerializer implements SignalSerializer<LogRecordData> {
+  @Nullable private static LogRecordDataSerializer instance;
+
+  private LogRecordDataSerializer() {}
+
+  static LogRecordDataSerializer get() {
+    if (instance == null) {
```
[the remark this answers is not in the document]
LikeTheSalad: Thanks, I've made the changes

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/serialization/serializers/LogRecordDataSerializer.java
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  @Override
+  public byte[] serialize(Collection<LogRecordData> logRecordData) {
+    LogsData proto = ProtoLogsDataMapper.getInstance().toProto(logRecordData);
+    try (ByteArrayOutputStream out = new ByteArrayOutputStream()) {
+      proto.writeDelimitedTo(out);
+      return out.toByteArray();
+    } catch (IOException e) {
+      throw new IllegalArgumentException(e);
```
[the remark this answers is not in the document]
LikeTheSalad: Thanks, I don't have a strong opinion on this, I've just changed it

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/files/ReadableFile.java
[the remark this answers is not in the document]
LikeTheSalad: I've added a comment explaining why it was done this way, and also, the animalsniffer check will complain if Files.copy is used

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/files/ReadableFile.java
```diff
@@ -0,0 +1,153 @@
...
+    try (InputStream in = new BufferedInputStream(new FileInputStream(from));
+        OutputStream out = new FileOutputStream(to, false)) {
```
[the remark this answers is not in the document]
LikeTheSalad: Thanks, I've removed it

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/files/ReadableFile.java
```diff
@@ -0,0 +1,153 @@
...
+    if (isClosed.compareAndSet(false, true)) {
+      unconsumedResult = null;
+      fileTransferUtil.close();
+      reader.close();
+      temporaryFile.delete();
+    }
+  }
+
+  private static void copyFile(File from, File to) throws IOException {
+    try (InputStream in = new BufferedInputStream(new FileInputStream(from));
```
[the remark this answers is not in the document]
LikeTheSalad: Good point, I've removed it.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/Storage.java
LikeTheSalad: It's removed now.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/files/ReadableFile.java
LikeTheSalad: Thanks, I've added an explanation in the class's doc

trask (approved):
thx @LikeTheSalad!

and thx @zeitlinger @breedx-splk @jack-berg @laurit for reviewing!!

PR #924, opened 2023-06-15
generate error/fault metrics by aws sdk status code

scaugrated:
**Description:**

A short term solution to generate error/fault metrics by aws sdk status code.

When the AWS SDK calls an API then returns non-200 status code, the AWS SDK simply throws an exception. The exception thrown by the AWS SDK is stored within the produced spans and is accessible in the AwsSpanMetricsProcessor, where we generate Fault/Error metrics. 

We attempt to get the throwable object and call its getStatusCode method to obtain the status code, then generate metrics based on the corresponding status code.

**Existing Issue(s):**

[http.status_code attribute in span was not being populated when AWS APIs were returning non-200 status codes](https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/8795)

**Testing:**

unit tests

**Documentation:**
N/A

**Outstanding items:**
N/A

### DIFF
--- a/aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsSpanMetricsProcessor.java
+++ b/aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsSpanMetricsProcessor.java
@@ -15,7 +15,11 @@ import io.opentelemetry.sdk.resources.Resource;
 import io.opentelemetry.sdk.trace.ReadWriteSpan;
 import io.opentelemetry.sdk.trace.ReadableSpan;
 import io.opentelemetry.sdk.trace.SpanProcessor;
+import io.opentelemetry.sdk.trace.data.EventData;
 import io.opentelemetry.sdk.trace.data.SpanData;
+import io.opentelemetry.sdk.trace.internal.data.ExceptionEventData;
+import java.lang.reflect.Method;
+import javax.annotation.Nullable;
 import javax.annotation.concurrent.Immutable;
 
 /**
@@ -92,7 +96,7 @@ public final class AwsSpanMetricsProcessor implements SpanProcessor {
 
     // Only record metrics if non-empty attributes are returned.
     if (!attributes.isEmpty()) {
-      recordErrorOrFault(span, attributes);
+      recordErrorOrFault(spanData, attributes);
       recordLatency(span, attributes);
     }
   }
@@ -102,10 +106,14 @@ public final class AwsSpanMetricsProcessor implements SpanProcessor {
     return true;
   }
 
-  private void recordErrorOrFault(ReadableSpan span, Attributes attributes) {
-    Long httpStatusCode = span.getAttribute(HTTP_STATUS_CODE);
+  private void recordErrorOrFault(SpanData spanData, Attributes attributes) {
+    Long httpStatusCode = spanData.getAttributes().get(HTTP_STATUS_CODE);
     if (httpStatusCode == null) {
-      return;
+      httpStatusCode = getAwsStatusCode(spanData);
+
+      if (httpStatusCode == null || httpStatusCode < 100L || httpStatusCode > 599L) {
+        return;
+      }
     }
 
     if (httpStatusCode >= ERROR_CODE_LOWER_BOUND && httpStatusCode <= ERROR_CODE_UPPER_BOUND) {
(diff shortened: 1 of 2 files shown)


aws-xray/src/test/java/io/opentelemetry/contrib/awsxray/AwsSpanMetricsProcessorTest.java
```diff
@@ -56,6 +61,18 @@ private enum ExpectedStatusMetric {
 
   private AwsSpanMetricsProcessor awsSpanMetricsProcessor;
 
+  static class ThrowableWithMethodGetStatusCode extends Throwable {
```
thpierce, line 64: Lets add a test covering a Throwable with a `statusCode` method and a Throwable with neither `statusCode` nor `getStatusCode` methods.

vastin (approved):
LGTM

wangzlei (approved):

scaugrated:
Hi @trask , all other approval requirements met, please take a look when you have a chance.

srprash (approved):

scaugrated:
Hi @jack-berg , all other approval requirements met, please take a look when you have a chance.

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsSpanMetricsProcessor.java
```diff
@@ -116,6 +125,57 @@ private void recordErrorOrFault(ReadableSpan span, Attributes attributes) {
...
+   * AmazonServiceExceptions (V1) or AwsServiceExceptions (V2) because the throwable comes from a
+   * separate class loader and attempts to cast will fail with ClassCastException.
+   *
+   * <p>TODO: Short term workaround. This can be completely removed once
+   * https://github.com/open-telemetry/opentelemetry-java-contrib/issues/919 is resolved.
+   */
+  @Nullable
+  private static Long getAwsStatusCode(SpanData spanData) {
+    InstrumentationScopeInfo scopeInfo = spanData.getInstrumentationScopeInfo();
+    if (scopeInfo == null) {
```
jack-berg, line 144: This is unnecessary as spanData.getInstrumentationScopeInfo() is not annotated with `@Nullable` and does not return null.

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsSpanMetricsProcessor.java
```diff
@@ -116,6 +125,57 @@ private void recordErrorOrFault(ReadableSpan span, Attributes attributes) {
...
+      return null;
+    }
+
+    String scopeName = scopeInfo.getName();
+    if (scopeName == null || !scopeName.contains("aws-sdk")) {
```
jack-berg, line 149: Again, no null check required for `scopeInfo.getName()`.

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/AwsSpanMetricsProcessor.java
```diff
@@ -116,6 +125,57 @@ private void recordErrorOrFault(ReadableSpan span, Attributes attributes) {
...
+      return null;
+    }
+
+    for (EventData event : spanData.getEvents()) {
+      if (event instanceof ExceptionEventData) {
+        ExceptionEventData exceptionEvent = (ExceptionEventData) event;
+        Throwable throwable = exceptionEvent.getException();
+
+        try {
+          Method method = throwable.getClass().getMethod("getStatusCode", new Class<?>[] {});
```
jack-berg, line 159: Code owners for this are @wangzlei and @srprash. Seems like a bad idea to do reflection on the hot path.

The javadoc for the method above mentions that this is a short term fix until [this issue](https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/8795) is resolved. I won't block this if the code owners want to proceed with this solution, but it would be nice to see a solution proposed to address the root of the problem.

scaugrated: yes, this short term solution is not prefect but acceptable from our side, @wangzlei and @srprash have approved this change. We will deprecate this solution while delivering long term solution.

scaugrated:
Hi @trask , @jack-berg , could you please take a look when you have a chance.

trask (approved):

Issue #910, opened 2023-06-08
[jmx-metrics] Metrics Gatherer reports old data if target becomes unavailable


dehaansa:
### Component(s)

jmx-metrics

## Description
When monitoring a JMX target with the JMX metrics gatherer, if that target becomes unresponsive or unavailable the most recently collected metrics continue to report via the configured exporter as if they are new metrics.

## Steps to Reproduce
Start a simple java executable like the following
```
cat <<EOF > hello.java
class Sleepy {
  public static void main(String args[]) throws InterruptedException {
    while (true) {
      Thread.sleep(10000);
    }
  }
}
EOF
javac hello.java

java -ea -Dcom.sun.management.jmxremote.port=9010 -Dcom.sun.management.jmxremote.authenticate=false -Dcom.sun.management.jmxremote.ssl=false -Djava.rmi.server.hostname=127.0.0.1 Sleepy > /dev/null 2>&1 &
```

Configure the JMX Metric gatherer to monitor said target.
```
otel.jmx.service.url = service:jmx:rmi:///jndi/rmi://localhost:9010/jmxrmi
otel.jmx.target.system = jvm
otel.jmx.interval.milliseconds = 5000
otel.jmx.remote.registry.ssl=false
otel.metrics.exporter = prometheus
otel.exporter.prometheus.host = localhost
otel.exporter.prometheus.port = 4317
```

After seeing metrics collected at the prometheus endpoint, then stop the Sleepy java process. Observe that the metrics are still reported (with fresh timestamps) at the prometheus endpoint.

While the prometheus endpoint would make some sense to continue to show old data points (with correct timestamps), the real use case where this issue was encountered was utilizing the jmx receiver in opentelemetry-collector-contrib, where we would not expect old, duplicate metrics to be exported via OTLP and sent to the receiver. To observe that behavior, the jmx receiver would be configured like the following (note - the receiver can only be used with a released version of the JMX Metrics gatherer, or the collector must be built with special build flags to allow use of a non-released version):

```
exporters:
  file:
    path: ./output.json
receivers:
  jmx:
    collection_interval: 5s
    endpoint: localhost:9010
    jar_path: /path/to/opentelemetry-java-contrib-jmx-metrics.jar
    target_system: jvm
service:
  pipelines:
    metrics:
      exporters:
        - file
      receivers:
        - jmx
```

## Expected Result
[...]

PR #949, opened 2023-07-11
[jmx-metrics] Collect in callback

dehaansa:
**Description:**
Update the jmx metrics gatherer to collect the actual metric values being reported in the callback.

This was initially started to resolve an issue where once a target went down, the same metric values continued to report.  The existing code registers its async callback(s) (as many as have been successful collections, so a potential memory leak) to report the value last collected when the groovy script(s) are executed, so when the SDK triggers its `collectAll()` method, those callbacks are executed and the last collected value is reported. Now, on execution of groovy scripts callbacks are registered with the MBeans that currently match the filter in the script, but the actual metric collection happens inside of the callback that is executed when the SDK goes to export.

**Existing Issue(s):** https://github.com/open-telemetry/opentelemetry-java-contrib/issues/910
May be relevant to several other issues including memory leaks recently reported & issues with duplicate data errors reporting.

**Testing:** Local testing against JVM targets. Trying to wrap my head around adding unit tests that would cover this scenario, but as long as the integration tests work that should cover the behavior.

Have validated locally that the following situations work as expected
- metrics gatherer starts with target unavailable, then target becomes available. No metrics collected initially, once target comes available metrics flow as expected
- metrics gatherer starts with target available, target becomes unavailable, then later becomes available again. Metrics collected initially, collection returns no metrics while target is unavailable, then metrics collected as expected again once target is again available

**Documentation:**

Should not be a user-facing change that would require documentation. The target of this refactor was to work with the existing JMX scripts without any issues.

**Outstanding items:**

- [x] Add thorough comments to the new code


jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/GroovyMetricEnvironment.java
```diff
@@ -331,55 +471,99 @@ public void registerDoubleUpDownCounterCallback(
...
+      final int descriptorHash, final Consumer<ObservableDoubleMeasurement> updater) {
+    doubleUpdaterRegistry.putIfAbsent(descriptorHash, new AtomicReference<>());
     AtomicReference<Consumer<ObservableDoubleMeasurement>> existingUpdater =
-        doubleUpdaterRegistry.get(descriptor.hashCode());
+        doubleUpdaterRegistry.get(descriptorHash);
     existingUpdater.set(updater);
     return doubleResult -> {
-      Consumer<ObservableDoubleMeasurement> existing = existingUpdater.get();
+      Consumer<ObservableDoubleMeasurement> existing =
+          doubleUpdaterRegistry.get(descriptorHash).get();
```
rmfitzpatrick, line 554: Is this indirection necessary?

jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/InstrumentHelper.groovy
```diff
@@ -58,81 +62,54 @@ class InstrumentHelper {
...
-        def tupleToUpdates = [:] // tuple is of form (instrument, instrumentName, description, unit)
-
-        values.each { collectedValue ->
-            def mbean = collectedValue.getV1()
-            def attribute = collectedValue.getV2()
-            def value = collectedValue.getV3()
+        // Look at the first mbean collected to evaluate if the attributes requested are
+        // composite data types or simple. Composite types require different parsing to
+        // end up with multiple recorders in the same callback.
+        def bean = mbeans.first()
```
rmfitzpatrick, line 86: Are we sure it's guaranteed the first will be reflective of the rest?

jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/GroovyMetricEnvironment.java
```diff
@@ -219,19 +227,43 @@ public LongHistogram getLongHistogram(
...
+                InstrumentType.OBSERVABLE_GAUGE,
+                InstrumentValueType.DOUBLE)
+            .hashCode();
+
+    // Only build the instrument if it isn't already in the registry
+    ObservableMeasurement obs = instrumentOnceRegistry.get(descriptorHash);
+    if (obs == null) {
+      obs = meter.gaugeBuilder(name).setDescription(description).setUnit(unit).buildObserver();
+      instrumentOnceRegistry.put(descriptorHash, obs);
+      // If an updater was not provided, the measurement is expected to be added
```
rmfitzpatrick, line 251: Does this describe an intentional case or is it for safety?

dehaansa: This is an intentional case, providing for the instruments for composite data to be grouped into the same batchCallback.

jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/InstrumentHelper.groovy
dehaansa: it's not guaranteed 100%, but it would break metric conventions if some mbeans in a group were composite data and some were not (IE you would have `jvm.memory.pool` metrics & the `jvm.memory.pool.(committed|used|etc)` metrics if that instrument contained some composite & some simple.

We could add some complexity in order to split mbeans by type, or maybe we just log an error and bail if not all of the mbeans in an instrument match?

jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/GroovyMetricEnvironment.java
dehaansa: Storing it into the local variable before calling accept? No, I just didn't refactor that from the existing code. Will do.

jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/GroovyMetricEnvironment.java
```diff
@@ -331,56 +389,95 @@ public void registerDoubleUpDownCounterCallback(
...
+    }
+  }
+
+  private <T extends ObservableMeasurement> Consumer<T> proxiedObserver(
+      final int descriptorHash,
+      final Map<Integer, AtomicReference<Consumer<T>>> registry,
+      final Consumer<T> updater) {
+    registry.putIfAbsent(descriptorHash, new AtomicReference<>());
+    registry.get(descriptorHash).set(updater);
+    return longResult -> registry.get(descriptorHash).get().accept(longResult);
```
rmfitzpatrick, line 481: if this is for all

```suggestion
    return result -> registry.get(descriptorHash).get().accept(result);
```

jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/GroovyMetricEnvironment.java
```diff
@@ -331,56 +389,95 @@ public void registerDoubleUpDownCounterCallback(
...
-            name, description, unit, instrumentType, InstrumentValueType.DOUBLE);
-    doubleUpdaterRegistry.putIfAbsent(descriptor.hashCode(), new AtomicReference<>());
-    AtomicReference<Consumer<ObservableDoubleMeasurement>> existingUpdater =
-        doubleUpdaterRegistry.get(descriptor.hashCode());
-    existingUpdater.set(updater);
-    return doubleResult -> {
-      Consumer<ObservableDoubleMeasurement> existing = existingUpdater.get();
-      existing.accept(doubleResult);
-    };
+  private <T extends ObservableMeasurement> T registerCallback(
```
rmfitzpatrick, line 420: This helped a lot with readability.

jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/InstrumentHelper.groovy
dehaansa: Okay. As the batch callback needs to register all of its expected instruments, this might be a little complex, will take a pass at it.

dehaansa: I believe this is improved, let me know!

akats7:
Hey @dehaansa, I tested this change under the same conditions that was causing the rampant heap usage and looks like this fully resolved the issue.

jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/InstrumentHelper.groovy
```diff
@@ -52,140 +56,201 @@ class InstrumentHelper {
...
-                def tuple = new Tuple(instrument, instrumentName, description, unit)
-                logger.fine("Recording ${instrumentName} - ${instrument.method} w/ ${value} - ${labels}")
-                if (!tupleToUpdates.containsKey(tuple)) {
-                    tupleToUpdates[tuple] = []
-                }
-                tupleToUpdates[tuple].add(prepareUpdateClosure(instrument, value, labels))
+            } catch (AttributeNotFoundException ignored) {
+                logger.fine("Attribute ${attribute} not found on any of the collected mbeans")
+            } catch (InvalidAttributeValueException ignored) {
+                logger.info("Attribute ${attribute} was not consistently CompositeData for " +
```
rmfitzpatrick, line 100: I think this may warrant a ~breaking change changelog entry or readme update in the chance that users run into it

rmfitzpatrick (approved):
🥇 thanks!

jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/InstrumentHelper.groovy
dehaansa: I added a note to the readme, from what I can see it looks like changelog entries are added outside the scope of the PR introducing the change. If I'm wrong though, I'm happy to add whatever's required for a changelog entry.

jmx-metrics/README.md
```diff
@@ -128,6 +128,11 @@ mutually exclusive with `otel.jmx.groovy.script`. The currently supported target
     [`CompositeData`](https://docs.oracle.com/javase/7/docs/api/javax/management/openmbean/CompositeData.html) instances,
     each key of their `CompositeType` `keySet` will be `.`-appended to the specified `instrumentName`, whose resulting
     instrument will be updated for each respective value.
+  - If the underlying MBean(s) held by the provided MBeanHelper are a mixed set of
+    [`CompositeData`](https://docs.oracle.com/javase/7/docs/api/javax/management/openmbean/CompositeData.html) instances
+    and simple values, the InstrumentHelper will not attempt to collect the metric as it would generate
+    metrics identified with the `instrumentName` and also the `instrumentName` with the `keySet` `.`-appended,
+    which breaks OpenTelemetry metric conventions.
```
breedx-splk, line 135: Trying to help split up the long run on:

```suggestion
  - If the underlying MBean(s) held by the provided MBeanHelper are a mixed set of
    [`CompositeData`](https://docs.oracle.com/javase/7/docs/api/javax/management/openmbean/CompositeData.html) instances
    and simple values, the InstrumentHelper will not attempt to collect the metric. This is to prevent generating
    metrics identified with the `instrumentName` and also the `instrumentName` with the `keySet` `.`-appended,
    which breaks OpenTelemetry metric conventions.
```

dehaansa: Thanks for that improvement, committed 👍

dehaansa:
@trask I see this was tagged for release, anything left that it needs to be merged?

trask (approved):

PR #957, opened 2023-07-14
disk-buffering refactoring

breedx-splk:
This is a few things I noticed during #913 that I didn't want to necessarily draw out the PR life. They are mostly cosmetic, but the addition of the `DiskExportBuilder` also helps us stick to the creational patterns established in the other otel java repos.

* Test methods can be package access (improve readability)
* Remove work being done in constructors, embrace DI instead
* Add `DiskExportBuilder` and wire it up
* Use `Clock` from core repo and remove `StorageClock`.


LikeTheSalad (approved):
Thanks for your PR! I like the builder approach and the overall cleaning up/simplification.

There's just one thing I'd like to point out, which is not necessarily a showstopper, though I believe it's important to mention it since it might cause not-so-obvious issues later, which is that it would be nice if we somehow could get the time in millis value anywhere without having to write the nanos->millis conversion every time. I mention it because I noticed that this process is repeated a couple of times across the changes. Since it always has to be done anyway, I'd like to avoid having to repeat ourselves in case something has to change in the future or in case we need this value for a future feature and we might forget that we need to apply this conversion every time. On the same note, it's probably better to have some sort of testing utils that take care of the reverse conversion as well, since that's also repeated several times and always has to be done.

Other than that, I don't see any issues with merging this PR. Thanks again!

breedx-splk:
> would be nice if we somehow could get the time in millis value anywhere without having to write the nanos->millis conversion every time

Yeah, I agree that it's slightly annoying. I will PR an addition to see if the core repo is agreeable to having this on the otel clock for uses like this. In the meantime, I introduced a static helper method that hopefully helps mitigate the visual duplication anyway.

trask (approved):

PR #960, opened 2023-07-16
Added transformation closure to MBeanHelper

### LINKED ISSUE #901: Enable ability to overwrite mbean attribute value
### Component(s)

jmx-metrics

We have a use case where we would like to create metrics for string attributes, it makes sense to not provide a string callback, it would be useful to be able to force a value in the instrument helper (just set the value equal to '1') or even better it would be if the InstrumentHelper also accepted a callback/closure that allows you to build logic to set the metric value to a long, double, or counter, when the mbean attribute itself is a different type.

I would like the InstrumentHelper constructor to accept a closure or callback function, that if set will be used to update the metric value based on the logic in the closure

e.g.,

```
def metric = otel.mbeans("RANDOM_MBEAN")
otel.instrument(metric, "metric_based_off_string",
        "description", "by",
        "string-metric",  checkValue ,otel.&longValueCallback)
```
```
def checkValue = { value ->
  if (value == "running") {
    return 1
  } else {
      return 0 
  }
} 
```

_No response_

### Additional context

_No response_

akats7:
This enhancement enables non-numeric mbean attributes to be transformed into numeric values that can be consumed by the instrument callbacks. The MbeanHelper component has been updated to also accept a map of closures for the attributes being targeted. 

Example: 

```
def exampleMBean = otel.mbean(
  "example.mbean", ["CustomAttrFromString": { mbean -> (mbean.getProperty("Attribute") == "running") ? 1 : 0 }]
)
otel.instrument(exampleMBean, "my-metric", "CustomAttrFromString", otel.&longUpDownCounterCallback)
```
**Existing Issue(s):**

Resolves #901 

**Testing:**

Added unit tests to MbeanHelperTest to include transformation scenarios

**Documentation:**

Updated README to include usage instructions


rmfitzpatrick:
Do you think my suggestion of https://github.com/open-telemetry/opentelemetry-java-contrib/issues/901#issuecomment-1582643958 would be more expressive for general use cases? The closures accept an mbean instead of an attribute value.

jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/MBeanHelper.groovy
```diff
@@ -106,7 +125,9 @@ class MBeanHelper {
         return [ofInterest, attributes].combinations().collect { pair ->
             def (bean, attribute) = pair
             try {
-                new Tuple3(bean, attribute, bean.getProperty(attribute))
+                def extractedAttribute = bean.getProperty(attribute)
```
rmfitzpatrick, line 128: I would expect this to throw the AttributeNoteFoundException for custom ones. As is this would require all passed maps to be value-ovewriting processors instead of allowing a general computed value capability.

akats7:
> Do you think my suggestion of [#901 (comment)](https://github.com/open-telemetry/opentelemetry-java-contrib/issues/901#issuecomment-1582643958) would be more expressive for general use cases? The closures accept an mbean instead of an attribute value.

Hey @rmfitzpatrick, that's valid, I can make that adjustment.

jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/MBeanHelper.groovy
akats7: Right, I guest when I first conceptualized this I only thought about it through the lens of transforming existing attributes. Do you think it would be good to allow the instrument helpers to target attributes that don't actually exist on the mbeans?

akats7: @rmfitzpatrick wanted to get your thoughts

akats7:
Hey @rmfitzpatrick, I addressed you comments and just added another test case for custom attributes.

cc. @breedx-splk @dehaansa @Mrod1598

jmx-metrics/src/test/java/io/opentelemetry/contrib/jmxmetrics/MBeanHelperTest.java
```diff
@@ -32,6 +36,7 @@
 @Timeout(value = 10, unit = SECONDS)
 class MBeanHelperTest {
 
+  // private static final Logger logger = Logger.getLogger(MBeanHelperTest.class.getName());
```
breedx-splk, line 39: Remove dead code.

jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/MBeanHelper.groovy
```diff
@@ -87,9 +104,10 @@ class MBeanHelper {
         }
 
         def ofInterest = isSingle ? [mbeans[0]]: mbeans
+
         return ofInterest.collect {
             try {
-                it.getProperty(attribute)
+                attributeTransformation.containsKey(attribute) ? attributeTransformation[attribute](it) : it.getProperty(attribute)
```
breedx-splk, line 110: Small point -- I'm not sure about groovy, but in java calling `containsKey()` immediately followed by `get()` could (in early versions at least) incur two lookups. It's probably fine, and the size of the number of beans in our maps here is so small to render this moot....however, if you did factor out a method at least you could avoid the code duplication below. Something like `getPropertyValue(bean, attribute)`

breedx-splk (approved):
I like it. I think this is a worthy change that we can continue improving on if needed. Thanks!

jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/MBeanHelper.groovy
akats7: Hey @breedx-splk, yep I refactored this when resolving the conflict with the callback change

jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/MBeanHelper.groovy
```diff
@@ -100,13 +117,18 @@ class MBeanHelper {
         def ofInterest = isSingle ? [mbeans[0]]: mbeans
         return [ofInterest, attributes].combinations().collect { pair ->
             def (bean, attribute) = pair
-            new Tuple3(bean, attribute, getBeanAttribute(bean, attribute))
+            new Tuple3(bean, attribute, getBeanAttributeTransform(bean, attribute))
         }
     }
 
+    Object getBeanAttributeTransform(GroovyMBean bean, String attribute){
```
dehaansa, line 124: Should this be `getBeanAttribute*or*Transform`?

dehaansa (approved):
Tiny naming nit, otherwise looks great.

jmx-metrics/src/test/java/io/opentelemetry/contrib/jmxmetrics/MBeanHelperTest.java
```diff
@@ -116,6 +119,77 @@ void multiple() throws Exception {
...
+    Map<String, Closure<?>> map =
+        Stream.of(
+                new Object[][] {
+                  {
+                    "SomeAttribute",
+                    Eval.me(
+                        "{mbean -> mbean.getProperty(\"SomeAttribute\") == 'someValue' ? 'newValue' : 'someValue'}")
+                  },
+                  {
+                    "AnotherAttribute",
```
rmfitzpatrick, line 159: Do you mind updating this to try to get a nonexistent property for error handling vetting? I think it repeats existing added coverage.

rmfitzpatrick (approved):

jmx-metrics/src/test/java/io/opentelemetry/contrib/jmxmetrics/MBeanHelperTest.java
akats7: In its current state I think that'll throw an AttributeNotFound exception, do we want to handle that error in the same manner as if a nonexistent property was passed to an instrument helper? I just followed the pattern of labelFuncs and the other closure params where it doesn't look like theres any additional error handling

rmfitzpatrick: I think the same error behavior makes sense and it could be helpful to add coverage to better enforce that moving forward.

trask (approved):

PR #965, opened 2023-07-20
implement kafka-exporter

naser-ayat:
**Description:**

The `kafka-exporter` module is added. This module contains `KafkaSpanExporter` (an implementation of the `SpanExporter` interface) that can be used for sending `SpanData` to Kafka.

**Existing Issue(s):**
Solves #145 

**Testing:**

The module contains both unit tests and an integration test.

**Documentation:**

A README file documents the usage


trask (commented):
hi @naser-ayat!

are you willing to be a component owner for this going forward?

ideally we would like to have two component owners, maybe try posting in the `#otel-java` slack channel to see if anyone else is interested as well?

kafka-exporter/src/main/java/io/opentelemetry/contrib/kafka/KafkaSpanExporter.java
```diff
@@ -0,0 +1,188 @@
...
+      Producer<String, Collection<SpanData>> producer,
+      ExecutorService executorService,
+      long timeoutInSeconds) {
+    this.topicName = topicName;
+    this.producer = producer;
+    this.executorService = executorService;
+    this.timeoutInSeconds = timeoutInSeconds;
+  }
+
+  public KafkaSpanExporter(
```
vincentfree, line 54: Instead of adding more constructors you might want to opt in for using `static factory methods` or a builder like pattern. I would think static factory methods make the code much clearer. You could use some default names like `create` to replace the main constructor, and I've used things like basic, and with timeout etc etc as other options.

This gives the replacements names and makes it easier to work with.
But looking at all of the options you have a Builder might be a good option as well. probably with some default/required fields to start with

kafka-exporter/src/main/java/io/opentelemetry/contrib/kafka/SpanDataDeserializer.java
```diff
@@ -0,0 +1,26 @@
...
+import io.opentelemetry.proto.collector.trace.v1.ExportTraceServiceRequest;
+import java.util.Objects;
+import org.apache.kafka.common.errors.SerializationException;
+import org.apache.kafka.common.serialization.Deserializer;
+
+public class SpanDataDeserializer implements Deserializer<ExportTraceServiceRequest> {
+  @Override
+  public ExportTraceServiceRequest deserialize(String topic, byte[] data) {
+    if (Objects.isNull(data)) {
+      return ExportTraceServiceRequest.getDefaultInstance();
```
vincentfree, line 18: Would you return an empty event? I think if it's not a valid response it should throw. Empty events hide the fact that nothing worthy of processing can be returned

naser-ayat:
> 

We're currently discussing this at ING and will provide an update soon. Additionally, I've sent a message to the Slack channel to see if anyone else is interested.

naser-ayat:
> hi @naser-ayat!
> 
> are you willing to be a component owner for this going forward?
> 
> ideally we would like to have two component owners, maybe try posting in the `#otel-java` slack channel to see if anyone else is interested as well?

kafka-exporter/src/main/java/io/opentelemetry/contrib/kafka/SpanDataDeserializer.java
naser-ayat: In the Deserialiser interface documentation, it states that the 'data' parameter, which represents serialized bytes, may be null. Implementations are advised to handle null by either returning a value or null instead of throwing an exception. That's why the exception is not thrown. I personally opted for returning an empty object to simplify dealing with null values in the callers. If you prefer to receive null values, please let me know.

kafka-exporter/src/main/java/io/opentelemetry/contrib/kafka/KafkaSpanExporter.java
naser-ayat: A builder is added.

kafka-exporter/src/main/java/io/opentelemetry/contrib/kafka/SpanDataDeserializer.java
vincentfree: Returning a empty value creates a valid return value that in turn needs to be validated at the client side. This hides the fact that the the incoming event could not be deserialized. 

In other words it transforms the result to a successful event with no content. Default null handling of client side libraries might be a better solution here. I understand that null doesn't seem like the best return value but based what you said about the docs this is the better return type here imo

spockz:
> > hi @naser-ayat!
> > are you willing to be a component owner for this going forward?
> > ideally we would like to have two component owners, maybe try posting in the `#otel-java` slack channel to see if anyone else is interested as well?
> 
> We're currently discussing this at ING and will provide an update soon. Additionally, I've sent a message to the Slack channel to see if anyone else is interested.

I'm interested in being a component owner.

.github/component_owners.yml
```diff
@@ -58,3 +58,6 @@ components:
     - trask
   static-instrumenter:
     - anosek-an
+  kafka-exporter:
+    - spockz
```
spockz, line 62: Approved.

.github/component_owners.yml
```diff
@@ -58,3 +58,6 @@ components:
...
+    - vincentfree
```
trask, line 63: @naser-ayat did you want to be a component owner as well?

naser-ayat: Two component owners should be enough. Thanks for the suggestion.

naser-ayat: @vincentfree Could you please approve the PR?

trask: it would be great to get approvals from both of the component owners

vincentfree: Approved on my part 👍

trask: @vincentfree can you approve via github? thx!

image

trask: @spockz ^^ can you approve via github also?

vincentfree (approved):

.github/component_owners.yml
vincentfree: Done, sorry was going to do an actual review but I did that before so that's fine now. finally came around to click that button :)

spockz (approved):

trask (approved):

PR #986, opened 2023-07-28
[disk-buffering] Migrate StorageFile to FileOperations

breedx-splk:
I think it's hard to justify imposing a class hierarchy for a single instance field. This instead prefers an interface and eliminates the direct field access.


LikeTheSalad (approved):
Thank you. LGTM

trask (approved):

PR #1008, opened 2023-08-21
Removing protobuf dependency

LikeTheSalad:
**Description:**

Bug fix  - Since this module has a dependency on [OTel Java proto](https://github.com/open-telemetry/opentelemetry-proto-java) it means that it also gets [Protobuf's](https://protobuf.dev/) java lib as a transitive dependency. The protobuf-java library is not ideal for Android projects due to its size which is why there's another version of it, called [protobuf-javalite](https://github.com/protocolbuffers/protobuf/blob/main/java/lite.md), that aims to be more suitable for Android use cases due to its smaller size.

Given that the `protobuf-javalite` dependency is better for Android projects, some projects might have it included already which causes a problem if they later add the `disk-buffering` dependency as well, since it comes with the "regular protobuf dependency". This is an issue because both, the regular and lite versions of the protobuf libs, contain classes that have the exact same fully qualified names, which makes Gradle raise a compilation error due to the "duplicated classes" it finds across different dependencies.

This was discussed in the latest Java SIG meeting where a couple of interesting ideas were proposed to avoid this issue such as:

* Shadowing the regular protobuf lib's classes to prevent them from clashing with the lite ones later. The problem with this option is the final lib size, as well as the added size across both, the lite version and the "regular" version of the protobuf libraries that would get added into an app that already has the lite version as a dependency.
* Adding instructions on how to shadow these classes on the consumer project. This would make it a bit complicated to integrate the `disk-buffering` dependency into a project that already has the protobuf-javalite dependency. Ideally, I'd like this lib to be easy to use.
* It was briefly mentioned that we could also make the [OTel Java proto](https://github.com/open-telemetry/opentelemetry-proto-java) artifacts to use the lite version only. Though my concern with that approach would be that, in case a project has the "protobuf-java regular" version already as a dependency, then the same kind of issue would happen but the other way around.

The approach I'm proposing to go with in this PR is to not use any of the protobuf dependencies, neither the "regular" nor the "lite" one, in `disk-buffering`. Instead, we can use [Wire](https://square.github.io/wire/), which is a third-party tool (already being used in other [OTel Java projects](https://github.com/open-telemetry/opentelemetry-java/blob/c0f0ff8bca099c79010b80e4e9981f89f08b10dd/exporters/otlp/common/build.gradle.kts#L8)) that takes care of generating the Java sources from the proto files without having any dependency on Google's protobuf libraries. As a bonus side of using this tool, we get smaller generated sources since Wire makes them a bit more android friendly by reducing the number of methods for each generated class.

**Existing Issue(s):**

#1009

**Testing:**

Unit, integration, manual tests.

**Documentation:**

No documentation changes are needed as part of this PR since the changes are only internal.

**Outstanding items:**


zeitlinger:
are all of those serializer classes hand-written?

LikeTheSalad:
> are all of those serializer classes hand-written?

Yes. They used to be auto-generated using MapStruct, however, that caused the PR to be bigger since MapStruct couldn't reuse the existing builders/creators for all the implementations of the OTel API in the Java SDK.

The changes made to all of them here are minimal though, it's mostly accessing fields directly instead of getters (because Wire doesn't create those to reduce the number of methods), and also renaming some setters to the style added by Wire.

zeitlinger (approved):

mateuszrzeszutek (approved):

trask:
@LikeTheSalad no need to add change log in this repo, just update the PR title to what you would like displayed and we'll pull it in semi-automatically before making the next release

PR #1014, opened 2023-08-23
Signals processing

LikeTheSalad:
**Description:**

Feature addition - Tools for globally processing signal data items in order to allow for:
- Modification
- Filtering

As part of the [Splunk Android SDK donation](https://github.com/open-telemetry/community/issues/1400), some of the items requested from the community were related to moving generic tools used in the Splunk code over to the java-contrib repo. Some of these tools enabled [filtering spans](https://github.com/open-telemetry/opentelemetry-android/blob/main/instrumentation/src/main/java/io/opentelemetry/android/export/FilteringSpanExporterBuilder.java) based on predefined predicates as well as [adding attributes](https://github.com/open-telemetry/opentelemetry-android/blob/c911c819f2922657eb85a178c90f8d2318f6cadb/instrumentation/src/main/java/io/opentelemetry/android/GlobalAttributesSpanAppender.java#L15) to them.

The proposed tools in this PR aren't exact replicas of the classes used in the Splunk SDK although they aim to enable the same use cases (signal modification and filtering) not only for spans but also for logs and metrics in a quite generic way that should allow for more use cases to be supported, albeit in a less verbose way, but flexible enough to be further extended to add more specific use cases on top, such as "rejectSpansNamed()" and "rejectSpansWithAttributesMatching()" for example, as some of the ones currently available [in here](https://github.com/open-telemetry/opentelemetry-android/blob/main/instrumentation/src/main/java/io/opentelemetry/android/export/FilteringSpanExporterBuilder.java).

There are no defined guidelines on how to accomplish this functionality, so what I'm proposing here is just a way to do so that I consider simple and flexible enough to accommodate different use cases. So I'm open to any alternatives or adjustments that could improve this feature in any way!

**How it works**

There's an [Interceptor](https://github.com/LikeTheSalad/opentelemetry-java-contrib/blob/signals-processing/processors/src/main/java/io/opentelemetry/contrib/interceptor/api/Interceptor.java) interface that can modify or eliminate a signal right before it gets exported. This example shows how to modify and/or filter a span out from exporting:

```java
    InMemorySpanExporter myExporter = InMemorySpanExporter.create();
    ComposableInterceptor interceptor = new ComposableInterceptor();
    Tracer tracer = createTracerWithExporter(new InterceptableSpanExporter(myExporter, interceptor));

    // This interceptor will modify all the spans by adding a new attr.
    interceptor.add(
        spanData -> {
          ModifiableSpanData modified = new ModifiableSpanData(spanData);
          modified.attributes.put("global.attr", "from interceptor");
          return modified;
        });

    // This interceptor will prevent some spans from getting exported:
    interceptor.add(
        spanData -> {
          if (spanData.getName().endsWith("deleted")) {
            return null;
          }
          return spanData;
        });

     // All the spans will go through both interceptors when getting exported
    tracer.spanBuilder("My Span").startSpan()//etc...
```

You can find more examples in the [tests](https://github.com/LikeTheSalad/opentelemetry-java-contrib/tree/signals-processing/processors/src/test/java/io/opentelemetry/contrib/interceptor) added as part of this PR.

**Existing Issue(s):**

**Testing:**

Unit tests

**Documentation:**

**Outstanding items:**


.github/component_owners.yml
```diff
@@ -47,6 +47,9 @@ components:
     - HaloFour
   noop-api:
     - jack-berg
+  processors:
+    - LikeTheSalad
+    - ?
```
breedx-splk, line 52: You can add me here. I think the literal `?` breaks things.

processors/src/main/java/io/opentelemetry/contrib/interceptor/common/Interceptable.java
```diff
@@ -0,0 +1,46 @@
...
+package io.opentelemetry.contrib.interceptor.common;
+
+import io.opentelemetry.contrib.interceptor.api.Interceptor;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+/** Base class to reuse the code related to intercepting signals. */
```
breedx-splk, line 15: I strongly believe that we should not use object inheritance as a mechanism for reusing code.

breedx-splk (changes requested):
Let's use composition here instead. Recommend:

* Remove `Interceptor` interface.
* Rename `Interceptable` to `Interceptor` and make it `final`
* Remove `extends ....` from each of the exporters.
* Add constructor arg and field of type `Interceptor` to each exporter
* Call `interceptor.interceptAll()` in the export method.

processors/src/main/java/io/opentelemetry/contrib/interceptor/common/Interceptable.java
```diff
@@ -0,0 +1,46 @@
...
+public class Interceptable<T> {
+  private final Set<Interceptor<T>> interceptors = new HashSet<>();
```
breedx-splk, line 17: I could think of a few contrived scenarios where order might matter. I'd use a list.

LikeTheSalad:
> Let's use composition here instead. Recommend:
> 
> Rename Interceptable to MultiInterceptor or something and make it final It might also implement Interceptor<Collection<T>>
> Remove extends .... from each of the exporters.
> Add constructor arg and field of type MultiInterceptor (or the interface) to each exporter
> Call interceptor.interceptAll() in the export method.

Sounds good, I've applied the changes you mentioned with some small diffs, for the multi interceptor name I went with "ComposableInterceptor" and for the exporters' constructor arg for an interceptor I went with the interface `Interceptor` instead of any implementation to allow users to use any kind of implementation they'd like. Also, the `interceptAll` method was made a default Interceptor method since it seems appropriate as a reusable functionality regardless of the implementation.

processors/src/main/java/io/opentelemetry/contrib/interceptor/api/Interceptor.java
```diff
@@ -0,0 +1,41 @@
...
+
+  /**
+   * Intercepts a signal.
+   *
+   * @param item The signal object.
+   * @return The received signal modified (or null for excluding this signal from getting exported).
+   *     If there's no operation needed to be done for a specific signal, it should be returned as
+   *     is.
+   */
+  T intercept(T item);
```
mateuszrzeszutek, line 26:
```suggestion
  @Nullable
  T intercept(T item);
```

processors/src/main/java/io/opentelemetry/contrib/interceptor/common/ComposableInterceptor.java
```diff
@@ -0,0 +1,33 @@
...
+
+package io.opentelemetry.contrib.interceptor.common;
+
+import io.opentelemetry.contrib.interceptor.api.Interceptor;
+import java.util.ArrayList;
+import java.util.List;
+
+/** Allows to run an item through a list of interceptors in the order they were added. */
+public final class ComposableInterceptor<T> implements Interceptor<T> {
+  private final List<Interceptor<T>> interceptors = new ArrayList<>();
```
mateuszrzeszutek, line 14: Typically, when using the `BatchSpanProcessor` sending the telemetry data happens on a separate thread; consider using `CopyOnWriteArrayList` instead
```suggestion
  private final List<Interceptor<T>> interceptors = new CopyOnWriteArrayList<>();
```

processors/src/main/java/io/opentelemetry/contrib/interceptor/common/ComposableInterceptor.java
```diff
@@ -0,0 +1,33 @@
...
+
+  public void add(Interceptor<T> interceptor) {
+    if (!interceptors.contains(interceptor)) {
+      interceptors.add(interceptor);
+    }
```
mateuszrzeszutek, line 19: Assuming `interceptors` is an instance of `CopyOnWriteArrayList`:
```suggestion
    interceptors.addIfAbsent(interceptor);
```

processors/src/main/java/io/opentelemetry/contrib/interceptor/InterceptableLogRecordExporter.java
```diff
@@ -0,0 +1,39 @@
...
+package io.opentelemetry.contrib.interceptor;
+
+import io.opentelemetry.contrib.interceptor.api.Interceptor;
+import io.opentelemetry.sdk.common.CompletableResultCode;
+import io.opentelemetry.sdk.logs.data.LogRecordData;
+import io.opentelemetry.sdk.logs.export.LogRecordExporter;
+import java.util.Collection;
+
+/** Intercepts logs before delegating them to the real exporter. */
+public class InterceptableLogRecordExporter implements LogRecordExporter {
```
mateuszrzeszutek, line 15: Nit: all of these exporters can be `final`
```suggestion
public final class InterceptableLogRecordExporter implements LogRecordExporter {
```

processors/src/test/java/io/opentelemetry/contrib/interceptor/InterceptableLogRecordExporterTest.java
```diff
@@ -0,0 +1,153 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class InterceptableLogRecordExporterTest {
+  private InMemoryLogRecordExporter memoryLogRecordExporter;
+  private Logger logger;
+  private ComposableInterceptor<LogRecordData> interceptor;
+
+  @BeforeEach
+  public void setUp() {
```
mateuszrzeszutek, line 35: Nit: JUnit 5 tests & lifecycle methods can be package-private
```suggestion
  void setUp() {
```

LikeTheSalad:
Thanks @mateuszrzeszutek ! I've applied your suggestions.

mateuszrzeszutek (approved):
👍

trask:
@breedx-splk could you give this another look? thx

breedx-splk (approved):

PR #1028, opened 2023-09-05
Add some libraries missing from the main README

breedx-splk:
Top-level README was missing some of the newer libraries. This adds links to them.


trask (approved):
thx!

PR #1032, opened 2023-09-08
Add xray propagators that prioritizes xray environment variable

tylerbenson:
This will allow support for AWS's environment variable with no additional config keys and no changes to instrumentation required.

I included a propagator that did just the env var propagation without the standard xray propagation as well as one that does both but prioritizes xray env propagation first.


rapphil:
This is related to https://github.com/open-telemetry/semantic-conventions/pull/277

No decision was made yet.

tylerbenson:
This PR has been updated to match the [latest spec update](https://github.com/open-telemetry/semantic-conventions/pull/354) and is now ready for review.

tylerbenson:
@srprash @wangzlei please take a look.

rapphil (approved):

aws-xray-propagator/src/main/java/io/opentelemetry/contrib/awsxray/propagator/AwsXrayLambdaPropagator.java
```diff
@@ -0,0 +1,100 @@
...
+ *
+ * <p>To register the X-Ray propagator together with default propagator when using the SDK:
+ *
+ * <pre>{@code
+ * OpenTelemetrySdk.builder()
+ *   .setPropagators(
+ *     ContextPropagators.create(
+ *         TextMapPropagator.composite(
+ *             W3CTraceContextPropagator.getInstance(),
+ *             AwsXrayEnvPropagator.getInstance())))
```
wangzlei, line 32: AwsXrayLambdaPropagator

aws-xray-propagator/src/main/java/io/opentelemetry/contrib/awsxray/propagator/AwsXrayLambdaPropagator.java
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  @Override
+  public List<String> fields() {
+    return xrayPropagator.fields();
+  }
+
+  @Override
+  public <C> void inject(Context context, @Nullable C carrier, TextMapSetter<C> setter) {
+    xrayPropagator.inject(context, carrier, setter);
```
wangzlei, line 58: Does AwsXrayLambdaPropagator need inject() implementation?

aws-xray-propagator/src/main/java/io/opentelemetry/contrib/awsxray/propagator/AwsXrayLambdaPropagator.java
```diff
@@ -0,0 +1,100 @@
...
+
+    String traceHeader = System.getProperty(AWS_TRACE_HEADER_PROP);
+    if (isEmptyOrNull(traceHeader)) {
+      traceHeader = System.getenv(AWS_TRACE_HEADER_ENV_KEY);
+    }
+    if (isEmptyOrNull(traceHeader)) {
+      return xrayContext;
+    }
+    return xrayPropagator.extract(
+        xrayContext,
```
wangzlei, line 77: Considering 2 things: 
1. AWS Lambda supports w3c in the future
We don't know how Lambda support w3c, but one possible is using new env variables for w3c trace context and baggage. We can think about rename AwsXrayLambdaPropagator to be more generic `AwsLambdaPropagator`.
2. decoupling AWS Lambda propagator and AWS xray propagator
The current AWSXrayLambdaPropagator combines both Lambda propagator and XRayPropagator, can we think about only handle Lambda environment variable or system property? Then let customer set "propagator=awslambda, xray" if needed. The AwsLambdaPropagator inject() can be a No-Op method.

aws-xray-propagator/src/main/java/io/opentelemetry/contrib/awsxray/propagator/AwsXrayLambdaConfigurablePropagator.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+public final class AwsXrayLambdaConfigurablePropagator implements ConfigurablePropagatorProvider {
+  @Override
+  public TextMapPropagator getPropagator(ConfigProperties config) {
+    return AwsXrayLambdaPropagator.getInstance();
+  }
+
+  @Override
+  public String getName() {
+    return "xray-lambda";
```
wangzlei, line 24: Seems it is from spec, we may want to update spec to be `awslambda` or `aws-lambda`.

aws-xray-propagator/src/main/java/io/opentelemetry/contrib/awsxray/propagator/AwsXrayLambdaPropagator.java
tylerbenson: 1. When you add support for tracecontext/baggage, I suggest following a similar pattern and creating a `AwsW3CTraceContextLambdaPropagator`/`AwsW3CBaggageLambdaPropagator`.
2. This was originally the plan, but I ran into a major problem when implementing.  The fix was to combine the propagators.  See https://github.com/open-telemetry/semantic-conventions/pull/778 for more detail.

aws-xray-propagator/src/main/java/io/opentelemetry/contrib/awsxray/propagator/AwsXrayLambdaConfigurablePropagator.java
tylerbenson: If we implement support for w3c like I suggested above, using your naming suggestion would not allow a distinction between tracecontext-lambda vs tracecontext, and would also not allow users to choose just xray or just tracecontext.

wangzlei: Correct me if I am wrong, the current solution for w3c in Lambda works for API GW -> Lambda case, only if the http request has w3c trace headers?

aws-xray-propagator/src/main/java/io/opentelemetry/contrib/awsxray/propagator/AwsXrayLambdaPropagator.java
wangzlei: Thanks for your great help for Lambda support. Would you help me understand the `problem when implementing.`?
Your original plan sounds reasonable that if user prefer Lambda env variable, uses propagator order "lambda, w3c, xray". If user prefer http header, uses propagator in order "w3c, xray, lambda". It means "lambda" propagator is for AWS Lambda specific environment and no need handle injection.

tylerbenson: > problem when implementing

My initial POC implemented the original spec which separated the propagators as you suggested.  After adding the clause mentioned in that PR, the spec became broken which I didn't discover until I had time to implement the final spec.

If you need more detail of how it was broken, please review the description in https://github.com/open-telemetry/semantic-conventions/pull/778 and let me know if that's unclear.

wangzlei: Hi, @tylerbenson 
If customer's application is: client-> API GW -> Lambda -> S3, both API GW and Lambda are Active tracing enabled. Could you help draw the expected trace graph? I am trying to see if the Lambda spec make sense to customer.

wangzlei: The solution seems reasonable. Other than this PR, does [here](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/aws-lambda/aws-lambda-core-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambdacore/v1_0/internal/AwsLambdaFunctionInstrumenterFactory.java#L26) need modification if user uses xray-lambda propagator?

Additionally, there are two other matters:
1. The current AWS services do not support the W3C trace header. If AWS service Active tracing is enabled, there will be inconsistency between two headers, which could be considered conflicting if the customer uses both the xray-lambda propagator and Active tracing mode. To avoid this conflict, can we say:
   * If the user wants to use X-Ray, the propagator should be set to "xray-lambda" (without w3c).
   * If the user does not use X-Ray, the propagator can be set to "w3c, xray". It's even uncertain if the xray propagator has any significance here. Perhaps only w3c is necessary?

2. If the user does not use the xray service but chooses to prioritize the w3c propagator, it will block the Lambda Lineage feature. This is not related to the X-Ray service; it's because this feature adds a lineage field to the xray header at the Lambda FrontEnd (Service) level. If the propagation from the environment variable is skipped, Lambda Lineage will not work. This feature can be extended to w3c baggage in the future when AWS Lambda supports w3c.

Both of the above points are due to the fact that AWS managed services currently do not support the W3C trace header. As far as I understand, but not guaranteed, the AWS w3c trace context support solution would be to adopt a similar approach to the OTel propagator list. During extraction, select one based on priority, and during injection, inject multiple formats with consistent content to ensure the above two issues do not occur. Lambda should have a built-in propagation mechanism, where regardless of whether the user adopts Active tracing mode or passthrough mode, and regardless of what format of trace header is included in the incoming request, the Lambda FrontEnd propagates the correct trace context to the user Lambda handler function through environment variables. The difference between Passthrough mode and Active mode is that the former is the original trace context from the incoming request, while the latter is the trace context from the Lambda-vended span. In both cases, the same Lineage information is included.

Looking forward to hear your opinion. Appreciate!

tylerbenson: Yes, once this PR is merged, I'll work on the changes required in the instrumentation repo.

The propagators list is ordered, with the last one winning.  By specifying the xray/xray-lambda propagator last, it takes priority over traceparent. If the user decides to prioritize traceparent, that's up to them.

Interesting point about the recursive loop detection. Perhaps this should be documented on the OTel side somewhere... @rapphil WDYT?  I assume that it will still work if we encourage people to prioritize xray/xray-lambda over the w3c headers, right?

I'm having a hard time following that last part.  Perhaps it would be easier to discuss in the SIG meeting?

Is there anything else remaining blocking this PR?  I'm happy to continue discussing, but I'd like to get this merged so I can work on the changes needed in instrumentation.

wangzlei: @tylerbenson 
Does [here](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/aws-lambda/aws-lambda-core-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambdacore/v1_0/internal/) need modification if user uses xray-lambda propagator? Else the Lambda vended span is both parent and link of OTel span?

The last part please refer to [OTEL_SDK_DISABLED](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#general-sdk-configuration). If user does not enable active tracing, the trace context can pass though the AWS service.

wangzlei: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/aws-lambda/aws-lambda-core-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambdacore/v1_0/internal/AwsLambdaFunctionInstrumenterFactory.java#L26

tylerbenson: I was planning on removing that line since the span linking is no longer required by the spec. Are you expecting something different?

I don't understand how [OTEL_SDK_DISABLED](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#general-sdk-configuration) is relevant here. Are you suggesting this propagator interferes with that?

wangzlei: AWS services would like to support the similar functionality as [OTEL_SDK_DISABLED](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#general-sdk-configuration) . Take the example of Lambda, if user does not enable Active tracing, the environment variable would have the same trace context with from API GW request. It means OTel in Lambda only need xxxx-Lambda propagator in the future.

tylerbenson: > If Lambda is root, it adds a trace context with sampled flag 0.

Why would you/we want this?

tylerbenson:
@wangzlei I don't think anything in our discussion above represents a blocker for merging this PR. Is there anything that I'm missing?

If you'd like to continue discussing how this should all work, we should probably include the rest of the SIG instead of in this repo.  Please open an issue in the `opentelemetry-lambda` repo or come attend the SIG meeting.

aws-xray-propagator/src/main/java/io/opentelemetry/contrib/awsxray/propagator/AwsXrayLambdaPropagator.java
wangzlei: > If Lambda is root, it adds a trace context with sampled flag 0.

It is the existing behavior, would be update to be same as [OTEL_SDK_DISABLED](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#general-sdk-configuration) does.

laurit (approved):

aws-xray-propagator/src/main/java/io/opentelemetry/contrib/awsxray/propagator/AwsXrayLambdaPropagator.java
tylerbenson: @wangzlei FYI: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/10930

Issue #1044, opened 2023-09-17
Update maven-extension to latest semconv


trask:
### Component(s)

maven-extension

### Is your feature request related to a problem? Please describe.

https://github.com/open-telemetry/opentelemetry-java-contrib/blob/811d8d7e1cdcefb03ed2a20ae09312aa16a3f33a/maven-extension/build.gradle.kts#L64-L69

### Describe the solution you'd like

Update maven-extension to latest semconv

cyrille-leclerc:
Fixed in:
* https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1298

PR #1059, opened 2023-09-29
Adopt new `io.opentelemetry.semconv:opentelemetry-semconv` lib

cyrille-leclerc:
**Description:**

Adopt new `io.opentelemetry.semconv:opentelemetry-semconv` lib without yet replacing `http.url` and `http.method` by `url.full` and `http.request.method` as we will want to align with the OTel Java Auto Instrumentation and support `-Dotel.semconv-stability.opt-in=http...` to embrace the new semantic conventions

**Existing Issue(s):**

* https://github.com/open-telemetry/opentelemetry-java-contrib/issues/1044

**Testing:**

None as there are no change in the behavior of the instrumentation

**Documentation:**

None as there are no change in the behavior of the instrumentation

**Outstanding items:**

Not really outstanding but we will have to add support for the new HTTP Semantic Conventions, `url.full` and `http.request.method` and we want to do it the same way the Otel Java Auto Instr implementing `-Dotel.semconv-stability.opt-in=http...` as described in https://github.com/open-telemetry/opentelemetry-java-instrumentation/releases/tag/v1.27.0

### DIFF
--- a/maven-extension/src/main/java/io/opentelemetry/maven/handler/GoogleJibBuildHandler.java
+++ b/maven-extension/src/main/java/io/opentelemetry/maven/handler/GoogleJibBuildHandler.java
@@ -9,7 +9,7 @@ import io.opentelemetry.api.trace.SpanBuilder;
 import io.opentelemetry.api.trace.SpanKind;
 import io.opentelemetry.maven.MavenGoal;
 import io.opentelemetry.maven.semconv.MavenOtelSemanticAttributes;
-import io.opentelemetry.semconv.trace.attributes.SemanticAttributes;
+import io.opentelemetry.semconv.SemanticAttributes;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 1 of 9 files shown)


kenfinnigan (approved):

maven-extension/src/main/java/io/opentelemetry/maven/handler/GoogleJibBuildHandler.java
```diff
@@ -29,6 +32,7 @@ public List<MavenGoal> getSupportedGoals() {
         MavenGoal.create("com.google.cloud.tools", "jib-maven-plugin", "build"));
   }
 
+  @SuppressWarnings("deprecation") // until old http semconv are dropped in 2.0
```
trask, line 35: @mateuszrzeszutek @jack-berg @laurit we haven't discussed bumping the contrib repo to 2.0 when we bump the instrumentation repo, but it could make sense to keep the -instrumentation and -contrib repos version aligned

@srprash @wangzlei would you want to follow this approach instead of the clean break that was merged in #1050? (we have time before the next release to revert that change if you want)

SylvainJuge:
Hi @cyrille-leclerc do you think there are still changes in this PR that are relevant to include ?

Some of the changes were included in https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1299 and there are now a few conflicts, so it might need a bit of time (and work) to update and fix it.

PR #1061, opened 2023-09-29
Replace SimpleHttpClient (okhttp based) by JdkHttpClient (HttpURLConnection based)

rjbaucells:
**Description:**

Replaced existing `okhttp` based HttpClient (`SimpleHttpClient`) by a JDK 1.8 implementation using `HttpURLConnection`. `okhttp` adds too many dependencies to a project that performs simple HTTP calls to discover resource metadata in AWS ECS and EKS. The AWS endpoints supports HTTP/1.1 and there is no need to higher protocol version (HTTP/2).

See the `okhttp` dependency tree in a Java project:

image

Is is well known that having `okhttp` as a dependency makes very difficult (or impossible) to create a Graal native image of a consuming application.

Also, `okio` has known [vulnerabilities](https://nvd.nist.gov/vuln/detail/CVE-2023-3635)

OpenTelemetry OTLP exporter supports excluding `okhttp` based implementation in favor of a JDK-11 implementation via ServiceManager. I do not think the same solution is needed here since the HTTP client requirements for this project are very simple and not a high throughput implementation is required for a single HTTP request in the application lifespan.

Initial PR does not remove the `okhttp` implementation nor the dependency (made optional). An update is required before merging it to completely remove `okhttp`.

### DIFF
--- a/aws-resources/build.gradle.kts
+++ b/aws-resources/build.gradle.kts
@@ -16,7 +16,7 @@ dependencies {
   compileOnly("io.opentelemetry:opentelemetry-sdk-extension-autoconfigure")
 
   implementation("com.fasterxml.jackson.core:jackson-core")
-  implementation("com.squareup.okhttp3:okhttp")
+  compileOnly("com.squareup.okhttp3:okhttp")
 
   testImplementation("io.opentelemetry:opentelemetry-sdk-extension-autoconfigure")
   testImplementation("io.opentelemetry:opentelemetry-sdk-testing")
--- a/aws-resources/src/main/java/io/opentelemetry/contrib/aws/resource/Ec2Resource.java
+++ b/aws-resources/src/main/java/io/opentelemetry/contrib/aws/resource/Ec2Resource.java
@@ -138,7 +138,7 @@ public final class Ec2Resource {
 
   // Generic HTTP fetch function for IMDS.
   private static String fetchString(String httpMethod, URL url, String token, boolean includeTtl) {
-    SimpleHttpClient client = new SimpleHttpClient();
+    JdkHttpClient client = new JdkHttpClient();
     Map<String, String> headers = new HashMap<>();
 
     if (includeTtl) {
--- a/aws-resources/src/main/java/io/opentelemetry/contrib/aws/resource/EcsResource.java
+++ b/aws-resources/src/main/java/io/opentelemetry/contrib/aws/resource/EcsResource.java
@@ -44,11 +44,11 @@ public final class EcsResource {
   }
 
   private static Resource buildResource() {
-    return buildResource(System.getenv(), new SimpleHttpClient());
+    return buildResource(System.getenv(), new JdkHttpClient());
   }
 
   // Visible for testing
-  static Resource buildResource(Map<String, String> sysEnv, SimpleHttpClient httpClient) {
+  static Resource buildResource(Map<String, String> sysEnv, JdkHttpClient httpClient) {
     // Note: If V4 is set V3 is set as well, so check V4 first.
     String ecsMetadataUrl =
         sysEnv.getOrDefault(ECS_METADATA_KEY_V4, sysEnv.getOrDefault(ECS_METADATA_KEY_V3, ""));
@@ -65,7 +65,7 @@ public final class EcsResource {
   }
 
   static void fetchMetadata(
-      SimpleHttpClient httpClient, String url, AttributesBuilder attrBuilders) {
+      JdkHttpClient httpClient, String url, AttributesBuilder attrBuilders) {
     String json = httpClient.fetchString("GET", url, Collections.emptyMap(), null);
     if (json.isEmpty()) {
       return;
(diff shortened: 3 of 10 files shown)


trask:
check out https://github.com/open-telemetry/opentelemetry-java/pull/3991, in particular:

> notably this will prevent resource detection requests from being traced during agent startup due to https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/4642

cc @LikeTheSalad who is investigating options for suppress tracing of specific calls

jackshirazi:
+1 on eliminating this dependency for this specific case

trask:
there's work in-progress which could unblock this: https://github.com/open-telemetry/opentelemetry-java/pull/5918

jackshirazi:
I think that's to avoid having the call traced, whereas this PR is about removing the okhttp dependency completely for this specific resource?

trask:
the reverse of this PR (moving from HttpURLConnection to OkHttp) was done a while ago in https://github.com/open-telemetry/opentelemetry-java/pull/3991 (when this code lived in that repository)

I think really it's only a problem for the AWS distribution of the OTel Java agent, since that distribution pulls this component in, while the upstream Java agent doesn't ([yet](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/7927)).

The reason why OkHttp calls aren't traced in the Java agent is because OkHttp is loaded in the agent class loader, and instrumentation isn't applied in that class loader. While HttpURLConnection is loaded in the bootstrap class loader and so instrumentation is applied.

SylvainJuge:
Another aspect that might be relevant to deal with once this PR is merged is providing the ability to reuse the `SimpleHttpClient` for other resource providers implementations when adding those to the contrib repo: https://github.com/open-telemetry/opentelemetry-java-contrib/issues/1074.

rjbaucells:
> there's work in-progress which could unblock this: [open-telemetry/opentelemetry-java#5918](https://github.com/open-telemetry/opentelemetry-java/pull/5918)

Is there anything else preventing the PR from being merged? The above mentioned change was already merged...

trask:
@rjbaucells can you integrate that change into your PR?

rjbaucells:
> @rjbaucells can you integrate that change into your PR?

I looked at the details of PR [5918](https://github.com/open-telemetry/opentelemetry-java/pull/5918) and it provides a way to supress instrumentation of internal calls in "exporters" via setting the`suppress_internal_exporter_instrumentation` context variable to `true`.

Looking at the implementation of the JDK (11+) based sender, it is not suppressing the instrumentation of the requests; only the okHttp sender implementation is using it. I would assume the java agent will process calls to the JDK HttpClient the same way as the HttpURLConnection. Do we really need to suppress these calls for the HttpURLConnection?

trask:
> only the okHttp sender implementation is using it

right, so the okhttp sender is using it so that auto-instrumentation of OkHTTP won't capture those http requests

in the same way, I think the code here could use it so that auto-instrumentation of HttpURLConnection won't capture these http requests

rjbaucells:
The Context value in the above mentioned PR is `suppress_internal_exporter_instrumentation`. Looking at the name it is very specific to "exporters". I did not find the place the agent is reading the boolean context value to skip the call instrumentation.

Is there a generic Context value I can set prior to the HTTP call that will instruct the agent to skip the instrumentation?

trask:
check out https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/9739

driverpt:
any ETA on this one ?

SylvainJuge:
The https://github.com/open-telemetry/opentelemetry-java/pull/6546 PR was just merged and will be available with 1.41.0 SDK release, so the issue about using an exporter-related context entry ([here](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1061#issuecomment-1855932215)) is solved.

The logical next steps are thus:
- release of sdk 1.41.0
- updating this PR to use the new internal API introduced in https://github.com/open-telemetry/opentelemetry-java/pull/6546

This of course assumes that @rjbaucells has time to dedicate to this, let us know if you need any help here.

driverpt:
When is SDK 1.41.0 expected to be released?

SylvainJuge:
I don't know if there is an official release schedule, but looking at past releases hint that we have roughly one release per month: https://github.com/open-telemetry/opentelemetry-java/releases, so it would likely be available within 2/3 weeks.

rjbaucells:
I have the PR already updated, waiting for SDK 1.41.0 to push it.

trask:
> I don't know if there is an official release schedule

check out https://github.com/open-telemetry/opentelemetry-java/blob/main/RELEASING.md#release-cadence, so Aug release target will be Fri, Aug 9

driverpt:
Rebase it!

driverpt:
Oh. I was not aware of that. Apologies then.

Do you know +/- an ETA for this to be "releaseable"?

Thanks

trask:
I think this can move forward now using the new internal API that was introduced in https://github.com/open-telemetry/opentelemetry-java/pull/6546

driverpt:
Hi, I think that's a question for @rjbaucells . He's the owner of the code

SylvainJuge:
> Hi, I think that's a question for @rjbaucells . He's the owner of the code

Sorry for the noise, I failed to see that you weren't the original PR author :-).

PR #1102, opened 2023-11-17
Add Error handling for closure parameters

akats7:
In #960 we discussed the way we handle attributes that do not exist when extracted from user defined closures, currently this will throw an AttributeNotFoundException, which does not seem like the correct pattern. This should should just be handled gracefully and in the case if the transformation just avoid creating the metric and in the case of the labels, omit that label.

**Existing Issue(s):**

N/A

**Testing:**

### DIFF
--- a/jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/MBeanHelper.groovy
+++ b/jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/MBeanHelper.groovy
@@ -122,8 +122,16 @@ class MBeanHelper {
     }
 
     Object getBeanAttributeWithTransform(GroovyMBean bean, String attribute){
-      def transformationClosure = attributeTransformation.get(attribute);
-      return transformationClosure != null ? transformationClosure(bean) : getBeanAttribute(bean, attribute)
+        def transformationClosure = attributeTransformation.get(attribute);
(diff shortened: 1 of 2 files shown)


jmx-metrics/src/main/groovy/io/opentelemetry/contrib/jmxmetrics/MBeanHelper.groovy
```diff
@@ -123,7 +123,16 @@ class MBeanHelper {
 
     Object getBeanAttributeWithTransform(GroovyMBean bean, String attribute){
       def transformationClosure = attributeTransformation.get(attribute);
-      return transformationClosure != null ? transformationClosure(bean) : getBeanAttribute(bean, attribute)
+      if (transformationClosure != null){
```
breedx-splk, line 126: nit: I like inverting this and returning `null` early because it helps to reduce the indent levels (reader burden).

breedx-splk (approved):
Looks good to me. @akats7 it looks like you need to sign the CLA, also please format using `./gradlew spotlessApply`. Thanks!

akats7:
> Looks good to me. @akats7 it looks like you need to sign the CLA, also please format using `./gradlew spotlessApply`. Thanks!

Hey @breedx-splk, thanks will do, I must be missing some metadata on a commit, I’ll rebase.

PR #1108, opened 2023-11-21
Add zstd compressor implementation for OTLP exporters

jack-berg:
Implementation of the `Compressor` SPI I propose in https://github.com/open-telemetry/opentelemetry-java/pull/5990

This PR adds support for a zstd compression, which is one of the algorithms supported by the collector: https://github.com/open-telemetry/opentelemetry-collector/blob/73fa163951aeee538bd42f561d3a96205ee251d0/config/configgrpc/README.md

I think compressors SPIs like this belong outside the core repo because they're not explicitly specified, and there may be disagreement about how to package up the code which performs the compression. For example, this artifact proposed in this PR has a dependency on `com.github.luben:zstd-jni:1.5.5-10`, which uses JNI to improve performance. Some might prefer a pure java version, or prefer the dependency be shaded. I want to provide something people can easily use, without providing a solution for every combination of user requirements (users can always provide their own, quite easily).


compressors/build.gradle.kts
```diff
@@ -0,0 +1,10 @@
+subprojects {
+  // https://github.com/gradle/gradle/issues/847
+  group = "io.opentelemetry.compressors"
+  val proj = this
+  plugins.withId("java") {
+    configure<BasePluginExtension> {
+      archivesName.set("opentelemetry-compressor-${proj.name}")
+    }
+  }
+}
```
trask, line 10: is this file needed? we don't seem to need parent module `runtime-attach`

jack-berg: Runtime attach gets around this by fully qualifying the names of the nested modules. For example, it has:

```
/runtime-attach
  /runtime-attach
    ...
  /runtime-attach-core
```

While this file allows us to avoid the redundant naming and still produce artifacts with the correct names (i.e. `opentelemetry-compressor-zstd`:
```
/compressors
  /zstd
```

This is the pattern we use in `opentelemetry-java` (e.g. see [opentelemetry-java/exporters/build.gradle.kts](https://github.com/open-telemetry/opentelemetry-java/blob/main/exporters/build.gradle.kts)), but I don't have a strong opinion. We can accomplish the same thing either way.

compressors/zstd/src/main/java/io/opentelemetry/contrib/compressor/zstd/ZstdCompressor.java
```diff
@@ -0,0 +1,32 @@
...
+  private ZstdCompressor() {}
+
+  public static ZstdCompressor getInstance() {
+    return INSTANCE;
+  }
+
+  @Override
+  public String getEncoding() {
+    return "zstd";
+  }
```
trask, line 26: (unrelated to this PR) was there reason for this method to be in `Compressor` and not `CompressorProvider` (e.g. similar to `ConfigurableSpanExporterProvider.getName()`)

buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
```diff
@@ -10,11 +10,17 @@ plugins {
...
 
 group = "io.opentelemetry.contrib"
 
-base.archivesName.set("opentelemetry-${project.name}")
+base {
+  // May be set already by a parent project, only set if not.
+  if (!archivesName.get().startsWith("opentelemetry-")) {
+    archivesName.set("opentelemetry-$name")
+  }
+}
```
trask, line 21: (just in case we don't need the `compressor` parent module, we can revert this?)

trask (approved):

compressors/build.gradle.kts
```diff
@@ -0,0 +1,10 @@
+subprojects {
+  // https://github.com/gradle/gradle/issues/847
+  group = "io.opentelemetry.compressors"
```
trask, line 3: will this publish it under groupId `io.opentelemetry.compressors`?

compressors/zstd/src/main/java/io/opentelemetry/contrib/compressor/zstd/ZstdCompressor.java
jack-berg: Yeah so this is a good question which I encountered implementing. My thinking:
- We want some sort of name that we can reference when we call `setCompression(String)` on the exporters
- We also want a value to use for the `Content-Encoding` header
- I think it will always be sufficient to use the same name for `setCompression` and `Content-Encoding`.

By the time the Compressor is resolved from CompressorProvider, the compressor wouldn't be able to access a `CompressorProvider#getName` any more. So my thinking was, put the identifier down in Compressor instead of CompressorProvider, so its always accessible to the Compressor where its needed for `Content-Encoding`. Count on the fact that `CompressorProvider#getInstance()` _should_ return singletons, and therefore you can call `CompressorProvider.getInstance().getName()` to get the name identifier of the `CompressorProvider`.

I also thought about how [ContextStorageProvider](https://github.com/open-telemetry/opentelemetry-java/blob/main/context/src/main/java/io/opentelemetry/context/ContextStorageProvider.java) also doesn't have a `#getName()` method, since its identified by its FQCN for configuration. Compressor is different because we definitely want to be able to use a short name instead of FQCN, but that's a little bit of the prior art and what I was thinking about.

buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
jack-berg: Yes

compressors/build.gradle.kts
jack-berg: Oh this is something goofy. This just avoids a bug where nested modules have the same name. Like, if had `:foo:bar` and `:baz:bar`, there would be a conflict because some of the internals of gradle only identify the module by the last part of the module name, and we would have to modules named `bar`.

This still publishes under `io.opentelemetry.contrib`, but I think I can get rid of this actually because we don't have any other nested modules called `:zstd`. (In the core repo, we have multiple modules ending in `:testing-internal`.)

PR #1130, opened 2023-12-10
switch from acceptance to rejection threshold

oertl:
**Description:**

As agreed in the last Sampling SIG meeting (see https://github.com/open-telemetry/oteps/pull/235#issuecomment-1845647087), the sampling probability information will be encoded as "rejection threshold" instead of an "acceptance threshold". This PR reflects this decision.

### DIFF
--- a/consistent-sampling/src/main/java/io/opentelemetry/contrib/sampler/consistent56/ConsistentSamplingUtil.java
+++ b/consistent-sampling/src/main/java/io/opentelemetry/contrib/sampler/consistent56/ConsistentSamplingUtil.java
@@ -10,7 +10,9 @@ import com.google.errorprone.annotations.CanIgnoreReturnValue;
 public final class ConsistentSamplingUtil {
 
   private static final int RANDOM_VALUE_BITS = 56;
-  private static final long MAX_THRESHOLD = 1L << RANDOM_VALUE_BITS;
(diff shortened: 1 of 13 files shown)


consistent-sampling/src/main/java/io/opentelemetry/contrib/sampler/consistent56/ConsistentSamplingUtil.java
```diff
@@ -41,29 +43,24 @@ public static double calculateSamplingProbability(long threshold) {
...
   public static long calculateThreshold(double samplingProbability) {
     checkProbability(samplingProbability);
-    return Math.round(samplingProbability * 0x1p56);
+    return MAX_THRESHOLD - Math.round(samplingProbability * 0x1p56);
   }
 
   /**
    * Calculates the adjusted count from a given threshold.
    *
    * <p>Returns 1, if the threshold is invalid.
```
PeterF778, line 52: This is no longer correct ...

oertl: thanks, fixed that

PeterF778 (approved):
Looks great!

trask (approved):

PR #1146, opened 2024-01-04
Shadowing generated proto java sources

LikeTheSalad:
There was an issue raised in the `otel-android` Slack channel where a user needed to add the [OTel Android](https://github.com/open-telemetry/opentelemetry-android) lib along with the [opentelemetry-proto](https://github.com/open-telemetry/opentelemetry-proto-java) one in the same project. This caused a compilation issue due to duplicated proto classes across [opentelemetry-proto](https://github.com/open-telemetry/opentelemetry-proto-java) and [disk-buffering](https://github.com/open-telemetry/opentelemetry-java-contrib/tree/main/disk-buffering) (added in OTel Android).

These changes aim to avoid this issue by renaming the generated proto classes' package for the disk-buffering project.

### DIFF
--- a/disk-buffering/build.gradle.kts
+++ b/disk-buffering/build.gradle.kts
@@ -1,8 +1,10 @@
+import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
 import ru.vyarus.gradle.plugin.animalsniffer.AnimalSniffer
 
 plugins {
   id("otel.java-conventions")
   id("otel.publish-conventions")
+  id("com.github.johnrengelman.shadow")
   id("me.champeau.jmh") version "0.7.2"
   id("ru.vyarus.animalsniffer") version "1.7.1"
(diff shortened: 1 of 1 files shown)


trask:
cc @zeitlinger 

(automation isn't working due to https://github.com/open-telemetry/community/issues/1826)

disk-buffering/build.gradle.kts
```diff
@@ -61,6 +63,13 @@ wire {
   )
 }
 
+tasks.named("shadowJar", ShadowJar::class.java) {
```
trask, line 66: try this
```suggestion
tasks.named<ShadowJar>("shadowJar") {
```

trask (approved):

disk-buffering/build.gradle.kts
LikeTheSalad: that's nicer, thanks!

PR #1161, opened 2024-01-17
[disk-buffering] - Single responsibility for disk exporters

breedx-splk:
So I thought it was a little confusing that there were classes like `DiskExporter` (and all their type-specific friends) that did double duty -- that is they both wrote to AND read from the on-disk buffer.

The directionality of `DiskExporter` always suggested (to me at least) that export was done _to_ the disk. Even if wasn't your immediate instinct, at least you'll hopefully agree with me that decomposing into separate responsibility units has merit.

I offer up a decomposition which turns `DiskExporter` into `ToDiskExporter` and `FromDiskExporter`. I hope that it's much easier to follow which is which.

You might notice that there's asymmetry between the "to disk" type specific classes and the (lack of) type specific "from disk" classes. This is primarily due to the fact that the SDK classes (`SdkTracerProvider`, `SdkMeterProvider`, `SdkLoggerProvider`) need signal-specific implementation classes...which the impls here are. The "from disk" classes do not have this same problem because there is no sdk coupling, just api delegation.


breedx-splk:
Ah combining exporters into single package made them all internal. Lemme fix that.

zeitlinger (approved):

disk-buffering/DESIGN.md
```diff
@@ -1,11 +1,18 @@
...
+* [SpanToDiskExporter](src/main/java/io/opentelemetry/contrib/disk/buffering/SpanToDiskExporter.java))
+
+Each is responsible for writing a specific type of telemetry to disk storage for later 
+harvest/ingest.
+
+For later reading, there is:
+* [FromDiskExporter](src/main/java/io/opentelemetry/contrib/disk/buffering/FromDiskExporter.java)
+which can be created with `FromDiskExporter.builder()....build()`. 
+As explained in the [README](README.md), this has to be triggered manually by the consumer of 
+* this library and does not happen automaticall
```
LikeTheSalad, line 15:
```suggestion
* this library and does not happen automatically.
```

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/FromDiskExporterBuilder.java
```diff
@@ -0,0 +1,81 @@
...
+      public List<T> deserialize(byte[] source) {
+        return Collections.emptyList();
+      }
+    };
+  }
+
+  private final StorageBuilder storageBuilder = Storage.builder();
+
+  @CanIgnoreReturnValue
+  public FromDiskExporterBuilder<T> setFolderName(String folderName) {
```
LikeTheSalad, line 47: The folder name and the serializer must be the same as the ones used in each `ToDiskExporter` impl for it to work properly, so I'm a bit concerned about providing this builder publicly as it seems like it could be easy to misconfigure it. Instead, I would move it (and also `FromDiskExporter.java`) to the internal package and create public implementations for each signal in this package, similarly to what's done for the `ToDiskExporter`s.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/StorageConfiguration.java
```diff
@@ -8,11 +8,18 @@
...
 import java.util.concurrent.TimeUnit;
 
 /** Defines how the storage should be managed. */
 @AutoValue
 public abstract class StorageConfiguration {
+
+  /**
+   * The root storage location for buffered telemetry.
+   */
+  public abstract File getRootDir();
```
LikeTheSalad, line 21: This is cool! After using this tool a couple of times, I just ended up passing the same root file everywhere which wasn't too nice, this is much cleaner.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/StorageConfiguration.java
LikeTheSalad, line 1: Somehow I just noticed that this is in the internal package 🤦‍♂️ it shouldn't. I guess I'll address it after this PR is merged.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/exporter/ToDiskExporterBuilder.java
```diff
@@ -34,71 +33,55 @@ public List<T> deserialize(byte[] source) {
...
-  DiskExporterBuilder() {}
+  ToDiskExporterBuilder() {}
 
-  @CanIgnoreReturnValue
-  public DiskExporterBuilder<T> setRootDir(File rootDir) {
-    this.rootDir = rootDir;
-    return this;
-  }
+//  @CanIgnoreReturnValue
+//  public ToDiskExporterBuilder<T> setRootDir(File rootDir) {
```
LikeTheSalad, line 45: I think this shouldn't be commented out?

LikeTheSalad (commented):
This new concept is great, thank you!

Just a couple of comments but overall I think we should go with it.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/FromDiskExporterBuilder.java
breedx-splk: Ok, I gave that a shot....and it ended up working out and now there's nice parity between the `ToDisk` and the `FromDisk` per signal types. Thanks for the idea!

Unfortunately, it ended up being a LOT of work to get the tests shored up. For starters, we were several versions behind on the protobufs...and I got us on the latest. The latest, tho, includes the `TraceFlags` as part of the protos, which we hadn't accounted for. There's also asymmetry in several places because the `TraceFlags` are _not_ serialized into protobufs for the Exemplars (and maybe also for parent context, I forget).

I any case, I had to shim in a bunch of slightly different expected results, which contain the `TraceFlags.getSampled()` which is what we hard-code to on the de-serialization side. I think overall though it probably cleaned up the tests a little and definitely encouraged some reuse.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/StorageConfiguration.java
breedx-splk: Yeah please this one is already super heavy. :) Thank you!

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/exporter/ToDiskExporterBuilder.java
breedx-splk: The root dir is now part of the storage configuration, so it doesn't belong on this builder any longer (see the `build()` method for when it's used).

I'll remove this commented/boneyard tho.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/FromDiskExporterBuilder.java
LikeTheSalad: Oh man, I know how painful dealing with these serializations is, especially when it comes to metrics! 😩 thank you for taking the time to update it 🙏

LikeTheSalad (approved):

breedx-splk:
@open-telemetry/java-maintainers we have component-owners approvals. ❤️ 🙏🏻 Thanks.

trask (approved):

PR #1167, opened 2024-01-23
[disk-buffering] Split serializer

breedx-splk:
Single responsibility principle -- Split the `SignalSerializer` interface into `SignalSerializer` and `SignalDeserializer`. This cleans up a few things after the last refactor.

Note: The tests are currently testing both directions (serialize/deserialize), so I didn't feel like it was really necessary to decompose those in this PR. If someone else did that I wouldn't complain, but it felt a bit like overkill.

### DIFF
--- a/disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/serialization/deserializers/MetricDataDeerializer.java
+++ b/disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/serialization/deserializers/MetricDataDeerializer.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.contrib.disk.buffering.internal.serialization.deserializers;
+
+import io.opentelemetry.contrib.disk.buffering.internal.serialization.mapping.metrics.ProtoMetricsDataMapper;
+import io.opentelemetry.proto.metrics.v1.MetricsData;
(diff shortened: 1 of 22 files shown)


disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/serialization/deserializers/MetricDataDeerializer.java
```diff
@@ -0,0 +1,31 @@
...
+
+package io.opentelemetry.contrib.disk.buffering.internal.serialization.deserializers;
+
+import io.opentelemetry.contrib.disk.buffering.internal.serialization.mapping.metrics.ProtoMetricsDataMapper;
+import io.opentelemetry.proto.metrics.v1.MetricsData;
+import io.opentelemetry.sdk.metrics.data.MetricData;
+import java.io.IOException;
+import java.util.List;
+
+public final class MetricDataDeerializer implements SignalDeserializer<MetricData> {
```
LikeTheSalad, line 14:
```suggestion
public final class MetricDataDeserializer implements SignalDeserializer<MetricData> {
```

LikeTheSalad (approved):
LGTM!

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/serialization/deserializers/MetricDataDeerializer.java
breedx-splk: Whoopsie! Thanks for the catch....will fix.

trask (approved):

PR #1178, opened 2024-02-01
remove baggage propagation from xray propagator

jamesmoessis:
**Description:**

This PR removes baggage propagation from the X-Ray propagator.

**Existing Issue(s):**

We've seen strong incompatibilities between the OpenTelemetry/W3C baggage specification, and that which X-Ray allows. 

The primary incompatibility is that the limit of 256 characters in the Amazon Trace header. This is incompatible with the [W3C baggage limit](https://www.w3.org/TR/baggage/#limits) of 8192 bytes. Attempting to truncate the baggage to fit within this limit risks corrupting the entries and should not be attempted. 

We've seen issues with baggage propagation in X-Ray, for example here: https://github.com/aws-observability/aws-otel-java-instrumentation/issues/563

Additionally, other language's xray propagator do not attempt to attach baggage, for example: 
* [OpenTelemetry JS X-Ray Propagator](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/propagators/opentelemetry-propagator-aws-xray/src/AWSXRayPropagator.ts#L60)
* [OpenTelemetry Python X-Ray Propagator](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/propagator/opentelemetry-propagator-aws-xray/src/opentelemetry/propagators/aws/aws_xray_propagator.py)

**Testing:**

Removed tests related to baggage.

### DIFF
--- a/aws-xray-propagator/src/main/java/io/opentelemetry/contrib/awsxray/propagator/AwsXrayPropagator.java
+++ b/aws-xray-propagator/src/main/java/io/opentelemetry/contrib/awsxray/propagator/AwsXrayPropagator.java
@@ -5,9 +5,6 @@
 
 package io.opentelemetry.contrib.awsxray.propagator;
 
-import io.opentelemetry.api.baggage.Baggage;
-import io.opentelemetry.api.baggage.BaggageBuilder;
-import io.opentelemetry.api.baggage.BaggageEntry;
 import io.opentelemetry.api.internal.StringUtils;
 import io.opentelemetry.api.trace.Span;
 import io.opentelemetry.api.trace.SpanContext;
@@ -21,7 +18,6 @@ import io.opentelemetry.context.propagation.TextMapPropagator;
 import io.opentelemetry.context.propagation.TextMapSetter;
 import java.util.Collections;
 import java.util.List;
-import java.util.function.BiConsumer;
 import java.util.logging.Logger;
 import javax.annotation.Nullable;
 
@@ -112,49 +108,19 @@ public final class AwsXrayPropagator implements TextMapPropagator {
     char samplingFlag = spanContext.isSampled() ? IS_SAMPLED : NOT_SAMPLED;
     // TODO: Add OT trace state to the X-Ray trace header
 
-    StringBuilder traceHeader = new StringBuilder();
-    traceHeader
-        .append(TRACE_ID_KEY)
-        .append(KV_DELIMITER)
-        .append(xrayTraceId)
-        .append(TRACE_HEADER_DELIMITER)
-        .append(PARENT_ID_KEY)
-        .append(KV_DELIMITER)
-        .append(parentId)
-        .append(TRACE_HEADER_DELIMITER)
-        .append(SAMPLED_FLAG_KEY)
-        .append(KV_DELIMITER)
-        .append(samplingFlag);
+    String traceHeader = TRACE_ID_KEY +
+            KV_DELIMITER +
+            xrayTraceId +
+            TRACE_HEADER_DELIMITER +
+            PARENT_ID_KEY +
+            KV_DELIMITER +
+            parentId +
+            TRACE_HEADER_DELIMITER +
+            SAMPLED_FLAG_KEY +
+            KV_DELIMITER +
+            samplingFlag;
 
-    Baggage baggage = Baggage.fromContext(context);
-    // Truncate baggage to 256 chars per X-Ray spec.
-    baggage.forEach(
-        new BiConsumer<String, BaggageEntry>() {
-
-          private int baggageWrittenBytes;
-
-          @Override
-          public void accept(String key, BaggageEntry entry) {
-            if (key.equals(TRACE_ID_KEY)
-                || key.equals(PARENT_ID_KEY)
-                || key.equals(SAMPLED_FLAG_KEY)) {
-              return;
-            }
-            // Size is key/value pair, excludes delimiter.
-            int size = key.length() + entry.getValue().length() + 1;
-            if (baggageWrittenBytes + size > 256) {
-              return;
-            }
-            traceHeader
-                .append(TRACE_HEADER_DELIMITER)
-                .append(key)
-                .append(KV_DELIMITER)
-                .append(entry.getValue());
-            baggageWrittenBytes += size;
-          }
-        });
-
-    setter.set(carrier, TRACE_HEADER_KEY, traceHeader.toString());
+    setter.set(carrier, TRACE_HEADER_KEY, traceHeader);
   }
 
   @Override
@@ -179,9 +145,6 @@ public final class AwsXrayPropagator implements TextMapPropagator {
     String traceId = TraceId.getInvalid();
     String spanId = SpanId.getInvalid();
     Boolean isSampled = false;
-
-    BaggageBuilder baggage = null;
-    int baggageReadBytes = 0;
 
     int pos = 0;
     while (pos < traceHeader.length()) {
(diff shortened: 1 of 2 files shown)


wangzlei:
Thanks @jamesmoessis raise the issue. 

Other than baggage length 8192 bytes problem, could you help me understand what is other problem? awsxraypropagator is a short term solution before x-ray fully migrate to w3c, we want to understand the customer impact.
https://github.com/aws-observability/aws-otel-java-instrumentation/issues/563#issuecomment-1924450848

jamesmoessis:
> Other than baggage length 8192 bytes problem, could you help me understand what is other problem? 

It's just the length problem. 

> awsxraypropagator is a short term solution before x-ray fully migrate to w3c, we want to understand the customer impact.

The customer impact is that the xray propagator causes 400s to AWS and has the potential to corrupt baggage. I'm excited for when Amazon fully adopt W3C, but until then, the xraypropagator's handling of baggage is harmful, and it's specific to the Java implementation. For the short-term solution, there should be no baggage.

jamesmoessis:
@wangzlei any updates or thoughts? 

Are we happy causing 400s and corrupting baggage? It's a blocker for using AWS SDK instrumentation since the propagator is hardcoded into it.

jack-berg:
@jamesmoessis I added it to the agenda for the 4/11/24 Java SIG. Hopefully we can figure out how to proceed.

jamesmoessis:
Thankyou @jack-berg appreciate it. I would come to the SIG but unfortunately it's at 2am for me 😴  I'll keep an eye on the SIG meeting notes too

aws-xray-propagator/src/main/java/io/opentelemetry/contrib/awsxray/propagator/AwsXrayPropagator.java
```diff
@@ -112,49 +108,20 @@ public <C> void inject(Context context, @Nullable C carrier, TextMapSetter<C> se
...
-          @Override
-          public void accept(String key, BaggageEntry entry) {
-            if (key.equals(TRACE_ID_KEY)
-                || key.equals(PARENT_ID_KEY)
-                || key.equals(SAMPLED_FLAG_KEY)) {
-              return;
-            }
-            // Size is key/value pair, excludes delimiter.
-            int size = key.length() + entry.getValue().length() + 1;
-            if (baggageWrittenBytes + size > 256) {
```
jack-berg, line 145: > The primary incompatibility is that the limit of 256 characters in the Amazon Trace header. This is incompatible with the [W3C baggage limit](https://www.w3.org/TR/baggage/#limits) of 8192 bytes. Attempting to truncate the baggage to fit within this limit risks corrupting the entries and should not be attempted.

It looks like the code is only trying to include baggage entires if the entire entry can be included without exceeding 256 bytes.

When you talk about corrupting entries, are you referring to some entries being present while others not? Can you help me understand why this is problematic?

aws-xray-propagator/src/main/java/io/opentelemetry/contrib/awsxray/propagator/AwsXrayPropagator.java
```diff
@@ -112,49 +108,20 @@ public <C> void inject(Context context, @Nullable C carrier, TextMapSetter<C> se
...
-        .append(PARENT_ID_KEY)
-        .append(KV_DELIMITER)
-        .append(parentId)
-        .append(TRACE_HEADER_DELIMITER)
-        .append(SAMPLED_FLAG_KEY)
-        .append(KV_DELIMITER)
-        .append(samplingFlag);
-
-    Baggage baggage = Baggage.fromContext(context);
-    // Truncate baggage to 256 chars per X-Ray spec.
```
jack-berg, line 130: I can't seem to find a spec document for the xray spec format. If there is a document, and it does talk about including baggage like this, we should likely follow the spec.

jack-berg, line 130: What do you think about this [comment](https://github.com/aws-observability/aws-otel-java-instrumentation/issues/563#issuecomment-1924450848):

> Regarding the length of the X-Ray header, I did not find a definition in X-Ray spec. However, based on some clues, it can be inferred that StepFunction likely has a contract with X-Ray, hence there is length checking in the StepFunction client API. So it is reasonable to truncate the baggage length to (256 - the length of the existing trace context).

My read of this is that including up to 256 baggage bytes is a bug that should be fixed, instead of removing baggage encoding altogether.

rapphil: I'm not aware of a formal specification as well, but I usually follow this reference: https://docs.aws.amazon.com/xray/latest/devguide/aws-xray.html#Tracing-header:~:text=sampling%20rules.-,Tracing%20header,-All%20requests%20are

It seems that there is no equivalent concept to baggage, at least in the reference above.

jack-berg: No reference to baggage here, so maybe its right to delete it.

trask:
discussed in today's SIG meeting:
* we'd like to use this PR as a forcing function to ensure we have responsive component owners for the aws xray propagator
* @rapphil is going to check inside AWS about finding new component owner(s)
* if it doesn't look like that's going to be fruitful in the next 2 weeks, we'll look for new component owners within the broader OpenTelemetry community

wangzlei:
We have internal discussion with @rapphil today. To avoid conflicts with the baggage propagator, xray propagator can update the code to use a dedicated key instead of ImplicitContextKeyed Baggage https://github.com/open-telemetry/opentelemetry-java-contrib/blob/15a956b40044badb2282f2ec7d7a3f31ee7fa6e3/aws-xray-propagator/src/main/java/io/opentelemetry/contrib/awsxray/propagator/AwsXrayPropagator.java#L245

aws-xray-propagator/src/main/java/io/opentelemetry/contrib/awsxray/propagator/AwsXrayPropagator.java
```diff
@@ -241,9 +199,6 @@ private static <C> Context getContextFromHeader(
     if (spanContext.isValid()) {
       context = context.with(Span.wrap(spanContext));
     }
-    if (baggage != null) {
-      context = context.with(baggage.build());
```
wangzlei, line 245: Can we use a new key "XRAY_HEADER_ADDITIONAL_FIELDS"
```
context = context.with(XRAY_HEADER_ADDITIONAL_FIELDS ,baggage.build());
```
FYI https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1178#issuecomment-2050711021

aws-xray-propagator/src/main/java/io/opentelemetry/contrib/awsxray/propagator/AwsXrayPropagator.java
wangzlei: Right. [#563](https://github.com/aws-observability/aws-otel-java-instrumentation/issues/563) is StepFunction receives a xray header over 256. AWS Lambda is working on a new encoding algorithm to reducing the xray header to be less than 256, then we can fix the bug by truncating xray header to be 256, not truncating baggage to 256.

wangzlei: refer to the comment, using key "XRAY_HEADER_ADDITIONAL_FIELDS"

jamesmoessis:
Sorry folks, haven't had time to get back to this but I would like to as soon as I get the time. If someone else wants to pick this up and fix the problem, feel free. Otherwise I will try to find some time.

PR #1255, opened 2024-04-04
stacktrace span processor

SylvainJuge:
**Description:**

We'd like to contribute a span processor that allows to capture `code.stacktrace` when the span ends. The main technical challenge is that the SDK does not allow to simply modify spans on end.

This is described in detail in the following github discussion https://github.com/open-telemetry/opentelemetry-java-instrumentation/discussions/10973 

This was originally contributed in our own distribution in https://github.com/elastic/elastic-otel-java with only minor changes to make it fit.

**Existing Issue(s):**

GH discussion: https://github.com/open-telemetry/opentelemetry-java-instrumentation/discussions/10973 

**Testing:**

Unit tests have been provided

**Documentation:**

Simple description and usage is provided in the module README.md.

**Outstanding items:**


.github/component_owners.yml
```diff
@@ -66,3 +66,7 @@ components:
   kafka-exporter:
     - spockz
     - vincentfree
+  span-stacktrace:
+    - jackshirazi
+    - jonaskunz
```
jkwatson, line 71: we certainly need Jack and Jonas to comment on this PR, to make sure they are on board with supporting it.

jackshirazi: Yup, totally onboard to support this

span-stacktrace/src/main/java/io/opentelemetry/contrib/stacktrace/StackTraceSpanProcessor.java
```diff
@@ -0,0 +1,101 @@
...
+public class StackTraceSpanProcessor extends AbstractSimpleChainingSpanProcessor {
+
+  // TODO : remove this once semconv 1.24.0 is available
+  static final AttributeKey<String> SPAN_STACKTRACE = AttributeKey.stringKey("code.stacktrace");
+
+  private static final Logger logger = Logger.getLogger(StackTraceSpanProcessor.class.getName());
+
+  private final long minSpanDurationNanos;
+
+  private final Function<ReadableSpan, Boolean> filterFunction;
```
JonasKunz, line 28: [Predicate](https://docs.oracle.com/javase/8/docs/api/java/util/function/Predicate.html) should be the preferred standard type here

JonasKunz (approved):

trask:
@jackshirazi @JonasKunz when you have a chance, can you review and approve? thanks

JonasKunz (approved):

jackshirazi (approved):

span-stacktrace/src/main/java/io/opentelemetry/contrib/stacktrace/util/AbstractSimpleChainingSpanProcessor.java
```diff
@@ -0,0 +1,139 @@
...
+
+/**
+ * A @{@link SpanProcessor} which in addition to all standard operations is capable of modifying and
+ * optionally filtering spans in the end-callback.
+ *
+ * <p>This is done by chaining processors and registering only the first processor with the SDK.
+ * Mutations can be performed in {@link #doOnEnd(ReadableSpan)} by wrapping the span in a {@link
+ * MutableSpan}
+ */
+public abstract class AbstractSimpleChainingSpanProcessor implements SpanProcessor {
```
trask, line 24: are the util package classes part of the public API? if not, we use the convention of putting them somewhere under an `.internal.` package

SylvainJuge: For now, they aren't meant to be public, so we can definitely move them to an `.internal.` package. We would need to make them public only in the case where we need to have another implementation of it in a separate module, but there is no such yet.

In practice it also depends on the progress we might see on having a beforeEnd span processor hook, but that's definitely not granted nor in any known time frame, so safer.

So, to summarize I will rename `...contrib.stacktrace.util` package to `...contrib.stacktrace.internal` to fit this convention.

SylvainJuge: Fixed in 7f09b322ed2f43c66e7c8f3cebbeaa149666d090

SylvainJuge:
@open-telemetry/java-instrumentation-approvers I think it's ready to be reviewed (and maybe merged) now, let me know if there is anything left to fix.

JonasKunz (approved):

trask (approved):

PR #1290, opened 2024-04-24
Add baggage span processor component

### LINKED ISSUE #1277: Request for new component: Baggage Span Processor
### Component(s)

processors

The Honeycomb distro currently provides a Span processor that takes items from the baggage and adds those items as attributes onto spans. 

https://github.com/honeycombio/honeycomb-opentelemetry-java/blob/main/common/src/main/java/io/honeycomb/opentelemetry/sdk/trace/spanprocessors/BaggageSpanProcessor.java

_No response_

### Additional context

_No response_

MikeGoldsmith:
**Description:**

Adds the baggage span processor as a new component.

**Existing Issue(s):**

- Closes #1277 

**Testing:**

Added unit tests to verify behaviour

**Documentation:**

Added README. Let me know if / where additional docs are required.

**Outstanding items:**

I'm unsure how contrib project components are used by the java agent, is there additional work to enable that?


baggage-processor/README.md
```diff
@@ -0,0 +1,19 @@
+# OpenTelemetry Baggage Span Processor
+
+The BaggageActivityProcessor reads entries stored in Baggage from the parent context
```
cijothomas, line 3:
```suggestion
The BaggageSpanProcessor reads entries stored in Baggage from the parent context
```

baggage-processor/README.md
```diff
@@ -0,0 +1,19 @@
...
+and adds the baggage keys and values to the `Activity` as tags(attributes) on start.
```
cijothomas, line 4:
```suggestion
and adds the baggage keys and values to the `Span` as attributes on start.
```

zeitlinger:
What about making the baggage items that are added as spans tags configurable, with the option to provide `*` for all - like here: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/355003538a4dcf7ec58b6d9be42086c53238bdac/instrumentation/log4j/log4j-appender-2.17/library/README.md?plain=1#L101

MikeGoldsmith:
> What about making the baggage items that are added as spans tags configurable, with the option to provide `*` for all - like here: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/355003538a4dcf7ec58b6d9be42086c53238bdac/instrumentation/log4j/log4j-appender-2.17/library/README.md?plain=1#L101

We had a similar discussion in the .NET contrib project but thought it was more complicated than just using a set of prefixes so created an issue to continue the discussion. The plain processor that copies all baggage entries (like using `*` in your example) is likely to be accepted first.
- https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/1695

I'd be happy to open a similar issue for Java contrib too if you feel it would be beneficial. It would be additive to the current state of the processor now too.

zeitlinger (approved):

MikeGoldsmith:
Ping @open-telemetry/java-instrumentation-approvers - hoping to get feedback on this new component. We (Honeycomb) have users who like this functionality, and would like to enable more OpenTelemetry users to benefit from it.

baggage-processor/README.md
```diff
@@ -0,0 +1,19 @@
+# OpenTelemetry Baggage Span Processor
+
+The BaggageSpanProcessor reads entries stored in Baggage from the parent context
+and adds the baggage keys and values to the `Span` as attributes on start.
+
+Add this activity processor to a tracer provider.
```
trask, line 6:
```suggestion
Add this span processor to a tracer provider.
```

baggage-processor/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+Warning!
+
+To repeat: a consequence of adding data to Baggage is that the keys and values
+will appear in all outgoing trace context headers from the application.
+
+Do not put sensitive information in Baggage.
```
trask, line 13: +💯

baggage-processor/build.gradle.kts
```diff
@@ -0,0 +1,16 @@
...
+  id("otel.publish-conventions")
+}
+
+description = "OpenTelemetry Baggage Span Processor"
+otelJava.moduleName.set("io.opentelemetry.contrib.baggage.processor")
+
+dependencies {
+  api("io.opentelemetry:opentelemetry-api")
+  api("io.opentelemetry:opentelemetry-sdk")
+  api("io.opentelemetry:opentelemetry-sdk-extension-autoconfigure-spi")
```
trask, line 13: Is this used?

```suggestion
```

MikeGoldsmith: Not currently, it can be removed. I would like to add support for it so the java agent could also utilise this component too, but we can do that as a follow up PR.

MikeGoldsmith:
Thanks for the feedback @trask - I've applied your suggestions.

baggage-processor/src/main/java/io/opentelemetry/contrib/baggage/processor/BaggageSpanProcessor.java
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * This span processor copies attributes stored in {@link Baggage} into each newly created {@link
+ * io.opentelemetry.api.trace.Span}.
+ */
+public class BaggageSpanProcessor implements SpanProcessor {
+  @Override
+  public void onStart(Context parentContext, ReadWriteSpan span) {
+    Baggage.fromContext(parentContext)
+        .forEach((s, baggageEntry) -> span.setAttribute(s, baggageEntry.getValue()));
```
tylerbenson, line 22: Consider allowing a subset of baggage entries instead via an optional constructor argument that defines a filter.

MikeGoldsmith: Hey @tylerbenson - I've opened the following issue to track adding a user-defined function to determine if an attribute should be copied. I've been discussing options with a few contrib SIGs as a way to limit sensitive data being copied.
- https://github.com/open-telemetry/opentelemetry-java-contrib/issues/1295

The user-defined function is the most flexible as it can be configured as a allow or deny list using prefixes or whole matches.

The main concern with that approach is how it could be used with the Java Agent. An allow or deny list provided by system settings or env variables would be simplest solution but it's open to discussion. I'm aiming for consistency across language implementations so decided to add the processor without that feature (plus clear docs) while gathering feedback from each SIG.

zeitlinger: in the Java SIG, we discussed how to integrate this component into the javaagent.

We decided that the configuration (what baggage items to include) can and should be done here in the contrib repo as well, using an [AutoConfigurationCustomizer](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk-extensions/autoconfigure-spi/src/main/java/io/opentelemetry/sdk/autoconfigure/spi/AutoConfigurationCustomizer.java)

I'm happy to work with you together on this part - as a follow up PR :smile: 

See https://docs.google.com/document/d/1WK9h4p55p8ZjPkxO75-ApI9m0wfea6ENZmMoFRvXSCw/edit#bookmark=id.lm2raghsdnup

MikeGoldsmith: That's great to hear @zeitlinger - thank you for reporting back.

Yep, happy to work with you on making this processor accessible in the java agent 👍🏻

MikeGoldsmith:
@open-telemetry/java-contrib-maintainers @open-telemetry/java-contrib-approvers Is there anything else you would like to see in this PR?

@zeitlinger has agreed to help make it accessible with the Java Agent in a follow-up PP - see https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1290#discussion_r1590808231

tylerbenson:
> Is there anything else you would like to see in this PR?
@MikeGoldsmith I think I'd still like to see a way to filter somehow. If you want to limit that to just a name filter, or add full predicate function support that's up to you. I don't think this should be merged without that though.

MikeGoldsmith:
> > Is there anything else you would like to see in this PR?
> 
> @MikeGoldsmith I think I'd still like to see a way to filter somehow. If you want to limit that to just a name filter, or add full predicate function support that's up to you. I don't think this should be merged without that though.

I took your thumbs up on my reply to the original request for that as acceptance we'll add a filter option in a follow-up PR. If you feel strongly it should not accepted with out it, I can look at adding something.

I'm disappointed it's blocked in that case, all the other SDK contrib projects accepted without the filter with the expectation it will come afterwards. We're aiming for consistency with each implementation.

tylerbenson:
@MikeGoldsmith to be clear, I'm not a maintainer or approver here.  I'm just expressing what I view as the MVP.

My main concern with this feature without the filter is that it would easily add things to the span that would be inappropriate, either by dramatically increasing span sizes (and and the resulting payload), or including something sensitive that shouldn't be there. (Nobody knows all the crazy stuff people put into baggage or for what reasons.)

For this reason, my preferred usage of this feature would be that of an allow list where only things explicitly ok'd would be added, rather than filtering out things that aren't desired.

baggage-processor/src/main/java/io/opentelemetry/contrib/baggage/processor/BaggageSpanProcessor.java
```diff
@@ -0,0 +1,59 @@
...
+   * Creates a new {@link BaggageSpanProcessor} that copies all baggage entries into the newly
+   * created {@link io.opentelemetry.api.trace.Span}.
+   */
+  public BaggageSpanProcessor() {}
+
+  /**
+   * Creates a new {@link BaggageSpanProcessor} that copies only baggage entries with keys that pass
+   * the provided filter into the newly created {@link io.opentelemetry.api.trace.Span}.
+   */
+  public BaggageSpanProcessor(Function<String, Boolean> keyFilter) {
```
tylerbenson, line 32: I would recommend making the filter a [BiPredicate<String,BaggageEntry>](https://docs.oracle.com/javase/8/docs/api/java/util/function/BiPredicate.html). The entry can be ignored, but maybe someone might want to filter on the contents of the baggage.

trask: (a BiPredicate constructor could always be added later)
```suggestion
  public BaggageSpanProcessor(Predicate<String> keyFilter) {
```

trask (approved):

baggage-processor/src/main/java/io/opentelemetry/contrib/baggage/processor/BaggageSpanProcessor.java
MikeGoldsmith: Added in https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1290/commits/331f2e2de9429a59333c538950c81f9b9e130869

baggage-processor/README.md
```diff
@@ -0,0 +1,43 @@
...
+```java
+import io.opentelemetry.contrib.baggage.processor;
+// ...
+
+Tracer tracer = SdkTracerProvider.builder()
+  .addSpanProcessor(new BaggageSpanProcessor())
+  .build()
+```
+
+A predicate function that determines what baggage entries are copied can be provided as a contructor paramter.
```
MikeGoldsmith, line 31:
```suggestion
A predicate function that determines what baggage entries are copied can be provided as a constructor parameter.
```

MikeGoldsmith:
I guess the only outstanding query regarding the baggage key predicate and it's default behaviour:
1) defaults to allow all baggage keys, optional constructor parameter to set alternative
  - pros: current behaviour, easier for people to pick up and use
  - cons: not secure by default, eg all baggage entries are copied to every span
2) defaults to not allow any baggage keys, required constructor parameter
  - pros: secure by default, user must provide key selector that adds items they care about
  - cons: not as easy to use, not as plug-and-play ready

What are your thoughts @trask @tylerbenson @zeitlinger @cijothomas?

MikeGoldsmith:
After discussion with other SIGs, I think we should try to be secure by default by requiring a key filter as a constructor parameter.

When we move forward with getting it to work with the Java Agent, we'll have to make sure it's clear it's expected to provide key filter in whatever way that works out to be. eg env var / system property that sets allowed prefixes.

MikeGoldsmith:
Can't see why https://github.com/open-telemetry/opentelemetry-java-contrib/commit/775b1449498f573a59c9c04922e78f32ae05d1bb is failing the CLA. I'm pretty sure both my and @cijothomas's email addresses are correct.

trask:
closing and re-opening to try to resolve CLA failure

tylerbenson (approved):
Thanks for considering my feedback.

trask:
I've seen this a couple of times recently, opened a support ticket with EasyCLA to investigate this PR

MikeGoldsmith:
> I've seen this a couple of times recently, opened a support ticket with EasyCLA to investigate this PR: https://jira.linuxfoundation.org/plugins/servlet/desk/portal/4/SUPPORT-26928

Thanks @trask 👍🏻

laurit:
@MikeGoldsmith I think easycla also looks at the `Co-authored-by` lines in the commit message. Probably the problematic one is `Co-authored-by: Cijo Thomas <cithomas@microsoft.com>`. I'd try replacing it with `Co-authored-by: Cijo Thomas <cijo.thomas@gmail.com>` or removing it.

MikeGoldsmith:
> @MikeGoldsmith I think easycla also looks at the `Co-authored-by` lines in the commit message. Probably the problematic one is `Co-authored-by: Cijo Thomas <cithomas@microsoft.com>`. I'd try replacing it with `Co-authored-by: Cijo Thomas <cijo.thomas@gmail.com>` or removing it.

Looks like that did the trick, thanks @laurit!

baggage-processor/src/main/java/io/opentelemetry/contrib/baggage/processor/BaggageSpanProcessor.java
SylvainJuge: Hi! Have the work on adding this to the agent already started somewhere ? I have found https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/4520 issue that was opened a while ago before this contribution, but no follow-up PR to implement it. I'd be happy to help making progress here.

trask: hey @SylvainJuge! check out #1330

SylvainJuge: I've just opened https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/11697 to add it to the agent.

PR #1298, opened 2024-05-02
Inline incubating attributes + central semconv-incubating dependency

SylvainJuge:
**Initial intent:**

- use semconv-incubating for code.stacktrace in `span-stacktrace` module
- centralize version dependency of semconv-incubating across modules.

The non-incubating `semconv` dependency is transitively provided by the `opentelemetry-instrumentation-bom-alpha` BOM, which unfortunately does not include the `semconv-incubating` artifact.

As a possible follow-up improvement, we could include it in the instrumentation BOM as it should align with the version of `semconv` by default, also it would help the maintainers of distributions by allowing to directly pull both the `semconv` and `semconv-incubating` dependencies from the BOM instead of having to keep two versions in sync.

------
**Updated description:**

- inline all the incubating semconv attributes in production code
- keep a single centralized semconv version
- only depend on incubating semconv in tests, this implicitly validates inlined attributes fit the incubating semantic conventions.


jackshirazi (approved):

gcp-resources/build.gradle.kts
```diff
@@ -15,7 +15,7 @@ dependencies {
   implementation("com.google.cloud.opentelemetry:detector-resources-support:0.28.0")
 
   implementation("io.opentelemetry.semconv:opentelemetry-semconv")
-  implementation("io.opentelemetry.semconv:opentelemetry-semconv-incubating:1.25.0-alpha")
+  implementation("io.opentelemetry.semconv:opentelemetry-semconv-incubating")
```
jack-berg, line 18: All these usages of `opentelemetry-semconv-incubating` go against our [stated advice](https://github.com/open-telemetry/semantic-conventions-java?tab=readme-ov-file#published-released):

> Library instrumentation SHOULD NOT depend on this.

> Libraries can use this for testing, but should make copies of the attributes to avoid possible runtime errors from version conflicts.

I think we should follow our own advice to avoid unresolvable transitive version conflicts.

SylvainJuge: Thanks for the reminder, I completely missed it.

I'm definitely missing the context on the intent of this advice, so maybe there is something to clarify here.

For now, the way I understand things:

The `opentelemetry-semconv` artifact only contains stable semantic conventions.
The `opentelemetry-semconv-incubating` artifact only contains incubating semantic conventions.

Consumers of semantic conventions who need/want to avoid breaking change:

- should have a direct dependency on `opentelemetry-semconv`.
- might have a test dependency on `opentelemetry-semconv-incubating`.
- should inline the incubating semconv attributes in the production code, this is where they trade maintainability for stability.
  
Consumers of semantic conventions who can tolerate/handle breaking changes:

- should have a direct dependency on `opentelemetry-semconv`
- should have a direct dependency on `opentelemetry-semconv-incubating` which provides an easy way to find usages and refactor in the IDE.

For example:
- in the instrumentation agent, we already use the incubating artifact and can embrace breaking changes quite easily due to the large set of contributors who can help. However, the goal is to have as much stable as possible and sometime the incubating metrics (like the JVM metrics) are opt-in through configuration, however this option is not always possible at attribute level.
- for a module in contrib that has only two contributors and is rarely updated, keeping it stable is probably a better option, so having some duplication is a good compromise.

To me, by adding a dependency on `opentelemetry-semconv-incubating` on a module in contrib, you opt-in for some breaking changes.

Maybe here having a single version used for all the contrib modules is a bold move and having each module handle its own version (even if they are currently aligned) is a better compromise, so each module can be updated at its own pace.

SylvainJuge:
@jack-berg I now understand that having direct dependencies on the incubating attributes artifact can be problematic and why we'd like to avoid direct references to them in the published artifacts.

However, for the developer/maintainer in this repository there are definitely some benefits to have it, and I don't think that manually duplicating the incubating attributes is something we can enforce as it makes later maintenance more tedious: whenever new attributes are promoted/modified in a new semconv version, you have to `grep` the code to modify and update it, whereas with direct dependencies we can get automatic upgrades or compile time errors that are easier to manage.

Now that we have a separate artifact for incubating dependencies, we could maybe shade it in every sub-project of this repo where it's used, so it would make any artifact depending on it effectively immune to be used with a different incubating attributes artifact in the classpath.

JonasKunz (approved):

dependencyManagement/build.gradle.kts
```diff
@@ -33,6 +34,7 @@ val CORE_DEPENDENCIES = listOf(
   "com.google.errorprone:error_prone_core:${errorProneVersion}",
   "io.github.netmikey.logunit:logunit-jul:2.0.0",
   "io.opentelemetry.proto:opentelemetry-proto:1.0.0-alpha",
+  "io.opentelemetry.semconv:opentelemetry-semconv-incubating:${contribIncubatingVersion}",
```
laurit, line 37: perhaps we should also list `opentelemetry-semconv` here like it is done in the instrumentation repo https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/373224c4c09918de7eaa2aa24e456f94f3975c2f/dependencyManagement/build.gradle.kts#L52

SylvainJuge: That's a good idea, relying on the indirect dependency on the incubating artifact can be confusing.

SylvainJuge: fixed in 18869374843cfbdd60e4e08988a4ae0d87d5cd56

JonasKunz (approved):

dependencyManagement/build.gradle.kts
```diff
@@ -23,6 +23,7 @@ val errorProneVersion = "2.27.1"
 val prometheusVersion = "0.16.0"
 val mockitoVersion = "4.11.0"
 val slf4jVersion = "2.0.13"
+val contribIncubatingVersion = "1.25.0-alpha"
```
laurit, line 26: could you rename this to `semConvVersion`

SylvainJuge: fixed in 67fc874dd6297e2851edb906a6f274764a216606

laurit: thanks

laurit (approved):

jackshirazi (approved):

jack-berg (approved):

PR #1306, opened 2024-05-07
[disk-buffering] Fix mocking in test code

breedx-splk:
These problems are blocking the build on #1294 from passing.


breedx-splk:
cc @LikeTheSalad

LikeTheSalad (approved):
It's odd, though it's not the first time I see issues with mocks depending on what syntax is used so.. Thanks for taking the time!

laurit (approved):

PR #1311, opened 2024-05-09
[jfr-connection] Pin JMC version for testing

breedx-splk:
Supersedes #1287 
Supersedes #1286 

See [this discussion](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1287#issuecomment-2101577239) in #1287 about the root problem and how we're now going to just pin this jmc version to work around this.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -61,6 +61,12 @@
     {
       "matchPackagePrefixes": ["com.diffplug.spotless"],
       "groupName": "spotless packages"
+    },
+    {
+      // pinned version for compatibility with java 8 JFR parsing
+      "matchPackagePrefixes": ["org.openjdk.jmc"],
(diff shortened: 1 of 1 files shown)


.github/renovate.json5
```diff
@@ -61,6 +61,12 @@
     {
       "matchPackagePrefixes": ["com.diffplug.spotless"],
       "groupName": "spotless packages"
+    },
+    {
+      // pinned version for compatibility with java 8 JFR parsing
+      "matchPackagePrefixes": ["org.openjdk.jmc"],
+      "matchCurrentVersion": "8.3.1",
```
laurit, line 68: usually we'd use `"matchUpdateTypes": ["major"]` but this is fine too

laurit (approved):

PR #1330, opened 2024-05-31
add config support for BaggageSpanProcessor

zeitlinger:
Add config support for BaggageSpanProcessor to add it to java agent - simply by adding the library.

### DIFF
--- a/baggage-processor/README.md
+++ b/baggage-processor/README.md
@@ -42,8 +42,18 @@ Pattern pattern = Pattern.compile("^key.+");
 new BaggageSpanProcessor(baggageKey -> pattern.matcher(baggageKey).matches())
 ```
 
+## Usage with SDK auto-configuration
+
+If you are using the OpenTelemetry SDK auto-configuration, you can add the span processor this
+library to configure the span processor.
+
+| Property                                  | Description                                                                                     |
+|-------------------------------------------|-------------------------------------------------------------------------------------------------|
+| otel.traces.baggage-to-attributes.include | Add baggage items to span attributes, e.g. `key1,key2` or `*` to add all baggage items as keys. |
+
 ## Component owners
 
 - [Mike Golsmith](https://github.com/MikeGoldsmith), Honeycomb
+- [Gregor Zeitlinger](https://github.com/zeitlinger), Grafana
 
 Learn more about component owners in [component_owners.yml](../.github/component_owners.yml).
--- a/baggage-processor/README.md
+++ b/baggage-processor/README.md
@@ -49,7 +49,7 @@ library to configure the span processor.
 
 | Property                                  | Description                                                                                     |
 |-------------------------------------------|-------------------------------------------------------------------------------------------------|
-| otel.traces.baggage-to-attributes.include | Add baggage items to span attributes, e.g. `key1,key2` or `*` to add all baggage items as keys. |
+| otel.traces.baggage-to-attributes.include | Add baggage entries as span attributes, e.g. `key1,key2` or `*` to add all baggage items as keys. |
 
 ## Component owners
 
--- a/baggage-processor/README.md
+++ b/baggage-processor/README.md
@@ -47,9 +47,9 @@ new BaggageSpanProcessor(baggageKey -> pattern.matcher(baggageKey).matches())
 If you are using the OpenTelemetry SDK auto-configuration, you can add the span processor this
 library to configure the span processor.
 
-| Property                                  | Description                                                                                     |
-|-------------------------------------------|-------------------------------------------------------------------------------------------------|
-| otel.traces.baggage-to-attributes.include | Add baggage entries as span attributes, e.g. `key1,key2` or `*` to add all baggage items as keys. |
+| Property                                                         | Description                                                                                       |
+|------------------------------------------------------------------|---------------------------------------------------------------------------------------------------|
+| otel.java.experimental.span-attributes.copy-from-baggage.include | Add baggage entries as span attributes, e.g. `key1,key2` or `*` to add all baggage items as keys. |
 
 ## Component owners
 
--- a/baggage-processor/src/main/java/io/opentelemetry/contrib/baggage/processor/BaggageSpanProcessorCustomizer.java
+++ b/baggage-processor/src/main/java/io/opentelemetry/contrib/baggage/processor/BaggageSpanProcessorCustomizer.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.contrib.baggage.processor;
+
+import io.opentelemetry.sdk.autoconfigure.spi.AutoConfigurationCustomizer;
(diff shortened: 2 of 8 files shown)


zeitlinger:
@jack-berg @trask please check

MikeGoldsmith (approved):
Looks great, thanks @zeitlinger!

baggage-processor/README.md
```diff
@@ -42,8 +42,18 @@ Pattern pattern = Pattern.compile("^key.+");
...
 ```
 
+## Usage with SDK auto-configuration
+
+If you are using the OpenTelemetry SDK auto-configuration, you can add the span processor this
+library to configure the span processor.
+
+| Property                                  | Description                                                                                     |
+|-------------------------------------------|-------------------------------------------------------------------------------------------------|
+| otel.traces.baggage-to-attributes.include | Add baggage items to span attributes, e.g. `key1,key2` or `*` to add all baggage items as keys. |
```
MikeGoldsmith, line 52: I'm unsure about `baggage-to-attributes` but also unsure if there's a convention for the property name so can't really offer a better name. Maybe `copy-baggage-attributes`?

Small suggestion for property description.
```suggestion
| otel.traces.baggage-to-attributes.include | Add baggage entries as span attributes, e.g. `key1,key2` or `*` to add all baggage items as keys. |
```

zeitlinger: I want to check what other languages have done

MikeGoldsmith: I don't believe other languages have an automatic configuration component that can use the baggage span processor yet.

breedx-splk (approved):
LGTM

baggage-processor/README.md
jack-berg, line 52:
```suggestion
| otel.java.experimental.traces.baggage-to-attributes.include | Add baggage items to span attributes, e.g. `key1,key2` or `*` to add all baggage items as keys. |
```
Including java prefix because this is language specific. Following guidance [here](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#language-specific-environment-variables).

Including experimental prefix because this whole processor and config is experimental. See [autoconfigure docs](https://github.com/open-telemetry/opentelemetry-java/tree/main/sdk-extensions/autoconfigure) for examples.

baggage-processor/src/main/java/io/opentelemetry/contrib/baggage/processor/BaggageSpanProcessorCustomizer.java
```diff
@@ -0,0 +1,41 @@
...
+import io.opentelemetry.sdk.autoconfigure.spi.AutoConfigurationCustomizer;
+import io.opentelemetry.sdk.autoconfigure.spi.AutoConfigurationCustomizerProvider;
+import io.opentelemetry.sdk.autoconfigure.spi.ConfigProperties;
+import io.opentelemetry.sdk.trace.SdkTracerProviderBuilder;
+import java.util.List;
+
+public class BaggageSpanProcessorCustomizer implements AutoConfigurationCustomizerProvider {
+  @Override
+  public void customize(AutoConfigurationCustomizer autoConfigurationCustomizer) {
+    autoConfigurationCustomizer.addTracerProviderCustomizer(
```
jack-berg, line 17: For a moment I thought this wouldn't work because this customizers are [applied after the batch exporter is registered](https://github.com/open-telemetry/opentelemetry-java/blob/788347e262b34ccf287a5b32a9976eb43e170c5f/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdkBuilder.java#L473-L475), but then I realized that order of registration doesn't matter for span processors which want to take advantage of modifying the span on start.

We would need to make changes in the SDK if this was a `BaggageLogRecordProcessor`: Either providing a hook to customize the logger provider before batch processor is added, or by extending `SdkLoggerProviderBuilder` with a method to insert a processor at the front of the queue.

zeitlinger: > Either providing a hook to customize the logger provider before batch processor is added

I think this is a great idea - would make sense for all signals

BTW, I've made the test end-to-end to cover that "onStast" is actually called

baggage-processor/src/test/java/io/opentelemetry/contrib/baggage/processor/BaggageSpanProcessorCustomizerTest.java
```diff
@@ -0,0 +1,149 @@
...
+  @Test
+  void test_customizer() {
+    assertCustomizer(Collections.emptyMap(), span -> span.hasTotalAttributeCount(0));
+    assertCustomizer(
+        Collections.singletonMap(
+            "otel.java.experimental.span-attributes.copy-from-baggage.include", "key"),
+        span -> span.hasAttribute(AttributeKey.stringKey("key"), "value"));
+  }
+
+  private static void assertCustomizer(
```
zeitlinger, line 54: @jack-berg should we put this method in the sdk? it's the only way (I know of) to test the autoconfig module.

zeitlinger: I gave it a try: https://github.com/open-telemetry/opentelemetry-java/pull/6496

- it uses optional dependencies
- awaitility only for traces can probably be improved
- no tests yet

baggage-processor/src/main/java/io/opentelemetry/contrib/baggage/processor/BaggageSpanProcessorCustomizer.java
```diff
@@ -0,0 +1,42 @@
...
+    if (keys.isEmpty()) {
+      return;
+    }
+
+    sdkTracerProviderBuilder.addSpanProcessor(createProcessor(keys));
+  }
+
+  static BaggageSpanProcessor createProcessor(List<String> keys) {
+    if (keys.size() == 1 && keys.get(0).equals("*")) {
+      return new BaggageSpanProcessor(BaggageSpanProcessor.allowAllBaggageKeys);
```
trask, line 38: unrelated, I'd suggest turning `allowAllBaggageKeys` into a static method returning `BaggageSpanProcessor`, which matches the pattern we've used elsewhere, e.g. https://github.com/open-telemetry/opentelemetry-java/blob/4680146024b83892c14e5df4e08158f1f602c75c/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/samplers/Sampler.java#L29

and then you can do the simpler here:

```suggestion
      return BaggageSpanProcessor.allowAllBaggageKeys();
```

trask (approved):

baggage-processor/src/main/java/io/opentelemetry/contrib/baggage/processor/BaggageSpanProcessorCustomizer.java
zeitlinger: I like it - added and deprecated the old field

Issue #1410, opened 2024-08-12
Closing a JFR recording may throw an Exception


dsgrieve:
### Component(s)

jfr-connection

## Description
Calling the Recording#close() method may throw an exception if the FlightRecorder MXBean closeRecording function fails. 
The behavior of the original microsoft/jfr-streaming code was to swallow the exception. The behavior of the contributed code is different and this has caused a regression in an application that uses this module. 

## Expected Result
The close() method on a Recording should not throw an exception. 

### Component version

1.37.0

dsgrieve:
PR #1412

trask:
Resolved by #1412

PR #1412, opened 2024-08-13
jfr-connection: Recording close should not throw exception

dsgrieve:
**Description:**
Recording close method would throw an exception if the FlightRecorderConnection closeRecording method threw an exception. This is a regression from the microsoft/jfr-streaming code that was contributed.
 
**Existing Issue(s):**

Closing a JFR recording may throw an Exception [#1410](https://github.com/open-telemetry/opentelemetry-java-contrib/issues/1410)
**Testing:**

Existing tests were examined and no changes were necessary. 

**Documentation:**

No changes to documentation were needed. However, several methods in FlightRecorderMXBeanConnection stated that the method was not available in Java 8. The javadoc for these methods was changed to more correctly state that the command is not available through the Diagnostic Command MBean.

**Outstanding items:**

None.

### DIFF
--- a/jfr-connection/src/main/java/io/opentelemetry/contrib/jfr/connection/Recording.java
+++ b/jfr-connection/src/main/java/io/opentelemetry/contrib/jfr/connection/Recording.java
@@ -270,8 +270,14 @@ public class Recording implements AutoCloseable {
         connection.stopRecording(id);
       } catch (IOException | JfrConnectionException ignored) {
         // Stopping the recording is best-effort
-      } finally {
+      }
+    }
+    if (oldState == State.STOPPED || oldState == State.RECORDING) {
+      try {
         connection.closeRecording(id);
+      } catch (IOException | JfrConnectionException | UnsupportedOperationException ignored) {
+        // Closing the recording is best-effort
+        // FlightRecorderDiagnosticCommandConnection close throws UnsupportedOperationException
       }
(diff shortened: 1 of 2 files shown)


jfr-connection/src/main/java/io/opentelemetry/contrib/jfr/connection/Recording.java
```diff
@@ -270,8 +270,14 @@ public void close() throws IOException, JfrConnectionException {
...
         // Stopping the recording is best-effort
-      } finally {
+      }
+    }
+    if (oldState == State.STOPPED || oldState == State.RECORDING) {
+      try {
         connection.closeRecording(id);
+      } catch (IOException | JfrConnectionException | UnsupportedOperationException ignored) {
+        // Closing the recording is best-effort
+        // FlightRecorderDiagnosticCommandConnection close throws UnsupportedOperationException
```
breedx-splk, line 280: Consider logging a warning. This shouldn't happen very often, and if it does it's nice to have some indication or remnant of a deeper underlying problem. Root cause analysis can be difficult without this.

breedx-splk (approved):
This seems good to me. Maybe it was just an oversight or a mistake, but it does indeed seem odd that it was swallowing the exception for `stopRecording()` but not `closeRecording()`. Thanks!

PR #1414, opened 2024-08-14
add span stacktrace config option

SylvainJuge:
**Description:**

Span stacktrace module can't use autoconfiguration for now as it relies on a strict ordering of the span processors. The addition of `OnEnding` callback would unblock this but it's not available yet (https://github.com/open-telemetry/opentelemetry-java/pull/6367).

While autoconfiguration is still not supported, this PR adds a way to provide a `ConfigProperties` for configuration, which allows to
- define configuration option and default value
- implement configuration in a consistent way across usages

Configuration option name `otel.java.experimental.span-stacktrace.min.duration`
Default value: `5ms`

### DIFF
--- a/span-stacktrace/README.md
+++ b/span-stacktrace/README.md
@@ -47,6 +47,7 @@ Even if autoconfiguration is not yet supported, usages of this module must use t
 allow consistent configuration across usages.
 
 The following constants are provided as a convenience:
+
 - `StackTraceSpanProcessor.CONFIG_MIN_DURATION`
 - `StackTraceSpanProcessor.CONFIG_MIN_DURATION_DEFAULT`
 
--- a/span-stacktrace/README.md
+++ b/span-stacktrace/README.md
@@ -54,7 +54,7 @@ The following constants are provided as a convenience:
 This means the default for this feature will capture a stacktrace for all spans that last 5ms or more,
 disabling it must be allowed by setting a negative value to `otel.span.stacktrace.min.duration`.
 Setting `otel.span.stacktrace.min.duration` to zero means a stacktrace will be captured for all
-spans without span duration limit.
+spans.
 
 ## Component owners
 
--- a/span-stacktrace/README.md
+++ b/span-stacktrace/README.md
@@ -50,6 +50,9 @@ The following constants are provided as a convenience:
 - `StackTraceSpanProcessor.CONFIG_MIN_DURATION`
 - `StackTraceSpanProcessor.CONFIG_MIN_DURATION_DEFAULT`
 
+It means default this feature will capture a stacktrace for all spans that last 5ms or more,
+disabling it must be allowed by setting `otel.span.stacktrace.min.duration` zero or less.
+
 ## Component owners
 
 - [Jack Shirazi](https://github.com/jackshirazi), Elastic
--- a/span-stacktrace/README.md
+++ b/span-stacktrace/README.md
@@ -51,9 +51,9 @@ The following constants are provided as a convenience:
 - `StackTraceSpanProcessor.CONFIG_MIN_DURATION`
 - `StackTraceSpanProcessor.CONFIG_MIN_DURATION_DEFAULT`
 
-It means default this feature will capture a stacktrace for all spans that last 5ms or more,
+This means the default for this feature will capture a stacktrace for all spans that last 5ms or more,
 disabling it must be allowed by setting a negative value to `otel.span.stacktrace.min.duration`.
-Setting `otel.span.stacktrace.min.duration` to zero means stacktrace will be captured for all
+Setting `otel.span.stacktrace.min.duration` to zero means a stacktrace will be captured for all
 spans without span duration limit.
 
 ## Component owners
--- a/span-stacktrace/README.md
+++ b/span-stacktrace/README.md
@@ -43,7 +43,7 @@ OpenTelemetrySdk sdk = OpenTelemetrySdk.builder().setTracerProvider(tracerProvid
 ### Configuration
 
 Even if autoconfiguration is not yet supported, usages of this module must use the
-`otel.span.stacktrace.min.duration` configuration option (in nanoseconds, defaults to 5ms) to
+`otel.span.stacktrace.min.duration` configuration option (defaults to 5ms) to
 allow consistent configuration across usages.
 
 The following constants are provided as a convenience:
--- a/span-stacktrace/README.md
+++ b/span-stacktrace/README.md
@@ -40,6 +40,16 @@ SdkTracerProvider tracerProvider = SdkTracerProvider.builder()
 OpenTelemetrySdk sdk = OpenTelemetrySdk.builder().setTracerProvider(tracerProvider).build();
 ```
 
+### Configuration
+
+Even if autoconfiguration is not yet supported, usages of this module must use the
+`otel.span.stacktrace.min.duration` configuration option (in nanoseconds, defaults to 5ms) to
+allow consistent configuration across usages.
+
+The following constants are provided as a convenience:
+- `StackTraceSpanProcessor.CONFIG_MIN_DURATION`
+- `StackTraceSpanProcessor.CONFIG_MIN_DURATION_DEFAULT`
+
 ## Component owners
 
 - [Jack Shirazi](https://github.com/jackshirazi), Elastic
--- a/span-stacktrace/README.md
+++ b/span-stacktrace/README.md
@@ -27,6 +27,10 @@ will be ignored.
 InMemorySpanExporter spansExporter = InMemorySpanExporter.create();
 SpanProcessor exportProcessor = SimpleSpanProcessor.create(spansExporter);
 
+Map<String, String> configMap = new HashMap<>();
+configMap.put("otel.span.stacktrace.min.duration", "1ms");
+ConfigProperties config = DefaultConfigProperties.createFromMap(configMap);
+
 Predicate<ReadableSpan> filterPredicate = readableSpan -> {
   if(readableSpan.getAttribute(AttributeKey.stringKey("ignorespan")) != null){
     return false;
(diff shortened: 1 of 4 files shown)


span-stacktrace/README.md
```diff
@@ -40,6 +40,16 @@ SdkTracerProvider tracerProvider = SdkTracerProvider.builder()
 OpenTelemetrySdk sdk = OpenTelemetrySdk.builder().setTracerProvider(tracerProvider).build();
 ```
 
+### Configuration
+
+Even if autoconfiguration is not yet supported, usages of this module must use the
+`otel.span.stacktrace.min.duration` configuration option (in nanoseconds, defaults to 5ms) to
```
JonasKunz, line 46: nanoseconds, is that intended?

JonasKunz (approved):

span-stacktrace/README.md
SylvainJuge: Yes, the span duration (latency) is already measured in nanoseconds so it's simpler to keep the same unit.

SylvainJuge: In fact here when the value is parsed as `java.time.Duration` from configuration it can be set to any unit, so only the default value unit needs to be specified.

span-stacktrace/README.md
```diff
@@ -40,6 +40,22 @@ SdkTracerProvider tracerProvider = SdkTracerProvider.builder()
...
+Even if autoconfiguration is not yet supported, usages of this module must use the
+`otel.span.stacktrace.min.duration` configuration option (defaults to 5ms) to
+allow consistent configuration across usages.
+
+The following constants are provided as a convenience:
+
+- `StackTraceSpanProcessor.CONFIG_MIN_DURATION`
+- `StackTraceSpanProcessor.CONFIG_MIN_DURATION_DEFAULT`
+
+It means default this feature will capture a stacktrace for all spans that last 5ms or more,
```
jackshirazi, line 54:
```suggestion
This means the default for this feature will capture a stacktrace for all spans that last 5ms or more,
```

span-stacktrace/README.md
```diff
@@ -40,6 +40,22 @@ SdkTracerProvider tracerProvider = SdkTracerProvider.builder()
...
+disabling it must be allowed by setting a negative value to `otel.span.stacktrace.min.duration`.
+Setting `otel.span.stacktrace.min.duration` to zero means stacktrace will be captured for all
```
jackshirazi, line 56:
```suggestion
Setting `otel.span.stacktrace.min.duration` to zero means a stacktrace will be captured for all
```

span-stacktrace/README.md
```diff
@@ -40,6 +40,22 @@ SdkTracerProvider tracerProvider = SdkTracerProvider.builder()
...
+spans without span duration limit.
```
jackshirazi, line 57: what do you mean by `without span duration limit`?

SylvainJuge: I mean that every span will have a stack trace without first checking the span duration to know if the stack trace overhead is worth it.

jackshirazi: so every span. The additional words seem to be irrelevant? If you agree ...
```suggestion
spans.
```

span-stacktrace/README.md
```diff
@@ -40,6 +40,22 @@ SdkTracerProvider tracerProvider = SdkTracerProvider.builder()
...
+allow consistent configuration across usages.
+
+The following constants are provided as a convenience:
+
+- `StackTraceSpanProcessor.CONFIG_MIN_DURATION`
+- `StackTraceSpanProcessor.CONFIG_MIN_DURATION_DEFAULT`
+
+This means the default for this feature will capture a stacktrace for all spans that last 5ms or more,
+disabling it must be allowed by setting a negative value to `otel.span.stacktrace.min.duration`.
+Setting `otel.span.stacktrace.min.duration` to zero means a stacktrace will be captured for all
```
jackshirazi, line 56:
```suggestion
Setting `otel.span.stacktrace.min.duration` to zero means a stacktrace will be captured for every
```

span-stacktrace/README.md
```diff
@@ -40,6 +40,22 @@ SdkTracerProvider tracerProvider = SdkTracerProvider.builder()
...
+spans.
```
jackshirazi, line 57:
```suggestion
Setting `otel.span.stacktrace.min.duration` to zero means a stacktrace will be captured with every
span.
```

span-stacktrace/README.md
```diff
@@ -20,26 +20,38 @@ section below to configure it.
...
-    .addSpanProcessor(new StackTraceSpanProcessor(exportProcessor, 1000, filterPredicate))
+    .addSpanProcessor(new StackTraceSpanProcessor(exportProcessor, config, filterPredicate))
     .build();
 
 OpenTelemetrySdk sdk = OpenTelemetrySdk.builder().setTracerProvider(tracerProvider).build();
 ```
 
+### Configuration
+
+The `otel.span.stacktrace.min.duration` configuration option (defaults to 5ms) allows to configure
```
jackshirazi, line 49:
```suggestion
The `otel.span.stacktrace.min.duration` configuration option (defaults to 5ms) allows configuring
```

jackshirazi (approved):

span-stacktrace/README.md
```diff
@@ -20,26 +20,38 @@ section below to configure it.
...
-the spans that have a duration >= 1000 ns. The spans that have an `ignorespan` string attribute
+the spans that have a duration >= 1 ms. The spans that have an `ignorespan` string attribute
 will be ignored.
 
 ```java
 InMemorySpanExporter spansExporter = InMemorySpanExporter.create();
 SpanProcessor exportProcessor = SimpleSpanProcessor.create(spansExporter);
 
+Map<String, String> configMap = new HashMap<>();
+configMap.put("otel.span.stacktrace.min.duration", "1ms");
```
trask, line 31:
```suggestion
configMap.put("otel.java.experimental.span-stacktrace.min.duration", "1ms");
```

jackshirazi: Is this really needed?

SylvainJuge: it's only necessary if you want to override the default, but that's optional if you think it's simpler let's remove it.

SylvainJuge: Sorry I completely misread the discussion here, so please ignore my previous msg.

There are quite a few other config options patterns in the contrib repo (non-exhaustive list)
- `otel.jmx.*` used by the current JMX gatherer, but unlikely to conflict with anything due to the nature of the gatherer.
- `otel.inferred.spans.*` for inferred spans, which means we would have to migrate those for consistency
- `otel.java.experimental.*` pattern used only by the baggage to span attribute module until we also use it here.

So it seems that so far the `otel.<feature>` where `<feature>` somehow matches the module name is the most common option.

What I don't like with the `experimental` is that is puts a "stability/maturity" semantic into the configuration. For example, the inferred spans are quite mature because we migrated it from our own agent.

SylvainJuge: There are rules defined or about to be defined in semantic conventions for attribute and metric names, could we make most config options also fit the existing ones ?

For example, despite the shared ownership of contrib repo, would it make sense to include those in the documentation for Java https://opentelemetry.io/docs/languages/java/configuration/, it already contains:
- general SDK configuration with implementation details for Java with `otel.` prefix
- Java specific configuration for the SDK and Instrumentation (with the `otel.java.` prefix

With that scheme, I think that maybe having a `otel.java.<feature>.` pattern as prefix for contrib would be nice because:
- we can have a clear list of features weither they are in contrib or not (with appropriate links)
- from the person in charge of configuration, the fact it's part of contrib is an implementation detail, if the feature is later moved to a separate artifact is not relevant to how it's configured.

SylvainJuge: After discussing this on our side (Elastic), we think we should separate this discussion about config and feature maturity from this PR, we will follow-up in future Java SIG meetings.

As a consequence, for now we'll just update it to `otel.java.experimental.span-stacktrace.min.duration` and we can enhance it later.

SylvainJuge:
As discussed yesterday SIG we can't mark the feature as stable in part because it relies on an unstable semconv attribute `code.stacktrace`.

While we work on making this `code.stacktrace` a stable semconv attribute, we should still be able to merge this PR with an experimental configuration option `otel.java.experimental.span-stacktrace.min.duration`, so at least add support for some configuration.

So here I think this PR is ready for review and we should be able to merge it as-is until we manage to make progress on making the feature stable.

@trask @jack-berg does it sounds like a good next step for this ?

span-stacktrace/README.md
```diff
@@ -20,26 +20,38 @@ section below to configure it.
...
-    .addSpanProcessor(new StackTraceSpanProcessor(exportProcessor, 1000, filterPredicate))
+    .addSpanProcessor(new StackTraceSpanProcessor(exportProcessor, config, filterPredicate))
     .build();
 
 OpenTelemetrySdk sdk = OpenTelemetrySdk.builder().setTracerProvider(tracerProvider).build();
 ```
 
+### Configuration
+
+The `otel.java.experimental.span-stacktrace.min.duration` configuration option (defaults to 5ms) allows configuring
```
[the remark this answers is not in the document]
SylvainJuge: I'm fine to remove `-` and replace it with `.`, for now we just need to have one configuration option to start, we will of course need to revise this later when making this feature stable.

I think the suggestion to use `-` in the feature name came from what we currently have for `otel.java.experimental.span-attributes.copy-from-baggage.include` in the [baggage-processor](https://github.com/open-telemetry/opentelemetry-java-contrib/tree/main/baggage-processor), which by the way isn't really consistent as we have `otel.java.experimental.span-attributes` as prefix for a module named `baggage-processor`.

As long as things are not marked as stable, I think those minor inconsistencies are fine and acceptable, especially in the contrib repo where things are mostly alpha.

-----

However we probably need to have some proper definition somewhere about those configuration options, for example to at least define what are the main namespaces like `otel.java.*`, `otel.javaagent.*` or `otel.java.experimental.*` or even define a global registry with all the possible options to ensure there isn't any conflict between all the repositories. Also, this might be something also happening in other platforms & otel components SIGs.

However that's definitely something that will take a while and should be covered separately, and I know that it sounds like "let's start another SIG about configuration options" :smile: .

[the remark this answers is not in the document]
jack-berg: Ha - well unfortunately I doubt we'll get any additional clarity around property naming anytime soon. There's already a dedicated config sig (which I lead), focussing on building out [declarative configuration](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/configuration#declarative-configuration) which aims to replace the system property / env var config scheme as the dominant mechanism for configuration. That's where all the configuration effort is being spent these days.

I think you're right to stick with precedents that already exist in this repo - I wasn't aware of those.

SylvainJuge:
Woops, I deleted the branch by accident, thus just re-opening it with this comment.

trask (approved):

Issue #1416, opened 2024-08-15
otel.resource.attributes and OTEL_RESOURCE_ATTRIBUTES doesn't work


marquiswang:
### Component(s)

maven-extension

## Description
Setting the otel.resource.attributes doesn't work, contrary to documentation.

I tested a few versions - this was working in 1.27, and stopped working in 1.28 and does not work in latest.

## Steps to Reproduce
Configure the `opentelemetry-maven-extension` (I'm exporting to Honeycomb, but I don't expect it to make a difference).

Run:
```
mvn clean -Dotel.resource.attributes="key1=val1,key2=val2" 
```

## Expected Result
These attributes are exported to the otel collector.

## Actual Result
These attributes are missing.

### Component version

v1.28.0 - v.37.0

### Log output

When building with 1.27 and `-X`, I get 
```
[DEBUG] ContextStorageClass=io.opentelemetry.context.ThreadLocalContextStorage, identity=375273627 classloader=ClassRealm[coreExtensi
on>io.opentelemetry.contrib:opentelemetry-maven-extension:1.27.0-alpha, parent: ClassRealm[plexus.core, parent: null]] codeLocation=f
ile:/Users/mwang/.m2/repository/io/opentelemetry/contrib/opentelemetry-maven-extension/1.27.0-alpha/opentelemetry-maven-extension-1.2
7.0-alpha.jar
[DEBUG] OpenTelemetry: Initialize OpenTelemetrySdkService v1.27.0-alpha...
[DEBUG] OpenTelemetry: OpenTelemetry SDK initialized with  Configuration: otel.traces.exporter=otlp, otel.exporter.otlp.endpoint=https://api.honeycomb.io, otel.resource.attributes=key1=val1,key2=val2, Resource: {key1="val1", key2="val2", service.name="maven", service.version="3.9.8", telemetry.sdk.language="java", telemetry.sdk.name="opentelemetry", telemetry.sdk.version="1.27.0"}
[DEBUG] OpenTelemetry: OpenTelemetry extension registered as execution listener. InitialExecutionListener: org.apache.maven.eventspy.internal.EventSpyExecutionListener@7e7fe6d
[DEBUG] OpenTelemetry: afterSessionStart
```

When building with 1.28, I just see:

```
[DEBUG] OpenTelemetry: Initialize OpenTelemetrySdkService v1.28.0-alpha...
[DEBUG] OpenTelemetry: OpenTelemetry SDK initialized
[...]

PR #1434, opened 2024-08-26
Load OTel SDK config from environment variables and system properties.…

cyrille-leclerc:
**Description:**

Load OTel SDK config from environment variables and system properties.

**Existing Issue(s):**

* #1416

**Testing:**

See added unit tests

**Documentation:**

< Describe the documentation added.
  Please be sure to modify relevant existing documentation if needed. >

**Outstanding items:**


maven-extension/src/main/java/io/opentelemetry/maven/OpenTelemetrySdkService.java
```diff
@@ -47,26 +55,51 @@ public OpenTelemetrySdkService() {
...
-    properties.put("otel.traces.exporter", "none");
-    properties.put("otel.metrics.exporter", "none");
-    properties.put("otel.logs.exporter", "none");
-
     AutoConfiguredOpenTelemetrySdk autoConfiguredOpenTelemetrySdk =
         AutoConfiguredOpenTelemetrySdk.builder()
             .setServiceClassLoader(getClass().getClassLoader())
-            .addPropertiesSupplier(() -> properties)
+            .addPropertiesCustomizer(configProperties -> {
+              // Change default of "otel.[traces,metrics,logs].exporter" from "otlp" to "none"
```
SylvainJuge, line 62: [minor] Could you elaborate a bit in a comment on why we need to override the default here ? For example, if I understand correctly when there is no endpoint configured then we assume the extension should silently skip exporting signals.

cyrille-leclerc: Thanks for the review Sylvain, please see updated comment

trask:
@cyrille-leclerc just a heads up looks like there's a failing test

cyrille-leclerc:
Thanks @trask , unit test fixed.

trask:
cc @kenfinnigan in case you have a chance to review

cyrille-leclerc:
@SylvainJuge could you by any chance review this PR?

maven-extension/src/main/java/io/opentelemetry/maven/OpenTelemetrySdkService.java
```diff
@@ -47,26 +56,69 @@ public OpenTelemetrySdkService() {
...
+                            "otlp", configProperties.getString("otel.metrics.exporter", "otlp"))
+                        && configProperties.getString("otel.exporter.otlp.metrics.endpoint")
+                            == null) {
+                      properties.put("otel.metrics.exporter", "none");
+                    }
+                    if (Objects.equals(
+                            "otlp", configProperties.getString("otel.logs.exporter", "otlp"))
+                        && configProperties.getString("otel.exporter.otlp.logs.endpoint") == null) {
+                      properties.put("otel.logs.exporter", "none");
+                    }
```
SylvainJuge, line 89: [minor] it would be more readable to create a helper method to check this as it takes quite an effort to read and understand it that we check two config options to not be explicitly set and set one of them in the modified config, all of that 3 times with close but slightly different variables for logs, traces and metrics.

maven-extension/src/main/java/io/opentelemetry/maven/OpenTelemetrySdkService.java
```diff
@@ -47,26 +56,69 @@ public OpenTelemetrySdkService() {
...
+                })
             .disableShutdownHook()
             .build();
 
+    if (this.resource == null) {
+      this.resource = Resource.empty();
+    }
+    if (this.configProperties == null) {
+      this.configProperties = DefaultConfigProperties.createFromMap(Collections.emptyMap());
+    }
```
SylvainJuge, line 116: [minor] in order to avoid any ambiguity and possible concurrent modification by the SDK init (I'm pretty sure it's synchronous but we'd better be safe than sorry), it would be better to set `this.resource` and `this.configProperties` before the call to `AutoConfiguredOpenTelemetrySdk.builder()`.

maven-extension/src/test/java/io/opentelemetry/maven/OpenTelemetrySdkServiceTest.java
```diff
@@ -5,42 +5,117 @@
...
 
-  /** Verify default `service.name` */
+  /** Verify default config */
   @Test
-  @Disabled
   public void testDefaultConfiguration() {
-    testConfiguration("maven");
+    System.clearProperty("otel.exporter.otlp.endpoint");
+    System.clearProperty("otel.service.name");
+    System.clearProperty("otel.resource.attributes");
```
SylvainJuge, line 22: [minor] As an alternative you could use JUnit `@ClearSystemProperty` and `@SetSystemProperty` as it helps to prevent global state side-effects and allows to preserve existing values (if any), see [docs](https://junit-pioneer.org/docs/system-properties/) for more details.

cyrille-leclerc:
Thanks @SylvainJuge , I'll improve ASAP

maven-extension/src/test/java/io/opentelemetry/maven/OpenTelemetrySdkServiceTest.java
cyrille-leclerc: Great idea.
JUnit Pioneer requires Java 11 ([here](https://github.com/junit-pioneer/junit-pioneer/blob/v2.3.0/build.gradle.kts#L30)) when we support Java 8.
I captured a TODO to simplify the code once otel-java-contrib bumps to Java 11.

trask: just fyi, we don't have any plans to bump the whole repo to Java 11, but we can bump an individual module to Java 11 if it's warranted, e.g.

https://github.com/open-telemetry/opentelemetry-java-contrib/blob/d1cc766de244c9737e1c5b3859b723e82003c078/static-instrumenter/agent-instrumenter/build.gradle.kts#L24-L26

maven-extension/src/main/java/io/opentelemetry/maven/OpenTelemetrySdkService.java
cyrille-leclerc: Please review the helper method I introduced

maven-extension/src/main/java/io/opentelemetry/maven/OpenTelemetrySdkService.java
```diff
@@ -47,32 +55,94 @@ public OpenTelemetrySdkService() {
...
-    properties.put("otel.traces.exporter", "none");
-    properties.put("otel.metrics.exporter", "none");
-    properties.put("otel.logs.exporter", "none");
-
     AutoConfiguredOpenTelemetrySdk autoConfiguredOpenTelemetrySdk =
         AutoConfiguredOpenTelemetrySdk.builder()
             .setServiceClassLoader(getClass().getClassLoader())
-            .addPropertiesSupplier(() -> properties)
+            .addPropertiesCustomizer(
+                OpenTelemetrySdkService::requireExplicitConfigOfTheOtlpExporter)
```
cyrille-leclerc, line 62: @SylvainJuge please review introduced helper method

maven-extension/src/main/java/io/opentelemetry/maven/OpenTelemetrySdkService.java
```diff
@@ -47,32 +55,90 @@ public OpenTelemetrySdkService() {
...
-    // The impacts are
-    // * If no otel exporter settings are passed, then the Maven extension will not export
-    //   rather than exporting on OTLP GRPC to http://localhost:4317
-    // * If OTEL_EXPORTER_OTLP_ENDPOINT is defined but OTEL_[TRACES,METRICS,LOGS]_EXPORTER,
-    //   is not, then don't export
-    Map<String, String> properties = new HashMap<>();
-    properties.put("otel.traces.exporter", "none");
-    properties.put("otel.metrics.exporter", "none");
-    properties.put("otel.logs.exporter", "none");
+    this.resource = Resource.empty();
```
cyrille-leclerc, line 58: @SylvainJuge please review initialization to fallback values done at the beginning for the code block

maven-extension/src/test/java/io/opentelemetry/maven/OpenTelemetrySdkServiceTest.java
cyrille-leclerc: Thanks @trask , I updated the comment in the code indicating that there was no plan to bump to Java 11.

maven-extension/src/test/java/io/opentelemetry/maven/OpenTelemetrySdkServiceTest.java
```diff
@@ -5,42 +5,121 @@
...
-  /** Verify overwritten `service.name` */
+  /** Verify overwritten `service.name`,`key1` and `key2` */
   @Test
-  @Disabled
-  public void testOverwrittenConfiguration() {
+  public void testOverwrittenResourceAttributes() {
     System.setProperty("otel.service.name", "my-maven");
-    try {
-      testConfiguration("my-maven");
+    System.setProperty("otel.resource.attributes", "key1=val1,key2=val2");
```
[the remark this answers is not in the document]
cyrille-leclerc: Could we postpone this improvement? I lack time to clean this up unfortunately.

maven-extension/src/main/java/io/opentelemetry/maven/OpenTelemetrySdkService.java
```diff
@@ -47,32 +55,90 @@ public OpenTelemetrySdkService() {
...
+                  return this.resource;
+                })
             .disableShutdownHook()
             .build();
 
     this.openTelemetrySdk = autoConfiguredOpenTelemetrySdk.getOpenTelemetrySdk();
 
+    logger.debug("OpenTelemetry: OpenTelemetrySdkService initialized, resource:{}", resource);
+
+    // TODO should we replace `getBooleanConfig(name)` by `configProperties.getBoolean(name)`?
```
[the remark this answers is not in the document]
cyrille-leclerc: Fixed

maven-extension/src/test/java/io/opentelemetry/maven/OpenTelemetrySdkServiceTest.java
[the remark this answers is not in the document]
SylvainJuge: Yes, we can definitely postpone that to later improvement. Just to keep it safe once this test class has executed adding an `@AfterAll` method to ensure there are no side effects on things that execute after.

```
System.clearProperty("otel.exporter.otlp.endpoint");
System.clearProperty("otel.service.name");
System.clearProperty("otel.resource.attributes");
```

maven-extension/src/main/java/io/opentelemetry/maven/AutoConfigureUtil2.java
```diff
@@ -0,0 +1,35 @@
...
+
+  private AutoConfigureUtil2() {}
+
+  /**
+   * Returns the {@link Resource} that was autoconfigured.
+   *
+   * <p>Inspired by {@link
+   * io.opentelemetry.sdk.autoconfigure.internal.AutoConfigureUtil#getConfig(AutoConfiguredOpenTelemetrySdk)}
+   */
+  public static Resource getResource(
```
SylvainJuge, line 24: [minor] can you elaborate a bit why do we call this method and what it allows to do here ? In addition, is there a follow-up action to make this method publicly accessible in the SDK if there are other similar use-cases ?

cyrille-leclerc: Great catch. Comment added in he code of the `OpenTelemetrySdkService`. The use cases are:
* Debugging & troubleshooting helping find traces if they are missing reporting the `service.name` and `service.namespace`
* Testing: before this PR, we misconfigured the AutoConfigureOtlSdk and overwriting resource attributes through system properties or env vars failed.

> In addition, is there a follow-up action to make this method publicly accessible in the SDK if there are other similar use-cases ?
Valid point, I didn't bring thes us cases to the Java SIG as I did in the past for other SDK config topics. We have the same debugging/troubleshooting use case in the Jenkins OTel plugin.

maven-extension/src/test/java/io/opentelemetry/maven/OpenTelemetrySdkServiceTest.java
```diff
@@ -5,42 +5,132 @@
...
-    // openTelemetrySdkService.autoConfiguredOpenTelemetrySdk.getResource();
-    //      assertThat(resource.getAttribute(ResourceAttributes.SERVICE_NAME))
-    //          .isEqualTo(expectedServiceName);
-    //    } finally {
-    //      openTelemetrySdkService.dispose();
-    //      GlobalOpenTelemetry.resetForTest();
-    //      GlobalEventEmitterProvider.resetForTest();
-    //    }
+  @AfterAll
+  static void afterAll() {
```
cyrille-leclerc, line 128: @SylvainJuge please review

maven-extension/src/test/java/io/opentelemetry/maven/OpenTelemetrySdkServiceTest.java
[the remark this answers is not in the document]
cyrille-leclerc: Great suggestion. Please review below.

SylvainJuge (approved):
LGTM, just make sure to open an issue/PR to make the `AutoConfiguredOpenTelemetrySdk#getResource` publicly accessible, linking to this PR would help provide context why it's needed.

cyrille-leclerc:
@SylvainJuge thanks, FYI I created:
* https://github.com/open-telemetry/opentelemetry-java/issues/6935

trask (approved):

Issue #1362, opened 2024-07-03
Refactor & merge JMX gatherer & insight implementations


SylvainJuge:
As discussed in June 20th Java SIG meeting, we have two distinct ways to capture JMX metrics:
- with instrumentation agent through the [JMX insight](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/jmx-metrics/javaagent/README.md) feature
- with contrib through the [JMX gatherer](https://github.com/open-telemetry/opentelemetry-java-contrib/tree/main/jmx-metrics) (in this repository)

As a consequence, we have:
- JMX gatherer is using groovy scripts for metrics definition
- JMX insight is using YAML files for metrics definition
- both JMX insight and JMX gatherer allow to provide custom user-provided configuration using respectively a groovy script and a yaml file
- JMX gatherer provides ready-to-use metrics configuration for some target systems (tomcat, kafka, cassandra, ...)
- JMX insight provides ready-to-use metrics configuration for a few of the target systems supported by JMX gatherer.

In addition to the duplication of code, the metrics definition themselves are duplicated in two different incompatible formats (YAML and Groovy scripts), which makes the maintenance even harder and further complicates their evolution. For example trying to change definitions in instrumentation like in [opentelemetry-java-instrumentation#11621](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/11621) becomes more complicated than it should.

JMX gatherer allows to capture JMX metrics from outside of the JVM and is used in the collector-contrib repository. Usages of it can be found by searching for usages of [`otel.jmx.service.url`](https://github.com/search?q=%22otel.jmx.service.url%22&type=code) on GH.

JMX gatherer is used in the [opentelemetry-collector-contrib](https://github.com/open-telemetry/opentelemetry-collector-contrib) repository by the [`jmxreceiver`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/eabe829fcb634e72b1155faced62b7a33592e8dd/receiver/jmxreceiver/README.md?plain=1#L73)

### Proposal

- reuse YAML-based implementation of JMX insight (instrumentation) for JMX gatherer implementation
- replace target system metrics definition with YAML equivalents
- drop support for user-provided groovy scripts, they would have to be replaced by YAML equivalent
- for now keep the metrics definitions identical to their current state to preserve compatibility
[...]

PR #1445, opened 2024-09-05
[WIP] JMX scraper

SylvainJuge:
This PR introduces a mew `jmx-scrapper` module that aims to provide replacement of `jmx-metrics` with the following characteristics:
- JMX MBeans and attributes mapping to metrics is defined in YAML
- YAML parsing logic is common with instrumentation and provided by the [`instrumenation/jmx-metrics`](https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation/jmx-metrics/library) module library.
- Groovy configuration is not supported
- Metrics are directly to an OTLP collector endpoint, there is no prometheus HTTP endpoint.

Introducing a new module allows to make future transition from `jmx-metrics` easier once all the currently supported target systems are included, the `jmx-metrics` is currently used by the opentelemetry collector through the `jmxreceiver` feature.

This PR is a shared contribution of @robsunday and @SylvainJuge, opened as a draft to:
- provide progress status
- allow early feedback on the implementation

### Scope

- support one target system: `tomcat` with mapping equivalent to the current `tomcat.groovy`.
- support for other systems will be done as separate PRs
- alignment with definitions in integration is out of scope (see https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/12158 for details).

### Implementation checklist
- [x] bootstrap a new module and pick a name: `jmx-scrapper`
- [x] add ability to connect to a remote JMX tree (partial copy if current impl. + unit tests), no advanced SSL support


SylvainJuge:
It seems that manually squashing update commits made github actions assign lots of people, sorry for the noise here. @jack-berg can you reset this PR assignees ?

breedx-splk:
I know this is a work in progress, but if you can find any way to try and do a series of partial PRs to keep the footprint smaller for each review that would be helpful @SylvainJuge. 😁

SylvainJuge:
@breedx-splk I have re-framed this PR to only focus on "bootstrapping" this new module, the other implementation steps have been moved to https://github.com/open-telemetry/opentelemetry-java-contrib/issues/1362 description.

This PR is already getting quite large, so it's probably better to review and merge it before going further, I've thus marked it as ready for review.

Also, further progress is blocked unless:
- instrumentatation 2.9.0 is released https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/12178 has been merged but missed the 2.8.0 release by a couple of commits, so we will likely have to wait for ~1 month unless a patch release is needed.
- we can use and resolve a snapshot version in the dependencies

For now, no artifact is published, so maybe having a snapshot dependency could be a temporary compromise that we could merge in `main`. This would help us making progress by being able to merge small chunks into `main` rather than an "integration branch" from a fork as we did for this PR.

jmx-scraper/build.gradle.kts
```diff
@@ -0,0 +1,91 @@
+plugins {
+  application
+  id("com.github.johnrengelman.shadow")
+
+  id("otel.java-conventions")
+
+  // publishing disabled until component is ready to be used
+  // id("otel.publish-conventions")
```
trask, line 8: add TODO to make sure we don't forget to remove SNAPSHOT dependency

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/JmxScraper.java
```diff
@@ -0,0 +1,138 @@
...
+   */
+  @SuppressWarnings({"SystemOut", "SystemExitOutsideMain"})
+  public static void main(String[] args) {
+    JmxScraperConfig config;
+    JmxScraper jmxScraper = null;
+    try {
+      JmxScraperConfigFactory factory = new JmxScraperConfigFactory();
+      config = JmxScraper.createConfigFromArgs(Arrays.asList(args), factory);
+      jmxScraper = new JmxScraper(config);
+
```
robsunday, line 46: Don't we need a shutdown hook to make sure all the metrics gathered are sent out? It was there but now it is removed.

jmx-metrics/build.gradle.kts
```diff
@@ -13,6 +13,7 @@ application.mainClass.set("io.opentelemetry.contrib.jmxmetrics.JmxMetrics")
 
 repositories {
   mavenCentral()
+  mavenLocal()
```
robsunday, line 16: Is it intentional change? Do we have to modify jmx-metrics?

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/JmxScraper.java
```diff
@@ -0,0 +1,138 @@
...
+          "Usage: java -jar <path_to_jmxscraper.jar> "
+              + "-config <path_to_config.properties or - for stdin>");
+      System.exit(1);
+    } catch (ConfigurationException e) {
+      System.err.println(e.getMessage());
+      System.exit(1);
+    }
+
+    try {
+      Objects.requireNonNull(jmxScraper).start();
```
robsunday, line 58: requireNonNull call is redundant and can be removed.

jmx-metrics/build.gradle.kts
breedx-splk: We talked about this in the SIG -- it's was a stop-gap measure in order to consume fresh builds of the agent artifacts...but do we still need it here? I thought we agreed that snapshots would work.

jmx-scraper/README.md
```diff
@@ -0,0 +1,8 @@
+# JMX Metric Scraper
+
+This utility provides a way to query JMX metrics and export them to an OTLP endpoint.
+The JMX MBeans and their metrics mapping is defined in YAML and is reusing implementation from
```
breedx-splk, line 4:
```suggestion
The JMX MBeans and their metric mappings are defined in YAML and reuse implementation from
```

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/JmxScraper.java
```diff
@@ -0,0 +1,138 @@
...
+   * Main method to create and run a {@link JmxScraper} instance.
+   *
+   * @param args - must be of the form "-config {jmx_config_path,'-'}"
+   */
+  @SuppressWarnings({"SystemOut", "SystemExitOutsideMain"})
+  public static void main(String[] args) {
+    JmxScraperConfig config;
+    JmxScraper jmxScraper = null;
+    try {
+      JmxScraperConfigFactory factory = new JmxScraperConfigFactory();
```
breedx-splk, line 43: This factory is only used inside of the `createConfigFromArgs` method, so it doesn't need to be created up here and passed in -- it can just be created in that method.

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/JmxScraper.java
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  /**
+   * Create {@link JmxScraperConfig} object basing on command line options
+   *
+   * @param args application commandline arguments
+   */
+  static JmxScraperConfig createConfigFromArgs(List<String> args, JmxScraperConfigFactory factory)
+      throws ArgumentsParsingException, ConfigurationException {
+    if (!args.isEmpty() && (args.size() != 2 || !args.get(0).equalsIgnoreCase("-config"))) {
```
breedx-splk, line 72: I think it'd be nice to throw an exception early in the `args.isEmpty()` case. It makes this `if` simpler and prevents later trying to deal with config loading from an unspecified config. Being explicit is better here.

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/JmxScraper.java
```diff
@@ -0,0 +1,138 @@
...
+        loadPropertiesFromStdin(loadedProperties);
+      } else {
+        loadPropertiesFromPath(loadedProperties, path);
+      }
+    }
+
+    return factory.createConfig(loadedProperties);
+  }
+
+  private static void loadPropertiesFromStdin(Properties props) throws ConfigurationException {
```
breedx-splk, line 89: Passing a mutable argument just to have it mutated (populated) is an anti-pattern in Java. Better to remove the property and have this method return a `Properties` instance that the caller can use.

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/JmxScraper.java
```diff
@@ -0,0 +1,138 @@
...
+    try (InputStream is = new DataInputStream(System.in)) {
+      props.load(is);
+    } catch (IOException e) {
+      throw new ConfigurationException("Failed to read config properties from stdin", e);
+    }
+  }
+
+  private static void loadPropertiesFromPath(Properties props, String path)
```
breedx-splk, line 97: same comment about mutating input param.

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/JmxScraper.java
```diff
@@ -0,0 +1,138 @@
...
+    }
+  }
+
+  JmxScraper(JmxScraperConfig config) throws ConfigurationException {
+    String serviceUrl = config.getServiceUrl();
+    int interval = config.getIntervalMilliseconds();
+    if (interval < 0) {
+      throw new ConfigurationException("interval must be positive");
+    }
+    this.client = JmxRemoteClient.createNew(serviceUrl);
```
breedx-splk, line 112: nitpick: constructors shouldn't do work. This constructor is doing both validation and creation of true dependencies (violation of DI principles). This can be worked around with a static factory method.

It's also odd to me that the interval is verified here, but neither the config nor the interval are passed to the client. That feels wrong.

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/client/JmxRemoteClient.java
```diff
@@ -0,0 +1,149 @@
...
+import javax.management.remote.JMXConnectorFactory;
+import javax.management.remote.JMXServiceURL;
+import javax.security.auth.callback.Callback;
+import javax.security.auth.callback.CallbackHandler;
+import javax.security.auth.callback.NameCallback;
+import javax.security.auth.callback.PasswordCallback;
+import javax.security.auth.callback.UnsupportedCallbackException;
+import javax.security.sasl.RealmCallback;
+
+public class JmxRemoteClient {
```
breedx-splk, line 29: nit: Since this class serves as a builder-pattern style object whose purpose is to create a `JMXConnector` at some point, we could call it `JMXConnectorBuilder`. It's further complicated by the fact that the thing you get from calling `connect()` still has `connect()` on it. :) Also the existence of `JMXConnectorFactory` makes it worse. Heh, these 20+ year old names that we're stuck with....

Anyway, I think `JMXConnectorBuilder` is slightly more agreeable.

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/client/JmxRemoteClient.java
```diff
@@ -0,0 +1,149 @@
...
+
+  private static final Logger logger = Logger.getLogger(JmxRemoteClient.class.getName());
+
+  private final JMXServiceURL url;
+  @Nullable private String userName;
+  @Nullable private String password;
+  @Nullable private String profile;
+  @Nullable private String realm;
+  private boolean sslRegistry;
```
breedx-splk, line 38: Since it's not assigned in the constructor nor initialized, technically also nullable:

```suggestion
  @Nullable private boolean sslRegistry;
```

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/client/JmxRemoteClient.java
```diff
@@ -0,0 +1,149 @@
...
+
+  private JmxRemoteClient(JMXServiceURL url) {
+    this.url = url;
+  }
+
+  public static JmxRemoteClient createNew(String host, int port) {
+    return new JmxRemoteClient(buildUrl(host, port));
```
breedx-splk, line 45: VERY stupid nit, but prefer chaining rather than duplicating calls to the constructors:

```suggestion
    return new createNew(buildUrl(host, port));
```

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/client/JmxRemoteClient.java
```diff
@@ -0,0 +1,149 @@
...
+                  } else if (callback instanceof RealmCallback) {
+                    ((RealmCallback) callback).setText(realm);
+                  } else {
+                    throw new UnsupportedCallbackException(callback);
+                  }
+                }
+              });
+    } catch (ReflectiveOperationException e) {
+      logger.log(Level.WARNING, "SASL unsupported in current environment: " + e.getMessage());
+    }
```
breedx-splk, line 113: Factor out the top half of this method to a new method called `buildEnv()` to improve readability/length.

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/client/JmxRemoteClient.java
```diff
@@ -0,0 +1,149 @@
...
+
+    try {
+      if (sslRegistry) {
+        return doConnectSslRegistry(url, env);
+      } else {
```
breedx-splk, line 118: remove redundant `else`

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/config/JmxScraperConfig.java
```diff
@@ -0,0 +1,72 @@
...
+  int intervalMilliseconds;
+  String metricsExporterType = "";
+
+  String otlpExporterEndpoint = "";
+
+  String username = "";
+  String password = "";
+  String realm = "";
+  String remoteProfile = "";
+  boolean registrySsl;
```
breedx-splk, line 25: This feels like a symptom of groovy porting or something. I don't _really_ have an issue with it, it's just not idiomatic and will give java devs pause.

What do we think about making these private and adding getters, or maybe introducing `@AutoValue`?

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/JmxScraper.java
SylvainJuge: I've removed the `requireNonNull`.

For the shutdown hook, I don't think it's needed anymore as the polling and sending is now managed by the SDK (but I might be wrong here, and we will eventually discover this with end-to-end tests with short-lived JVMs).

From what I understand, the task scheduling in `io.opentelemetry.instrumentation.jmx.engine.BeanFinder#discoverBeans` handles the discovery and registration of JMX beans metrics, and the `io.opentelemetry.sdk.metrics.export.PeriodicMetricReader` in the SDK deals with exporing the metrics at a regular interval so this should be already covered thanks to SDK auto-configuration in `io.opentelemetry.sdk.autoconfigure.AutoConfiguredOpenTelemetrySdkBuilder` that registers `io.opentelemetry.sdk.OpenTelemetrySdk#close` to be called as shutdown hook.

SylvainJuge: In fact the `requirenonNull` is needed because nullaway check fails build otherwise.

jmx-metrics/build.gradle.kts
SylvainJuge: Good point, it's a leftover from using a local snapshot.

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/client/JmxRemoteClient.java
SylvainJuge: I agree, I've renamed it to `JmxConnectorBuilder` and renamed `connect()` to `build()` even if it actually establishes the connection.

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/client/JmxRemoteClient.java
SylvainJuge: It is a `boolean` with small `b` thus it defaults `false` and can't be `null`.

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/config/JmxScraperConfig.java
SylvainJuge: I have refactored and it's not using a static factory method instead of a separate class. I am not familiar with `@AutoValue` usage for such configuration classes so that could be another interesting option as well.

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/client/JmxRemoteClient.java
breedx-splk: Derp. Oh Java. :)

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/config/JmxScraperConfig.java
```diff
@@ -0,0 +1,244 @@
...
+
+  static final String OTLP_ENDPOINT = "otel.exporter.otlp.endpoint";
+
+  static final String JMX_USERNAME = "otel.jmx.username";
+  static final String JMX_PASSWORD = "otel.jmx.password";
+  static final String JMX_REMOTE_PROFILE = "otel.jmx.remote.profile";
+  static final String JMX_REALM = "otel.jmx.realm";
+
+  static final List<String> AVAILABLE_TARGET_SYSTEMS =
+      Arrays.asList(
```
breedx-splk, line 38: Should wrap this with `Collections.unmodifiableList()` so that the list cannot accidentally be modified.

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/config/JmxScraperConfig.java
```diff
@@ -0,0 +1,244 @@
...
+      throw new ConfigurationException(
+          String.format(
+              "%s must specify targets from %s", config.targetSystems, AVAILABLE_TARGET_SYSTEMS));
+    }
+
+    if (isBlank(config.otlpExporterEndpoint)
+        && (!isBlank(config.metricsExporterType)
+            && config.metricsExporterType.equalsIgnoreCase("otlp"))) {
+      throw new ConfigurationException(OTLP_ENDPOINT + " must be specified for otlp format.");
+    }
```
breedx-splk, line 238: Can we default the endpoint to the same default that the agent uses? OTLP http+protobuf to localhost:4318? Seems like a reasonable/sane default to me.

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/util/StringUtils.java
```diff
@@ -0,0 +1,22 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.contrib.jmxscraper.util;
+
+import javax.annotation.Nullable;
+
+public final class StringUtils {
```
breedx-splk, line 10: Seems silly, but elsewhere we take extra care to ensure that utilities like this are in an "internal" package and have a javadoc that says that they are internal and may change or be deleted at any time....so that users are discouraged from using them directly. We could do that here too.

jmx-scraper/src/test/java/io/opentelemetry/contrib/jmxscraper/config/JmxScraperConfigTest.java
```diff
@@ -0,0 +1,407 @@
...
+    validProperties.setProperty(JmxScraperConfig.CUSTOM_JMX_SCRAPING_CONFIG, "");
+    validProperties.setProperty(JmxScraperConfig.TARGET_SYSTEM, "tomcat, activemq");
+    validProperties.setProperty(JmxScraperConfig.METRICS_EXPORTER_TYPE, "otel");
+    validProperties.setProperty(JmxScraperConfig.INTERVAL_MILLISECONDS, "1410");
+    validProperties.setProperty(JmxScraperConfig.REGISTRY_SSL, "true");
+    validProperties.setProperty(JmxScraperConfig.OTLP_ENDPOINT, "http://localhost:4317");
+    validProperties.setProperty(JmxScraperConfig.JMX_USERNAME, "some-user");
+    validProperties.setProperty(JmxScraperConfig.JMX_PASSWORD, "some-password");
+    validProperties.setProperty(JmxScraperConfig.JMX_REMOTE_PROFILE, "some-profile");
+    validProperties.setProperty(JmxScraperConfig.JMX_REALM, "some-realm");
```
breedx-splk, line 35: I like static imports for these to improve readability.

jmx-scraper/src/test/java/io/opentelemetry/contrib/jmxscraper/config/JmxScraperConfigTest.java
```diff
@@ -0,0 +1,407 @@
...
+    Properties properties = (Properties) validProperties.clone();
+    properties.setProperty(JmxScraperConfig.INTERVAL_MILLISECONDS, "abc");
+
+    // When and Then
+    assertThatThrownBy(() -> JmxScraperConfig.fromProperties(properties, new Properties()))
+        .isInstanceOf(ConfigurationException.class)
+        .hasMessage("Failed to parse otel.jmx.interval.milliseconds");
+  }
+
+  //  @ClearSystemProperty(key = "otel.metric.export.interval")
```
breedx-splk, line 252: Either remove the commented test code or add a TODO that explains why it's here and the path forward. Thanks.

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/TargetSystemIntegrationTest.java
```diff
@@ -0,0 +1,163 @@
...
+    scraper =
+        new JmxScraperContainer(otlpEndpoint)
+            .withNetwork(network)
+            .withService(TARGET_SYSTEM_NETWORK_ALIAS, JMX_PORT);
+
+    scraper = customizeScraperContainer(scraper);
+    scraper.start();
+
+    // TODO: replace with real assertions
+    assertThat(otlpServer.getMetrics()).isEmpty();
```
breedx-splk, line 124: yeah so this is kinda weird. I appreciate the TODO, so we can circle back on it...but it's odd that we're doing all this setup to verify that nothing happened. :)

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/JmxConnectorBuilder.java
```diff
@@ -0,0 +1,161 @@
...
+                  }
+                }
+              });
+    } catch (ReflectiveOperationException e) {
+      logger.log(Level.WARNING, "SASL unsupported in current environment: " + e.getMessage());
+    }
+    return env;
+  }
+
+  @SuppressWarnings("BanJNDI")
```
breedx-splk, line 138: 🙈

breedx-splk (approved):
I think it's completely great to see this coming along....nice work! Happy to get this core in so additional work can proceed. 🤘🏻

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/TargetSystemIntegrationTest.java
robsunday: Refactored a bit. Added abstract method to validate target system metrics. Method will be implemented in dedicated integration test.

robsunday (approved):

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/config/JmxScraperConfig.java
SylvainJuge: Should now be fixed by https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1445/commits/2c95782292d06eb29c41f266466e3648ca025fc8

PR #1446, opened 2024-09-05
add stability status to libraries

jackshirazi:
**Description:**

Library stability was discussed and we noted that one library is stable. I think we should show that status on the README. I don't know if this is the best way to do it. I've set everything else to `beta` which again maybe isn't the best way

**Existing Issue(s):**

n/a

**Testing:**

viewed the README as a file and checked the links

**Documentation:**

This is the doc

**Outstanding items:**

none

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,31 +9,37 @@ feature or via instrumentation, this project is hopefully for you.
 
 ## Provided Libraries
 
-| Status | Library                                                           |
-| ------ | ----------------------------------------------------------------- |
-| beta   | [AWS Resources](./aws-resources/README.md)                        |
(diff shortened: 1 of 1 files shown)


breedx-splk:
In the SIG we talked about using `alpha` to match the published artifacts, and adding a small link to the otep (0232) with a clarifying statement about production workloads.

jackshirazi:
Changed statuses and added discussion points per Sep 12 2024 SIG meeting and also the one from Aug 29 2024

trask (approved):

breedx-splk (approved):

PR #1455, opened 2024-09-13
[disk-buffering] Add debug mode for verbose logging.

breedx-splk:
Previously, the disk buffering code would ALWAYS log when it attempted to read from disk and when it was exporting data from disk. This gets a bit chatty and may not be what users want to see, so we need a way to silence it.

This PR introduces "debug mode" which is effectively verbose logging for troubleshooting purposes. It adds additional logging statements and logs the storage configuration in `StorageBuilder.build()`.

### DIFF
--- a/disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/serialization/deserializers/MetricDataDeserializer.java
+++ b/disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/serialization/deserializers/MetricDataDeserializer.java
@@ -6,6 +6,7 @@
 package io.opentelemetry.contrib.disk.buffering.internal.serialization.deserializers;
 
 import io.opentelemetry.contrib.disk.buffering.internal.serialization.mapping.metrics.ProtoMetricsDataMapper;
+import io.opentelemetry.contrib.disk.buffering.internal.utils.SignalTypes;
 import io.opentelemetry.proto.metrics.v1.MetricsData;
 import io.opentelemetry.sdk.metrics.data.MetricData;
 import java.io.IOException;
@@ -31,6 +32,6 @@ public final class MetricDataDeserializer implements SignalDeserializer<MetricDa
 
   @Override
   public String signalType() {
-    return "metrics";
+    return SignalTypes.metrics.name();
   }
 }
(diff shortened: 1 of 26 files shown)


disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/serialization/deserializers/MetricDataDeserializer.java
```diff
@@ -28,4 +28,9 @@ public List<MetricData> deserialize(byte[] source) {
       throw new IllegalArgumentException(e);
     }
   }
+
+  @Override
+  public String signalType() {
+    return "metrics";
```
LikeTheSalad, line 34: Do you think we could reuse the same names already provided when building the exporters, like in the case of metrics [this one](https://github.com/open-telemetry/opentelemetry-java-contrib/blob/15e430779defecfedebfaa20f101f537d6a5849c/disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/MetricFromDiskExporter.java#L24), for example. Or if it's not feasible, then I think it'd be nice to at least have them as constants so that we don't have to repeat them.

LikeTheSalad (commented):
Thank you. Just a few comments, though I like the idea!

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/Storage.java
```diff
@@ -95,8 +112,19 @@ private ReadableResult readAndProcess(Function<byte[], Boolean> processing, int
     }
   }
 
+  private void log(String msg) {
+    log(msg, Level.INFO);
+  }
+
+  private void log(String msg, Level level) {
```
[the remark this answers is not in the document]
breedx-splk: Yeah cool I think that worked out. Thanks for the idea.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/serialization/deserializers/MetricDataDeserializer.java
breedx-splk: Sure, I went ahead and pulled them into an enum.

LikeTheSalad (approved):
Thanks!

PR #1480, opened 2024-09-30
Integration tests for JMX Scraper

robsunday:


breedx-splk (commented):
I'm a bit confused by this. Unless I'm missing something, you shouldn't have to find an ephemeral/open/unused port on the box running the tests -- just use a fixed port for JMX, use that fixed port in the jvm args for both rmi and jmx, and then docker itself should assign an ephemeral port on the outside surface of the container.

That's what the `getMappedPort()` then does for you -- it lets you know what external port docker assigned to the fixed internal port. Make sense?

With that, I think you can just delete the `PortSelector` class.

SylvainJuge:
Here we have multiple challenges, and using a fixed port mapping is a way to work-around the JMX/RMI implementation details, this is clearly not ideal and I don't know if we have a better option to test JMX with RMI. I haven't investigated deeper but it seems that using `jmxmp` instead of `rmi` protocol instead would be a better option for most integration tests.

In short:
- setting `com.sun.management.jmxremote.port=9999` will make the target JVM listen on port `9999`
- testcontainers map the `9999` JMX port to a random port, let's say `12345`.
- we build a JMX URI with `localhost:12345` which allows to connect to the mapped port on the host
- when connecting to `localhost:12345` the server-side JMX implementation replies that a different `host:port` which in my case in Linux host is something like `172.20.0.xx:9999`.
- This `host:port` combination can be controlled with `com.sun.management.jmxremote.rmi.port` and `java.rmi.server.hostname` properties on the target JVM
- this behavior is related to the RMI protocol itself and is problematic with NAT as it forces to use only a single host:port to be used at once.

Maybe here another option could also be to make sure that we have a network that is always accessible from the host like it does on Linux could be a simpler option.

trask:
just in case it's helpful/relevant here: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/rmi/javaagent/src/test/java/io/opentelemetry/javaagent/instrumentation/rmi/RmiTest.java

SylvainJuge:
After failing to make things working with `jmxmp`, I think using a fixed port is an acceptable work-around for now:
- it's only needed to test the `JmxConnectorBuilder` class with host-to-container communication, packaging the test and running it in a container would be way overkill and harder to debug in the future.
- there is no impact on container-to-container communication for other integration tests.
- it's simpler to have the same implementation for both mac and linux, the fact it works on Linux and in CI by default is almost "by accident"

I've opened a PR on this PR to simplify even further and reuse an available port implementation already in classpath: https://github.com/robsunday/opentelemetry-java-contrib/pull/1

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/TargetSystemIntegrationTest.java
```diff
@@ -90,32 +85,20 @@ void afterEach() {
...
-        "Target system started, JMX port: {} mapped to {}:{}", JMX_PORT, targetHost, targetPort);
-
-    // TODO : wait for metrics to be sent and add assertions on what is being captured
-    // for now we just test that we can connect to remote JMX using our client.
-    try (JMXConnector connector = JmxConnectorBuilder.createNew(targetHost, targetPort).build()) {
-      assertThat(connector.getMBeanServerConnection()).isNotNull();
-    } catch (IOException e) {
-      throw new RuntimeException(e);
-    }
-
```
SylvainJuge, line 114: [for reviewer] this part had to go anyway as it's now replaced by actual integration tests in #1485 , here it only checked the connection from the host as a proxy of "target JVM is working with JMX settings".

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/TargetSystemIntegrationTest.java
```diff
@@ -90,32 +85,20 @@ void afterEach() {
 
   @Test
   void endToEndTest() {
+    int jmxPort = PortSelector.getAvailableRandomPort();
 
     target =
-        createTargetContainer(JMX_PORT)
-            .withLogConsumer(new Slf4jLogConsumer(logger))
+        createTargetContainer(jmxPort)
+            .withLogConsumer(new Slf4jLogConsumer(targetSystemLogger))
```
SylvainJuge, line 92: @robsunday here we don't need to find an available port on host as this will be used between containers anyway (and we don't connect anymore from docker host), see 
https://github.com/robsunday/opentelemetry-java-contrib/pull/2 to simplify this.

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/JmxConnectorBuilder.java
```diff
@@ -138,6 +138,7 @@ private Map<String, Object> buildEnv() {
   @SuppressWarnings("BanJNDI")
   private static JMXConnector doConnect(JMXServiceURL url, Map<String, Object> env)
       throws IOException {
+    logger.info("Connecting to " + url);
```
trask, line 141: maybe better to be debug, i'm not sure about jmx-scraper but we are generally pretty light on info logs (outside of tests)

```suggestion
    logger.debug("Connecting to {}", url);
```

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/TestAppContainer.java
```diff
@@ -58,9 +62,33 @@ public TestAppContainer withUserAuth(String login, String pwd) {
...
+    addFixedExposedPort(port, port);
+    return this;
+  }
+
   @Override
   public void start() {
-
+    //    properties.put("com.sun.management.jmxremote.local.only", "false");
+    //    properties.put("java.rmi.server.logCalls", "true");
+    //
```
trask, line 91: I'd suggest either remove or include a code comment explaining why you may want to uncomment

```suggestion
```

trask (approved):

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/JmxConnectorBuilder.java
robsunday: So this happens only once on application startup and the url is not printed anywhere else. I think it may be useful  for connection issue investigations because default Java error message for JMX is quite cryptic (like "Unable to connect Unable to connect to :0" for invalid host). Would you agree to leave it as is?

trask: > Would you agree to leave it as is?

👍

PR #1485, opened 2024-10-02
Reuse JMX Insights in Scraper + Tomcat support

SylvainJuge:
- **add snapshot dependency**
- **first working poc with tomcat**
- **add comments for future config enhancement**
- port tomcat YAML definiton from groovy
- port jvm YAML definiton from groovy

**Description:**

Part of #1362

**Testing:**

Integration tests from JMX Gatherer almost copied as-is for `jvm` and `tomcat` systems.

Manual testing on a local tomcat server started with following JVM options in `bin/setenv.sh`

```
#!/bin/bash
opts=''
opts="${opts} -Dcom.sun.management.jmxremote.port=12345"
opts="${opts} -Dcom.sun.management.jmxremote.local.only=false"
opts="${opts} -Dcom.sun.management.jmxremote.authenticate=false"
opts="${opts} -Dcom.sun.management.jmxremote.ssl=false"
export CATALINA_OPTS="${opts:-}"
```

Scraper started with command line (in the `./jmx-scraper` subfolder as working directory).
```
java \
-Dotel.metrics.exporter=otlp,logging \
-Dotel.exporter.otlp.endpoint=${OTEL_EXPORTER_OTLP_ENDPOINT} \
-Dotel.jmx.service.url=service:jmx:rmi:///jndi/rmi://localhost:12345/jmxrmi \
-Dotel.jmx.target.system=tomcat \
-jar ./build/libs/opentelemetry-jmx-scraper-1.40.0-alpha-SNAPSHOT.jar

```

The explicit `otel.exporter.otlp.endpoint` and `otel.metrics.exporter` Java system properties are needed due to a check in the configuration parsing (and they aren't yet read from the env variables), but this will later be refactored to use auto-configuration in a later PR to solve this.


jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/JmxScraper.java
```diff
@@ -109,29 +128,64 @@ private static Properties loadPropertiesFromPath(String path) throws Configurati
...
-    // MetricConfiguration metricConfig = new MetricConfiguration();
-    // TODO create JMX insight config from scraper config
-    // service.startRemote(metricConfig, () -> Collections.singletonList(connection));
+  private static MetricConfiguration getMetricConfig(JmxScraperConfig scraperConfig) {
+    MetricConfiguration config = new MetricConfiguration();
+    for (String system : scraperConfig.getTargetSystems()) {
+      try {
+        addRulesForSystem(system, config);
+      } catch (RuntimeException e) {
+        logger.warning("unable to load rules for system " + system + ": " + e.getMessage());
```
robsunday, line 169: [minor] - It would be nice to start all the messages (logged and in exception) from upper case letter

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/JmxScraper.java
```diff
@@ -109,29 +128,64 @@ private static Properties loadPropertiesFromPath(String path) throws Configurati
...
-      throw new IllegalStateException(e);
+  private static void addRulesForSystem(String system, MetricConfiguration conf) {
+    String yamlResource = system + ".yaml";
+    try (InputStream inputStream =
+        JmxScraper.class.getClassLoader().getResourceAsStream(yamlResource)) {
+      if (inputStream != null) {
+        RuleParser parserInstance = RuleParser.get();
+        parserInstance.addMetricDefsTo(conf, inputStream, system);
+      } else {
+        throw new IllegalStateException("no support for " + system);
```
robsunday, line 185: I think IllegalStateException was designed for something else. I think ConfigurationException would be better (or at least IllegalArgumentException).

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/JmxScraper.java
```diff
@@ -109,29 +128,64 @@ private static Properties loadPropertiesFromPath(String path) throws Configurati
...
+      }
+    } catch (Exception e) {
+      throw new IllegalStateException("error while loading rules for system " + system, e);
```
robsunday, line 188: Other exception would be better here as well

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/config/JmxScraperConfig.java
```diff
@@ -55,9 +55,9 @@ public class JmxScraperConfig {
   private String serviceUrl = "";
   private String customJmxScrapingConfigPath = "";
   private Set<String> targetSystems = Collections.emptySet();
-  private int intervalMilliseconds;
-  private String metricsExporterType = "";
-  private String otlpExporterEndpoint = "";
+  private int intervalMilliseconds; // TODO only used to set 'otel.metric.export.interval' from SDK
```
robsunday, line 58: What this TODO means here? Is there anything aditional to be implemented?

robsunday (approved):

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/JmxScraper.java
SylvainJuge: I'll leave this one as-is for now, as we don't have dedicated runtime exceptions, but this could be an improvement later.

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/config/JmxScraperConfig.java
SylvainJuge: Yes, basically this configuration option is only used to configure the metrics polling interval and there is already a dedicated configuration option in the SDK for this, so if someone is used to configure the SDK using this one we should reuse the existing config option (we'll deal with that when adding auto-configuration here).

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/JmxScraperContainer.java
```diff
@@ -76,6 +76,7 @@ public void start() {
     // for now only configure through JVM args
     List<String> arguments = new ArrayList<>();
     arguments.add("java");
+    arguments.add("-Dotel.metrics.exporter=otlp");
```
SylvainJuge, line 79: missing this one means the SDK does not export, we should probably set it by default when not set explicitly as it's the only option for exporting metrics.

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/TargetSystemIntegrationTest.java
```diff
@@ -104,26 +109,54 @@ void endToEndTest() {
...
+              List<ExportMetricsServiceRequest> receivedMetrics = otlpServer.getMetrics();
+              assertThat(receivedMetrics).describedAs("no metric received").isNotEmpty();
+
+              List<Metric> metrics =
+                  receivedMetrics.stream()
+                      .map(ExportMetricsServiceRequest::getResourceMetricsList)
+                      .flatMap(rm -> rm.stream().map(ResourceMetrics::getScopeMetricsList))
+                      .flatMap(Collection::stream)
+                      .filter(
+                          // TODO: disabling batch span exporter might help remove unwanted metrics
```
SylvainJuge, line 138: this filtering is needed because the logs and trace exporters do report some telemetry metrics.

SylvainJuge:
@robsunday I have just added support for both the JVM metrics and the Tomcat metrics with tests from Gatherer, so we have end-to-end testing now :tada:.

If you have some time for another review that would be great !

SylvainJuge:
In the context of integration tests, we always want to send data to otlp as otherwise we don't have a way to validate what it's captured, this is not meant to be a reusable test container outside of those integration tests.

For the filtering on the metrics, there was a filter on the metrics that are sent with a specific sdk version, now that we reuse code from instrumentation we do not get the project version as we used to have, but the version of the instrumentation project which is used as dependency.

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/JmxScraperContainer.java
SylvainJuge: No, this does not change the default, it's only that whenever we run JmxScraper in this docker setup we need to export metrics through OTLP, so it's not worth making it configurable for now unless there are other use-cases.

trask:
sorry @robsunday, I hit the wrong button and re-requested a review from you, which also cleared your approval, can you re-approve? and let me know if there are any outstanding discussions, and if not I'll merge, thanks!

trask:
@SylvainJuge sorry, can you fix up the merge conflicts from #1480? thanks!

robsunday (approved):

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/JmxScraper.java
```diff
@@ -35,15 +44,25 @@ public class JmxScraper {
    */
   @SuppressWarnings({"SystemOut", "SystemExitOutsideMain"})
   public static void main(String[] args) {
+
+    // enable SDK auto-configure if not explicitly set by user
+    // TODO: refactor this to use AutoConfiguredOpenTelemetrySdk
```
trask, line 49: 👍

trask (approved):

PR #1486, opened 2024-10-04
Rename param and description to proper value

steverao:
Rename param and description to proper value


breedx-splk (approved):
Copypasta strikes again!

PR #1497, opened 2024-10-15
Integration test activemq

robsunday:
Primary goal of this PR is to use JMX Metric Insight to gather ActiveMQ JMX properties while preserving compatibility with JMX Gatherer.

At the moment there are two lines of ActiveMQ product:
- Classic
- Artemis

This test covers Classic line. Artemis line have different set of JMX attributes, so it should have separate yaml and test (nice summary of differences is here: https://www.datadoghq.com/blog/activemq-architecture-and-metrics/). Artemis support may be included in the future.

Assertions in the ActiveMQ integration test are copied from opentelemetry-java/contrib/jmx-metrics while activemq.yaml is taken from opentelemetry-java-instrumentation/jmx-metrics (JMX Metric Insight)
Minor changes were appiled to both of them. These changes are described as comments in the files.


jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/ActiveMqIntegrationTest.java
```diff
@@ -0,0 +1,156 @@
...
+  @Override
+  protected void verifyMetrics() {
+    waitAndAssertMetrics(
+        metric ->
+            assertSumWithAttributes(
+                metric,
+                "activemq.consumer.count",
+                "The number of consumers currently reading from the broker.",
+                "consumers",
+                /* isMonotonic= */ false,
```
robsunday, line 56: [for reviewer] isMonotonic=false was added due to updatedDefinition of asserSumWithAttributes.

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/ActiveMqIntegrationTest.java
```diff
@@ -0,0 +1,156 @@
...
+                    attrs.containsOnly(
+                        entry("destination", "ActiveMQ.Advisory.MasterBroker"),
+                        entry("broker", "localhost"))),
+        metric ->
+            assertSumWithAttributes(
+                metric,
+                "activemq.producer.count",
+                "The number of producers currently attached to the broker.",
+                "producers",
+                /* isMonotonic= */ false,
```
robsunday, line 67: [for reviewer] isMonotonic=false was added due to updatedDefinition of asserSumWithAttributes.

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/ActiveMqIntegrationTest.java
```diff
@@ -0,0 +1,156 @@
...
+                    attrs.containsOnly(
+                        entry("destination", "ActiveMQ.Advisory.MasterBroker"),
+                        entry("broker", "localhost"))),
+        metric ->
+            assertSumWithAttributes(
+                metric,
+                "activemq.connection.count",
+                "The total number of current connections.",
+                "connections",
+                /* isMonotonic= */ false,
```
robsunday, line 78: [for reviewer] isMonotonic=false was added due to updatedDefinition of asserSumWithAttributes.

jmx-scraper/src/main/resources/activemq.yaml
robsunday, line 1: Metric names has been changed to match that used in jmx-metrics

jmx-scraper/src/main/resources/activemq.yaml
```diff
@@ -0,0 +1,69 @@
...
+        desc: The total number of messages delivered to consumers.
+      AverageEnqueueTime:
+        metric: message.wait_time.avg
+        unit: ms
+        type: gauge
+        desc: The average time a message was held on a destination.
+
+  - bean: org.apache.activemq:type=Broker,brokerName=*
+    metricAttribute:
+      broker: param(brokerName)
```
robsunday, line 54: This attribute was present in original activemq.yaml. Because of this some minor changes has to be made in assertions to check this metric attribute

jmx-scraper/src/main/resources/activemq/env
robsunday, line 1: Once we merge it will be good to update resources folder structure. I did not want to do too much in this PR.

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/ActiveMqIntegrationTest.java
```diff
@@ -0,0 +1,156 @@
...
+import org.testcontainers.utility.MountableFile;
+
+public class ActiveMqIntegrationTest extends TargetSystemIntegrationTest {
+
+  @Override
+  protected GenericContainer<?> createTargetContainer(int jmxPort) {
+    return new GenericContainer<>(
+            new ImageFromDockerfile()
+                .withDockerfileFromBuilder(
+                    builder -> builder.from("apache/activemq-classic:5.18.6").build()))
```
robsunday, line 26: Updated activemq docker image to the most recent ActiveMQ Classic version (from rmohr/activemq:5.15.9 used in JMX Metric Gatherer)

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/MetricAssertions.java
```diff
@@ -0,0 +1,139 @@
...
+      String name,
+      String description,
+      String unit,
+      Consumer<MapAssert<String, String>>... attributeGroupAssertions) {
+    assertSumWithAttributes(
+        metric, name, description, unit, /* isMonotonic= */ true, attributeGroupAssertions);
+  }
+
+  @SafeVarargs
+  static void assertSumWithAttributes(
```
robsunday, line 80: [for reviewer] This override method is added to support 'monotonic' flag and metric attributes at he same time.

jmx-scraper/src/main/resources/activemq.yaml
SylvainJuge: Maybe add an explicit comment like in the `tomcat.yaml` to explain the difference, adding one extra metric attribute should not have much impact and thus it should remain compatible.

```suggestion
      # minor divergence from activemq.groovy to capture broker name, making it closer to
      # the definition in JMX Insights
      broker: param(brokerName)
```

jmx-scraper/src/main/resources/activemq.yaml
SylvainJuge: You should probably update the PR description to reflect that the current goal is to preserve compatibility with `jmx-metrics` (JMX Gatherer).

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/ActiveMqIntegrationTest.java
SylvainJuge: For all of the 3 metrics that are non-monotonic here I think it was probably a bug in the original test of the groovy version. The fact that they all have `current` in their description means that they should represent a value that can go up or down, and thus can't be monotonic.

jmx-scraper/src/main/resources/activemq.yaml
SylvainJuge: And also, only the classic activemq is supported, not the Artemis one.

SylvainJuge (approved):

trask:
thanks, @robsunday mentioned on slack that he may push one more update, will wait for his confirmation

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/ActiveMqIntegrationTest.java
```diff
@@ -0,0 +1,156 @@
...
+        .withEnv("LOCAL_JMX", "no")
+        .withCopyFileToContainer(
+            MountableFile.forClasspathResource("activemq/env"), "/opt/apache-activemq/bin/env")
```
trask, line 29: can you add a comment why we need to customize activemq/env? thanks

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/ActiveMqIntegrationTest.java
```diff
@@ -26,6 +26,8 @@ protected GenericContainer<?> createTargetContainer(int jmxPort) {
                     builder -> builder.from("apache/activemq-classic:5.18.6").build()))
         .withEnv("LOCAL_JMX", "no")
         .withCopyFileToContainer(
+            // Overwrite default ActiveMQ configuration in order to let ActiveMQ use JMX options
+            // stored in $ACTIVEMQ_JMX_OPTS env variable defined below
```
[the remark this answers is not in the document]
robsunday: This is actually a very good point. Replacing file with JAVA_TOOL_OPTIONS was trivial and worked. '''env''' file was used in jmx-metrics and ported here, but really was not needed. 
Great suggestion!

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/TomcatIntegrationTest.java
```diff
@@ -30,7 +30,6 @@ protected GenericContainer<?> createTargetContainer(int jmxPort) {
                                 "https://tomcat.apache.org/tomcat-9.0-doc/appdev/sample/sample.war",
                                 "/usr/local/tomcat/webapps/ROOT.war")
                             .build()))
-        .withEnv("LOCAL_JMX", "no")
```
robsunday, line 33: This is not needed for Tomcat as well.

jmx-scraper/src/main/resources/activemq.yaml
```diff
@@ -10,11 +10,15 @@ rules:
     mapping:
       ProducerCount:
         metric: producer.count
+        # Unit name inherited from activemq.groovy file.
```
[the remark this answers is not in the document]
SylvainJuge: Yes, those are likely to remain "as is", and regarding their location having them in a dedicated folder might also make sense when we reorganize  but it's another topic.

I'm just trying to not to forget about this, but a TODO here is probably not the best choice, it's probably better to add this to the issue where we listed all the supported systems for feature parity: https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/12158 so I've updated it to ensure we don't forget about it.

SylvainJuge (approved):

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/ActiveMqIntegrationTest.java
robsunday: Finally I got rid of this file at all

trask (approved):

PR #1499, opened 2024-10-16
span stacktrace refactor + autoconfig

SylvainJuge:
The SDK version [1.42.0](https://github.com/open-telemetry/opentelemetry-java/releases/tag/v1.42.0) introduced with ([#6367](https://github.com/open-telemetry/opentelemetry-java/pull/6367)) a new `ExtendedSpanProcessor.onEnding(ReadWriteSpan span)` that provide a hook to modify the span on end, which the `SpanProcessor.onEnd(ReadableSpan readableSpan)` does not allow.

This allows to simplify the current implementation of span stacktrace capture by removing all internal classes. This also allows to implement autoconfiguration which is also part of this PR.

When using autoconfiguration, this PR introduces the `otel.java.experimental.span-stacktrace.filter` configuration option which allows to provide the class name of the filter.

This is a breaking change and users of this module will see a compile-time error due to the change in the `StackTraceSpanProcessor` constructor arguments.


span-stacktrace/src/main/java/io/opentelemetry/contrib/stacktrace/StackTraceAutoConfig.java
```diff
@@ -0,0 +1,109 @@
...
+
+  private static final String CONFIG_FILTER = "otel.java.experimental.span-stacktrace.filter";
+
+  @Override
+  public void customize(AutoConfigurationCustomizer config) {
+    config.addTracerProviderCustomizer(
+        (providerBuilder, properties) -> {
+          long minDuration = getMinDuration(properties);
+          Predicate<ReadableSpan> filter = getFilterPredicate(properties);
+          providerBuilder.addSpanProcessor(new StackTraceSpanProcessor(minDuration, filter));
```
JonasKunz, line 38: I don't think that allowing the construction of the `StackTraceSpanProcessor` with a negative duration makes sense, because it is always No-Op.

I would suggest to change `StackTraceSpanProcessor` to only allow min-durations greater or equal zero.
Then, here in the auto configuration only construct and add the processor if the duration is non-negative.

span-stacktrace/src/test/java/io/opentelemetry/contrib/stacktrace/StackTraceAutoConfigTest.java
```diff
@@ -0,0 +1,88 @@
...
+import io.opentelemetry.sdk.trace.ReadableSpan;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.Predicate;
+import org.junit.jupiter.api.Test;
+
+public class StackTraceAutoConfigTest {
+
+  @Test
```
JonasKunz, line 20: I think we could improve the coverage of this test if you instead test on the level of the `TracerProviderCustomizer` using : This way you could record calls to `addSpanProcessor` and introspect the added processor for the correct configuration (or ensuring that no processor is added at all).

span-stacktrace/src/main/java/io/opentelemetry/contrib/stacktrace/StackTraceAutoConfig.java
SylvainJuge: good point, I've fixed that in d1f984a

span-stacktrace/src/test/java/io/opentelemetry/contrib/stacktrace/StackTraceAutoConfigTest.java
SylvainJuge: I've done that in 3072467832520b0ed37831ea6758b6c999a27e1c but in the other test class as it's more an "end-to-end" setup with a complete SDK, this one remains more a unit-test.

span-stacktrace/README.md
```diff
@@ -6,51 +6,23 @@ This module provides a `SpanProcessor` that captures the [`code.stacktrace`](htt
 Capturing the stack trace is an expensive operation and does not provide any value on short-lived spans.
 As a consequence it should only be used when the span duration is known, thus on span end.
 
-However, the current SDK API does not allow to modify span attributes on span end, so we have to
-introduce other components to make it work as expected.
+## Usage and configuration
 
-## Usage
+This extension supports autoconfiguration.
```
trask, line 11: maybe worth explaining this a bit more, e.g. explaining that just by having including it as a dependency, the Java SDK autoconfiguration module will automatically find and enable it

SylvainJuge: Good point, I've enhanced the autoconfiguration description in 8745a7f

JonasKunz (approved):

trask (approved):

PR #1505, opened 2024-10-17
Update to Java Instrumentation 2.9.0

trask:


laurit (approved):

SylvainJuge (approved):
Thanks for taking care of removing the snapshot dependency !

robsunday (approved):

PR #1515, opened 2024-10-22
Cassandra support and integration test added.

robsunday:
This PR adds support for Cassandra NoSQL DB JMX metrics. The goal of this PR is to port jmx-metrics functionality to jmx-scraper.
JMX Insights has no yaml file for Cassandra and cassandra.yaml file in this PR is created based on cassandra.groovy file from jmx-metrics.
Assertions are copied from jmx-metrics.
Cassandra version is updated from 3.11 used in jmx-scraper to 5.0.2 (the most recent one), however I verified that metrics are correctly gathered on 3.* and 4.* versions. In the future we may consider using multiple versions of tested target system, so we can easily make sure that all of them are supported by single yaml config.
Also jmx-scraper base container image is updated to the most recent java8 version that also supports arm64 arch (for Macs)

### DIFF
--- a/jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/JmxScraperContainer.java
+++ b/jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/JmxScraperContainer.java
@@ -28,7 +28,7 @@ public class JmxScraperContainer extends GenericContainer<JmxScraperContainer> {
   private final Set<String> customYamlFiles;
 
   public JmxScraperContainer(String otlpEndpoint) {
-    super("openjdk:8u272-jre-slim");
+    super("openjdk:8u342-jre-slim");
 
     String scraperJarPath = System.getProperty("shadow.jar.path");
     assertThat(scraperJarPath).isNotNull();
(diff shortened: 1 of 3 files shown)


jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/JmxScraperContainer.java
```diff
@@ -28,7 +28,7 @@ public class JmxScraperContainer extends GenericContainer<JmxScraperContainer> {
   private final Set<String> customYamlFiles;
 
   public JmxScraperContainer(String otlpEndpoint) {
-    super("openjdk:8u272-jre-slim");
+    super("openjdk:8u342-jre-slim");
```
robsunday, line 31: [for reviewer] Updated to most recent version of Java8 and to support arm64 arch

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/CassandraIntegrationTest.java
```diff
@@ -0,0 +1,172 @@
...
+    return scraper.withTargetSystem("cassandra");
+  }
+
+  @Override
+  protected void verifyMetrics() {
+    waitAndAssertMetrics(
+        metric ->
+            assertGauge(
+                metric,
+                "cassandra.client.request.range_slice.latency.50p",
```
breedx-splk, line 54: Bummer that cassandra doesn't provide a histogram. :)

breedx-splk (approved):
LGTM!

PR #1517, opened 2024-10-24
Add jetty.yaml to JMX scraper

SylvainJuge:
Adds support for `jetty` system by providing an `jetty.yaml` and related tests.

Part of #1362

Tests had to be slightly adapted from their groovy counterparts to make it work as expected, but captured metrics are compatible (some of them only have a few extra attributes).

Adds a `genericJmxJvmArguments` method to generate common JMX JVM arguments to remove some documentation.

Bonus:
- making cassandra startup faster as it's a single isolated node.
- fixing some minor inconsistencies with plural units and align with stable conventions: https://opentelemetry.io/docs/specs/semconv/general/metrics/#instrument-units


jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/MetricAssertions.java
```diff
@@ -136,4 +153,19 @@ private static void assertAttributedPoints(
                             KeyValue::getKey, keyValue -> keyValue.getValue().getStringValue())))
         .satisfiesExactlyInAnyOrder(assertions);
   }
+
+  @SuppressWarnings("unchecked")
+  private static void assertAttributedMultiplePoints(
```
SylvainJuge, line 158: [for reviewer] the difference with `assertAttributedPoints` is that we check all the provided assertions to all the data points, whereas `assertAttributedPoints` expects to have exactly one assertion per point.

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/TargetSystemIntegrationTest.java
```diff
@@ -196,4 +198,27 @@ public void export(
       sb.http(0);
     }
   }
+
+  protected static String genericJmxJvmArguments(int port) {
```
robsunday, line 202: Good change! However it is a bit over-engineered in my opinion. Simple StringBuilder and additional private method would be enough, no need for maps and streams. It is just my opinion :-)

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/JettyIntegrationTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+  @Override
+  protected GenericContainer<?> createTargetContainer(int jmxPort) {
+    GenericContainer<?> container =
+        new GenericContainer<>(
+            new ImageFromDockerfile()
+                .withDockerfileFromBuilder(
+                    builder ->
+                        builder
+                            .from("jetty:11")
```
robsunday, line 29: Did you consider upgrade to jetty 12?

robsunday (approved):

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/JettyIntegrationTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+  @Override
+  protected void verifyMetrics() {
+    waitAndAssertMetrics(
+        metric ->
+            assertSumWithAttributes(
+                metric,
+                "jetty.session.count",
+                "The number of sessions established in total.",
+                "{sessions}",
```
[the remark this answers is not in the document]
SylvainJuge: I agree with you here, it should be `{session}` and not `{sessions}`. I think however that it would be easier to leave it as-is to ensure maximum compatibility, maybe with a comment about this inconsistency as an attempt not to reproduce it in the future.

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/JettyIntegrationTest.java
SylvainJuge: No, ideally we should test all those metrics on multiple versions but it's not the case for now, but it would be a nice improvement for later :-).

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/CassandraIntegrationTest.java
```diff
@@ -18,24 +18,17 @@
...
             "JVM_EXTRA_OPTS",
-            " -Dcassandra.jmx.remote.port="
-                + jmxPort
-                + " -Dcom.sun.management.jmxremote.rmi.port="
-                + jmxPort
-                + " -Dcom.sun.management.jmxremote.local.only=false"
-                + " -Dcom.sun.management.jmxremote.ssl=false"
-                + " -Dcom.sun.management.jmxremote.authenticate=false")
+            genericJmxJvmArguments(jmxPort)
+                // making cassandra startup faster for single node, from ~1min to ~15s
```
breedx-splk, line 30: nice 👍🏻

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/TargetSystemIntegrationTest.java
breedx-splk: Yeah, I think I agree...I would have just hard coded a bunch of string concats...but I don't think it's a huge problem.

jmx-scraper/src/main/resources/jetty.yaml
```diff
@@ -0,0 +1,71 @@
...
+rules:
+
+  - bean: org.eclipse.jetty.io:context=*,type=managedselector,id=*
+    mapping:
+      selectCount:
+        metric: jetty.select.count
+        type: counter
+        # Unit should not use a plural form, left as-is for compatibility with jetty.groovy
+        # https://opentelemetry.io/docs/specs/semconv/general/metrics/#instrument-units
+        unit: "{operations}"
```
breedx-splk, line 12: Is there a reason that we're trying to keep this compatibility instead of fixing it?

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/JettyIntegrationTest.java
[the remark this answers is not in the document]
breedx-splk: Yeah, it concerns me that we'll never be able to fix this if we perpetuate it. Any ideas on when/how we might do that?

breedx-splk (approved):
Looks good to me, but offered a couple discussion points.

jmx-scraper/src/main/resources/jetty.yaml
SylvainJuge: Yes, I think this is better as a first step to stay very close to the original implementation (even if it's buggy).
The unit tests are almost copied as-is which helps us prevent any regressions and to ensure that we have the same behavior between the two implementations.

However, I think that it would be a good idea to remove this (and the other ones) as soon as possible with a separate PR that fixes the two implementations (this one in yaml and groovy) at once. This should not be a breaking change as the units are very unlikely to be relied on beyond documentation, the captured values remain the same.

SylvainJuge: Giving it a second thought on this, the "we'll never be able to fix this if we perpetuate it" changed my mind and I've fixed that directly in this PR with 187ca89.

I was thinking more about the process of replicating existing tests with some fidelity than fixing small inconsistencies while we are at it when we can do it in a non-breaking way.

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/JettyIntegrationTest.java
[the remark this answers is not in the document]
SylvainJuge: fixed in 187ca89

PR #1531, opened 2024-11-06
add wildfly.yaml to JMX scraper

SylvainJuge:
**Description:**

- add wildfly support for jmx scraper with dedicated YAML configuration
- adds wildfly tests using `jboss-client.jar` for using HTTP management API
- fixes support for login/pwd authentication
- adds a test webapp needed to test wildfly session metrics
- moves existing test app to dedicated module for simplicity (and making it similar to test webapp)
- fixes use of plural form in units in `wildfly.groovy` to fit the stable [convention for metric units](https://opentelemetry.io/docs/specs/semconv/general/metrics/#instrument-units).

Part of #1362

**Notes:**

In the original groovy implementation session metrics are not tested, this requires deploying and application to the container, this is not properly tested in the YAML implementation.

In the original groovy implementation the gatherer was run directly from the same docker container probably for simplicity as `jboss-client.jar` is directly available.
When using two containers, the jar needs to be copied from wildfly container to scraper container and it complicates authentication: login/pwd can be skipped if both processes run on the same host and can access the same filesystem.
Testing here is done using two distinct containers as this setup is a bit more complex and thus likely to fail (doing so here helped me find that we did not handle it properly).


jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/ActiveMqIntegrationTest.java
```diff
@@ -25,11 +25,12 @@ protected GenericContainer<?> createTargetContainer(int jmxPort) {
                     builder -> builder.from("apache/activemq-classic:5.18.6").build()))
         .withEnv("JAVA_TOOL_OPTIONS", genericJmxJvmArguments(jmxPort))
         .withStartupTimeout(Duration.ofMinutes(2))
-        .waitingFor(Wait.forListeningPort());
+        .waitingFor(Wait.forLogMessage(".*Apache ActiveMQ.*started.*", 1));
```
SylvainJuge, line 28: [for reviewer] waiting for open port seems to sometimes not be accurate enough to detect proper startup, which makes the scraper container start when the JMX port is not yet available. Waiting on log message is an attempt to make this a bit less flaky (time will tell if that's a good idea or not here).

robsunday: Wait.forListeningPort() requires withExposedPorts(jmxPort) to be called to work properly. I think Tomcat test works by a chance.

SylvainJuge: As discussed today I've implemented it for all other containers in 3ab2f28.

jmx-scraper/src/main/resources/wildfly.yaml
```diff
@@ -0,0 +1,104 @@
...
+rules:
+  - bean: jboss.as:deployment=*,subsystem=undertow
+    metricAttribute:
+      deployment: param(deployment)
+    prefix: wildfly.session.
+    unit: "{session}"
+    mapping:
+      sessionsCreated:
+        metric: count
+        type: counter
```
robsunday, line 12: minor - could be defined as a common type

jmx-scraper/src/main/resources/wildfly.yaml
```diff
@@ -0,0 +1,104 @@
...
+        metric: request.time
+        unit: ns
+        desc: The total amount of time spent on requests.
+      errorCount:
+        metric: request.server_error
+        unit: "{request}"
+        desc: The number of requests that have resulted in a 5xx response.
+      bytesSent:
+        metric: network.io
+        unit: "by"
```
robsunday, line 48: "By" seems to be unit recommented by semconv. Can we just use "By" here?

jmx-scraper/src/main/resources/wildfly.yaml
```diff
@@ -0,0 +1,104 @@
...
+        metricAttribute:
+          state: const(in)
+
+  - bean: jboss.as:subsystem=datasources,data-source=*,statistics=pool
+    metricAttribute:
+      data_source: param(data-source)
+    type: counter
+    prefix: wildfly.jdbc.
+    mapping:
+      ActiveCount:
```
robsunday, line 65: ActiveCount and IdleCount could use YAML references since they share a lot. The same applies to few props in the bean below. What do you think about it?

robsunday (approved):

trask (approved):

PR #1533, opened 2024-11-07
Allow customization of parent-override behaviour for inferred-spans

JonasKunz:
**Description:**

The inferred spans extension sometimes has to insert inferredspans inbetween the parent-child relationship of normal spans for correctness.

This is however not directly possible, that is why we represent this as a span-link on the inferred spans. This PR adds a configuration hook to fully customize this behaviour with a custom handler.

The main reason for adding this is because in our elastic vendor distro of the opentelemetry agent we are still shipping the pre-contribution version of inferred spans. We currently still have to do this, because there these special span links are marked with an `elastic.is_child` instead of just `is_child` attribute and older versions of our backend rely on this elastic-specific attribute.

With this change we can switch to this contributed version of inferred spans and add the `elastic.is_child` attribute on the span link via the customization hook.


jackshirazi (approved):

inferred-spans/src/main/java/io/opentelemetry/contrib/inferredspans/internal/CallTree.java
```diff
@@ -427,6 +431,7 @@ int spanify(
       @Nullable Span parentSpan,
       TraceContext parentContext,
       SpanAnchoredClock clock,
+      BiConsumer<SpanBuilder, SpanContext> normalSpanParentOverride,
```
SylvainJuge, line 434: [minor] here I would suggest to remove the `normal` prefix from the name as it's not really clear what "normal" means in this context, to me it should be either the default value or the overridden one.

SylvainJuge (approved):

inferred-spans/README.md
```diff
@@ -41,6 +41,8 @@ So if you are using an autoconfigured OpenTelemetry SDK, you'll only need to add
...
+| otel.inferred.spans.parent.override.handler <br/> OTEL_INFERRED_SPANS_PARENT_OVERRIDE_HANDLER | Defaults to a handler adding span-links to the inferred span                                                                                                                                                                                                      | Inferred spans sometimes need to be inserted as the new parent of a normal span, which is not directly possible because that span has already been sent. For this reason, this relationship needs to be represented differently, which normally is done by adding a span-link to the inferred span. This configuration can be used to override that behaviour by providing the fully qualified name of a class implementing `BiConsumer<SpanBuilder, SpanContext>`: The biconsumer will be invoked with the inferred span as first argument and the span for which the inferred one was detected as new parent as second argument |
```
trask, line 45: I wonder if you could use [ComponentProvider](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk-extensions/autoconfigure-spi/src/main/java/io/opentelemetry/sdk/autoconfigure/spi/internal/ComponentProvider.java), which is how file configuration is supporting configuring things like this (instead of fully-qualified class name)

cc @jack-berg

JonasKunz: Eventually this definitely would make sense I think. However, is the new structured config API already in a state so that we can fully replace the usage of the `AutoConfigurationCustomizerProvider` here? Because I don't think that we can simply mix in the `ComponentProvider` to my understanding

trask: makes sense

I sent a PR to your branch just to demonstrate what I was thinking: https://github.com/JonasKunz/opentelemetry-java-contrib/pull/1

but it does require depending on a bunch of internal / experimental classes, so really not sure if it's worth it yet

jack-berg: ComponentProvider is for providing implementations of SDK plugin extension points referenced in the declarative confg files. I.e. samplers, processors, exporters, etc. 

@trask it looks like you're using it to provide an implementation of ParentOverrideHandler (not an SDK plugin extension point), and using outside of declarative config.

trask: oh yeah, I was thinking maybe we could use component provider for custom declarative config stuff, but I guess declarative config has no way to know about what nodes represent components and what provider types to use

trask (approved):

PR #1538, opened 2024-11-12
JMX Scraper - YAML config and integration test for HBase

robsunday:
- Created YAML config for connecting JMX Insight to HBase created based on jmx-metrics groovy script
- Updated jmx-metrics units to match semconv rules

"Hadoop:service=HBase,name=Master,sub=AssignmentManager" metrics are defined as updowncounter, while in my opinion it should be just gauges. I kept updowncounters to minimize incompatibilities.
We may consider if updownocunters should be converted to gauges in this or the following PR


jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/MetricAssertions.java
```diff
@@ -41,18 +41,20 @@ static void assertSum(
     assertThat(metric.getName()).isEqualTo(name);
     assertThat(metric.getDescription()).isEqualTo(description);
     assertThat(metric.getUnit()).isEqualTo(unit);
-    assertThat(metric.hasSum()).isTrue();
+    assertThat(metric.hasSum()).withFailMessage("Invalid metric type").isTrue();
+    assertThat(metric.getSum().getIsMonotonic())
+        .withFailMessage("Metric should " + (isMonotonic ? "" : "not ") + "be monotonic")
```
robsunday, line 46: [for reviewer] Meaningful error reported when assertion fails

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/MetricAssertions.java
robsunday, line 1: [for reviewer] Some small improvements to report meaningful errors for few failed assertions

jmx-metrics/src/integrationTest/java/io/opentelemetry/contrib/jmxmetrics/target_systems/HbaseIntegrationTest.java
```diff
@@ -381,7 +381,7 @@ void endToEnd() {
                 metric,
                 "hbase.region_server.authentication.count",
                 "Number of client connection authentication failures/successes.",
-                "{authentication requests}",
+                "{authentication request}",
```
SylvainJuge, line 384: [question] is there anything mentioned about spaces or the fact that is must be a single word/noun in the semantic conventions ? I think this could be something to be further discussed later but I think the fact it's about authentication is already provided by the metric name.

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/MetricAssertions.java
```diff
@@ -27,7 +27,7 @@ static void assertGauge(Metric metric, String name, String description, String u
     assertThat(metric.getName()).isEqualTo(name);
     assertThat(metric.getDescription()).isEqualTo(description);
     assertThat(metric.getUnit()).isEqualTo(unit);
-    assertThat(metric.hasGauge()).isTrue();
+    assertThat(metric.hasGauge()).withFailMessage("Invalid metric type").isTrue();
```
SylvainJuge, line 30: [minor] assertion description us usually only used trying to understand why things fail, so rephrasing it like `metric with gauge expected` allows to understand what is missing/wrong without even looking at the assertion code, so things are easier to understand from the failed test output.

jmx-scraper/src/main/resources/hbase.yaml
```diff
@@ -0,0 +1,329 @@
...
+    prefix: hbase.region_server.
+    type: updowncounter
+    metricAttribute:
+      region_server: &hostname beanattr(tag\.Hostname)
+    mapping:
+      regionCount:
+        metric: region.count
+        unit: "{region}"
+        desc: The number of regions hosted by the region server.
+
```
SylvainJuge, line 48: [minor] the spacing between blocks is intentional to keep things a bit readable here, maybe adding an explicit "header comment" with for example the metric name/prefix/suffix would help to understand this is intentional and needs to be preserved when maintaining this.

jmx-scraper/src/main/resources/hbase.yaml
```diff
@@ -0,0 +1,329 @@
...
+        desc: Append operation max latency.
+      Append_min:
+        metric: operation.append.latency.min
+        desc: Append operation minimum latency.
+      Append_mean:
+        metric: operation.append.latency.mean
+        desc: Append operation mean latency.
+      Append_median:
+        metric: operation.append.latency.median
+        desc: Append operation median latency.
```
SylvainJuge, line 180: Not related to this PR, but I am pretty sure those are metrics [created with dropwizard](https://metrics.dropwizard.io/4.2.0/getting-started.html#reporting-via-jmx), we need to make sure those are mapped consistently for every supported system where they are used. The downside of those here is that there is no way for us to translate that easily to an otel histogram as we don't have access to the individual samples.

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/HBaseIntegrationTest.java
```diff
@@ -0,0 +1,441 @@
...
+import org.testcontainers.containers.wait.strategy.Wait;
+
+public class HBaseIntegrationTest extends TargetSystemIntegrationTest {
+  private static final int DEFAULT_MASTER_SERVICE_PORT = 16000;
+
+  @Override
+  protected GenericContainer<?> createTargetContainer(int jmxPort) {
+    return new GenericContainer<>("dajobe/hbase")
+        .withEnv("JAVA_HOME", "/usr/lib/jvm/java-8-openjdk-amd64")
+        .withEnv("HBASE_OPTS", "-XX:+UseConcMarkSweepGC")
```
SylvainJuge, line 27: [minor] Is the GC algorithm is really needed for the JMX metrics ? If yes then adding a comment to explain why could be relevant, otherwise we can simply remove it as we are not really doing a production ready setup here.

SylvainJuge (approved):

jmx-metrics/src/integrationTest/java/io/opentelemetry/contrib/jmxmetrics/target_systems/HbaseIntegrationTest.java
robsunday: I looked in the docs specifically for whitespaces on the unit labels, but I did not find anything. That's why I left it here

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/HBaseIntegrationTest.java
robsunday: Removed

jmx-scraper/src/main/resources/hbase.yaml
robsunday: Comments added

jmx-scraper/src/main/resources/hbase.yaml
```diff
@@ -7,6 +7,7 @@ rules:
     unit: "{server}"
     type: updowncounter
     mapping:
+      # Group of properties to build hbase.master.region_server.count metric
```
SylvainJuge, line 10:
```suggestion
      # hbase.master.region_server.count
```
[nitpicking] Maybe simplify this even further as everything here is a metric (and do the same with others that match `Group of properties to build .* metric` regex)

SylvainJuge (approved):

jmx-scraper/src/main/resources/hbase.yaml
robsunday: Done

SylvainJuge (approved):

jmx-metrics/src/main/resources/target-systems/hbase.groovy
```diff
@@ -16,45 +16,45 @@
 
 def beanMasterServer = otel.mbeans("Hadoop:service=HBase,name=Master,sub=Server")
 otel.instrument(beanMasterServer, "hbase.master.region_server.count",
-  "The number of region servers.", "{servers}",
+  "The number of region servers.", "{server}",
```
trask, line 19: I'm probably misremembering, but I thought the plan was to not change things in the existing `jmx-metrics` component, and fix them only in the `jmx-scraper` implementation?

SylvainJuge: That was our initial plan, however we (or at least I) changed a bit our mind about it for units because:
- units is just metadata, the values captured remain the same and it's quite unlikely any implementation strongly relies on those values (for example, it might break some UI/i18n, but the values remain the same). Changing the metric or its attribute names would not be the same minor impact.
- the [semconv conventions about this are stable](https://opentelemetry.io/docs/specs/semconv/general/metrics/#instrument-units), thus those changes will eventually have to be done at some point, and this makes one less difference to deal with when enhancing the mappings later.

We already did similar changes with other JMX scraper PRs, for example:
- https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1531
- https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1517

robsunday: I may only add that this change should be perfectly safe since the only changes were made in unit annotations that are discarded by the parsers anyway (according to [these docs](https://ucum.org/ucum#para-curly)).

trask: makes sense, thanks!

trask (approved):

PR #1545, opened 2024-11-16
Fix required status check

trask:
This fixes the problem that I was able to merge #1544 even though one of the tests was failing.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -153,7 +153,7 @@ jobs:
           GPG_PASSWORD: ${{ secrets.GPG_PASSWORD }}
 
   required-status-check:
-    if: github.event_name == 'pull_request'
+    if: github.event_name == 'pull_request' && always()
     needs:
       - build
       - integration-test


.github/workflows/build.yml
```diff
@@ -153,7 +153,7 @@ jobs:
           GPG_PASSWORD: ${{ secrets.GPG_PASSWORD }}
 
   required-status-check:
-    if: github.event_name == 'pull_request'
+    if: github.event_name == 'pull_request' && always()
```
breedx-splk, line 156: I'm not following how this helps. Isn't `&& always()` basically `&& true` which is always true?

trask: one would think...

check out https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#always

and you can see the difference below, `required-status-check` is run and fails (and so should block merging even after someone approves this PR), whereas in #1542 for example, `required-status-check` is skipped (and therefore didn't block merging)

trask:
I'm going to rebase now that #1546 has been merged, which should fix the integration tests and make `required-status-check` pass (allowing this PR to be merged)

breedx-splk (approved):

PR #1546, opened 2024-11-16
Revert "fix(deps): update dependency com.linecorp.armeria:armeria-bom to v1.31.0 (#1542)"

trask:
#1542 was merged even though there was a failing test, this PR reverts that change until we can figure out what the underlying issue is (renovate should dutifully re-open the update).

#1545 fixes CI to prevent merging in the future when there is a failing test.

### DIFF
--- a/aws-resources/src/test/java/io/opentelemetry/contrib/aws/resource/SimpleHttpClientTest.java
+++ b/aws-resources/src/test/java/io/opentelemetry/contrib/aws/resource/SimpleHttpClientTest.java
@@ -10,7 +10,6 @@ import static org.assertj.core.api.Assertions.assertThat;
 import com.google.common.collect.ImmutableMap;
 import com.linecorp.armeria.common.AggregatedHttpRequest;
 import com.linecorp.armeria.common.HttpResponse;
-import com.linecorp.armeria.common.TlsKeyPair;
 import com.linecorp.armeria.server.ServerBuilder;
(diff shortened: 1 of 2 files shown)


aws-resources/src/test/java/io/opentelemetry/contrib/aws/resource/SimpleHttpClientTest.java
```diff
@@ -69,7 +68,7 @@ static class HttpsServerTest {
         new ServerExtension() {
           @Override
           protected void configure(ServerBuilder sb) {
-            sb.tls(TlsKeyPair.of(certificate.privateKeyFile(), certificate.certificateFile()));
+            sb.tls(certificate.certificateFile(), certificate.privateKeyFile());
```
jack-berg, line 71: Check out what I did in core to fix this deprecation: https://github.com/open-telemetry/opentelemetry-java/pull/6877/files
```suggestion
            sb.tls(TlsKeyPair.of(certificate.privateKey(), certificate.certificate()));
```

jack-berg (approved):

aws-resources/src/test/java/io/opentelemetry/contrib/aws/resource/SimpleHttpClientTest.java
trask: will fix when renovate re-opens, just want to roll this back which should never have been merged in the first place (#1545)

PR #1554, opened 2024-11-19
[jmx-scraper] keep old servlet API and java8 compatibility

SylvainJuge:
- Pin servlet version to 5.0.0 to prevent update PRs like #1543 
- use java source and target code level instead of toolchain contraints, given there isn't any java 8 compatible toolchain available for aarch64, we can replace it with any other compatible version that can compile to Java 8 (and let gradle find a suitable one).

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -87,7 +87,7 @@
     },
     {
       // pinned version for compatibility
-      "matchFileNames": ["jmx-scraper/test-webapp/**"],
+      "matchFileNames": ["jmx-scraper/test-webapp/build.gradle.kts"],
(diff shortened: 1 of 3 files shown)


trask:
> I've opened https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1554 with the needed changes to the renovate config bundled with another change that allows building on recent ARM-based MacOS systems.

can you explain this part a bit more? why is it only a problem for these two modules? thanks!

trask:
can you check if this addresses the issue? #1555

SylvainJuge:
#1555 fixes the issue, so I'll close this PR. Thanks @robsunday for double-checking it now works as expected in MacOS + ARM.

PR #1566, opened 2024-11-22
Assertions refactoring

robsunday:
Preview of new metrics assertions framework.
Assertion errors are now reported in much more compact and descriptive form.
Error message always contain only data related to failing metric, as opposed to old assertions where all received metrics were listed and it was extremely hard to find what is wrong in large set of metrics.


jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/CassandraIntegrationTest.java
```diff
@@ -44,110 +41,74 @@ protected JmxScraperContainer customizeScraperContainer(
...
+            "us")
+        .assertCounter(
+            "cassandra.compaction.tasks.completed",
+            "Number of completed compactions since server [re]start",
+            "1")
+        .assertGauge(
+            "cassandra.compaction.tasks.pending",
+            "Estimated number of compactions remaining to perform",
+            "1")
+        .assertUpDownCounter(
```
robsunday, line 90: [question to revier] In original framework we had assertSum method with overloaded version accepting isMonotonic parameter added. This was little unclear to me for two reasons:
1) It was not obvious that assertSum without isMonotonic parameter was the same as assertSum(isMonotonic=true)
2) In YAML we used instrument types gauge, counter, updowncounter, while in assertions we used assertGauge, assertSum with and without isMonotonic parameter. 

That's why I decided to implement assertGauge, assertCounter and assertUpDownCounter method families here.
I know it may lead to more methods to be implemented but for me this new API is cleaner.

What do you think about this approach?

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/MetricsVerifier.java
robsunday, line 1: [question to revier] In original framework we had assertSum method with overloaded version accepting isMonotonic parameter added. This was little unclear to me for two reasons:

It was not obvious that assertSum without isMonotonic parameter was the same as assertSum(isMonotonic=true)
In YAML we used instrument types gauge, counter, updowncounter, while in assertions we used assertGauge, assertSum with and without isMonotonic parameter.
That's why I decided to implement assertGauge, assertCounter and assertUpDownCounter method families here.
I know it may lead to more methods to be implemented but for me this new API is cleaner.

What do you think about this approach?

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/JvmIntegrationTest.java
```diff
@@ -48,43 +44,147 @@ protected void verifyMetrics() {
...
+     List<String> gcLabels =
+       Arrays.asList(
+           "Code Cache",
+           "PS Eden Space",
+           "PS Old Gen",
+           "Metaspace",
+           "Compressed Class Space",
+           "PS Survivor Space");
+   List<String> gcCollectionLabels = Arrays.asList("PS MarkSweep", "PS Scavenge");
+  */
```
SylvainJuge, line 189: leftover from my PR, we should remove it

SylvainJuge (approved):

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/TargetSystemIntegrationTest.java
```diff
@@ -125,6 +125,8 @@ void endToEndTest(@TempDir Path tmpDir) {
     verifyMetrics();
   }
 
+  // TODO: This implementation is DEPRECATED and will be removed once all integration tests are
+  // migrated to MetricsVerifier
   protected void waitAndAssertMetrics(Iterable<Consumer<Metric>> assertions) {
```
SylvainJuge, line 130: [for reviewer] we can't use `@Deprecated` here without introducing compiler warnings and having to use `@SuppressWarnings` in test code, so migrating away from this will be done in multiple steps as part of https://github.com/open-telemetry/opentelemetry-java-contrib/issues/1573

SylvainJuge (approved):

PR #1567, opened 2024-11-23
Update develocity configuration

trask:
TODO delete `GRADLE_ENTERPRISE_ACCESS_KEY` github action secret after merging this

### DIFF
--- a/buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
@@ -71,13 +71,6 @@ tasks {
       showCauses = true
       showStackTraces = true
     }
-
-    develocity.testRetry {
-      // You can see tests that were retried by this mechanism in the collected test reports and build scans.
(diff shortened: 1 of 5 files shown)


buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
```diff
@@ -71,13 +71,6 @@ tasks {
       showCauses = true
       showStackTraces = true
     }
-
-    develocity.testRetry {
-      // You can see tests that were retried by this mechanism in the collected test reports and build scans.
-      if (System.getenv().containsKey("CI") || rootProject.hasProperty("retryTests")) {
-        maxRetries.set(5)
-      }
-    }
```
trask, line 80: we may be able to keep this, but we won't be able to see flaky tests across runs anymore, so maybe better to remove it (at least in this repo)

laurit: I think removing it is fine, if flaky tests are an issue we can work on them or reintroduce the retry.

laurit (approved):

PR #1581, opened 2024-12-06
Change to release after SDK and before Instrumentation

trask:
cc @jackshirazi

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -12,9 +12,9 @@ as the last step, which publishes a snapshot build to
 
 ## Release cadence
 
-This repository roughly targets monthly minor releases from the `main` branch on the Friday after
+This repository roughly targets monthly minor releases from the `main` branch on the Tuesday after
 the second Monday of the month (roughly a couple of days after the monthly minor release of
-[opentelemetry-java-instrumentation](https://github.com/open-telemetry/opentelemetry-java-instrumentation)).
+[opentelemetry-java](https://github.com/open-telemetry/opentelemetry-java)).
 
 ## Preparing a new major or minor release
 
--- a/dependencyManagement/build.gradle.kts
+++ b/dependencyManagement/build.gradle.kts
@@ -7,15 +7,13 @@ data class DependencySet(val group: String, val version: String, val modules: Li
 val dependencyVersions = hashMapOf<String, String>()
 rootProject.extra["versions"] = dependencyVersions
 
-val otelInstrumentationVersion = "2.10.0-alpha"
-
 val DEPENDENCY_BOMS = listOf(
   "com.fasterxml.jackson:jackson-bom:2.18.2",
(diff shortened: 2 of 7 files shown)


RELEASING.md
```diff
@@ -12,9 +12,9 @@ as the last step, which publishes a snapshot build to
 
 ## Release cadence
 
-This repository roughly targets monthly minor releases from the `main` branch on the Friday after
+This repository roughly targets monthly minor releases from the `main` branch on the Tuesday after
```
trask, line 15: I'm thinking to change the Instrumentation release date to the Thursday after this

dependencyManagement/build.gradle.kts
```diff
@@ -7,15 +7,13 @@ data class DependencySet(val group: String, val version: String, val modules: Li
...
-val otelInstrumentationVersion = "2.10.0-alpha"
-
 val DEPENDENCY_BOMS = listOf(
   "com.fasterxml.jackson:jackson-bom:2.18.2",
   "com.google.guava:guava-bom:33.3.1-jre",
   "com.linecorp.armeria:armeria-bom:1.31.2",
   "org.junit:junit-bom:5.11.3",
   "io.grpc:grpc-bom:1.68.2",
-  "io.opentelemetry.instrumentation:opentelemetry-instrumentation-bom-alpha:${otelInstrumentationVersion}",
+  "io.opentelemetry:opentelemetry-bom-alpha:1.44.1-alpha",
```
trask, line 16: I didn't update to 1.45.0 in order to ensure everything works smoothly when renovate updates it after this PR is merged

laurit (approved):

jmx-scraper/build.gradle.kts
```diff
@@ -22,7 +22,7 @@ dependencies {
   runtimeOnly("io.opentelemetry:opentelemetry-exporter-otlp")
   runtimeOnly("io.opentelemetry:opentelemetry-exporter-logging")
 
-  implementation("io.opentelemetry.instrumentation:opentelemetry-jmx-metrics")
+  implementation("io.opentelemetry.instrumentation:opentelemetry-jmx-metrics:2.10.0-alpha")
```
SylvainJuge, line 25: Moving contrib before instrumentation in the release allows to remove this circular dependency if we move the JMX insights code directly into contrib and then reuse it as-is in instrumentation. Do you think this would be something relevant to do ? It also changes a bit the "ownership" of the JMX-related code in the instrumentation by relying on the component owners of contrib instead.

trask: I think I'd prefer to go in the other direction and keep / move things that we release as part of the core java agent distribution into the java instrumentation repo

@jackshirazi would this address your concern about release ordering, and we could go back to releasing instrumentation repo first and then contrib repo?

Issue #1573, opened 2024-11-26
JMX Scraper: Refactoring integration tests assertion framework


robsunday:
### Component(s)

jmx-scraper

### Is your feature request related to a problem? Please describe.

Current implementation of integration tests assertion framework (MetricAssertions class) reports assertion failures in a very verbose and non-optimal way.
In case of assertion failure it prints out all received metrics and assertions multiple times. It makes finding the issue really hard. The more metrics is checked the harder it is and often the only way to find out what went wrong is selectively commenting out the code.
Also some assertions are missing details forcing developer to guess what went wrong.

### Describe the solution you'd like

New assertion framework should be created fulfilling the following requirements:

1. Print only the metric that failed the assertion
2. Provide meaningful description and the metric property details needed to identify root cause of assertion failure
3. Fail when any of expected metrics was not received
4. Fail when unexpected metric was received (there is no assertion defined for it). There should be mechanism to opt out form this default behavior.  It may be useful when it is very hard/impossible to force tested target to report some metric.
5. API must be easy to use, concise and reliable (it should be hard to miss some metric checks)

New framework will coexist for some time with the current one, so currently implemented integration tests may be ported incrementally in separate PRs.
Once all tests are ported the old framework will be removed from the code

SylvainJuge:
https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1566 relates to this.

The migration work will likely be spread over many PRs.

robsunday:
[#1580](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1580) - Attribute assertions updated

robsunday:
New framework is implemented

PR #1589, opened 2024-12-10
[jmx-scraper] Assertions refactoring - Tomcat integration test converted

robsunday:
**Description:**

1. Migrated **Tomcat integration test** to use the new assertion framework.
2. Update units to match [semconv](https://opentelemetry.io/docs/specs/semconv/general/metrics/#instrument-units).
3. Update units and integration test in JMX Metrics Gatherer to be aligned with semconv and consistent with JMX Scraper .

Part of https://github.com/open-telemetry/opentelemetry-java-contrib/issues/1573


jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/TomcatIntegrationTest.java
```diff
@@ -46,67 +46,76 @@ protected JmxScraperContainer customizeScraperContainer(
...
-                "requests",
-                attrs -> attrs.containsOnly(entry("proto_handler", "\"http-nio-8080\""))));
+  protected MetricsVerifier createMetricsVerifier() {
+    return MetricsVerifier.create()
+        .add(
+            "tomcat.sessions",
+            metric ->
+                metric
+                    .hasDescription("The number of active sessions")
+                    .hasUnit("sessions") // TODO: not aligned with semconv. Should be "{session}"
```
SylvainJuge, line 56: Like in the previous PRs where such units inconsistencies were found I think it's worth fixing them straight away as they aren't "breaking changes" most of the time and we will have to fix it at some point because it's now part of the stable semantic conventtions.

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/TomcatIntegrationTest.java
```diff
@@ -46,67 +46,76 @@ protected JmxScraperContainer customizeScraperContainer(
...
+                    .hasDescription("The total processing time")
+                    .hasUnit("ms")
+                    .isCounter()
+                    .hasDataPointsWithOneAttribute(attribute("proto_handler", "\"http-nio-8080\"")))
+        .add(
+            "tomcat.traffic",
+            metric ->
+                metric
+                    .hasDescription("The number of bytes transmitted and received")
+                    .hasUnit("by")
```
SylvainJuge, line 80:
```suggestion
                    .hasUnit("By")
```
See https://opentelemetry.io/docs/specs/semconv/general/metrics/#instrument-units https://unitsofmeasure.org/ucum#para-48 and

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/TomcatIntegrationTest.java
robsunday: So my understanding is that in this case it may be a kind of breaking change. This is not the annotation in curly braces that is supposed to be skipped by parser.
I've been hesitating if I should update it or not and finally decided to leave it, but if reviewers agree that it should be changed then I'll gladly update it.

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/TomcatIntegrationTest.java
robsunday: Yes, however the same doubts apply here

SylvainJuge (approved):
As discussed today, for now we can leave the units as is as those are not in curly braces and we are not just changing from plural to singular form, same for bytes we can leave this inconsistency with a TODO and deal with it later. "Later" here means likely when we revisit those definitions and align with instrumentation.

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/TomcatIntegrationTest.java
trask: I think it's ok to update it since we are updating units elsewhere (in OTel we've said changing units is breaking whether they're curly braces or not)

robsunday: Great! I'll make the fix then

SylvainJuge (approved):

trask (approved):

PR #1595, opened 2024-12-12
[jmx-scraper] add support for Solr

SylvainJuge:
**Description:**

Adds Solr to jmx-scraper

- fix units for solr jmx metrics (in jmx scraper and jmx gatherer)
- add solr tests + yaml metrics


jmx-scraper/src/main/resources/solr.yaml
```diff
@@ -0,0 +1,110 @@
...
+        unit: '{insert}'
+        desc: The number of inserts to a cache.
+      cumulative_lookups:
+        metric: lookup.count
+        unit: '{lookup}'
+        desc: The number of lookups to a cache.
+      ramBytesUsed:
+        type: updowncounter
+        metric: size
+        unit: 'By'
```
robsunday, line 107: Tere is some inconsistency in a way units are defined. Sometimes name is enclosed in single quotes, sometimes not (see solr.index.size metric definition above).
Also in other yaml files we used double quotes. I think it would be good to keep consistency

jmx-scraper/src/main/resources/solr.yaml
```diff
@@ -43,7 +43,7 @@ rules:
       Mean:
         metric: solr.request.time.average
         type: gauge
-        unit: 'ms'
+        unit: "ms"
```
robsunday, line 46: I believe we used double quotes only in cases where unit was annotated. Plain curly braces caused parse error.

robsunday (approved):

SylvainJuge:
This is currently failing due to a flaky wildfly test which is not related

```
WildflyIntegrationTest > endToEndTest(Path) FAILED
    org.opentest4j.AssertionFailedError: [no data point matched attribute set '[{server=default-server}, {listener=default}, {state=in}]' for metric 'wildfly.network.io'] 
    expected: true
     but was: false
        at app//io.opentelemetry.contrib.jmxscraper.assertions.MetricAssert.lambda$hasDataPointsWithAttributes$2(MetricAssert.java:262)
        at app//io.opentelemetry.contrib.jmxscraper.assertions.MetricAssert.checkDataPoints(MetricAssert.java:188)
        at app//io.opentelemetry.contrib.jmxscraper.assertions.MetricAssert.hasDataPointsWithAttributes(MetricAssert.java:231)
        at app//io.opentelemetry.contrib.jmxscraper.target_systems.WildflyIntegrationTest.lambda$createMetricsVerifier$8(WildflyIntegrationTest.java:164)
```

We will probably fix that in a separate PR, as it might require to relax a bit the metric assertions or wait for longer until the metric is properly reported.

jmx-scraper/src/main/resources/solr.yaml
```diff
@@ -0,0 +1,110 @@
...
+
+  - beans:
+      - solr:dom1=core,dom2=*,category=QUERY,scope=*,name=requestTimes
+      - solr:dom1=core,dom2=*,category=UPDATE,scope=*,name=requestTimes
+    mapping:
+      Mean:
+        metric: solr.request.time.average
+        type: gauge
+        unit: ms
+        desc: "The average time of a query, based on Solr's histogram configuration."
```
breedx-splk, line 47: nit: I like the word "duration" here in place of "time" -- slightly less room for ambiguity.

Also don't we want to normalize units to seconds?

breedx-splk (approved):
Looks good. Had just one small comment about unit and time vs. duration. Thanks!

jmx-scraper/src/main/resources/solr.yaml
SylvainJuge: The plan for now is to provide the same metrics as the ones provided by the jmx gatherer, so we keep the current metric names. But however the long term plan is being able to enhance them in the future and remove duplication between instrumentation and jmx-scraper.

For the normalization of value to seconds, while I see the benefit of consistency on the consumer side, it brings a few extra challenges here: we have to add a way to convert existing values (this could for example be done automatically from the provided unit when it's time-related), when capturing very small values like nanoseconds or microseconds it could introduce some challenges with float conversions and lack of precision, and it makes debugging and understanding harder because the value captured at the JMX level is documented in a unit and the produced data would be in another unit which could be misleading.

PR #1596, opened 2024-12-12
Update change log for upcoming release

trask:


CHANGELOG.md
```diff
@@ -2,6 +2,46 @@
...
+- Align JVM units to semconv
+  ([#1593](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1593))
+
+### JMX scraper - New 🌟
+
+The future of the [JMX metrics](./jmx-metrics/README.md) component,
+built on top of the
+[JMX metrics](https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation/jmx-metrics/javaagent#jmx-metric-insight)
+component from the opentelemetry-java-instrumentation repository.
+
```
trask, line 37: cc @SylvainJuge @robsunday

.github/scripts/get-version.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/bash -e
 
-grep -Po "val stableVersion = \"\K[0-9]+.[0-9]+.[0-9]+" version.gradle.kts
+grep "val stableVersion = " version.gradle.kts | grep -Eo "[0-9]+.[0-9]+.[0-9]+"
```
breedx-splk, line 3: I tested both on debian and macos...the `grep -p` one fails on a mac, so that seems like a worthy improvement when developing locally....

breedx-splk (approved):
👍🏻

.github/scripts/get-version.sh
trask: for some reason `-P` used to work for me (windows / git bash) but no longer 🤷‍♂️

SylvainJuge (approved):

CHANGELOG.md
```diff
@@ -2,6 +2,46 @@
...
+- Align HBase metric units to semconv
+  ([#1538](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1538))
+- Align Cassandra metric units to semconv
+  ([#1591](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1591))
+- Align Tomcat metric units to semconv
+  ([#1589](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1589))
+- Align JVM units to semconv
+  ([#1593](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1593))
+
```
robsunday, line 30: There is one more: https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1553

robsunday (changes requested):

CHANGELOG.md
trask: oh, it looks like these were included in the last release, but I missed adding them to the change log / release notes

I'll send a separate PR to update the change log for 1.41.0 and I'll manually update the release notes

.github/scripts/get-version.sh
laurit: `-P` is probably a GNU grep extension. Try running `grep --version`. Mac uses `grep (BSD grep, GNU compatible) 2.6.0-FreeBSD` `brew install grep` installs gnu grep as `ggrep`.

PR #1603, opened 2024-12-13
Revert "Change to release after SDK and before Instrumentation (#1581)"

trask:
(it also broke the release process)

but more importantly, I think the more general dependency direction we want to promote is having contrib components using `opentelemetry-instrumentation-api` and `opentelemetry-javaagent-extension-api`

I'd prefer to explore moving contrib components into the instrumentation repo if we are including them in the java agent distribution

cc @jackshirazi

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -12,9 +12,9 @@ as the last step, which publishes a snapshot build to
 
 ## Release cadence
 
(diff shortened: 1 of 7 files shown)


breedx-splk (approved):
Thanks. It's a little unfortunate, but let's get unblocked before discussing again for next time. :) 

Mentioning #1581 here for the easy click.

PR #1604, opened 2024-12-13
Inline incubating semconv attributes

trask:
Follow-up to @laurit's https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1585#discussion_r1878030973

### DIFF
--- a/aws-xray/build.gradle.kts
+++ b/aws-xray/build.gradle.kts
@@ -14,7 +14,7 @@ dependencies {
 
   implementation("com.squareup.okhttp3:okhttp")
   implementation("io.opentelemetry.semconv:opentelemetry-semconv")
-  implementation("io.opentelemetry.semconv:opentelemetry-semconv-incubating")
+  testImplementation("io.opentelemetry.semconv:opentelemetry-semconv-incubating")
 
   annotationProcessor("com.google.auto.service:auto-service")
   testImplementation("com.google.auto.service:auto-service")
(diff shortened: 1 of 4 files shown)


breedx-splk (commented):
I think I must have missed something in the comment...why is it desirable to copy/duplicate these constants?

trask:
https://github.com/open-telemetry/semantic-conventions-java?tab=readme-ov-file#published-releases

> This artifact has the [-alpha](https://github.com/open-telemetry/opentelemetry-java/blob/main/VERSIONING.md) and comes with no compatibility guarantees. Libraries can use this for testing, but should make copies of the attributes to avoid possible runtime errors from version conflicts.

breedx-splk:
> https://github.com/open-telemetry/semantic-conventions-java?tab=readme-ov-file#published-releases
> 
> > This artifact has the [-alpha](https://github.com/open-telemetry/opentelemetry-java/blob/main/VERSIONING.md) and comes with no compatibility guarantees. Libraries can use this for testing, but should make copies of the attributes to avoid possible runtime errors from version conflicts.

So the entire purpose here is to avoid taking a dependency on an artifact with `-alpha` in the suffix? I'd rather do that than have copies of the values that could deviate from the semconv over time....

Maybe we figure that the tests (which use the `-alpha`) give us enough confidence?

trask:
we're worried about diamond dependency problem where some users will end up depending (transitively) on different irreconcilable versions of the semconv alpha artifact

> Maybe we figure that the tests (which use the `-alpha`) give us enough confidence?

yeah, though it would be interesting to validate this

laurit:
The guidance given in https://github.com/open-telemetry/semantic-conventions-java/blob/main/README.md states that

> Libraries can use this for testing, but should make copies of the attributes to avoid possible runtime errors from version conflicts.

laurit (approved):

trask:
@breedx-splk I'm going to merge since this is the (current) convention in this repo, but we can revisit

Issue #1601, opened 2024-12-13
BaggageProcessorCustomizerTest sporadically failing


trask:
https://github.com/open-telemetry/opentelemetry-java-contrib/actions/runs/12319402440/job/34386355480?pr=1598

(the reason this didn't show up before is because we used to auto-retry tests)

```
> Task :baggage-processor:test

BaggageProcessorCustomizerTest > test_customizer() FAILED
    org.awaitility.core.ConditionTimeoutException: Assertion condition defined as a Lambda expression in io.opentelemetry.contrib.baggage.processor.BaggageProcessorCustomizerTest null within 1 seconds.
        at app//org.awaitility.core.ConditionAwaiter.await(ConditionAwaiter.java:167)
        at app//org.awaitility.core.AssertionCondition.await(AssertionCondition.java:119)
        at app//org.awaitility.core.AssertionCondition.await(AssertionCondition.java:31)
        at app//org.awaitility.core.ConditionFactory.until(ConditionFactory.java:1006)
        at app//org.awaitility.core.ConditionFactory.untilAsserted(ConditionFactory.java:790)
        at app//io.opentelemetry.contrib.baggage.processor.BaggageProcessorCustomizerTest.assertCustomizer(BaggageProcessorCustomizerTest.java:82)
        at app//io.opentelemetry.contrib.baggage.processor.BaggageProcessorCustomizerTest.test_customizer(BaggageProcessorCustomizerTest.java:53)

        Caused by:
        java.util.concurrent.TimeoutException
            at java.base/java.util.concurrent.FutureTask.get(FutureTask.java:204)
            at org.awaitility.core.Uninterruptibles.getUninterruptibly(Uninterruptibles.java:101)
            at org.awaitility.core.Uninterruptibles.getUninterruptibly(Uninterruptibles.java:81)
            at org.awaitility.core.ConditionAwaiter.await(ConditionAwaiter.java:103)
            ... 6 more
```

PR #1607, opened 2024-12-16
increase timeout for tests - to tolerate high loads

zeitlinger:
fixes https://github.com/open-telemetry/opentelemetry-java-contrib/issues/1601#event-15651342584

### DIFF
--- a/baggage-processor/src/test/java/io/opentelemetry/contrib/baggage/processor/BaggageProcessorCustomizerTest.java
+++ b/baggage-processor/src/test/java/io/opentelemetry/contrib/baggage/processor/BaggageProcessorCustomizerTest.java
@@ -78,7 +78,7 @@ class BaggageProcessorCustomizerTest {
     }
 
     await()
-        .atMost(Duration.ofSeconds(60))
+        .atMost(Duration.ofSeconds(10))
         .untilAsserted(
             () -> {
               assertThat(spanExporter.getFinishedSpanItems())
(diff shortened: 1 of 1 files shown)


baggage-processor/src/test/java/io/opentelemetry/contrib/baggage/processor/BaggageProcessorCustomizerTest.java
```diff
@@ -78,7 +78,7 @@ private static void assertCustomizer(
     }
 
     await()
-        .atMost(Duration.ofSeconds(1))
+        .atMost(Duration.ofSeconds(60))
```
laurit, line 81: In instrumentation repo `waitAndAssertTraces` uses awaitility under the hood with default timeout of 10s. While 60s does the job I probably would have started with a lower timeout, like the default 10 or maybe 20.

laurit (approved):

baggage-processor/src/test/java/io/opentelemetry/contrib/baggage/processor/BaggageProcessorCustomizerTest.java
zeitlinger, line 81:
```suggestion
        .atMost(Duration.ofSeconds(10))
```

zeitlinger: done

Issue #1605, opened 2024-12-14
HBaseIntegrationTest is sporadically failing


trask:
https://github.com/open-telemetry/opentelemetry-java-contrib/actions/runs/12323227431/job/34402775016?pr=1603

(the reason this didn't show up before is because we used to auto-retry tests)

```
> Task :jmx-scraper:integrationTest

HBaseIntegrationTest > endToEndTest(Path) FAILED
    org.opentest4j.AssertionFailedError: [no data point matched attribute set '[{state=compaction}, {region_server}]' for metric 'hbase.region_server.queue.length'] 
    expected: true
     but was: false
        at app//io.opentelemetry.contrib.jmxscraper.assertions.MetricAssert.lambda$hasDataPointsWithAttributes$2(MetricAssert.java:262)
        at app//io.opentelemetry.contrib.jmxscraper.assertions.MetricAssert.checkDataPoints(MetricAssert.java:188)
        at app//io.opentelemetry.contrib.jmxscraper.assertions.MetricAssert.hasDataPointsWithAttributes(MetricAssert.java:231)
        at app//io.opentelemetry.contrib.jmxscraper.target_systems.HBaseIntegrationTest.lambda$createMetricsVerifier$9(HBaseIntegrationTest.java:126)
```

SylvainJuge:
For reference the flaky tests retry has been removed in https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1567 because it relied on develocity.

trask:
we can bring back test retry, but we no longer have access to see flaky tests across builds, and so the concern is that we won't see them at all

in the instrumentation repo we mitigated this a bit by only removing retries on the daily builds, so the flaky tests will still bubble up to us on the daily builds

if flaky tests become enough of a problem in this repo we can do the same

SylvainJuge:
As discussed today with @robsunday today, we think most of those test failures tend to be related to some JMX Beans not being published/ready, and it will need some case-by-case investigation, so for now we think increasing the timeout would probably help mitigate that in the short term, for example with #1612.

PR #1610, opened 2024-12-16
Log the hbase attributes for one failing data point

breedx-splk:
Relates to #1605

Because we're having intermittent test failures, I thought it might be nice to see the attributes for the given data point(s) that are failing. The output looks something like this:

```
Dec 16, 2024 3:25:34 PM io.opentelemetry.contrib.jmxscraper.assertions.MetricAssert lambda$logAttributes$4
INFO: hbase.region_server.queue.length -> attrs = [{region_server=> string_value: "1fdbc6238e8a"},{state=> string_value: "flush"}]
Dec 16, 2024 3:25:34 PM io.opentelemetry.contrib.jmxscraper.assertions.MetricAssert lambda$logAttributes$4
INFO: hbase.region_server.queue.length -> attrs = [{region_server=> string_value: "1fdbc6238e8a"},{state=> string_value: "compaction"}]
```

and the idea then is that you can turn this on for a while until we can figure out what aspect of the test is failing. Also open to other ideas on how to narrow it down.

### DIFF
--- a/jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/HBaseIntegrationTest.java
+++ b/jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/HBaseIntegrationTest.java
@@ -120,6 +120,7 @@ public class HBaseIntegrationTest extends TargetSystemIntegrationTest {
             "hbase.region_server.queue.length",
             metric ->
                 metric
+                    .logAttributes()
                     .isUpDownCounter()
                     .hasDescription("The number of RPC handlers actively servicing requests.")
                     .hasUnit("{handler}")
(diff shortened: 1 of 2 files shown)


robsunday:
This is needed functionality and it is good you brought this up, however I added on [my PR](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1615/files#diff-e9edd24edea3d2e097beec851f6e3a39c57ffad1119e10280cdfffccbea10004R252-R253) dumping metric data points together with attributes in case of attribute matching failure. This will provide us with complete unprocessed input data for failing metric.
If my changes will produce too verbose output then we may switch to your solution, for now I'd suggest put this PR on hold if you don't mind.

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/HBaseIntegrationTest.java
```diff
@@ -120,6 +120,7 @@ protected MetricsVerifier createMetricsVerifier() {
             "hbase.region_server.queue.length",
             metric ->
                 metric
+                    .logAttributes()
```
robsunday, line 123: This is something that should be done automatically.  Maybe just call it inside hasDataPointsWithAttributes() ?

breedx-splk: Nah, this design assumed that you'd only want to turn this on in specific cases for certain metrics or whatever. Doing this automatically for every metric assertion would be too verbose.

breedx-splk:
Superseded by #1615

PR #1612, opened 2024-12-17
Improve containers startup and increase max duration of test

robsunday:
**Description:**

This is not the full fix for flaky integration tests but it should improve their stability. More PRs will come.
Part of fix for https://github.com/open-telemetry/opentelemetry-java-contrib/issues/1605

Because some of the metrics may require some more time to be reported. Maximum metrics assertion timeout has been increased to support this scenario. This timeout increase does not impact successful tests duration because such tests end as soon as all assertions are fulfilled.


jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/TargetSystemIntegrationTest.java
```diff
@@ -167,7 +167,7 @@ protected final void waitAndAssertMetrics(Consumer<Metric>... assertions) {
   protected void verifyMetrics() {
     MetricsVerifier metricsVerifier = createMetricsVerifier();
     await()
-        .atMost(Duration.ofSeconds(30))
+        .atMost(Duration.ofSeconds(60))
```
robsunday, line 170: Increase maximum assertions timeout to give tested target more time to stabilize and generate all the metrics

trask (approved):

PR #1615, opened 2024-12-18
Fix HBase test flakiness

robsunday:
**Description:**

Part of fix for https://github.com/open-telemetry/opentelemetry-java-contrib/issues/1605

HBase is doing a lot of things after all the ports are open and before it completes its initialization. Because of this the test may be flaky.

**Scope**

- Changed condition for detecting container start from checking ports to checking log. This may require us to update regexp when testing HBase in different version, but it improved stability a lot.
- Fine tuning some assertion messages, including dumping all data points with attributes when matching attributes fails
- Cleanup of TargetSystemIntegrationTest class

### DIFF
--- a/jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/TargetSystemIntegrationTest.java
+++ b/jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/TargetSystemIntegrationTest.java
@@ -111,13 +111,13 @@ public abstract class TargetSystemIntegrationTest {
             .withLogConsumer(new Slf4jLogConsumer(targetSystemLogger))
             .withNetwork(network)
             .withNetworkAliases(TARGET_SYSTEM_NETWORK_ALIAS);
-    target.start();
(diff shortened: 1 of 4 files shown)


jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/TargetSystemIntegrationTest.java
```diff
@@ -125,45 +123,6 @@ void endToEndTest(@TempDir Path tmpDir) {
     verifyMetrics();
   }
 
-  // TODO: This implementation is DEPRECATED and will be removed once all integration tests are
-  // migrated to MetricsVerifier
-  protected void waitAndAssertMetrics(Iterable<Consumer<Metric>> assertions) {
```
SylvainJuge, line 130: [for reviewer] this is not related to this PR, but is not used anymore.

SylvainJuge (approved):

laurit (approved):

breedx-splk (approved):

PR #1619, opened 2024-12-20
Bring back test retry for JMX Scraper tests

trask:
we can look into these later, but I'd like to bring back test retry on these for now to avoid needing to re-run CI often


breedx-splk (approved):

breedx-splk:
Opened tracking issue to follow up.

PR #1621, opened 2024-12-20
Update documentation around detecting k8s resource attributes

### LINKED ISSUE #1609: gcp-resources documentation doesn't align with the GCP Resource Detector
The [GCP Resource Detector](https://github.com/open-telemetry/opentelemetry-java-contrib/tree/main/gcp-resources#gcp-resource-detectors-for-opentelemetry) mentions - and contains docs for - k8s Resource attributes. However, none of them aside from `k8s.cluser.name` appear to be instrumented:

https://github.com/open-telemetry/opentelemetry-java-contrib/blob/d06d470d0651b4534b8ce49ef53ee1db85265daf/gcp-resources/src/main/java/io/opentelemetry/contrib/gcp/resource/IncubatingAttributes.java#L57-L58

It looks like they exist in the [GCP Resource Translator](https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/blob/a08a050905f52dd566ccb0a3ddaeba632fd83ffc/shared/resourcemapping/src/main/java/com/google/cloud/opentelemetry/resource/ResourceTranslator.java#L96-L105), so they're only missing in the Otel side. I noticed they're in the `IncubatingAttributes` so perhaps they're still in development?

psx95:
**Description:**

Updates documentation to clarify the resource attributes being detected for Kubernetes.

**Existing Issue(s):**

fixes #1609 

**Documentation/Testing:**

The documentation in README is now consistent with the spec and the [GCP Resource Detection Sample](https://github.com/open-telemetry/opentelemetry-java-examples/pull/561) has been updated to confirm the new documented behavior.

**PS: Removed redundant dependency declaration in the build.gradle.kts file.**


punya (approved):

gcp-resources/README.md
```diff
@@ -14,41 +14,37 @@ The following OpenTelemetry semantic conventions will be detected:
...
-| k8s.namespace.name      |                                              | downward API                                           |                                          |                                                |                                                |
-| k8s.container.name      |                                              | hardcoded (manual)                                     |                                          |                                                |                                                |
+| k8s.pod.name            |                                              | environment variable                                   |                                          |                                                |                                                |
+| k8s.namespace.name      |                                              | environment variable                                   |                                          |                                                |                                                |
```
aabmass, line 18: Aren't these still using the downward API ?

psx95: @dashpole could you shed some light here ? 

Going through the [downward API documentation](https://kubernetes.io/docs/concepts/workloads/pods/downward-api/) did not clear up what would be apt to put in the table. 

I think downward API explicitly means [downward API volume](https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/).

dashpole: These should use the downward API to set an environment variable (e.g. following https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/). Then the user needs to set OTEL_RESOURCE_ATTRIBUTES to use the environment variables produced by the downward API.

dashpole: (k8s.container.name is still a hardcoded environment variable), rather than one set by the downward API.

psx95: Thanks for the explanation @dashpole. It makes sense.

Regarding the k8s.container.name, IIUC, the user would still have to pass the value of the environment variable manually to the OTEL_RESOURCE_ATTRIBUTES for proper resource detection, correct ?

Just setting the an environment variable named "CONTAINER_NAME" would not be enough.

dashpole: yes, that is correct

trask (approved):

PR #1625, opened 2024-12-24
Fix flaky MicrometerDoubleCounterTest

### LINKED ISSUE #1624: MicrometerDoubleCounterTest is sporadically failing
https://github.com/open-telemetry/opentelemetry-java-contrib/actions/runs/12473971011/job/34815301925?pr=1622

(the reason this didn't show up before is because we used to auto-retry tests)

```
MicrometerDoubleCounterTest > observableWithAttributes() FAILED
    org.opentest4j.AssertionFailedError: 
    expected: 18.015547904759003
     but was: 20.0
        at app//io.opentelemetry.contrib.metrics.micrometer.internal.instruments.MicrometerDoubleCounterTest.observableWithAttributes(MicrometerDoubleCounterTest.java:333)
```

trask:
Resolves #1624

### DIFF
--- a/micrometer-meter-provider/src/test/java/io/opentelemetry/contrib/metrics/micrometer/internal/instruments/MicrometerDoubleCounterTest.java
+++ b/micrometer-meter-provider/src/test/java/io/opentelemetry/contrib/metrics/micrometer/internal/instruments/MicrometerDoubleCounterTest.java
@@ -324,7 +324,7 @@ public class MicrometerDoubleCounterTest {
     callbackRegistrar.run();
     assertThat(counter.count()).isEqualTo(20.0);
 
-    double expectedCount = 10.0;
+    double expectedCount = 20.0;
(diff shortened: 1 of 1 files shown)


micrometer-meter-provider/src/test/java/io/opentelemetry/contrib/metrics/micrometer/internal/instruments/MicrometerDoubleCounterTest.java
```diff
@@ -324,7 +324,7 @@ void observableWithAttributes() {
     callbackRegistrar.run();
     assertThat(counter.count()).isEqualTo(20.0);
 
-    double expectedCount = 10.0;
+    double expectedCount = 20.0;
```
trask, line 327: value needs to be at least 20 (as above) since the monotonic value can't decrease

other tests in this file were correctly using `20`

laurit (approved):

Issue #1632, opened 2025-01-02
BaggageSpanProcessor should use ExtendedSpanProcessor.onEnding()


jameyg42:
### Component(s)

processors

### Is your feature request related to a problem? Please describe.

Because `BaggageSpanProcessor` runs `onStart()`, newly added included Baggage keys need to be manually added to the current Span.

### Describe the solution you'd like

`BaggageSpanProcessor` should be modified to implement `ExtendedSpanProcessor` and included baggage keys should be added `onEnding()`.

jkwatson:
@zeitlinger @MikeGoldsmith any thoughts?

zeitlinger:
Yes, we should do that

PR #1633, opened 2025-01-06
use onEnding to add baggage keys that are added while the span is active

zeitlinger:
Fixes https://github.com/open-telemetry/opentelemetry-java-contrib/issues/1632

### DIFF
--- a/baggage-processor/src/main/java/io/opentelemetry/contrib/baggage/processor/BaggageSpanProcessor.java
+++ b/baggage-processor/src/main/java/io/opentelemetry/contrib/baggage/processor/BaggageSpanProcessor.java
@@ -9,14 +9,14 @@ import io.opentelemetry.api.baggage.Baggage;
 import io.opentelemetry.context.Context;
 import io.opentelemetry.sdk.trace.ReadWriteSpan;
 import io.opentelemetry.sdk.trace.ReadableSpan;
-import io.opentelemetry.sdk.trace.SpanProcessor;
+import io.opentelemetry.sdk.trace.internal.ExtendedSpanProcessor;
 import java.util.function.Predicate;
 
 /**
  * This span processor copies attributes stored in {@link Baggage} into each newly created {@link
  * io.opentelemetry.api.trace.Span}.
  */
-public class BaggageSpanProcessor implements SpanProcessor {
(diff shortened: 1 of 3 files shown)


trask:
I'm not sure there is any difference? (sort of related to https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/12952)

zeitlinger:
> I'm not sure there is any difference? (sort of related to [open-telemetry/opentelemetry-java-instrumentation#12952](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/12952))

can you elaborate?

trask:
> can you elaborate?

sorry, I'm wondering why this change is needed? can you give a concrete example of when this implementation would have different behavior from existing implementation?

zeitlinger:
> sorry, I'm wondering why this change is needed? can you give a concrete example of when this implementation would have different behavior from existing implementation?

When a baggage item is added after span start:

```java
  Span span = sdk.getTracer("test").spanBuilder("test").startSpan();

      try (Scope unused = span.makeCurrent()) {
        try (Scope scope = Baggage.current().toBuilder().put("key", "value").build()
            .makeCurrent()) {
          span.end();
        }
      }
```

trask:
yeah, but that doesn't work when we follow https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/12952

zeitlinger:
> yeah, but that doesn't work when we follow [open-telemetry/opentelemetry-java-instrumentation#12952](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/12952)

OK - so you're saying that this feature should not be implemented because it would not work for users that use the agent - and that would be surprising?

trask:
> so you're saying that this feature should not be implemented because it would not work for users that use the agent

yeah, I'm not convinced we should make this change since I think "best practice" is to end span after closing scope

also I realized this change will cause the processor to break for people who are passing context manually instead of using thread local

zeitlinger:
> > so you're saying that this feature should not be implemented because it would not work for users that use the agent
> 
> yeah, I'm not convinced we should make this change since I think "best practice" is to end span after closing scope
> 
> also I realized this change will cause the processor to break for people who are passing context manually instead of using thread local

I think that's a good argument - I'll add it to the issue

PR #1647, opened 2025-01-16
GCP Auth extension - allow users to specify quota project ID

### LINKED ISSUE #1645: GCP Auth Extension - allow users to explicitly specify quota project ID
### Component(s)

gcp-auth-extension

The GCP Auth extension currently relies only on Application Default Credentials to infer the Quota project ID to be used for the export. This means users cannot explicitly specify a quota project ID that should be used for the export.

The feature request is for following changes:
 - Expose an optional argument for the users to explicitly specify quota project ID to be used.
    - If the argument is not specified, it is inferred from the Application Default Credentials.
    - If the resolved value of the quota project ID is empty or null, the quota user project header: `X-Goog-User-Project` will not be added. 
 - Make `GOOGLE_CLOUD_PROJECT` required option.

_No response_

### Additional context

_No response_

psx95:
**Description:**

This PR provides an optional argument that lets the users specify a quota project ID.

**Existing Issue(s):**

Fixes #1645 

**Testing:**
 - Unit tests verifying `google.cloud.project` as required field added.
 - Manually tested the extension JAR with test applications.

**Documentation:**

Updates to the README reflecting the user facing changes added.

**Outstanding items:**

Add parameterized unit tests to ensure behavior around null or empty values for the project IDs.

### DIFF
--- a/gcp-auth-extension/README.md
+++ b/gcp-auth-extension/README.md
@@ -39,7 +39,7 @@ Here is a list of configurable options for the extension:
   - This is a required option, the agent configuration will fail if this option is not set.
 - `GOOGLE_CLOUD_QUOTA_PROJECT`: Environment variable that represents the Google Cloud Quota Project ID which will be charged for the GCP API usage. To learn more about a *quota project*, see [here](https://cloud.google.com/docs/quotas/quota-project).
   - Can also be configured using `google.cloud.quota.project` system property.
-  - If this option is not configured, the extension would infer GCP Quota Project ID from the application default credentials. For more information on application default credentials, see [here](https://cloud.google.com/docs/authentication/application-default-credentials).
+  - If this option is not configured, the extension will use the Quota Project ID found in the Application Default Credentials (ADC), if available. For more information on application default credentials, see [here](https://cloud.google.com/docs/authentication/application-default-credentials).
 
 ## Usage
 
--- a/gcp-auth-extension/README.md
+++ b/gcp-auth-extension/README.md
@@ -36,7 +36,7 @@ Here is a list of configurable options for the extension:
 
 - `GOOGLE_CLOUD_PROJECT`: Environment variable that represents the Google Cloud Project ID to which the telemetry needs to be exported.
   - Can also be configured using `google.cloud.project` system property.
-  - This is a required option, the telemetry export to GCP will fail.
+  - This is a required option, the agent configuration will fail if this option is not set.
 - `GOOGLE_CLOUD_QUOTA_PROJECT`: Environment variable that represents the Google Cloud Quota Project ID which will be charged for the GCP API usage. To learn more about a *quota project*, see [here](https://cloud.google.com/docs/quotas/quota-project).
   - Can also be configured using `google.cloud.quota.project` system property.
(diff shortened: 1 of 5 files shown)


gcp-auth-extension/src/test/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProviderTest.java
aabmass, line 1: IIUC there was a bug before that we would send an empty (null?) quota project header if it wasn't configured which made requests fail.

Is there a test that confirms we don't add the quota project header if it is not configured?

gcp-auth-extension/README.md
```diff
@@ -36,7 +36,10 @@ Here is a list of configurable options for the extension:
 
 - `GOOGLE_CLOUD_PROJECT`: Environment variable that represents the Google Cloud Project ID to which the telemetry needs to be exported.
   - Can also be configured using `google.cloud.project` system property.
-  - If this option is not configured, the extension would infer GCP Project ID from the application default credentials. For more information on application default credentials, see [here](https://cloud.google.com/docs/authentication/application-default-credentials).
+  - This is a required option, the agent configuration will fail if this option is not set.
+- `GOOGLE_CLOUD_QUOTA_PROJECT`: Environment variable that represents the Google Cloud Quota Project ID which will be charged for the GCP API usage. To learn more about a *quota project*, see [here](https://cloud.google.com/docs/quotas/quota-project).
+  - Can also be configured using `google.cloud.quota.project` system property.
+  - If this option is not configured, the extension would infer GCP Quota Project ID from the application default credentials. For more information on application default credentials, see [here](https://cloud.google.com/docs/authentication/application-default-credentials).
```
aabmass, line 42: nit the extension doesn't infer the quota project id right? It just omits it to get the default behavior.

psx95: If not configured (set by the user), the quota project ID from the application default credentials is used as fallback.

If the inferred/resolved value of the quota project ID is null or empty, the header is omitted.

aabmass: OK I think my confusion is "infer" to me implies that the extension will always generate some kind quota project from ADC (not let it be null). I assumed it would try to set with ADC's project ID and always send

I think something like "it will use quota project found with ADC if available" and mention that this config is not required.

aabmass (approved):

gcp-auth-extension/README.md
psx95: I will update the wording to be more clear.

gcp-auth-extension/src/test/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProviderTest.java
psx95: I have added tests to ensure the expected behavior in regards to resolving the Quota project ID.

dashpole (approved):

trask (approved):

PR #1651, opened 2025-01-17
jmx scraper with sdk autoconfig

SylvainJuge:
Refactor jmx scraper configuration to use autoconfiguration.

This provides the following benefits:
- allow using environment variables in addition to java system properties
- override the default configuration using SDK API : config provider and config customizer

Now only the the `otel.metric.export.interval` SDK configuration should be used to define the metric sampling and export interval, compatibility is provided with existing `otel.jmx.interval.milliseconds` when needed.

When `otel.metric.export.interval` is not set, the default of 1 minute is applied which remains consistent with the SDK implementation and specification.

The ability to provide configuration through standard input, external properties and the JVM system properties have been preserved.

### To be discussed
- [x] Should we align the default exporter to be `otlp` to align with the SDK default value, or should we keep the `logging` as we have currently when it's not set. Setting `logging` here by default mostly benefits onboarding as users can see what is being captured, but requires to explicitly include `otlp` which is set by default for the SDK autoconfiguration.
  - Answer: let's align
  
### Follow-up tasks

In instrumentation (JMX insights), the `otel.jmx.config` option provides the path to a list of custom YAML metrics definitions, we currently use `otel.jmx.custom.scraping.config` which does not allow to use a list. Dealing with this was left outside of this PR and will be handled by this #1662.


jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/JmxScraperContainer.java
```diff
@@ -24,7 +24,6 @@ public class JmxScraperContainer extends GenericContainer<JmxScraperContainer> {
   private final String endpoint;
   private final Set<String> targetSystems;
   private String serviceUrl;
-  private int intervalMillis;
```
SylvainJuge, line 27: [for reviewer] this is never changed for tests, so we can use the minimal hard-coded value of 1s.

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/JmxScraper.java
```diff
@@ -43,58 +46,77 @@ public class JmxScraper {
...
+    for (Map.Entry<Object, Object> entry : properties.entrySet()) {
+      String key = entry.getKey().toString();
+      String value = entry.getValue().toString();
+      if (key.startsWith("javax.net.ssl.keyStore") || key.startsWith("javax.net.ssl.trustStore")) {
+        if (System.getProperty(key) == null) {
+          System.setProperty(key, value);
+        }
+      }
+    }
+  }
```
SylvainJuge, line 112: [for reviewer] propagation of system properties is only used to set the java keystore/truststore options from the program arguments/standard input to the global JVM settings, so it was simpler to move it here. In practice this will likely be tested when we add tests with custom keystore/truststore by setting those configuration options from the standard input or the properties file.

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/config/ConfigurationException.java
```diff
@@ -1,18 +0,0 @@
-/*
- * Copyright The OpenTelemetry Authors
- * SPDX-License-Identifier: Apache-2.0
- */
-
-package io.opentelemetry.contrib.jmxscraper.config;
-
-public class ConfigurationException extends Exception {
```
SylvainJuge, line 8: [for reviewer] there is already a similarly named runtime exception in the SDK autoconfiguration, so reusing it makes more sense and avoid ambiguity.

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/config/JmxScraperConfig.java
```diff
@@ -116,133 +111,50 @@ public boolean isRegistrySsl() {
...
-      properties.setProperty(key, defaultValue);
+  public static JmxScraperConfig fromConfig(ConfigProperties config) {
+    JmxScraperConfig scraperConfig = new JmxScraperConfig();
+
+    Duration exportInterval = config.getDuration(METRIC_EXPORT_INTERVAL);
+    if (exportInterval == null || exportInterval.isNegative() || exportInterval.isZero()) {
+      // if not explicitly set, use default value of 1 minute as defined in specification
+      scraperConfig.samplingInterval = Duration.ofMinutes(1);
+    } else {
+      scraperConfig.samplingInterval = exportInterval;
```
SylvainJuge, line 127: [for reviewer] we have to do this because we don't have access to the default value that will be used by the SDK when an explicit value is not present.

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/config/PropertiesCustomizer.java
```diff
@@ -0,0 +1,64 @@
...
+
+  private static final Logger logger = Logger.getLogger(PropertiesCustomizer.class.getName());
+
+  private static final String METRICS_EXPORTER = "otel.metrics.exporter";
+
+  @Nullable private JmxScraperConfig scraperConfig;
+
+  @Override
+  public Map<String, String> apply(ConfigProperties config) {
+    Map<String, String> result = new HashMap<>();
```
SylvainJuge, line 29: [for reviewer] the returned map contains the configuration items that will be explicitly set/overriden so that they can be later read by the SDK initialization. In addition to that we use this to build an instance of `JmxScraperConfig` from the provided `ConfigProperties` that will later be used to configure the jmx scraper.

breedx-splk:
Regarding the question of default logger -- I think we should continue to be consistent with the sdk and default to exporting otlp to localhost. 👍🏻  If a user needs to see details/logging they should be able to opt-into that I think.

SylvainJuge:
> Regarding the question of default logger -- I think we should continue to be consistent with the sdk and default to exporting otlp to localhost. 👍🏻 If a user needs to see details/logging they should be able to opt-into that I think.

I agree here, staying consistent with the SDK defaults would provide a more consistent experience.

Also:
- using "logging" mostly benefits the onboarding/setup phase, which is relatively short compared to the time it's being used with a real exporter, so we should make the most common use-case the easiest to use. Logging here is mostly relevant to investigate/debug.
- this makes the jmx scraper "do nothing by default", which makes it appear "broken by default unless you configure it", in particular when the logs/standard outpout is not directly visible, it would be easy to not notice that nothing is being sent to OTLP.
- if it was "otlp,logging" by default it would have at least do something useful by default, but then having to explicitly disable logging for production use is an extra burden.

Thus I'll align with the SDK unless we have a clear use-case that justifies it.

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/config/PropertiesSupplier.java
```diff
@@ -0,0 +1,28 @@
...
+  private final Properties properties;
+
+  public PropertiesSupplier(Properties properties) {
+    this.properties = properties;
+  }
+
+  @Override
+  public Map<String, String> get() {
+    Map<String, String> map = new HashMap<>();
+    properties.forEach((k, v) -> map.put((String) k, (String) v));
```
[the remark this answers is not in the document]
breedx-splk: `putAll()` even! 😁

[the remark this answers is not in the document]
SylvainJuge: Here it's mostly because `Properties` is a `Map<Object,Object>` and not a `Map<String,String>`, I don't think we can cast it directly.

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/JmxScraper.java
```diff
@@ -116,23 +138,24 @@ static Properties parseArgs(List<String> args)
...
 
-  private static Properties loadPropertiesFromStdin() throws ConfigurationException {
+  private static Properties loadPropertiesFromStdin() throws ArgumentsParsingException {
     Properties properties = new Properties();
     try (InputStream is = new DataInputStream(System.in)) {
       properties.load(is);
       return properties;
     } catch (IOException e) {
-      throw new ConfigurationException("Failed to read config properties from stdin", e);
+      throw new ArgumentsParsingException("Failed to read config properties from stdin", e);
```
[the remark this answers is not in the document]
SylvainJuge: as discussed today, a dedicated exception has been added in 646e156cb49d5114624f95578c439706aa2e2a2f, removing the `Parsing` from the name allows to indicate there is a problem with the provided parameters or reading configuration from them.

jmx-scraper/README.md
```diff
@@ -21,20 +21,70 @@ Minimal configuration required
...
+## Configuration reference
 
-### Extra libraries in classpath
+| config option                     | description                                                                                  |
+|-----------------------------------|----------------------------------------------------------------------------------------------|
+| `otel.jmx.service.url`            | mandatory JMX URL to connect to the remote JVM                                               |
+| `otel.jmx.target.system`          | comma-separated list of systems to monitor, mandatory unless a custom configuration is used  |
+| `otel.jmx.custom.scraping.config` | path to a custom YAML metrics definition, mandatory when `otel.jmx.target.system` is not set |
+| `otel.jmx.username`               | user name for JMX connection                                                                 |
+| `otel.jmx.password`               | password for JMX connection                                                                  |
```
breedx-splk, line 45: Doesn't have to be in this PR of course, but it might be nice to have an indication of which configs are required/optional and/or what the default values are.

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/InvalidArgumentException.java
```diff
@@ -0,0 +1,23 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.contrib.jmxscraper;
+
+/**
+ * Exception indicating something is wrong with the provided arguments or reading the configuration
+ * from them
+ */
+public class InvalidArgumentException extends Exception {
```
breedx-splk, line 12: Is this really providing any benefit over the built-in jdk `IllegalArgumentException`?

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/config/JmxScraperConfig.java
```diff
@@ -116,133 +114,50 @@ public boolean isRegistrySsl() {
...
-    }
-
-    if (config.intervalMilliseconds < 0) {
-      throw new ConfigurationException(INTERVAL_MILLISECONDS + " must be positive.");
+          "at least one of '" + JMX_TARGET_SYSTEM + "' or '" + JMX_CUSTOM_CONFIG + "' must be set");
     }
+    targetSystem.forEach(
+        s -> {
+          if (!AVAILABLE_TARGET_SYSTEMS.contains(s)) {
+            throw new ConfigurationException("unsupported target system: '" + s + "'");
```
breedx-splk, line 149: Doesn't this get hidden because it's buried within a closure within `forEach`?

Could consider switching to `if(!targetSystem.stream().allMatch(AVAILABLE_TARGET_SYSTEMS.contains(s)) ...` or similar? Maybe a filter if the detailed logging of which system is necessary.

breedx-splk (approved):
Had a few minor comments but LGTM! Stoked to have this autoconfig power.

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/InvalidArgumentException.java
SylvainJuge: The main benefit is that it's a checked exception and that it forces the caller to take care of it. However, given the very limited scope this has, we could definitely replace it with the `IllegalArgumentException` which is a runtime exception if you think that makes things less confusing.

jmx-scraper/README.md
SylvainJuge: I tried to do that with dedicated column but this wasn't looking great because
- there is only a single mandatory option for the remote URL
- other arguments are mandatory depending on the context (authentication enabled on remote JVM, if there is already a target system set or a custom yaml file used).

This is why I decided to document this mostly in the config option descriptions, I've clarified that for jmx authentication in d237a56

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/config/JmxScraperConfig.java
SylvainJuge: The exception does not get hidden here, this case is covered by the `JmxScraperConfigTest#shouldFailValidation_invalidTargetSystem` test case.

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/InvalidArgumentException.java
robsunday: Yeah, I think IllegalArgumentException was designed to be thrown as result of method parameters validation, not the application cmd arguments.
So to avoid confusion maybe let's call it differently, like "InvalidOptionException". Any other ideas?

PR #1659, opened 2025-01-20
Add test to required PR checks

trask:
I was surprised to find a PR that allowed me to merge even though a (sporadic) test was failing


jaydeluca (approved):

breedx-splk (approved):
Woah, how were we getting away with this not included before? Was it a temporary thing or just a long-term oversight? Weird.

PR #1661, opened 2025-01-20
Show stdout in tests

trask:
Same as other Java repos. Makes debugging easier.


buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
```diff
@@ -67,9 +67,7 @@ tasks {
 
     testLogging {
       exceptionFormat = TestExceptionFormat.FULL
-      showExceptions = true
-      showCauses = true
-      showStackTraces = true
```
trask, line 72: removed these because they all default to `true` (and to match Java instrumentation repo)

jaydeluca (approved):

breedx-splk (approved):

PR #1670, opened 2025-01-24
JMX Scraper: Kafka server, producer and consumer YAMLs and integration tests added

robsunday:
Three target systems connected to Kafka have been added:

1. Server
2. Producer
3. Consumer

To fully test Kafka metrics there is a need to spin multiple containers, so these three components can correctly interact and metrics can be correctly generated.
Base integration test class `TargetSystemIntegrationTest` was modified to support initialization of multiple containers.

There are still some TODOs left that will be fixed in subsequent PRs

### DIFF
--- a/jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/TargetSystemIntegrationTest.java
+++ b/jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/TargetSystemIntegrationTest.java
@@ -22,6 +22,7 @@ import java.nio.file.Path;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.ExecutionException;
@@ -56,6 +57,7 @@ public abstract class TargetSystemIntegrationTest {
 
   private static Network network;
   private static OtlpGrpcServer otlpServer;
+  private Collection<GenericContainer<?>> prerequisiteContainers;
   private GenericContainer<?> target;
   private JmxScraperContainer scraper;
 
@@ -86,12 +88,23 @@ public abstract class TargetSystemIntegrationTest {
 
   @AfterEach
   void afterEach() {
-    if (target != null && target.isRunning()) {
-      target.stop();
-    }
     if (scraper != null && scraper.isRunning()) {
       scraper.stop();
     }
+
+    if (target != null && target.isRunning()) {
+      target.stop();
+    }
+
+    if (prerequisiteContainers != null) {
+      prerequisiteContainers.forEach(
+          container -> {
+            if (container.isRunning()) {
+              container.stop();
+            }
+          });
+    }
+
     if (otlpServer != null) {
       otlpServer.reset();
     }
@@ -103,14 +116,31 @@ public abstract class TargetSystemIntegrationTest {
 
   @Test
   void endToEndTest(@TempDir Path tmpDir) {
+    startContainers(tmpDir);
+    verifyMetrics();
+  }
+
+  protected void startContainers(Path tmpDir) {
+    prerequisiteContainers = createPrerequisiteContainers();
 
     target =
         createTargetContainer(JMX_PORT)
             .withLogConsumer(new Slf4jLogConsumer(targetSystemLogger))
             .withNetwork(network)
             .withNetworkAliases(TARGET_SYSTEM_NETWORK_ALIAS);
+
+    // If there are any containers that must be started before target then initialize them.
+    // Then make target depending on them, so it is started after dependencies
+    for (GenericContainer<?> container : prerequisiteContainers) {
+      container.withNetwork(network);
+      target.dependsOn(container);
+    }
+
+    // Target container must be running before scraper container is customized.
+    // It is necessary to allow interactions with the container, like file copying etc.
     target.start();
 
+    // Create and initialize scraper container
     scraper =
         new JmxScraperContainer(otlpEndpoint, scraperBaseImage())
             .withLogConsumer(new Slf4jLogConsumer(jmxScraperLogger))
@@ -119,14 +149,13 @@ public abstract class TargetSystemIntegrationTest {
 
     scraper = customizeScraperContainer(scraper, target, tmpDir);
     scraper.start();
-
-    verifyMetrics();
   }
 
   protected void verifyMetrics() {
     MetricsVerifier metricsVerifier = createMetricsVerifier();
     await()
         .atMost(Duration.ofSeconds(60))
+        .pollInterval(Duration.ofSeconds(1))
         .untilAsserted(
             () -> {
               List<ExportMetricsServiceRequest> receivedMetrics = otlpServer.getMetrics();
@@ -156,6 +185,10 @@ public abstract class TargetSystemIntegrationTest {
   protected JmxScraperContainer customizeScraperContainer(
       JmxScraperContainer scraper, GenericContainer<?> target, Path tempDir) {
     return scraper;
+  }
+
+  protected Collection<GenericContainer<?>> createPrerequisiteContainers() {
+    return Collections.emptyList();
   }
 
   private static class OtlpGrpcServer extends ServerExtension {
--- a/jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/kafka/KafkaConsumerIntegrationTest.java
+++ b/jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/kafka/KafkaConsumerIntegrationTest.java
@@ -1,0 +1,146 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.contrib.jmxscraper.target_systems.kafka;
+
+import static io.opentelemetry.contrib.jmxscraper.assertions.DataPointAttributes.attribute;
+import static io.opentelemetry.contrib.jmxscraper.assertions.DataPointAttributes.attributeGroup;
+import static io.opentelemetry.contrib.jmxscraper.assertions.DataPointAttributes.attributeWithAnyValue;
+import static io.opentelemetry.contrib.jmxscraper.target_systems.kafka.KafkaContainerFactory.createKafkaConsumerContainer;
+import static io.opentelemetry.contrib.jmxscraper.target_systems.kafka.KafkaContainerFactory.createKafkaContainer;
+import static io.opentelemetry.contrib.jmxscraper.target_systems.kafka.KafkaContainerFactory.createKafkaProducerContainer;
+import static io.opentelemetry.contrib.jmxscraper.target_systems.kafka.KafkaContainerFactory.createZookeeperContainer;
+
+import io.opentelemetry.contrib.jmxscraper.JmxScraperContainer;
+import io.opentelemetry.contrib.jmxscraper.target_systems.MetricsVerifier;
+import io.opentelemetry.contrib.jmxscraper.target_systems.TargetSystemIntegrationTest;
+import java.nio.file.Path;
+import java.util.Arrays;
+import java.util.Collection;
+import org.slf4j.LoggerFactory;
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.containers.output.Slf4jLogConsumer;
+import org.testcontainers.containers.wait.strategy.Wait;
+
+public class KafkaConsumerIntegrationTest extends TargetSystemIntegrationTest {
+
+  @Override
+  protected Collection<GenericContainer<?>> createPrerequisiteContainers() {
+    GenericContainer<?> zookeeper =
+        createZookeeperContainer()
+            .withLogConsumer(new Slf4jLogConsumer(LoggerFactory.getLogger("zookeeper")))
+            .withNetworkAliases("zookeeper");
+
+    GenericContainer<?> kafka =
+        createKafkaContainer()
+            .withLogConsumer(new Slf4jLogConsumer(LoggerFactory.getLogger("kafka")))
+            .withNetworkAliases("kafka")
+            .dependsOn(zookeeper);
+
+    GenericContainer<?> kafkaProducer =
+        createKafkaProducerContainer()
+            .withLogConsumer(new Slf4jLogConsumer(LoggerFactory.getLogger("kafka-producer")))
+            .withNetworkAliases("kafka-producer")
+            .dependsOn(kafka);
+
+    return Arrays.asList(zookeeper, kafka, kafkaProducer);
+  }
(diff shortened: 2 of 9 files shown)


jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/TargetSystemIntegrationTest.java
```diff
@@ -119,14 +149,13 @@ void endToEndTest(@TempDir Path tmpDir) {
...
     scraper.start();
-
-    verifyMetrics();
   }
 
   protected void verifyMetrics() {
     MetricsVerifier metricsVerifier = createMetricsVerifier();
     await()
         .atMost(Duration.ofSeconds(60))
+        .pollInterval(Duration.ofSeconds(1))
```
robsunday, line 158: [for reviewer] Just to decrease CPU usage a bit if we do not have all metrics available immediately.

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/kafka/KafkaConsumerIntegrationTest.java
```diff
@@ -0,0 +1,146 @@
...
+    return MetricsVerifier.create()
+        .add(
+            "kafka.consumer.fetch-rate",
+            metric ->
+                metric
+                    .hasDescription("The number of fetch requests for all topics per second")
+                    .hasUnit("{request}")
+                    .isGauge()
+                    .hasDataPointsWithOneAttribute(
+                        attributeWithAnyValue("client.id"))) // changed to follow semconv
```
robsunday, line 76: [for reviewer] This is one of the changes to get closer to semconv standards

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/kafka/KafkaConsumerIntegrationTest.java
```diff
@@ -0,0 +1,146 @@
...
+                    .hasUnit("{message}")
+                    .isGauge()
+                    .hasDataPointsWithOneAttribute(attributeWithAnyValue("client.id")))
+        .add(
+            "kafka.consumer.total.bytes-consumed-rate",
+            metric ->
+                metric
+                    .hasDescription(
+                        "The average number of bytes consumed for all topics per second")
+                    .hasUnit("By")
```
robsunday, line 91: [for reviewer] Changed from 'by' to follow semconv

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/kafka/KafkaConsumerIntegrationTest.java
```diff
@@ -0,0 +1,146 @@
...
+                    .isGauge()
+                    .hasDataPointsWithOneAttribute(attributeWithAnyValue("client.id")))
+        .add(
+            "kafka.consumer.total.records-consumed-rate",
+            metric ->
+                metric
+                    .hasDescription(
+                        "The average number of records consumed for all topics per second")
+                    .hasUnit("{record}")
```
robsunday, line 109: [for reviewer] Multiple units in this file have been changed from "1" to annotated form to follow semconv

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/kafka/KafkaContainerFactory.java
```diff
@@ -0,0 +1,66 @@
...
+    return new GenericContainer<>(KAFKA_DOCKER_IMAGE)
+        //        .withCopyFileToContainer(
+        //            MountableFile.forClasspathResource("kafka-producer.sh"),
+        //            "/usr/bin/kafka-producer.sh")
+        //        .withCommand("/usr/bin/kafka-producer.sh")
+        .withCommand(
+            "sh",
+            "-c",
+            "echo 'Sending messages to test-topic-1'; "
+                + "i=1; while true; do echo \"Message $i\"; sleep .25; i=$((i+1)); done | /opt/bitnami/kafka/bin/kafka-console-producer.sh --bootstrap-server "
```
robsunday, line 46: [for reviewer] Topic is automatically created when messages are sent to it - no need to manually create it

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/kafka/KafkaIntegrationTest.java
```diff
@@ -0,0 +1,244 @@
...
+      JmxScraperContainer scraper, GenericContainer<?> target, Path tempDir) {
+    return scraper.withTargetSystem("kafka");
+  }
+
+  @Override
+  protected MetricsVerifier createMetricsVerifier() {
+    AttributeMatcherGroup[] requestTypes = {
+      // attribute values are changed from lowercase to PascalCase
+      // because it is impossible to make them lowercase with YAML config
+      // TODO: What about const value attributes defined in YAML that are lowercase?
```
robsunday, line 51: [for reviewer] We need to decide what to do with attribute names for Kafka.
Currently attributes extracted from bean names are in PascalCase, while the attributes defined by const(...) are lowercase (see YAMLs)
Maybe that is fine?

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/kafka/KafkaProducerIntegrationTest.java
```diff
@@ -0,0 +1,163 @@
...
+    return Arrays.asList(zookeeper, kafka);
+  }
+
+  @Override
+  protected GenericContainer<?> createTargetContainer(int jmxPort) {
+    return createKafkaProducerContainer()
+        .withEnv("JMX_PORT", Integer.toString(jmxPort))
+        .withExposedPorts(jmxPort)
+        .waitingFor(Wait.forListeningPorts(jmxPort));
+    //        .waitingFor(Wait.forLogMessage(".*Sending messages to test-topic-1.*", 1));
```
robsunday, line 50: [for reviewer] Will be cleaned

jmx-scraper/src/main/resources/kafka.yaml
```diff
@@ -0,0 +1,215 @@
...
+      - kafka.server:type=DelayedOperationPurgatory,name=PurgatorySize,delayedOperation=Produce
+      - kafka.server:type=DelayedOperationPurgatory,name=PurgatorySize,delayedOperation=Fetch
+    metricAttribute:
+      type: param(delayedOperation)
+    mapping:
+      Value:
+        metric: kafka.purgatory.size
+        type: updowncounter
+        desc: The number of requests waiting in purgatory
+        unit: "{request}"
```
robsunday, line 121: [for reviewer] Please verify all annotations of units. In the original files they were either in plular form, or were missing (unit: "1" was used instead)

jmx-scraper/src/main/resources/kafka-consumer.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+  - bean: kafka.consumer:client-id=*,type=consumer-fetch-manager-metrics
+    metricAttribute:
+      client.id: param(client-id)
+    prefix: kafka.consumer.
+    type: gauge
+    mapping:
+      fetch-rate:
+        desc: The number of fetch requests for all topics per second
+        unit: '{request}'
```
robsunday, line 13: [for reviewer] Please verify all annotations of units. In the original files they were either in plular form, or were missing (unit: "1" was used instead)

jmx-scraper/src/main/resources/kafka-producer.yaml
```diff
@@ -0,0 +1,46 @@
...
+        unit: ns
+      outgoing-byte-rate:
+        desc: The average number of outgoing bytes sent per second to all servers
+        unit: By
+      request-latency-avg:
+        desc: The average request latency
+        unit: ms
+      request-rate:
+        desc: The average number of requests sent per second
+        unit: '{request}'
```
robsunday, line 21: [for reviewer] Please verify all annotations of units. In the original files they were either in plular form, or were missing (unit: "1" was used instead)

jmx-scraper/src/main/resources/kafka.yaml
```diff
@@ -0,0 +1,215 @@
...
+      Count:
+        metric: kafka.message.count
+        type: counter
+        desc: The number of messages received by the broker
+        unit: "{message}"
+
+  # TODO: optimize kafka.request.count - use refs
+  - bean: kafka.server:type=BrokerTopicMetrics,name=TotalFetchRequestsPerSec
+    metricAttribute:
+      # TODO: Should we convert all constant values to be PascalCase for consistency with values from bean params?
```
SylvainJuge, line 16: I think this is something that we need to evaluate when trying to improve the metrics in the "define and align" phase that will follow, for example those metric attributes are not prefixed with a `kafka.*` namespace which will be another breaking change so we'd better do those all at once in a single step.

For now we should stay as close as we can from the groovy definition, leaving a TODO here will help remember to take care of this later.

jmx-scraper/src/main/resources/kafka-producer.yaml
```diff
@@ -0,0 +1,46 @@
...
+        unit: By
+      compression-rate:
+        desc: The average compression rate of record batches for a topic, defined as the average ratio of the compressed batch size divided by the uncompressed size
+        unit: '{ratio}'
+      record-error-rate:
+        desc: The average per-second number of record sends that resulted in errors for a topic
+      record-retry-rate:
+        desc: The average per-second number of retried record sends for a topic
+      record-send-rate:
+        desc: The average number of records sent per second for a topic
```
SylvainJuge, line 46: [minor] because the unit is different for a few metrics means it's probably simpler to always describe the unit at the same level instead of only overriding when it's different.

jmx-scraper/src/main/resources/kafka-consumer.yaml
```diff
@@ -0,0 +1,44 @@
...
+      client.id: param(client-id)
+      topic: param(topic)
+    prefix: kafka.consumer.
+    unit: By
+    type: gauge
+    mapping:
+      bytes-consumed-rate:
+        desc: The average number of bytes consumed per second
+      fetch-size-avg:
+        desc: The average number of bytes fetched per request
```
SylvainJuge, line 41: [minor] maybe move the unit to the mapping level for making it easier to read.

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/kafka/KafkaConsumerIntegrationTest.java
SylvainJuge: I think this is definitely something that we need to handle in the next phase to merge and align the metrics definitions (leaving a TODO in the yaml seems a good start to not forget about it). If we can keeping it close to the current groovy implementation is probably the safest move.

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/kafka/KafkaIntegrationTest.java
SylvainJuge: This is a case where the YAML implementation does not allow to apply any transformation to the JMX attributes or parameters in the object name, so we can either:
1. get by this inconsistency between the groovy and yaml implementations, but it might have to be properly documented
2. enhance the YAML implementation in instrumentation to implement some simple string modifications to apply this kind of transformation
3. modify the original groovy implementation for the sake of consistency.

I am not really sure there will ever be other cases where we need this, so option 2 would modify the implementation only to deal with this particular corner case. Also, I tend to think that in the future YAML definitions for JMX metrics we should aim for simplicity and "capture the things as they are exposed" rather than try to make them fit a particular shape, so keeping the parameter value as-is would likely be my recommendation in those cases.

For the third option, this would introduce a breaking change in the existing implementation, so I think we should probably avoid it.

As a conclusion I think here we should probably take the option 1 and document it as a known difference with the groovy implementation.

jmx-scraper/src/main/resources/kafka-consumer.yaml
robsunday: That is correct. I wanted to do this in next PR because this one was already quite big, but it is safer to do it now - in this PR, so I'll do it

jmx-scraper/src/main/resources/kafka.yaml
robsunday: I'll get rid of it since in the other place we agreed to rather proceed with unchanged values retrieved from monitored beans

jmx-scraper/src/main/resources/kafka-producer.yaml
robsunday: Yes, in this case I agree with your opinion - it is easier to grasp

SylvainJuge (approved):

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/kafka/KafkaIntegrationTest.java
```diff
@@ -0,0 +1,241 @@
...
+    //                    .hasDescription("Log flush time - 50th percentile")
+    //                    .hasUnit("ms")
+    //                    .isGauge()
+    //                    .hasDataPointsWithoutAttributes())
+    //        .add(
+    //            "kafka.logs.flush.time.99p",
+    //            metric ->
+    //                metric
+    //                    .hasDescription("Log flush time - 99th percentile")
+    //                    .hasUnit("ms")
```
breedx-splk, line 236: A tracking issue might be more visible than these comments. I think you should open an issue.

breedx-splk (approved):
Looks good to me. Wowza that's a lot of metrics. :) Makes one wish that kafka had native otel support ootb... 🤔

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/kafka/KafkaIntegrationTest.java
robsunday: Done: https://github.com/open-telemetry/opentelemetry-java-contrib/issues/1726

robsunday:
> Looks good to me. Wowza that's a lot of metrics. :) Makes one wish that kafka had native otel support ootb... 🤔

Yes, and newer Kafka versions have surely even more.

trask:
@robsunday can you merge main into your PR branch? that will fix the required `gradle-wrapper-validation` check, thanks!

PR #1672, opened 2025-01-24
[disk-buffering] Simplify MetricToDiskExporter creation

breedx-splk:
This makes the MetricToDiskExporter more consistent with the other exporters, and removes the requirement to have an additional mapping for temporality. Instead, we can just delegate to the MetricExporter that we are passed, because `MetricExporter` implements `AggregationTemporalitySelector`.  Furthermore, it eliminates the confusing and incorrect possibility where the to-disk exporter could have returned a temporality different than its delegate. Doh!


LikeTheSalad (approved):
Thank you! 🙌

trask (approved):

PR #1675, opened 2025-01-27
JMX Scraper: YAML file and integration test hadoop

robsunday:
**Scope**

1. YAML file for Hadoop added
2. Metrics definitions updated to be aligned with semconv in JMX Metrics Gatherer and in JMX Scraper
3. Integration test for Hadoop added for JMX Scraper


jmx-scraper/src/integrationTest/resources/hadoop-env.sh
```diff
@@ -0,0 +1,99 @@
...
+# For Kerberos debugging, an extended option set logs more invormation
+# export HADOOP_OPTS="-Djava.net.preferIPv4Stack=true -Dsun.security.krb5.debug=true -Dsun.security.spnego.debug"
+export HADOOP_OPTS="$HADOOP_OPTS -Djava.net.preferIPv4Stack=true"
+
+# Command specific options appended to HADOOP_OPTS when specified
+export HADOOP_NAMENODE_OPTS="-Dhadoop.security.logger=${HADOOP_SECURITY_LOGGER:-INFO,RFAS} -Dhdfs.audit.logger=${HDFS_AUDIT_LOGGER:-INFO,NullAppender} $HADOOP_NAMENODE_OPTS"
+export HADOOP_NAMENODE_OPTS="-Dcom.sun.management.jmxremote $HADOOP_NAMENODE_OPTS"
+export HADOOP_NAMENODE_OPTS="$HADOOP_NAMENODE_OPTS -Dcom.sun.management.jmxremote.authenticate=false"
+export HADOOP_NAMENODE_OPTS="$HADOOP_NAMENODE_OPTS -Dcom.sun.management.jmxremote.ssl=false"
+export HADOOP_NAMENODE_OPTS="$HADOOP_NAMENODE_OPTS -Dcom.sun.management.jmxremote.port=9999 -Dcom.sun.management.jmxremote.rmi.port=9999"
```
SylvainJuge, line 48: [minor] would it be possible to avoid having this script file by just setting the `HADOOP_NAMENODE_OPTS` with appropriate JMX settings directly ?

jmx-scraper/src/main/resources/hadoop.yaml
```diff
@@ -0,0 +1,61 @@
+---
+rules:
+  - bean: Hadoop:service=NameNode,name=FSNamesystem
+    prefix: hadoop.name_node.
+    metricAttribute:
+      node_name: beanattr(tag\.Hostname)
+    mapping:
+      CapacityUsed:
+        metric: capacity.usage
+        type: updowncounter
```
SylvainJuge, line 10: [minor] type is always `updowncounter` we could move it to the bean-level definition

SylvainJuge (approved):

jmx-scraper/src/main/resources/hadoop.yaml
robsunday: Good observation

jmx-scraper/src/integrationTest/resources/hadoop-env.sh
breedx-splk: What about all those other env vars too tho?

jmx-scraper/src/main/resources/hadoop.yaml
breedx-splk: Woah, you can do that? I guess it just applies then to all mappings? The mappings can still override the bean-level settings I assume?

jmx-scraper/src/main/resources/hadoop.yaml
```diff
@@ -0,0 +1,61 @@
...
+      TotalLoad:
+        metric: file.load
+        type: updowncounter
+        unit: "{operation}"
+        desc: The current number of concurrent file accesses.
+      NumLiveDataNodes:
+        metric: &metric data_node.count
+        type: updowncounter
+        unit: &unit "{node}"
+        desc: &desc The number of data nodes reporting to the name node.
```
breedx-splk, line 52: I see these odd looking `&unit` and `&desc` in the hbase yaml as well, and maybe another one, but I don't see mention of what it is used for or the value it provides. Is it a mistake? I couldn't find any mention in the [instrumentation docs](https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation/jmx-metrics/javaagent) either.

jmx-scraper/src/main/resources/hadoop.yaml
SylvainJuge: Yes, you can set defaults at "bean" level then override on a per-attribute basis.

jmx-scraper/src/integrationTest/resources/hadoop-env.sh
SylvainJuge: As discussed with @robsunday yesterday, this looks to be a modified copy of a shell script of the hadoop distribution, which might have been modified to configure the JMX configuration.

The current startup script might allow to easily provide this configuration, but quite often it's not possible to easily override through a single env variable, and `sed` surgery isn't the easiest thing to maintain.

jmx-scraper/src/main/resources/hadoop.yaml
SylvainJuge: Those are called "yaml anchors", they allow to reference any existing attribute value as-is to prevent duplication and see that things are linked together. For example if a metric is being split across two mbean attributes, those must have a common description and unit otherwise it would be two distinct metrics.

This is a feature provided by the yaml library used in instrumentation, it would be a good idea to document it to prevent future confusion.

jmx-scraper/src/integrationTest/resources/hadoop-env.sh
robsunday: So I tried few ideas how to get rid of this custom script, but without a luck. All tips/guides I've found on the internet recommend modification of hadoop-env.sh (as it is done in our test).
At first glance it looks like passing your own env var HADOOP_NAMENODE_OPTS to docker should work.
Unfortunately, Hadoop launches multiple processes during startup and executes this file many times to setup env variables and the ossue is that initial content of HADOOP_NAMENODE_OPTS passed to docker is somehow lost.
Only hardcoding JMX stuff in hadoop-env.sh worked for me, but if you know about reasonable alternate solution I'll be happy to use it.

jmx-scraper/src/main/resources/hadoop.yaml
robsunday: What is in your opinion the best place to explain it? Anchors and aliases are standard YAML feature so we may just mention somewhere that we use to improve maintainability.

This feature is already used in few our YAMLs

jmx-scraper/src/integrationTest/resources/hadoop-env.sh
SylvainJuge: Thanks for trying to remove it, we should be fine to leave it as-is since we don't have a better option.

jmx-scraper/src/main/resources/hadoop.yaml
SylvainJuge: Maybe documenting this directly into the [instrumentation/jmx-metrics](https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation/jmx-metrics) where this feature comes from would be a good start.

breedx-splk: I wouldn't sweat it, since it's mostly me having just been ignorant of yaml anchors. It can wait. Thanks for answering my question tho!

breedx-splk (approved):

PR #1678, opened 2025-01-28
jmx scraper align custom yaml config

### LINKED ISSUE #1662: [jmx-scraper] align custom configuration with instrumentation
This is a follow-up task of https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1651

In instrumentation (JMX insights), the `otel.jmx.config` option provides the path to a list of custom YAML metrics definitions, we currently use `otel.jmx.custom.scraping.config` which does not allow to use a list.

### Proposal
- rename `otel.jmx.custom.scraping.config` configuration option to `otel.jmx.config`
- add support for a list of configuration files
- providing compatibility with `otel.jmx.custom.scraping.config` is not required as this option only exists for jmx-scraper and is not expected to be widely used, the equivalent in jmx-gatherer was only supporting groovy.

SylvainJuge:
Fixes #1662 

**Description:**

Make jmx scraper reuse the same `otel.jmx.config` option to provide custom yaml metrics definitions.

Compatibility with the current `otel.jmx.custom.scraping.config` option is provided with a warning.

### DIFF
--- a/jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/config/JmxScraperConfig.java
+++ b/jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/config/JmxScraperConfig.java
@@ -150,8 +150,11 @@ public class JmxScraperConfig {
     // providing compatibility with the deprecated 'otel.jmx.custom.scraping.config' config option
     String jmxConfigDeprecated = config.getString(JMX_CONFIG_LEGACY);
     if (jmxConfigDeprecated != null) {
-      logger.warning(JMX_CONFIG_LEGACY + " deprecated option is used, replacing with '" + JMX_CONFIG
-          + "' is recommended");
+      logger.warning(
+          JMX_CONFIG_LEGACY
+              + " deprecated option is used, replacing with '"
+              + JMX_CONFIG
+              + "' is recommended");
       List<String> list = new ArrayList<>(jmxConfig);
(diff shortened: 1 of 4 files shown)


jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/config/JmxScraperConfig.java
```diff
@@ -76,9 +85,8 @@ public String getServiceUrl() {
     return serviceUrl;
   }
 
-  @Nullable
-  public String getCustomJmxScrapingConfigPath() {
-    return customJmxScrapingConfigPath;
+  public List<String> getJmxConfig() {
+    return jmxConfig;
```
breedx-splk, line 89: I like wrapping with immutable so it's clear that a caller can't change our internal state.
```suggestion
    return Collections.unmodifyableList(jmxConfig);
```

breedx-splk (approved):
One small comment but looks good to me. Good on the alignment!

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/config/JmxScraperConfig.java
SylvainJuge: Good point, I've applied your suggestion in 911eefe when the attribute is set rather than in the getter and also covered the target system config option.

Issue #1681, opened 2025-01-30
[jmx-scraper] support SSL connection with variants


SylvainJuge:
We need to add support for SSL with the following variants
- server certificate
- client certificate
- RMI registry with/without SSL
- JMX remote profile and realm (see jmx gatherer for details)

Part of #1362 

I did some experiments, and there are some unexpected challenges, so I think it's better to describe an implementation proposal to discuss/explain what the implementation will look like.

### Implementation proposal:

Currently, testing is split in two parts:
- connection to the remote JVM with `JmxConnectorBuilderTest`, where testing for those features would naturally fit
- target systems are tested without authentication nor SSL because we only need to check for the captured metrics, not how we connect to the JVM

In `JmxConnectorBuilderTest`, a target JVM is created and we connect using a local `JmxConnectorBuilder` instance, which works fine but has some challenges due to docker NAT and JMX/RMI protocols which forces us to find a random available port and then use it for communication between the host and the target JVM (in particular with macOS where the docker containers run in a separate JVM).

In the target systems tests we don't have this issue because the containers communicate to each other directly  without any port mapping.

Moving the tests in `JmxConnectorBuilderTest` to run within a container just complicates debugging without having much benefit for now, but the current "fixed mapped port" strategy we used so far is not possible with SSL as JMX and RMI need to have non-identical ports (otherwise we get "port already in use" errors).

Also, given the complexity of the communication with JMX (authentication, SSL, client and server certificates and their respective keystores, ...) it is also expected that end-users will experience challenges when trying to make it working. So I think we should think about adding a "test connection" feature in the jmx-scraper executable CLI and then test it with a regular integration test, that would only use container-to-container communication. The minor downside here is that the `JmxConnectorBuilder` testing and debugging will be a bit harder as it will be in a JVM running in a container, but it's an acceptable compromise.

The steps of this plan would be:
- [ ] create a new "test connection" feature to CLI, add an integration test for it with current implementation that would replace `JmxConnectionBuilderTest`
- [ ] add support for server certificate
- [ ] add support for client certificate
- [ ] add support for RMI registry with/without SSL
- [ ] add support for remote profile and realm

SylvainJuge:
This is now complete and released with version 1.44.0.

PR #1684, opened 2025-01-31
jmx scraper connection test

SylvainJuge:
Relates to #1681 

This PR adds a new CLI argument to enable end-users to test JMX connection.

JMX connection is expected to be tedious to setup and will benefit from a fast feedback loop as there are multiple things that can be problematic during setup:
- authentication credentials
- client and server SSL certificates
- SSL enabled on RMI registry
- ...

Also, testing connection through this mean allow us to validate configuration usage with CLI interface and avoids having any complication due to JMX/RMI port mapping which is not NAT friendly (open connection on a port, then another one is used afterwards).


jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/JmxConnectionTest.java
```diff
@@ -0,0 +1,133 @@
...
+      app.start();
+      try (JmxScraperContainer scraper = customizeScraper.apply(scraperContainer())) {
+        scraper.start();
+        waitTerminated(scraper);
+        checkConnectionLogs(scraper, /* expectedOk= */ true);
+      }
+    }
+  }
+
+  private static void checkConnectionLogs(JmxScraperContainer scraper, boolean expectedOk) {
```
SylvainJuge, line 86: [for reviewer] we have to parse and check scraper logs in order to know if the connection is OK or not, this is because we can't access the process exit value when the program completes.

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/TestAppContainer.java
```diff
@@ -61,28 +61,6 @@ public TestAppContainer withUserAuth(String login, String pwd) {
...
-   * Configures app container for host-to-container access, port will be used as-is from host to
-   * work-around JMX in docker. This is optional on Linux as there is a network route and the
-   * container is accessible, but not on Mac where the container runs in an isolated VM.
-   *
-   * @param port port to use, must be available on host.
-   * @return this
-   */
-  @CanIgnoreReturnValue
-  public TestAppContainer withHostAccessFixedJmxPort(int port) {
-    // To get host->container JMX connection working docker must expose JMX/RMI port under the same
```
SylvainJuge, line 74: [for reviewer] this is no longer necessary as we only rely on container-to-container communication and we don't have to use docker NAT to access the JMX endpoint.

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/JmxConnectorBuilderTest.java
```diff
@@ -1,95 +0,0 @@
...
-
-import java.io.IOException;
-import javax.management.ObjectName;
-import javax.management.remote.JMXConnector;
-import org.junit.jupiter.api.AfterAll;
-import org.junit.jupiter.api.BeforeAll;
-import org.junit.jupiter.api.Test;
-import org.testcontainers.containers.Network;
-
-public class JmxConnectorBuilderTest {
```
SylvainJuge, line 18: [for reviewer] this is replaced by `JmxConnectionTest`.

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/JmxScraper.java
```diff
@@ -52,8 +54,11 @@ public static void main(String[] args) {
     // set log format
     System.setProperty("java.util.logging.SimpleFormatter.format", "%1$tF %1$tT %4$s %5$s%n");
 
+    List<String> effectiveArgs = new ArrayList<>(Arrays.asList(args));
```
robsunday, line 57: In my opinion this approach is not so straightforward. Maybe it would be better to use some flexible commandline arg parse library instead?

SylvainJuge: For now we just have a few possible arguments, so I think this is fine.

If we start to introduce more I agree that the complexity to handle it with such a simple approach won't work. I think for now we could add a comment to explain that adding extra complexity here might be a good idea to bring a CLI argument parsing library.

robsunday: Ok. Let's do it that way

robsunday (approved):

robsunday (approved):

trask (approved):

PR #1685, opened 2025-02-03
[baggage-processor] Change description to fit purpose

breedx-splk:
The description of the `baggage-processor` module suggested that it was for spans. It contains the ability to add attributes to both spans and logs, so let's change the description to make it more fitting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,11 +10,11 @@ feature or via instrumentation, this project is hopefully for you.
 ## Provided Libraries
 
 | Status* | Library                                                           |
-| ------- | ----------------------------------------------------------------- |
+| ------- |-------------------------------------------------------------------|
 | beta    | [AWS Resources](./aws-resources/README.md)                        |
 | stable  | [AWS X-Ray SDK Support](./aws-xray/README.md)                     |
 | alpha   | [AWS X-Ray Propagator](./aws-xray-propagator/README.md)           |
-| alpha   | [Baggage Span Processor](./baggage-processor/README.md)           |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -10,11 +10,11 @@ feature or via instrumentation, this project is hopefully for you.
 ## Provided Libraries
 
 | Status* | Library                                                           |
-| ------- | ----------------------------------------------------------------- |
+| ------- |-------------------------------------------------------------------|
 | beta    | [AWS Resources](./aws-resources/README.md)                        |
 | stable  | [AWS X-Ray SDK Support](./aws-xray/README.md)                     |
 | alpha   | [AWS X-Ray Propagator](./aws-xray-propagator/README.md)           |
-| alpha   | [Baggage Span Processor](./baggage-processor/README.md)           |
+| alpha   | [Baggage to Attributes Processor](./baggage-processor/README.md) |
```
trask, line 17: I don't think we have precedence for the term "attributes processor", maybe:

```suggestion
| alpha   | [Baggage Processors](./baggage-processor/README.md) |
```

breedx-splk: I think "Baggage to Attributes Processor" is more accurate/descriptive -- it's literally what it does, it takes baggage and puts it into attributes (on spans or logs). I don't mind changing it, because "Baggage Processors" is still better than "Baggage Span Processors", and it matches the literal name of the repo...and maybe people get twitchy about using an established term like "Processors" when not applied directly to a signal?

trask (approved):

PR #1688, opened 2025-02-04
Update publishing for GCP Auth Extension

psx95:
**Description:**

Adds explicit classifier for shaded variant of the dependency. This allows both - shaded and unshaded variant of the dependency.

**Testing:**

Ran `./gradlew publishToMavenLocal` task locally to examine the generated `.pom` and `.module` files. 

**Documentation:**

Should be updated after the next release.

### DIFF
--- a/gcp-auth-extension/build.gradle.kts
+++ b/gcp-auth-extension/build.gradle.kts
@@ -60,12 +60,19 @@ tasks {
   }
 
   shadowJar {
-    archiveClassifier.set("")
+    archiveClassifier.set("shadow")
   }
 
   jar {
-    // Disable standard jar
-    enabled = false
+    /**
+     * We need to publish both - shaded and unshaded variants of the dependency
+     * Shaded dependency is required for use with the Java agent.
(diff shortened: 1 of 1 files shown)


gcp-auth-extension/build.gradle.kts
```diff
@@ -60,12 +60,19 @@ tasks {
...
 
   jar {
-    // Disable standard jar
-    enabled = false
+    /**
+     * We need to publish both - shaded and unshaded variants of the dependency
+     * Shaded dependency is required for use with the Java agent.
+     * Unshaded dependency can be used with OTel Autoconfigure module.
+     *
+     * Not overriding the classifier to empty results in an implicit classifier 'plain' being
```
trask, line 72: does this mean the classifier in that case would be the text `plain`? thanks

trask: got it thanks!

trask (approved):

breedx-splk (approved):

PR #1699, opened 2025-02-06
Fix scorecard analysis

trask:
Should fix

> Unable to resolve action `ossf/scorecard-action@v2`, unable to find version `v2`


breedx-splk (approved):
hrmph.

breedx-splk:
> hrmph.

tell me more about how much you like pinning, ossf? 🙃

PR #1701, opened 2025-02-06
Add owasp dependency check workflow

trask:


.github/workflows/owasp-dependency-check-daily.yml
```diff
@@ -0,0 +1,48 @@
...
+    # daily at 1:30 UTC
+    - cron: "30 1 * * *"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  analyze:
+    runs-on: otel-linux-latest-8-cores
```
trask, line 16: it's hanging when I run on my fork (where I don't have large runners available)

I can't test this here until after merging

trask: nevermind, I think I figured it out

laurit (approved):

PR #1709, opened 2025-02-07
Update test dependencies

trask:
OSSF scorecard is dinging us for vulnerabilities in the old spring version (it appears to only look at pom file dependencies)


maven-extension/src/test/resources/projects/springboot_1/pom.xml
```diff
@@ -6,7 +6,7 @@
   <parent>
     <groupId>org.springframework.boot</groupId>
     <artifactId>spring-boot-starter-parent</artifactId>
-    <version>2.6.1</version>
+    <version>3.4.1</version>
```
trask, line 9: one version below latest to test that the `ignorePresets` above will keep us updated going forward

breedx-splk (approved):

PR #1718, opened 2025-02-08
Add FOSSA license scanning

trask:


dependencyManagement/build.gradle.kts
```diff
@@ -2,96 +2,60 @@ plugins {
...
-    for (dependency in CORE_DEPENDENCIES) {
-      api(dependency)
-      val split = dependency.split(':')
-      dependencyVersions[split[0]] = split[2]
-    }
-    for (dependency in DEPENDENCIES) {
-      api(dependency)
-      val split = dependency.split(':')
-      dependencyVersions[split[0]] = split[2]
-    }
```
trask, line 95: getting rid of this was unrelated, but we don't use the `dependencyVersions` constants, and inlining the dependencies directly into this section seems so much more straightforward

laurit (approved):

PR #1710, opened 2025-02-07
jmx-scraper add ssl support

SylvainJuge:
~~This is based on #1684 which is not merged yet, I'll rebase to minimize the diff once it's merged.~~

Part of #1681

This adds the following features with proper integration tests for JMX with SSL:
- SSL enabled on server, server certificate must be trusted on the client
- SSL enabled on server with SSL RMI registry
- SSL enabled on server with client SSL certificate

Using other advanced JAAS configuration options for authentication is not part of this PR and will be implemented in a separate PR.

PR #1741, opened 2025-02-17
jmx-scraper add missing custom yaml support

SylvainJuge:
While working on #1710 I found that the `io.opentelemetry.contrib.jmxscraper.JmxScraperContainer#withCustomYaml` was not used, which means using custom yaml feature was not properly tested with integration tests.

It turns out that the feature was not only not tested but also not implemented, so this PR fixes both.

Testing extends the tests for the `jvm` target system to avoid creating duplication and also allows to test usage when combined with `otel.jmx.target.system`.

### DIFF
--- a/jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/JvmIntegrationTest.java
+++ b/jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/JvmIntegrationTest.java
@@ -30,7 +30,10 @@ public class JvmIntegrationTest extends TargetSystemIntegrationTest {
   @Override
   protected JmxScraperContainer customizeScraperContainer(
       JmxScraperContainer scraper, GenericContainer<?> target, Path tempDir) {
-    return scraper.withTargetSystem("jvm");
+    return scraper
+        .withTargetSystem("jvm")
+        // also testing custom yaml
+        .withCustomYaml("custom-metrics.yaml");
   }
 
   @Override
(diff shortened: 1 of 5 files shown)


jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/JvmIntegrationTest.java
```diff
@@ -30,7 +30,10 @@ protected GenericContainer<?> createTargetContainer(int jmxPort) {
   @Override
   protected JmxScraperContainer customizeScraperContainer(
       JmxScraperContainer scraper, GenericContainer<?> target, Path tempDir) {
-    return scraper.withTargetSystem("jvm");
+    return scraper
+        .withTargetSystem("jvm")
+        // also testing custom yaml
+        .withCustomYaml("custom-metrics.yaml");
```
robsunday, line 36: As we discussed today, most use cases will be for extending existing target with some custom metrics, but anyway - I think we need one more integration test with no target provided, just custom yaml.
This would prove that we are handling this scenario correctly as well.

robsunday (approved):
Approved, however it would be nice to add one more test.

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/JvmIntegrationTest.java
SylvainJuge: I've added a dedicated test class 0ef71dae795f41abdd961c0b05093bf33042c94e, there is minor duplication of jvm target system integration test, but it makes it simpler to understand than trying to share code between the two.

trask (approved):

PR #1749, opened 2025-02-20
jmx-scraper replicate and document jmxmp/sasl config

SylvainJuge:
Part of #1681

The JMX Gatherer provides options to use the JMXMP and SASL authentication options, which allow to extend the simple login/pwd authentication and delegate to a "real" authentication system like LDAP. While that sounds simple with words is more challenging in practice because this is something sparsely documented and the JMXMP client and server implementation is not even shipped with the JDK.

I was not able to reproduce a working client and server for end-to-end integration test, so for now it means it will be only documented and provided as-is. The current JMX Gatherer implementation is not tested either, and we can adjust the implementation and hopefully enable testing through user feedback.

In summary this PR contains:
- adding the JMXMP library from Terracotta (the same that is used with gatherer)
- document the configuration options

The implementation of those config options is already included and had been copied as-is from gatherer code.


settings.gradle.kts
```diff
@@ -14,6 +14,13 @@ dependencyResolutionManagement {
   repositories {
     mavenCentral()
     mavenLocal()
+    // terracotta repository for jmxmp connector
+    maven {
+      setUrl("https://repo.terracotta.org/maven2")
+      content {
+        includeGroupByRegex("""org\.terracotta.*""")
+      }
+    }
```
SylvainJuge, line 23: [for reviewer] moving this here allows to avoid duplication between gatherer and scraper to access this repository.

robsunday (approved):

trask (approved):

PR #1750, opened 2025-02-20
jmx-scraper test config sources

### LINKED ISSUE #1687: [jmx-scraper] test all the possible ways to provide configuration
Currently, the JMX scraper tests are all using JVM system properties for configuration.

However, there are other ways to provide the configuration and they aren't tested end-to-end:
- standard input
- environment variables
- external properties file.

In order to test those, we should at least attempt to cover the minimal configuration for each of the 4 ways to configure it, we don't have to test every configuration option with every possible way to provide it.

In addition, doing that test work based on "test connnection" implemented in #1684 would be relevant as it allows to test the whole CLI program as the user would experience it. Work on this issue would then depend on #1684 to be merged though.

SylvainJuge:
Fixes #1687

Adds tests to cover all the possible ways to provide configuration for jmx-scraper:
- standard input
- properties file
- system properties
- environment variables

Only the `JmxConnectionTest` has been parameterized to cover those as the connection with the remote JVM is the most complicated part as the JVM system environment need to be propagated from properties to global system properties.

The test fail when we remove the `JmxScraper#propagateToSystemProperties` method which handles system properties propagation.

Adding those tests did not uncover any new issue, thus they help prevent any future regression.


jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/JmxScraperContainer.java
```diff
@@ -182,83 +199,185 @@ public JmxScraperContainer withSslRmiRegistry() {
...
+   * @return this
+   */
+  @CanIgnoreReturnValue
+  public JmxScraperContainer withConfigSource(ConfigSource source) {
+    this.configSource = source;
+    return this;
+  }
+
   @Override
   public void start() {
```
robsunday, line 215: This method got very long and bit hard to grasp. I'd split it's content into few smaller methods.
Proposal:
1) lines 217-252 initConfig()
2) lines 256-327 configureCmd(cmd)
3) lines 335-353 generateScraperLaunchScript
...

SylvainJuge: good point, I've split in 3 methods in 2b97c46

robsunday (approved):

trask:
@SylvainJuge I'll merge once merge conflict is resolved, thanks!

trask (approved):

PR #1764, opened 2025-02-24
Add instructions for Auth Extension shadow variant

psx95:
**Description:**

Update docs instructing users to use the shaded variant when using the extension with Java auto-instrumentation agent.

**Documentation:**

Updates `README.md` in `gcp-auth-extension`.

### DIFF
--- a/gcp-auth-extension/README.md
+++ b/gcp-auth-extension/README.md
@@ -48,6 +48,9 @@ Here is a list of configurable options for the extension:
 The OpenTelemetry Java Agent Extension can be easily added to any Java application by modifying the startup command to the application.
 For more information on Extensions, see the [documentation here](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/examples/extension/README.md).
 
+> [!IMPORTANT]
+> Make sure to download the 'shaded' variant of the Authentication Extension for use with OpenTelemetry Java auto-instrumentation agent. The shaded version is available under the classifier name `shadow`.
+
 Below is a snippet showing how to add the extension to a Java application using the Gradle build system.
 
 ```gradle


gcp-auth-extension/README.md
```diff
@@ -48,6 +48,9 @@ Here is a list of configurable options for the extension:
 The OpenTelemetry Java Agent Extension can be easily added to any Java application by modifying the startup command to the application.
 For more information on Extensions, see the [documentation here](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/examples/extension/README.md).
 
+> [!IMPORTANT]
+> Make sure to download the 'shaded' variant of the Authentication Extension for use with OpenTelemetry Java auto-instrumentation agent. The shaded version is available under the classifier name `shadow`.
+
```
breedx-splk, line 53: This document doesn't (yet?) describe how to download anything. Can we clarify that as well please? This change alone doesn't seem to provide enough, or I'm missing something.

psx95: I can further clarify it by adding links to the maven artifact download. 
I did something [similar here](https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/tree/main/exporters/auto#shaded-local-jar), but I thought it was too verbose.

The intention of this  documentation is to let the users know not to download the plain JAR from the [maven page](https://mvnrepository.com/artifact/io.opentelemetry.contrib/opentelemetry-gcp-auth-extension/1.44.0-alpha), but look for the `shadow` jar insteaad.

psx95: @breedx-splk, let me know if the additional instructions make the document a little more clear. 

Thank you for the review!

michaelsafyan: Should this clarify _why_? Or what `shaded` means in this context?

psx95: I don't think we need to explain `shaded` here. Shading is a known process in Java and we are just using standard practice here.

For *why*, I'm on the fence, I don't have strong opinion since it is kind of implicit. 

'Shading' is the process of bundling all required dependencies in the JAR.  When used with the Java agent, we pass extensions as JAR file and it is necessary that all dependencies are present on the classpath at load time which is why agent extensions bundle them.

trask: I think it's ok to just focus on documenting it from the perspective of an end user

if there's a need for documenting why, I'd probably put that in the gradle file where the shading is implemented

michaelsafyan: I'm not sure I would take it for granted that readers know what `shaded` means. For example, I've heard of "thin JARs", "source JARs", and "JARs with dependencies" .... never heard of "shaded". Someone who isn't deeply familiar with packaging and distribution of libraries may not necessarily be familiar with this terminology.

At the very least, link to a reference for this? Perhaps:

 - https://softwareengineering.stackexchange.com/questions/297276/what-is-a-shaded-java-dependency
 - https://maven.apache.org/plugins/maven-shade-plugin/

psx95: "Shaded JAR" is the same as "Fat JAR" or "Uber JAR".

I understand that someone unfamiliar with packaging and distribution might find the terminology confusing. To keep the README simple and focused on explaining how to use this dependency, as Trask suggested, I've added some additional documentation in the build file of this extension. 

I hope this would provide additional context to users interested in understanding the packaging & distribution here.

psx95: I'm marking this as resolved. We can always update the documentation, but I think this looks good for now.

Feel free to re-open this discussion.

aabmass (approved):

jsuereth (approved):

Issue #1772, opened 2025-02-28
[disk-buffering] `StorageConfiguration.Builder.setTemporaryFileProvider()` exposes internal class


bencehornak:
The public `StorageConfiguration.Builder.setTemporaryFileProvider` class requires an instance of the `TemporaryFileProvider` class, which is situated in the `io.opentelemetry.contrib.disk.buffering.**internal**.files` package:

https://github.com/open-telemetry/opentelemetry-java-contrib/blob/be32409fe07b4ec94f78c5b7919d864451e6b56d/disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/StorageConfiguration.java#L84

As a result, the users of the library need to import the internal class if they wish to provide a custom implementation.

My suggestion is to move the `TemporaryFileProvider` class to the `io.opentelemetry.contrib.disk.buffering` package to explicitly mark it as public API.

LikeTheSalad:
Thank you for letting us know, this was a miss from our side. I'll create a PR to move it to a public package.

LikeTheSalad:
I've created [this PR](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1781) for it.

bencehornak:
Closed via #1781, thanks a lot @LikeTheSalad!

PR #1781, opened 2025-03-06
Diskbuffering configuration package

LikeTheSalad:
**Description:**

Creates a public ~configuration~ `config` package to place all the configuration-related files, including the TemporaryFileProvider type that was placed in an internal package by mistake.

**Existing Issue(s):**

https://github.com/open-telemetry/opentelemetry-java-contrib/issues/1772


breedx-splk (approved):
Thanks. It's a good change....although I would have preferred `io.opentelemetry.contrib.disk.buffering.config` to `io.opentelemetry.contrib.disk.buffering.configuration` for brevity, but not a huge issue.

LikeTheSalad:
> Thanks. It's a good change....although I would have preferred `io.opentelemetry.contrib.disk.buffering.config` to `io.opentelemetry.contrib.disk.buffering.configuration` for brevity, but not a huge issue.

Sounds good. I've made the change.

PR #1782, opened 2025-03-06
jmx reuse instrumentation metrics

SylvainJuge:
Adds a new configuration for jmx scraper: `otel.jmx.target.source` that defines how the values of `otel.jmx.target.system` are resolved to their respective YAML definitions.

3 possible values:
- `auto` (default) use instrumentation yaml if available, fallback to embedded if possible
- `legacy` only use the embedded yaml (which is very close to JMX gatherer)
- `instrumentation` only use the instrumentation yaml without fallback

Thanks @robsunday for the idea of an automatic fallback, that will really help with the migration to having everything defined in instrumentation. In short, the metrics produced by default will be the ones from instrumentation repository (in the "library" part, not the "javaagent" as they currently are), so every release and update of instrumentation dependency will bring an updated set of metrics. For example this will be the case for `jvm` system that will be added with https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/13392 pull-request.

I have tested it locally with a snapshot build of https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/13392 to have the JVM metrics included in instrumentation, so as the time of writing this is "works on my computer certified".


jmx-scraper/src/test/java/io/opentelemetry/contrib/jmxscraper/config/JmxScraperConfigTest.java
```diff
@@ -138,74 +142,96 @@ void shouldFailValidation_missingConfigPathAndTargetSystem() {
...
-  //
-  //  @Test
-  //  @SetSystemProperty(key = "otel.metric.export.interval", value = "123")
-  //  void otelMetricExportIntervalRespected() {
-  //    JmxConfig config = new JmxConfig();
-  //    assertThat(config.intervalMilliseconds).isEqualTo(10000);
-  //    assertThat(config.properties.getProperty("otel.metric.export.interval")).isEqualTo("123");
-  //  }
-  //
+  @ParameterizedTest
```
SylvainJuge, line 159: [for reviewer] we could do the same tests with "real" yaml systems definitions, but this is not possible because we need to cover all the cases where a system is exclusively supported in `legacy` or `instrumentation` and when it's in both, in which case the priority will depend on the value of `otel.jmx.target.source`.

trask (approved):

PR #1794, opened 2025-03-14
Update renovate rate limit

trask:


breedx-splk (approved):
Any reasoning behind this number?

trask:
> Any reasoning behind this number?

greater than the default (2) and less than [`prConcurrentLimit`](https://docs.renovatebot.com/configuration-options/#prconcurrentlimit) which is 10

PR #1823, opened 2025-04-01
add logs filtering

oliver-zhang:
Description:

Sometimes we don't want to collect all the logs, we only need the logs associated with the call chain, this processor only keeps logs if they have an associated span which is sample

Existing Issue(s):

NA

Testing:

FilteringLogRecordProcessorTest

Documentation:

Outstanding items:

### DIFF
--- a/processors/README.md
+++ b/processors/README.md
@@ -27,9 +27,10 @@ logger_provider:
   processors:
       - event_to_span_event_bridge:
 ```
-## Filtering Log Processor
-`FilteringLogRecordProcessor` is a `LogRecordProcessor` that only keep logs  based on a predicate
 
+## Filtering Log Processor
+
+`FilteringLogRecordProcessor` is a `LogRecordProcessor` that only keep logs  based on a predicate
 
 ## Component owners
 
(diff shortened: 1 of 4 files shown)


processors/README.md
```diff
@@ -27,6 +27,9 @@ logger_provider:
   processors:
       - event_to_span_event_bridge:
 ```
+## Filtering Log Processor
+`FilteringLogRecordProcessor` is a `LogRecordProcessor` which only keeps logs if they have an associated span which is sample
```
LikeTheSalad, line 31: I think the description is a bit misleading because, if I understood correctly, the `FilteringLogRecordProcessor` processor emits logs based on a predicate. So it is not related to spans (which I think is great because that makes this processor more flexible).

LikeTheSalad (commented):
Thank you. I think it's a nice addition 👍  I just added a comment to clarify the readme. Please run `./gradlew spotlessApply` to address the CI check failures.

processors/README.md
oliver-zhang: Yes, it should have nothing to do with trace

LikeTheSalad (approved):
Thank you!

trask (approved):

PR #1851, opened 2025-04-14
Update renovate config

trask:
Adopt a couple of best practices from https://github.com/open-telemetry/sig-security/pull/97


.github/renovate.json5
```diff
@@ -3,23 +3,28 @@
...
+      // (patch updates are typically non-breaking)
+      "groupName": "all patch versions",
+      "matchUpdateTypes": ["patch"],
+      "schedule": ["before 8am every weekday"]
+    },
+    {
+      // avoids these Renovate PRs from trickling in throughout the week
+      // (consolidating the review process)
+      "matchUpdateTypes": ["minor", "major"],
+      "schedule": ["before 8am on Monday"]
```
breedx-splk, line 27: Hmmmm, I kinda like having them trickle in because I assume that it helps spread the work across approver availability more evenly...but I dunno. This still separates packges into separate PRs tho right, not one big PR monday monring?

breedx-splk (approved):
question about the monday PR, and some feedback, but I don't feel too strongly about it.

.github/renovate.json5
trask: yeah, that makes sense, but this also rings true

> Receiving PRs at any hour can increase the feeling of being "overwhelmed" by updates and possibly interrupt your flow during working hours, so many Renovate users also consider reducing Renovate's schedule to be outside their normal working hours, for example weeknights and weekends. 

https://docs.renovatebot.com/noise-reduction/

I kind of like the idea of dealing with renovate PRs in the morning and not having them trickle in, but not sure how to balance that across multiple maintainers and multiple definitions of morning

PR #1852, opened 2025-04-14
Remove unneeded opentelemetrybot from script

trask:
Now that this repo is completely moved over to the otelbot app.


breedx-splk (approved):
also dependabot

PR #1855, opened 2025-04-16
Add messaging wrappers to support lightweight manual instrumentation

Cirilla-zmh:
**Description:**

This PR introduces a complete implementation of a generic messaging process wrapper API, designed to facilitate the creation of process spans and the propagation of trace context in a standardized and lightweight manner. It builds upon the initial draft presented in PR #1803, which provides additional context for this work.

**Key features and contributions include:**

- Generic Process Wrapper API: A flexible and reusable API that supports the creation of process spans and ensures trace context propagation across different libraries and frameworks.
- Library Implementations:
  - Added support for `kafka-clients`.
  - Added support for `aliyun-mns-sdk` (Alibaba Cloud Simple Message Queue).
 
**Existing Issue(s):**

- https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/13340
- https://github.com/open-telemetry/opentelemetry-specification/issues/4426

**Testing:**

- [x] Base API integration tests.
- [x] Integration tests for kafka-clients.
- [x] Integration tests for aliyun-mns-sdk (pending for building mocked server).

**Documentation:**

For detailed usage instructions and API documentation, please refer to the updated [README.md](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1855/files#diff-e6b58ce5cdab148e0a07d4fba830098be45d6fcc568e8fba3c884a47fd18e9b4)

### DIFF
--- a/messaging-wrappers/api/src/test/java/io/opentelemetry/contrib/messaging/wrappers/UserDefinedMessageSystemTest.java
+++ b/messaging-wrappers/api/src/test/java/io/opentelemetry/contrib/messaging/wrappers/UserDefinedMessageSystemTest.java
@@ -39,7 +39,6 @@ import io.opentelemetry.instrumentation.api.incubator.semconv.messaging.Messagin
 import java.nio.charset.StandardCharsets;
 import java.util.Collections;
 import java.util.HashMap;
-
 import org.assertj.core.api.AbstractAssert;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
--- a/messaging-wrappers/kafka-clients/src/test/java/io/opentelemetry/contrib/messaging/wrappers/kafka/KafkaClientTest.java
+++ b/messaging-wrappers/kafka-clients/src/test/java/io/opentelemetry/contrib/messaging/wrappers/kafka/KafkaClientTest.java
@@ -13,7 +13,6 @@ import static io.opentelemetry.semconv.incubating.MessagingIncubatingAttributes.
 import static io.opentelemetry.semconv.incubating.MessagingIncubatingAttributes.MESSAGING_KAFKA_CONSUMER_GROUP;
 import static io.opentelemetry.semconv.incubating.MessagingIncubatingAttributes.MESSAGING_KAFKA_MESSAGE_OFFSET;
 import static io.opentelemetry.semconv.incubating.MessagingIncubatingAttributes.MESSAGING_MESSAGE_BODY_SIZE;
-import static io.opentelemetry.semconv.incubating.MessagingIncubatingAttributes.MESSAGING_MESSAGE_ID;
 import static io.opentelemetry.semconv.incubating.MessagingIncubatingAttributes.MESSAGING_OPERATION;
 import static io.opentelemetry.semconv.incubating.MessagingIncubatingAttributes.MESSAGING_SYSTEM;
 


messaging-wrappers/api/src/main/java/io/opentelemetry/contrib/messaging/wrappers/semconv/MessagingProcessRequest.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.List;
+import javax.annotation.Nullable;
+
+/**
+ * An interface to expose messaging properties for the pre-defined process wrapper.
+ *
+ * <p>Inspired from <a
+ * href=https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation-api-incubator/src/main/java/io/opentelemetry/instrumentation/api/incubator/semconv/messaging/MessagingAttributesGetter.java>MessagingAttributesGetter</a>.
+ */
+public interface MessagingProcessRequest {
```
trask, line 19: Can define separate extractors for each implementation instead of implementing this interface for each implementation? that would seem more similar to the existing instrumentation extractor design

Cirilla-zmh: > separate extractors for each implementation instead of implementing this interface for each implementation

Sorry, I don't think so.

If we do that, users have got to create an `AttributesExtractor` for each manual implementation, rather than reuse the `DefaultMessagingAttributesGetter`. Therefore, users should associate the attributes keys defined in semconv with the fields message - it's normally hard, and that's one of reasons why we give the process wrapper today.

Conversely, if we give this interface. We can take advantage of the self-explanatory nature of the function name and add java comments to tell users what this value should point to. This process seems more natural, especially for those users who are not such familiar to OTel.

trask: I don't disagree with your reasoning, but doesn't that same reasoning apply to all of the existing instrumentation api semconv attribute extractor design?

Cirilla-zmh: Maybe I didn't get your point. Are you suggesting a design like this:

Drop the interface of `MessagingProcessRequest` and create a interface of `MessagingAttribute` (or reuse that in [instrumentations](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation-api-incubator/src/main/java/io/opentelemetry/instrumentation/api/incubator/semconv/messaging/MessagingAttributesGetter.java))

If we do so, the complexity for manual users may increase slightly—they will need to implement the `MyMessagingGetter` and define a request in cases where the framework's provided message structure is insufficient to retrieve all attributes (similar to [KafkaProcessRequest](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/kafka/kafka-clients/kafka-clients-common-0.11/library/src/main/java/io/opentelemetry/instrumentation/kafkaclients/common/v0_11/internal/KafkaProcessRequest.java)). This is also why I didn't reuse the design of `MessagingAttributesGetter`—I believe it might introduce additional cognitive overhead.

However, if you prefer to maintain consistency in the design to make it easier for developers to manage, that makes sense as well. I can adjust my implementation to align with the instrumentation as much as possible.

trask: > If we do so, the complexity for manual users may increase slightly—they will need to implement the `MyMessagingGetter` and define a request in cases where the framework's provided message structure is insufficient to retrieve all attributes (similar to [KafkaProcessRequest](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/kafka/kafka-clients/kafka-clients-common-0.11/library/src/main/java/io/opentelemetry/instrumentation/kafkaclients/common/v0_11/internal/KafkaProcessRequest.java)). This is also why I didn't reuse the design of `MessagingAttributesGetter`—I believe it might introduce additional cognitive overhead.

yeah, it's a good point, but I'd prefer to stay consistent with existing instrumentation.

ideally most users of this will be using our pre-built implementations (e.g. for kafka and mns) and so hopefully won't even be aware of which design we use

Cirilla-zmh: It makes sense to me. I'll adjust this implementation.

Cirilla-zmh:
Update:
- Refactored the README by moving the given implementation instructions ahead of the manual setup ones.
- Regarding the generic parameters in `AttributesGetter`: I believe we have to keep them this way because we're reusing the definition of `AttributesExtractor` from the instrumentation. If we restrict the type to a specific interface, users wouldn't be able to define their own `AttributesExtractor` without resorting to explicit type casting. At this point, I think it's reasonable to provide an extensible and generic type here.
- Added tests for aliyun-mns-sdk ;)

cc@steverao @trask

Cirilla-zmh:
Some feedback: reducing the flexibility of AttributesGetter

Cirilla-zmh:
@trask Hey Trask, I've refactored the design, and we now have an implementation that's very similar to the one in
instrumentation. Could you please take a look at it?

messaging-wrappers/api/src/main/java/io/opentelemetry/contrib/messaging/wrappers/semconv/DefaultMessagingAttributesGetter.java
```diff
@@ -0,0 +1,93 @@
...
+ */
+
+package io.opentelemetry.contrib.messaging.wrappers.semconv;
+
+import io.opentelemetry.instrumentation.api.incubator.semconv.messaging.MessagingAttributesGetter;
+import java.util.List;
+import javax.annotation.Nullable;
+
+public enum DefaultMessagingAttributesGetter
+    implements MessagingAttributesGetter<MessagingProcessRequest, Void> {
```
Cirilla-zmh, line 13: Now we have an implementation that mirrors the one used in the instrumentation. Users working with the manual API should define their own MessageRequest by implementing the MessagingProcessRequest interface (see [TestMessageRequest](messaging-wrappers/api/src/test/java/io/opentelemetry/contrib/messaging/wrappers/impl/MessageRequest.java) for example).

If users are using a predefined implementation (e.g., kafka-clients), they can simply reuse the provided class, such as [KafkaProcessRequest](messaging-wrappers/kafka-clients/src/main/java/io/opentelemetry/contrib/messaging/wrappers/kafka/semconv/KafkaProcessRequest.java).

messaging-wrappers/api/src/main/java/io/opentelemetry/contrib/messaging/wrappers/MessagingProcessWrapper.java
```diff
@@ -0,0 +1,112 @@
...
+    for (AttributesExtractor<REQUEST, Void> extractor : this.attributesExtractors) {
+      extractor.onStart(builder, context, request);
+    }
+    return spanBuilder.setAllAttributes(builder.build()).startSpan();
+  }
+
+  protected void handleEnd(Span span, Context context, REQUEST request, @Nullable Throwable t) {
+    AttributesBuilder builder = Attributes.builder();
+    for (AttributesExtractor<REQUEST, Void> extractor : this.attributesExtractors) {
+      extractor.onEnd(builder, context, request, null, t);
```
Cirilla-zmh, line 94: Should we catch exceptions here? There may be unchecked exceptions thrown in abnormal cases, which could affect the user's logic.

We didn't do that in the instrumentation because there's a universal try-catch block in advice code generated by bytebuddy.

messaging-wrappers/api/src/main/java/io/opentelemetry/contrib/messaging/wrappers/semconv/MessagingProcessRequest.java
Cirilla-zmh: Should we eventually remove this interface? If we do, users will have to define their own AttributesExtractor. However, since very few users interact with the manual API, it’s acceptable to leave the design more complex. If you agree, I’ll proceed with that.

PR #1860, opened 2025-04-21
GCP Auth: update header injection implementation

psx95:
**Description:**
Updates the internal implementation such that the required headers are retrieved from the Google Auth Library instead of manually constructing and passing them.

Also fixes flaky unit tests which were affected because of external environment variables.

**Testing:**
 - Update the unit tests.
 - Manually tested token refresh with long running sample application (~2 hr) for both user credentials & service account credentials.  

**Outstanding items:**
 - Efficient mocking of environment variables in the unit & smoke tests.
 - Adding a smoke test for service account credentials.


gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
```diff
@@ -95,20 +98,36 @@ private static SpanExporter addAuthorizationHeaders(
...
+        gcpHeaders.entrySet().stream()
+            .collect(
+                Collectors.toMap(
+                    Map.Entry::getKey,
+                    entry ->
+                        entry.getValue().stream()
+                            .filter(Objects::nonNull) // Filter nulls
+                            .filter(s -> !s.isEmpty()) // Filter empty strings
+                            .collect(Collectors.joining(",")),
+                    (v1, v2) -> v2 // Merge function - take the last seen value
```
aabmass, line 119: Does the overload without a map function work? It seems like you would never get a key collision since you started with a map

aabmass, line 119: Does the overload without a merge function work? It seems like you would never get a key collision since you started with a map

aabmass (commented):
LGTM. Is there a test case verifying that no quota project is sent when using a service account from metadata server?

aabmass (approved):

psx95:
> LGTM. Is there a test case verifying that no quota project is sent when using a service account from metadata server?

The behavior being tested is - [if quota project header is not present in the metadata server, then it is not expected to be in the export headers](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1860/files#diff-82c91483f1ca6d30109d43bba9bbd077fdf193c94a8857a69771846d5d674792R341-R343).

The auth extension is agnostic to the specific sub-type of `GoogleCredentials`.

trask (approved):

Issue #1874, opened 2025-05-03
Instrument artifact transfer


cyrille-leclerc:
### Component(s)

maven-extension

### Is your feature request related to a problem? Please describe.

Instrument artifact transferts to get finer granularity in build execution.

### Describe the solution you'd like

Could be implemented implementing the [org.eclipse.aether.transfer.TransferListener](https://maven.apache.org/resolver/apidocs/org/eclipse/aether/transfer/TransferListener.html) API.
WARNING this could generate thousands of spans breaking readability of the produced Maven build traces.

TODO verify how the extension can implement an Aether component.

smurf667:
Hi @cyrille-leclerc 

Thanks for opening this issue following our brief email exchange.

I don't have an immediate solution to the challenge of having too many transfer spans cluttering the results. One possible approach could be to aggregate them, though this risks losing some useful details (such as timing and location). To address flexibility, I would foresee a configuration switch to enable or disable tracing of transfers.

I've put together a (naive) prototype that I'll share with you privately first. It leverages `org.eclipse.aether.transfer.TransferListener` to track transfers.

Here's a preview of what it could look like in the Jaeger UI:

![Example screenshot](https://github.com/user-attachments/assets/1e288dc9-6258-44a3-b3f4-2c7654ed9645)

cyrille-leclerc:
Thanks @smurf667!

**Would you mind creating a draft PR with your prototype?** I would like to test it.

> I don't have an immediate solution to the challenge of having too many transfer spans cluttering the results. One possible approach could be to aggregate them, though this risks losing some useful details (such as timing and location). To address flexibility, I would foresee a configuration switch to enable or disable tracing of transfers.

Some ideas that come to my mind:
* As you said, a configuration param to enable/disable tracing of artifact transfer
* A duration threshold to create artifact transfer spans
* Looking at aggregation

PR #1875, opened 2025-05-04
add option to record transferred artifacts

smurf667:
**Description:**

This adds a `TransferListener` to the `maven-extension` so that artifact transfers can be recorded optionally.

**Existing Issue(s):**

See #1874 

**Testing:**

DRAFT: No additional testing has been added yet. 

**Documentation:**

DRAFT: No additional documentation has been added yet.

**Outstanding items:**

- See above: Additional tests and documentation still need to be done.
- Filtering of transfers below a certain threshold (size, duration) is not implemented yet; I tried to add a custom `SpanExporter` in `OpenTelemetrySdkService` - is see it getting called and filtering, but when I used the Jaeger UI setup (see [blog](https://maarten.mulders.it/2024/03/measure-your-maven-build/#option-3-maven-opentelemetry-extension)) all spans were unexpectedly shown. The issue appears that the span must be created first (because the transfer events define start and end), but those not matching the thresholds should be ignored, and I don't know how to do that other than with the exporter.
- I am not deeply familiar with the OpenTelemetry API nor the Maven APIs, so some things may have to be rectified.
- There are some TODOs that should be addressed.


cyrille-leclerc:
Thanks for this great contribution.
I would like to discuss a few idas layed out on https://github.com/smurf667/opentelemetry-java-contrib/pull/1

cyrille-leclerc:
@smurf667 can you share why you considered adding a "wrapping transfer span".
Not that I didn't yet test a build starting with an empty MAven local repo to understand when the first fransfer spans would take place

maven-extension/src/main/java/io/opentelemetry/maven/OtelLifecycleParticipant.java
```diff
@@ -43,6 +58,10 @@ public final class OtelLifecycleParticipant extends AbstractMavenLifecyclePartic
    */
   @Override
   public void afterProjectsRead(MavenSession session) {
+    registerExecutionListener(session);
```
cyrille-leclerc, line 61: Why did we change something here? I didn't expect this PR to touch the execution listener. Did I miss something? I'll look at the code.

maven-extension/src/main/java/io/opentelemetry/maven/OtelTransferListener.java
```diff
@@ -0,0 +1,161 @@
...
+                MavenOtelSemanticAttributes.MAVEN_TRANSFER_TYPE, event.getRequestType().name());
+
+    repositoryUriMapping
+        .computeIfAbsent(
+            event.getResource().getRepositoryUrl(),
+            str -> {
+              try {
+                return str.isEmpty() ? Optional.empty() : Optional.of(new URI(str));
+              } catch (URISyntaxException e) {
+                return Optional.empty();
```
cyrille-leclerc, line 97: Super smart! I was wondering how to track parsing failures :-)

maven-extension/src/main/java/io/opentelemetry/maven/OtelTransferListener.java
```diff
@@ -0,0 +1,161 @@
...
+              }
+            })
+        .ifPresent(
+            uri -> {
+              spanBuilder.setAttribute(ServerAttributes.SERVER_ADDRESS, uri.getHost());
+              if (uri.getPort() != -1) {
+                spanBuilder.setAttribute(ServerAttributes.SERVER_PORT, uri.getPort());
+              }
+              // prevent ever increasing size
+              if (repositoryUriMapping.size() > 128) {
```
cyrille-leclerc, line 107: Very smart

maven-extension/src/main/java/io/opentelemetry/maven/OtelTransferListener.java
```diff
@@ -0,0 +1,161 @@
...
+
+  @Override
+  public void transferCorrupted(TransferEvent event) {
+    logger.info("OpenTelemetry: OtelTransferListener#transferCorrupted({})", event);
+
+    Optional.ofNullable(spanRegistry.removeSpan(event)).ifPresent(span -> fail(span, event));
+  }
+
+  void finish(Span span, TransferEvent event) {
+    switch (event.getRequestType()) {
```
cyrille-leclerc, line 141: @smurf667 did you identify a problem in this logic? I guess that the `GET_EXISTENCE` / `HTTP HEAD` will have a transfer size of `0` but who knows.

maven-extension/src/main/java/io/opentelemetry/maven/OtelTransferListener.java
```diff
@@ -0,0 +1,161 @@
...
+    String urlTemplate =
+        event.getResource().getRepositoryUrl()
+            + "$groupId/$artifactId/$version/$artifactId-$version.$classifier";
+
+    String spanName = httpRequestMethod + " " + urlTemplate;
+
+    SpanBuilder spanBuilder =
+        this.openTelemetrySdkService
+            .getTracer()
+            .spanBuilder(spanName)
```
cyrille-leclerc, line 80: We could add a comment indicating we chose to create "http client spans" assuming the underlying http lib is not instrumented (span kind client...). 
If we instrument the http lib, then we will have to change this spn type.

smurf667:
I just noticed that transfers were happening outside of goals (fair enough) and though it would be helpful to group them in this "wrapping transfer span". There is not concrete need to do that, and we can also leave them "unrooted".

I would like to hear your thoughts and the potential filtering of transfers based on duration and size (e.g. skip fast/small transfers to avoid clutter) - since the transfer is started by one event and only ended by another event, one would have to keep the span in memory, deferred until one could decide if it was worth to record. Sounds tricky.

Also, right now, there are no unit tests; not sure what would be reasonable here?

maven-extension/src/main/java/io/opentelemetry/maven/OtelTransferListener.java
cyrille-leclerc: Fix proposed in:
* https://github.com/smurf667/opentelemetry-java-contrib/pull/2

cyrille-leclerc (approved):
Thanks @smurf667 , it's a great improvement! 
I'm wondering if we could later introduce threshold based creation of transfer spans like "create a transfer span if transfer.duration > xxx ms". Let's see what's the community feedback.

smurf667:
Found two TODOs, which I removed. I think it's good now; I assume after review & approval the commits will be squashed.
If just not called `span.end()` is legal, then a size threshold could be easily implemented; not so sure about a time threshold. Maybe it can start like it is now, and if the feature is desired, it could be added later.

cyrille-leclerc:
> If just not called span.end() is legal, then a size threshold could be easily implemented; not so sure about a time threshold. Maybe it can start like it is now, and if the feature is desired, it could be added later.

I think that you are right and we may use this trick but it would be a challenge if the http library became instrumented with OTel as we would get a missing spn in the chain.

>  I think it's good now;

Looking at the build failure logs, I forgot to run `./gradlew :maven-extension:spotlessApply` on the last change I proposed 👼

trask (approved):

trask:
@smurf667 can you run `./gradlew spotlessApply` to fix CI? thanks!

cyrille-leclerc:
@smurf667 please review the suggestion to fix the build failure on the spotlessCheck.
* https://github.com/smurf667/opentelemetry-java-contrib/pull/3

PR #1890, opened 2025-05-17
Add spotless auto-fix for PRs

trask:
Basically the same as https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/13872 + https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/13879.

### DIFF
--- a/.github/workflows/auto-spotless-apply.yml
+++ b/.github/workflows/auto-spotless-apply.yml
@@ -16,73 +16,59 @@ jobs:
       contents: write
       pull-requests: write
     steps:
-      - id: download-patch
-        name: Download patch
-        uses: actions/github-script@v7.0.1
+      - name: Download patch
+        uses: actions/download-artifact@v4.3.0
         with:
-          # this script copied from
-          # https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#using-data-from-the-triggering-workflow
-          script: |
-            let allArtifacts = await github.rest.actions.listWorkflowRunArtifacts({
-               owner: context.repo.owner,
-               repo: context.repo.repo,
-               run_id: context.payload.workflow_run.id
-            });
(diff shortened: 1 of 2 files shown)


.github/workflows/auto-spotless-apply.yml
```diff
@@ -0,0 +1,109 @@
...
+        working-directory: ${{ runner.temp }}/artifacts
+        run: unzip patch.zip
+
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        if: steps.download-patch.outputs.exists == 'true'
+
+      - name: Check out PR branch
+        if: steps.download-patch.outputs.exists == 'true'
+        env:
+          GH_TOKEN: ${{ github.token }}
```
trask, line 72: This token won't retrigger workflows after updating the PR

breedx-splk: MAGIC!

.github/workflows/auto-spotless-check.yml
```diff
@@ -0,0 +1,53 @@
...
+        with:
+          cache-read-only: true
+
+      - name: Check out PR branch
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: gh pr checkout ${{ github.event.pull_request.number }}
+
+      - name: Spotless
+        run: ./gradlew spotlessApply
```
breedx-splk, line 38: nit: the workflow called `spotless-check` is the one that runs `spotlessApply` and the workflow called `spotless-apply` doesn't. It's only slightly confusing, not a huge deal.

trask: yeah, it's named more for what you see in the github U/X, I'll probably make it all more generic and rename things again later

Spotless Check runs on PRs, generates the patch file, then Spotless Apply runs afterwards and applies the patch file

PR #1891, opened 2025-05-19
Add metrics support to auth extension

psx95:
**Description:**

Feature addition - Adds metrics support for authenticated OTLP exports to GCP.
Also adds a new configuration `GOOGLE_OTEL_AUTH_TARGET_SIGNALS` to allow users to select signals to which authentication should be applied to. 

**Testing:**

 - Adds unit tests (focussed only on metrics).
 - Manually tested the extension with OTLP traces & metrics combined.

**Documentation:**

 - Updated the README. 

**Outstanding items:**

 - Update documentation.
 - Fix broken integration test.

### DIFF
--- a/gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
+++ b/gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
@@ -132,14 +132,11 @@ public class GcpAuthAutoConfigurationCustomizerProvider
   // This method evaluates if the endpoint provided by the user is a known GCP telemetry endpoint.
   private static boolean isKnownGcpTelemetryEndpoint(String endpoint) {
     String knownBaseEndpointRegex = "^https://telemetry\\.googleapis\\.com(?:[:/].*)?$";
-    String knownBaseSandboxEndpoint =
-        "^https://staging-telemetry\\.sandbox\\.googleapis\\.com(?:[:/].*)?$";
     String knownRegionalizedEndpointRegex =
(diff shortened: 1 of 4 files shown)


gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
```diff
@@ -79,6 +105,34 @@ public int order() {
...
+    if (isSignalTargeted(SIGNAL_TYPE_METRICS)) {
+      return addAuthorizationHeaders(exporter, credentials);
+    }
+    return exporter;
+  }
+
+  // Checks if the auth extension is configured to target the passed signal for authentication.
+  private static boolean isSignalTargeted(String checkSignal) {
+    String userSpecifiedTargetedSignals =
+        ConfigurableOption.GOOGLE_OTEL_AUTH_TARGET_SIGNALS.getConfiguredValueWithFallback(
```
laurit, line 127: Any reason why you didn't use `configProperties` to read the configuration? It is possible to add additional properties to `configProperties` like for example agent allows adding configuration in a properties file https://opentelemetry.io/docs/zero-code/java/agent/configuration/#configuration-file or the otel spring starter adds spring properties. By reading system properties and env yourself your extension won't play nice with these additional property soruces.

psx95: That's a good idea. This was an oversight, I wasn't thinking of use-cases like declarative configs when this extension was first added.

I'll tackle this in #1903 to keep this PR's scope limited.

PR #1904, opened 2025-05-20
Update ConfigurableOptions to read from ConfigProperties

### LINKED ISSUE #1903: Read extension configuration from ConfigProperties
### Component(s)

gcp-auth-extension

## Description

The GCP Auth extension currently reads properties manually using `System.getenv` and `System.getProperty`. The extension should instead rely on reading the configuration from `ConfigProperties` instead to ensure it is supported by declarative config.

The auth extension's implementation leverages OpenTelemetry `ConfigProperties` for configuration.

_No response_

### Additional context

_No response_

psx95:
**Changes**

 - Updates the GCP auth config to read from ConfigProperties instead of directly reading environment variables or system properties.
 - Adds log statements informing users of possible misconfiguration of the `GOOGLE_OTEL_AUTH_TARGET_SIGNALS`. 

**Existing Issue(s):**

Fixes #1903 

**Testing:**

 - Existing unit tests pass indicating that reading from `ConfigProperties` works as expected.
 - Manually tested this extension with sample application to ensure successful exports.

**Documentation:**

 - Internal implementation change only, no documentation updates needed.

**Outstanding items:**
 - ~Manual testing post merge of #1891 .~


gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/ConfigurableOption.java
```diff
@@ -60,7 +61,7 @@ public enum ConfigurableOption {
     this.userReadableName = userReadableName;
     this.environmentVariableName = this.name();
     this.systemPropertyName =
-        this.environmentVariableName.toLowerCase(Locale.ENGLISH).replace('_', '.');
```
jsuereth, line 63: Capitalization and i18n are achilles' heel of string manipulation.

I think *either* using ENGLISH or ROOT will lead to inconsistency and weirdness for users.  What is your goal for support here and why this change?

aabmass (approved):

gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/ConfigurableOption.java
psx95: I made this change to make it consistent with the [internal implementation of normalizing  environment variable keys](https://github.com/open-telemetry/opentelemetry-java/blob/90e030f213547a0116a6397455a046cf4019baf9/api/all/src/main/java/io/opentelemetry/api/internal/ConfigUtil.java#L69) within [the autoconfigure module](https://github.com/open-telemetry/opentelemetry-java/blob/90e030f213547a0116a6397455a046cf4019baf9/sdk-extensions/autoconfigure-spi/src/main/java/io/opentelemetry/sdk/autoconfigure/spi/internal/DefaultConfigProperties.java#L70). 

I was considering using the implementation directly, but decided against it since the classes are marked internal and not meant for "public" use.

jsuereth: I'd check with Java maintainers, but I'd actually encourage re-use if possible.  I'd rather have all i18n issues around capitlalization of env variables fixed in one place if we can...

psx95: @laurit, @trask  do you have a opinion here - should I just re-use the [normalizeEnvironmentVariableKey](https://github.com/open-telemetry/opentelemetry-java/blob/90e030f213547a0116a6397455a046cf4019baf9/api/all/src/main/java/io/opentelemetry/api/internal/ConfigUtil.java#L69) from the internal `ConfigUtil` of the SDK instead ?

trask: we have a (long-term) goal to reduce / eliminate our cross-artifact usage of internal APIs (see section "Various artifacts make use of shared internal code" in https://github.com/open-telemetry/opentelemetry-java/issues/6970), so I'd prefer to copy here, but let's give @jack-berg the final word 😄

psx95: Looking at arguments presented by @jack-berg in https://github.com/open-telemetry/opentelemetry-java/issues/6970, I feel ok to leave the duplicated code in this artifact. 

However, we could consider promoting `ConfigUtil` class to publish an API surface for public consumption sometime in the future, especially if there are other artifacts consuming the functions in that class.

Issue #1907, opened 2025-05-23
[disk-buffering] Duplicate export in Windows


jigarkb:
### Component(s)

disk-buffering

## Description

On Windows systems, the `exportStoredBatch` method returns `true` when a file exists in the buffer directory. However, once the file is fully read, the subsequent `file.delete()` call in `cleanUp()` fails silently—returning `false` but not handled by the code. As a result, the same file is repeatedly re-exported on each `exportStoredBatch` call until a new file is written, leading to redundant exports and significant load on backend APIs.

This behavior appears linked to how Windows handles file deletion. After `file.delete()` is called, the file’s metadata may reflect 0KB, but the contents persist until a subsequent I/O operation (e.g., file creation) triggers actual cleanup. This suggests possible lazy deletion by the Windows OS, though further investigation at the kernel level would be needed to confirm.

Reference: https://github.com/open-telemetry/opentelemetry-java-contrib/blob/v1.39.0/disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/files/ReadableFile.java#L124-L127 

## Steps to Reproduce
Deploy your code on Windows with `debugEnabled` set to true, and you'll consistently see the message `Now exporting batch of ...` repeated.

## Expected Result
The file must be deleted and should not be re-exported.

## Actual Result
The file is repeatedly processed and exported.

### Component version

v1.39.0

zeitlinger:
from @jigarkb 

Also, after diving deep into the code, I found that the FolderManager is not shared between FromDiskExporter and ToDiskExporter so I am not sure we will ever hit this [code block](https://github.com/open-telemetry/opentelemetry-java-contrib/blob/be32409fe07b4ec94f78c5b7919d864451e6b56d/disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/FolderManager.java#L80-L84). Let me know if I am missing something.

PR #1912, opened 2025-05-27
Shared storage

zeitlinger:
Possible fix for https://github.com/open-telemetry/opentelemetry-java-contrib/issues/1907

WIP - only covered spans so far


disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/SpanToDiskExporter.java
```diff
@@ -31,6 +32,7 @@ public class SpanToDiskExporter implements SpanExporter {
    * @return A new SpanToDiskExporter instance.
    * @throws IOException if the delegate ToDiskExporter could not be created.
    */
+  @Deprecated
   public static SpanToDiskExporter create(SpanExporter delegate, StorageConfiguration config)
```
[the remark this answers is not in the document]
LikeTheSalad: Makes sense 👍 I think the `getBody()` one was causing issues before because it's defined upstream (maybe those are sorted now), but apart from that, I don't see why we should keep deprecated methods for now.

zeitlinger:
@LikeTheSalad please review

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/LogRecordFromDiskExporter.java
```diff
@@ -19,15 +18,12 @@ public class LogRecordFromDiskExporter implements FromDiskExporter {
 
   private final FromDiskExporterImpl<LogRecordData> delegate;
 
-  public static LogRecordFromDiskExporter create(
-      LogRecordExporter exporter, StorageConfiguration config) throws IOException {
+  public static LogRecordFromDiskExporter create(LogRecordExporter exporter, Storage storage)
+      throws IOException {
     FromDiskExporterImpl<LogRecordData> delegate =
-        FromDiskExporterImpl.<LogRecordData>builder()
-            .setFolderName(SignalTypes.logs.name())
```
LikeTheSalad, line 26: It's been a while since I worked on this, so apologies because I didn't notice it earlier, though I just recalled that each signal's set of files needs to be in its own folder, separated from other signal files, and that's why this property was set, and also why we needed a storage object per signal. I know from a first glance it looks strange, as I also was confused for a moment, and I think it's because the current architecture is not straightforward. I'm planning to work on enhancing it soon to make it cleaner, but for now, I'm afraid that if we merge these changes, it can cause issues, especially around serialization.

In theory, we could share the same storage object across a single signal's `ToDiskExporter` and `FromDiskExporter` objects, however, those are instantiated independently by the consumer, so to make it work we would have to create some sort of factory that focuses on providing both `ToDiskExporter` and `FromDiskExporter` instances for a single signal where we would internally ensure that the same storage object is used. And then we would have to make these `create` methods package private to prevent users from calling them, so that they don't provide the wrong storage object. Though I'm not sure if that would add much value compared to the current way of using the lib.

zeitlinger: Good point - I changed the PR so that the storage builder takes the signal type as an argument.
This should make clear that the storage should not be used across signal types.

LikeTheSalad (approved):
Looks good to me 👍

trask (approved):

PR #1918, opened 2025-05-28
Implementing ExtendedLogRecordData

LikeTheSalad:
These changes are meant as a temporary workaround for [this issue](https://github.com/open-telemetry/opentelemetry-java/issues/7363).

We need to decide whether we are going to add support for extended attributes in disk buffering while they are incubating. However, in the meantime these changes will prevent the issue mentioned above where logs are completely broken due to the cast issue.


laurit (approved):
I think it is fine for a temporary workaround.

PR #1919, opened 2025-05-28
Simplify auto spotless workflow

trask:
Also should fix the edge case where PR is coming from `main` branch on fork: https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1855#issuecomment-2916665955

The problem before was that there was already a `main` branch checked out locally, but now it checks out remote branch directly.


laurit (approved):
I have a vague memory that as a maintainer I was not allowed to push into the main branch of the fork. If that is true then probably this won't work anyway.

trask:
> I have a vague memory that as a maintainer I was not allowed to push into the main branch of the fork. If that is true then probably this won't work anyway.

I just tested pushing to #1855 and seems to work: https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1855#event-17873060265

Issue #1902, opened 2025-05-20
Using closeable in opamp client api


LikeTheSalad:
Based on this comment:

https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1835#discussion_r2097120696_

PR #1928, opened 2025-05-30
OpAMP HTTP service

LikeTheSalad:
**Description:**

This is a continuation of the OpAMP implementation work. The changes here are focused on adding the RequestService interface along with its HTTP implementation.

The RequestService interface serves as an abstraction layer that accommodates both HTTP and WebSocket protocols, so that the OpampClient implementation doesn't have to know which protocol is used.

### DIFF
--- a/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/HttpErrorException.java
+++ b/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/HttpErrorException.java
@@ -1,0 +1,36 @@
+package io.opentelemetry.opamp.client.internal.connectivity.http;
+
+import java.util.Objects;
+
+@SuppressWarnings("serial")
+public class HttpErrorException extends Exception {
+  public final int errorCode;
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (!(o instanceof HttpErrorException)) {
+      return false;
+    }
+    HttpErrorException that = (HttpErrorException) o;
+    return errorCode == that.errorCode && Objects.equals(getMessage(), that.getMessage());
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(errorCode, getMessage());
+  }
+
+  /**
+   * Constructs an HTTP error related exception.
+   *
+   * @param errorCode The HTTP error code.
+   * @param message The HTTP error message associated with the code.
+   */
+  public HttpErrorException(int errorCode, String message) {
+    super(message);
+    this.errorCode = errorCode;
+  }
+}
--- a/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/HttpErrorException.java
+++ b/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/HttpErrorException.java
@@ -1,3 +1,8 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
 package io.opentelemetry.opamp.client.internal.connectivity.http;
 
 import java.util.Objects;
--- a/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/OkHttpSender.java
+++ b/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/OkHttpSender.java
@@ -12,7 +12,7 @@ import okio.BufferedSink;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
-public class OkHttpSender implements HttpSender {
+public final class OkHttpSender implements HttpSender {
   private final OkHttpClient client;
   private final String url;
 
--- a/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/OkHttpSender.java
+++ b/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/OkHttpSender.java
@@ -1,0 +1,124 @@
+package io.opentelemetry.opamp.client.internal.connectivity.http;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.util.concurrent.CompletableFuture;
+import java.util.function.Consumer;
+import okhttp3.MediaType;
+import okhttp3.OkHttpClient;
+import okhttp3.RequestBody;
+import okio.BufferedSink;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class OkHttpSender implements HttpSender {
+  private final OkHttpClient client;
+  private final String url;
+
+  public static OkHttpSender create(String url) {
+    return create(url, new OkHttpClient());
+  }
+
+  public static OkHttpSender create(String url, OkHttpClient client) {
+    return new OkHttpSender(url, client);
+  }
+
+  private OkHttpSender(String url, OkHttpClient client) {
+    this.url = url;
+    this.client = client;
+  }
+
+  @Override
+  public CompletableFuture<Response> send(Consumer<OutputStream> writer, int contentLength) {
+    CompletableFuture<Response> future = new CompletableFuture<>();
+    okhttp3.Request.Builder builder = new okhttp3.Request.Builder().url(url);
+    String contentType = "application/x-protobuf";
+    builder.addHeader("Content-Type", contentType);
+
+    RequestBody body = new RawRequestBody(writer, contentLength, MediaType.parse(contentType));
+    builder.post(body);
+
+    try {
+      okhttp3.Response response = client.newCall(builder.build()).execute();
+      if (response.isSuccessful()) {
+        if (response.body() != null) {
+          future.complete(new OkHttpResponse(response));
+        }
+      } else {
+        future.completeExceptionally(new HttpErrorException(response.code(), response.message()));
+      }
+    } catch (IOException e) {
+      future.completeExceptionally(e);
+    }
+
+    future.completeExceptionally(new IllegalStateException());
+
+    return future;
+  }
+
+  private static class OkHttpResponse implements Response {
+    private final okhttp3.Response response;
+
+    private OkHttpResponse(okhttp3.Response response) {
+      this.response = response;
+    }
+
+    @Override
+    public int statusCode() {
+      return response.code();
+    }
+
+    @Override
+    public String statusMessage() {
+      return response.message();
+    }
+
+    @Override
+    public InputStream bodyInputStream() {
+      if (response.body() != null) {
+        return response.body().byteStream();
+      }
+      throw new NullPointerException();
+    }
+
+    @Override
+    public String getHeader(String name) {
+      return response.headers().get(name);
+    }
+
+    @Override
+    public void close() {
+      response.close();
+    }
+  }
+
+  private static class RawRequestBody extends RequestBody {
+    private final Consumer<OutputStream> writer;
+    private final int contentLength;
+    private final MediaType contentType;
+
+    private RawRequestBody(
+        Consumer<OutputStream> writer, int contentLength, MediaType contentType) {
+      this.writer = writer;
+      this.contentLength = contentLength;
+      this.contentType = contentType;
+    }
+
+    @Nullable
+    @Override
+    public MediaType contentType() {
+      return contentType;
+    }
+
+    @Override
+    public long contentLength() {
+      return contentLength;
+    }
+
+    @Override
+    public void writeTo(@NotNull BufferedSink bufferedSink) {
+      writer.accept(bufferedSink.outputStream());
+    }
+  }
+}
--- a/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/OkHttpSender.java
+++ b/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/OkHttpSender.java
@@ -1,3 +1,8 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
 package io.opentelemetry.opamp.client.internal.connectivity.http;
 
 import java.io.IOException;
--- a/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/RetryAfterParser.java
+++ b/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/RetryAfterParser.java
@@ -1,3 +1,8 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
 package io.opentelemetry.opamp.client.internal.connectivity.http;
 
 import io.opentelemetry.opamp.client.internal.tools.SystemTime;
--- a/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/RetryAfterParser.java
+++ b/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/RetryAfterParser.java
@@ -1,0 +1,44 @@
+package io.opentelemetry.opamp.client.internal.connectivity.http;
+
+import io.opentelemetry.opamp.client.internal.tools.SystemTime;
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.time.Duration;
+import java.util.Locale;
+import java.util.regex.Pattern;
+
+public final class RetryAfterParser {
+  private final SystemTime systemTime;
+  public static final Pattern SECONDS_PATTERN = Pattern.compile("\\d+");
+  public static final Pattern DATE_PATTERN =
+      Pattern.compile(
+          "^([A-Za-z]{3}, [0-3][0-9] [A-Za-z]{3} [0-9]{4} [0-2][0-9]:[0-5][0-9]:[0-5][0-9] GMT)$");
+  private static final SimpleDateFormat dateFormat =
+      new SimpleDateFormat("EEE, dd MMM yyyy HH:mm:ss z", Locale.US);
+
+  public static RetryAfterParser getInstance() {
+    return new RetryAfterParser(SystemTime.getInstance());
+  }
+
+  RetryAfterParser(SystemTime systemTime) {
+    this.systemTime = systemTime;
+  }
+
+  public Duration parse(String value) {
+    if (SECONDS_PATTERN.matcher(value).matches()) {
+      return Duration.ofSeconds(Long.parseLong(value));
+    } else if (DATE_PATTERN.matcher(value).matches()) {
+      return Duration.ofMillis(toMilliseconds(value) - systemTime.getCurrentTimeMillis());
+    }
+    throw new IllegalArgumentException("Invalid Retry-After value: " + value);
+  }
+
+  @SuppressWarnings({"JavaUtilDate", "ThrowSpecificExceptions"})
+  private static long toMilliseconds(String value) {
+    try {
+      return dateFormat.parse(value).getTime();
+    } catch (ParseException e) {
+      throw new RuntimeException(e);
+    }
+  }
+}
--- a/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/delay/AcceptsDelaySuggestion.java
+++ b/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/delay/AcceptsDelaySuggestion.java
@@ -1,0 +1,31 @@
+/*
+ * Licensed to Elasticsearch B.V. under one or more contributor
+ * license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright
+ * ownership. Elasticsearch B.V. licenses this file to you under
+ * the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *	http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package io.opentelemetry.opamp.client.internal.request.delay;
+
+import java.time.Duration;
+
+/**
+ * A {@link PeriodicDelay} implementation that wants to accept delay time suggestions, as explained
+ * <a
+ * href="https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#throttling">here</a>,
+ * must implement this interface.
+ */
+public interface AcceptsDelaySuggestion {
+  void suggestDelay(Duration delay);
+}
--- a/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/delay/AcceptsDelaySuggestion.java
+++ b/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/delay/AcceptsDelaySuggestion.java
@@ -1,21 +1,3 @@
-/*
- * Licensed to Elasticsearch B.V. under one or more contributor
- * license agreements. See the NOTICE file distributed with
- * this work for additional information regarding copyright
- * ownership. Elasticsearch B.V. licenses this file to you under
- * the Apache License, Version 2.0 (the "License"); you may
- * not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- *	http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing,
- * software distributed under the License is distributed on an
- * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
- * KIND, either express or implied.  See the License for the
- * specific language governing permissions and limitations
- * under the License.
- */
 package io.opentelemetry.opamp.client.internal.request.delay;
 
 import java.time.Duration;
--- a/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/delay/AcceptsDelaySuggestion.java
+++ b/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/delay/AcceptsDelaySuggestion.java
@@ -1,3 +1,8 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
 package io.opentelemetry.opamp.client.internal.request.delay;
 
 import java.time.Duration;
--- a/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/delay/PeriodicTaskExecutor.java
+++ b/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/delay/PeriodicTaskExecutor.java
@@ -1,0 +1,77 @@
+package io.opentelemetry.opamp.client.internal.request.delay;
+
+import java.util.Objects;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+public final class PeriodicTaskExecutor {
+  private final ScheduledExecutorService executorService;
+  private final Lock delaySetLock = new ReentrantLock();
+  private final AtomicReference<Runnable> periodicTask = new AtomicReference<>();
+  private PeriodicDelay periodicDelay;
+  @Nullable private ScheduledFuture<?> scheduledFuture;
+
+  public static PeriodicTaskExecutor create(PeriodicDelay initialPeriodicDelay) {
+    return new PeriodicTaskExecutor(
+        Executors.newSingleThreadScheduledExecutor(), initialPeriodicDelay);
+  }
+
+  PeriodicTaskExecutor(
+      ScheduledExecutorService executorService, PeriodicDelay initialPeriodicDelay) {
+    this.executorService = executorService;
+    this.periodicDelay = initialPeriodicDelay;
+  }
+
+  public void start(@Nonnull Runnable periodicTask) {
+    this.periodicTask.set(periodicTask);
+    scheduleNext();
+  }
+
+  public void executeNow() {
+    executorService.execute(periodicTask.get());
+  }
+
+  public void setPeriodicDelay(PeriodicDelay periodicDelay) {
+    delaySetLock.lock();
+    try {
+      if (scheduledFuture != null) {
+        scheduledFuture.cancel(false);
+      }
+      this.periodicDelay = periodicDelay;
+      periodicDelay.reset();
+      scheduleNext();
+    } finally {
+      delaySetLock.unlock();
+    }
+  }
+
+  public void stop() {
+    executorService.shutdown();
+  }
+
+  private void scheduleNext() {
+    delaySetLock.lock();
+    try {
+      scheduledFuture =
+          executorService.schedule(
+              new PeriodicRunner(), periodicDelay.getNextDelay().toNanos(), TimeUnit.NANOSECONDS);
+    } finally {
+      delaySetLock.unlock();
+    }
+  }
+
+  private class PeriodicRunner implements Runnable {
+    @Override
+    public void run() {
+      Objects.requireNonNull(periodicTask.get()).run();
+      scheduleNext();
+    }
+  }
+}
--- a/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/delay/PeriodicTaskExecutor.java
+++ b/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/delay/PeriodicTaskExecutor.java
@@ -1,3 +1,8 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
 package io.opentelemetry.opamp.client.internal.request.delay;
 
 import java.util.Objects;
--- a/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
+++ b/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
@@ -26,12 +26,10 @@ public final class HttpRequestService implements RequestService, Runnable {
   private final PeriodicTaskExecutor executor;
   private final PeriodicDelay periodicRequestDelay;
   private final PeriodicDelay periodicRetryDelay;
-  private final Object runningLock = new Object();
   private final AtomicBoolean retryModeEnabled = new AtomicBoolean(false);
+  private final AtomicBoolean isRunning = new AtomicBoolean(false);
   @Nullable private Callback callback;
   @Nullable private Supplier<Request> requestSupplier;
-  private boolean isRunning = false;
-  private boolean isStopped = false;
   public static final PeriodicDelay DEFAULT_DELAY_BETWEEN_REQUESTS =
       PeriodicDelay.ofFixedDuration(Duration.ofSeconds(30));
 
@@ -75,27 +73,18 @@ public final class HttpRequestService implements RequestService, Runnable {
 
   @Override
   public void start(Callback callback, Supplier<Request> requestSupplier) {
-    synchronized (runningLock) {
-      if (isStopped) {
-        throw new IllegalStateException("RequestDispatcher has been stopped");
-      }
-      if (isRunning) {
-        throw new IllegalStateException("RequestDispatcher is already running");
-      }
+    if (isRunning.compareAndSet(false, true)) {
       this.callback = callback;
       this.requestSupplier = requestSupplier;
       executor.start(this);
-      isRunning = true;
+    } else {
+      throw new IllegalStateException("RequestDispatcher is already running");
     }
   }
 
   @Override
   public void stop() {
-    synchronized (runningLock) {
-      if (!isRunning || isStopped) {
-        return;
-      }
-      isStopped = true;
+    if (isRunning.compareAndSet(true, false)) {
       executor.executeNow();
       executor.stop();
     }
@@ -165,7 +154,7 @@ public final class HttpRequestService implements RequestService, Runnable {
       String retryAfterHeader = response.getHeader("Retry-After");
       Duration retryAfter = null;
       if (retryAfterHeader != null) {
-        // retryAfter = TODO parse header to duration
+        retryAfter = Duration.ofSeconds(Long.parseLong(retryAfterHeader));
       }
       enableRetryMode(retryAfter);
     }
--- a/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
+++ b/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
@@ -2,6 +2,7 @@ package io.opentelemetry.opamp.client.internal.request.service;
 
 import io.opentelemetry.opamp.client.internal.connectivity.http.HttpErrorException;
 import io.opentelemetry.opamp.client.internal.connectivity.http.HttpSender;
+import io.opentelemetry.opamp.client.internal.connectivity.http.RetryAfterParser;
 import io.opentelemetry.opamp.client.internal.request.Request;
 import io.opentelemetry.opamp.client.internal.request.delay.AcceptsDelaySuggestion;
 import io.opentelemetry.opamp.client.internal.request.delay.PeriodicDelay;
@@ -28,6 +29,7 @@ public final class HttpRequestService implements RequestService, Runnable {
   private final PeriodicDelay periodicRetryDelay;
   private final AtomicBoolean retryModeEnabled = new AtomicBoolean(false);
   private final AtomicBoolean isRunning = new AtomicBoolean(false);
+  private final RetryAfterParser retryAfterParser;
   @Nullable private Callback callback;
   @Nullable private Supplier<Request> requestSupplier;
   public static final PeriodicDelay DEFAULT_DELAY_BETWEEN_REQUESTS =
@@ -57,18 +59,21 @@ public final class HttpRequestService implements RequestService, Runnable {
         requestSender,
         PeriodicTaskExecutor.create(periodicRequestDelay),
         periodicRequestDelay,
-        periodicRetryDelay);
+        periodicRetryDelay,
+        RetryAfterParser.getInstance());
   }
 
   HttpRequestService(
       HttpSender requestSender,
       PeriodicTaskExecutor executor,
       PeriodicDelay periodicRequestDelay,
-      PeriodicDelay periodicRetryDelay) {
+      PeriodicDelay periodicRetryDelay,
+      RetryAfterParser retryAfterParser) {
     this.requestSender = requestSender;
     this.executor = executor;
     this.periodicRequestDelay = periodicRequestDelay;
     this.periodicRetryDelay = periodicRetryDelay;
+    this.retryAfterParser = retryAfterParser;
   }
 
   @Override
@@ -154,7 +159,7 @@ public final class HttpRequestService implements RequestService, Runnable {
       String retryAfterHeader = response.getHeader("Retry-After");
       Duration retryAfter = null;
       if (retryAfterHeader != null) {
-        retryAfter = Duration.ofSeconds(Long.parseLong(retryAfterHeader));
+        retryAfter = retryAfterParser.parse(retryAfterHeader);
       }
       enableRetryMode(retryAfter);
     }
--- a/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
+++ b/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
@@ -1,3 +1,8 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
 package io.opentelemetry.opamp.client.internal.request.service;
 
 import io.opentelemetry.opamp.client.internal.connectivity.http.HttpErrorException;
--- a/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
+++ b/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
@@ -1,0 +1,222 @@
+package io.opentelemetry.opamp.client.internal.request.service;
+
+import io.opentelemetry.opamp.client.internal.connectivity.http.HttpErrorException;
+import io.opentelemetry.opamp.client.internal.connectivity.http.HttpSender;
+import io.opentelemetry.opamp.client.internal.request.Request;
+import io.opentelemetry.opamp.client.internal.request.delay.AcceptsDelaySuggestion;
+import io.opentelemetry.opamp.client.internal.request.delay.PeriodicDelay;
+import io.opentelemetry.opamp.client.internal.request.delay.PeriodicTaskExecutor;
+import io.opentelemetry.opamp.client.internal.response.Response;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.time.Duration;
+import java.util.Objects;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Consumer;
+import java.util.function.Supplier;
+import javax.annotation.Nullable;
+import opamp.proto.AgentToServer;
+import opamp.proto.ServerErrorResponse;
+import opamp.proto.ServerErrorResponseType;
+import opamp.proto.ServerToAgent;
+
+public final class HttpRequestService implements RequestService, Runnable {
+  private final HttpSender requestSender;
+  private final PeriodicTaskExecutor executor;
+  private final PeriodicDelay periodicRequestDelay;
+  private final PeriodicDelay periodicRetryDelay;
+  private final Object runningLock = new Object();
+  private final AtomicBoolean retryModeEnabled = new AtomicBoolean(false);
+  @Nullable private Callback callback;
+  @Nullable private Supplier<Request> requestSupplier;
+  private boolean isRunning = false;
+  private boolean isStopped = false;
+  public static final PeriodicDelay DEFAULT_DELAY_BETWEEN_REQUESTS =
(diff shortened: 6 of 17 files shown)


LikeTheSalad:
cc @tigrannajaryan

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/HttpErrorException.java
```diff
@@ -0,0 +1,41 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.opamp.client.internal.connectivity.http;
+
+import java.util.Objects;
+
+@SuppressWarnings("serial")
```
laurit: usually we just add `private static final long serialVersionUID = 1L;` instead of suppressing the warning

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/HttpErrorException.java
```diff
@@ -0,0 +1,41 @@
...
+      return false;
+    }
+    HttpErrorException that = (HttpErrorException) o;
+    return errorCode == that.errorCode && Objects.equals(getMessage(), that.getMessage());
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(errorCode, getMessage());
+  }
```
laurit: implementing `equals` and `hashCode` for exceptions is unusual, is it really useful?

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/HttpErrorException.java
```diff
@@ -0,0 +1,41 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.opamp.client.internal.connectivity.http;
+
+import java.util.Objects;
+
+@SuppressWarnings("serial")
+public class HttpErrorException extends Exception {
+  public final int errorCode;
```
laurit: it is more common to have a private field with a public getter

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/OkHttpSender.java
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  @Override
+  public CompletableFuture<Response> send(Consumer<OutputStream> writer, int contentLength) {
+    CompletableFuture<Response> future = new CompletableFuture<>();
+    okhttp3.Request.Builder builder = new okhttp3.Request.Builder().url(url);
+    String contentType = "application/x-protobuf";
+    builder.addHeader("Content-Type", contentType);
+
+    RequestBody body = new RawRequestBody(writer, contentLength, MediaType.parse(contentType));
```
laurit: `MediaType.parse(application/x-protobuf")` could be a constant

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/OkHttpSender.java
```diff
@@ -0,0 +1,129 @@
...
+    builder.post(body);
+
+    try {
+      okhttp3.Response response = client.newCall(builder.build()).execute();
```
laurit: perhaps it would be better to use `enqueue(Callback)` here

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/OkHttpSender.java
```diff
@@ -0,0 +1,129 @@
...
+    public String statusMessage() {
+      return response.message();
+    }
+
+    @Override
+    public InputStream bodyInputStream() {
+      if (response.body() != null) {
+        return response.body().byteStream();
+      }
+      throw new NullPointerException();
```
laurit: imo `IllegalStateException` would be a better fit for an unexpected state (as far as I can tell this object is not constructed with null body). It might be better to move this validation into the constructor to make it easier to see that the null body is not expected.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/RetryAfterParser.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.Locale;
+import java.util.regex.Pattern;
+
+public final class RetryAfterParser {
+  private final SystemTime systemTime;
+  public static final Pattern SECONDS_PATTERN = Pattern.compile("\\d+");
+  public static final Pattern DATE_PATTERN =
+      Pattern.compile(
+          "^([A-Za-z]{3}, [0-3][0-9] [A-Za-z]{3} [0-9]{4} [0-2][0-9]:[0-5][0-9]:[0-5][0-9] GMT)$");
+  private static final SimpleDateFormat dateFormat =
```
laurit: `SimpleDateFormat` is not thread safe

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/RetryAfterParser.java
```diff
@@ -0,0 +1,49 @@
...
+  public Duration parse(String value) {
+    if (SECONDS_PATTERN.matcher(value).matches()) {
+      return Duration.ofSeconds(Long.parseLong(value));
+    } else if (DATE_PATTERN.matcher(value).matches()) {
+      return Duration.ofMillis(toMilliseconds(value) - systemTime.getCurrentTimeMillis());
+    }
+    throw new IllegalArgumentException("Invalid Retry-After value: " + value);
+  }
+
+  @SuppressWarnings({"JavaUtilDate", "ThrowSpecificExceptions"})
```
laurit: usually we throw a more specific exception such as `IllegalStateException` instead of suppressing this warning

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/RetryAfterParser.java
```diff
@@ -0,0 +1,49 @@
...
+
+  RetryAfterParser(SystemTime systemTime) {
+    this.systemTime = systemTime;
+  }
+
+  public Duration parse(String value) {
+    if (SECONDS_PATTERN.matcher(value).matches()) {
+      return Duration.ofSeconds(Long.parseLong(value));
+    } else if (DATE_PATTERN.matcher(value).matches()) {
+      return Duration.ofMillis(toMilliseconds(value) - systemTime.getCurrentTimeMillis());
```
laurit: if the server or client time is wrong this could result in negative duration, would that be ok?

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/delay/AcceptsDelaySuggestion.java
```diff
@@ -0,0 +1,18 @@
...
+
+import java.time.Duration;
+
+/**
+ * A {@link PeriodicDelay} implementation that wants to accept delay time suggestions, as explained
+ * <a
+ * href="https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#throttling">here</a>,
+ * must implement this interface.
+ */
+public interface AcceptsDelaySuggestion {
```
laurit, line 16: currently it is only implemented in tests?

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
```diff
@@ -0,0 +1,221 @@
...
+      RetryAfterParser retryAfterParser) {
+    this.requestSender = requestSender;
+    this.executor = executor;
+    this.periodicRequestDelay = periodicRequestDelay;
+    this.periodicRetryDelay = periodicRetryDelay;
+    this.retryAfterParser = retryAfterParser;
+  }
+
+  @Override
+  public void start(Callback callback, Supplier<Request> requestSupplier) {
```
laurit, line 94: The javadoc in `RequestService` says that for http `start` doesn't do anything. Here this method starts periodically sending requests which imo contradicts what the javadoc says.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
```diff
@@ -0,0 +1,221 @@
...
+  }
+
+  private void disableRetryMode() {
+    if (retryModeEnabled.compareAndSet(true, false)) {
+      executor.setPeriodicDelay(periodicRequestDelay);
+    }
+  }
+
+  @Override
+  public void sendRequest() {
```
laurit, line 116: so the purpose of this method is to manually trigger sending the request without waiting for the next periodic request?

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/delay/PeriodicTaskExecutor.java
```diff
@@ -0,0 +1,82 @@
...
+public final class PeriodicTaskExecutor {
+  private final ScheduledExecutorService executorService;
+  private final Lock delaySetLock = new ReentrantLock();
+  private final AtomicReference<Runnable> periodicTask = new AtomicReference<>();
+  private PeriodicDelay periodicDelay;
+  @Nullable private ScheduledFuture<?> scheduledFuture;
+
+  public static PeriodicTaskExecutor create(PeriodicDelay initialPeriodicDelay) {
+    return new PeriodicTaskExecutor(
+        Executors.newSingleThreadScheduledExecutor(), initialPeriodicDelay);
```
laurit: should the executor use daemon threads?

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/HttpErrorException.java
LikeTheSalad: Thanks, it's updated now.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/HttpErrorException.java
LikeTheSalad: I had to check because I didn't remember the reason, though it seems like it was only for test purposes, so I've updated the tests to not rely on these implementations.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/HttpErrorException.java
LikeTheSalad: Thanks, it's updated now.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/OkHttpSender.java
LikeTheSalad: Thanks, it's updated now.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/OkHttpSender.java
LikeTheSalad: Thanks, it's updated now.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/OkHttpSender.java
LikeTheSalad: Thanks, it's updated now.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/RetryAfterParser.java
LikeTheSalad: Thanks, I've replaced it with DateTimeFormatter.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/RetryAfterParser.java
LikeTheSalad: This is no longer needed after the simpledateformat replacement, so it's removed now.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/RetryAfterParser.java
LikeTheSalad: This is a great point. I've added validations for it and made the response optional to signal the service that it can ignore these kinds of time delay suggestions.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/delay/AcceptsDelaySuggestion.java
LikeTheSalad: Yes. I'm planning to implement it in a future `PeriodicDelay` implementation that's made for retries only. However, that's not yet here, so if you think we should wait until then to add this interface, then I'll remove it from this PR.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
LikeTheSalad: Thanks, I've updated the javadoc.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
LikeTheSalad: Yes. My understanding is that this is needed in HTTP when something in the agent changes, so that the client notifies the server right away, leaving the periodic polling mechanism more as a means to check if the server has something to send to the client (this is what I recall being my interpretation of the spec back when I created this api). And for the WebSocket impl this is needed because there is no periodic polling (unless we're asked by the server to retry connecting later), as the server can send messages to the client without waiting for it to send a request first.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/delay/PeriodicTaskExecutor.java
LikeTheSalad: I don't see why not. I guess in an ideal scenario people should remember closing the client to stop the executor, but there's no harm in making them daemon threads just in case. So I've just updated it.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/OkHttpSender.java
```diff
@@ -0,0 +1,140 @@
...
+
+    client
+        .newCall(builder.build())
+        .enqueue(
+            new Callback() {
+              @Override
+              public void onResponse(@NotNull Call call, @NotNull okhttp3.Response response) {
+                if (response.isSuccessful()) {
+                  if (response.body() != null) {
+                    future.complete(new OkHttpResponse(response));
```
laurit: If the response body is `null` now the future will never complete

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/OkHttpSender.java
```diff
@@ -0,0 +1,140 @@
...
+
+    return future;
+  }
+
+  private static class OkHttpResponse implements Response {
+    private final okhttp3.Response response;
+
+    private OkHttpResponse(okhttp3.Response response) {
+      if (response.body() == null) {
+        throw new IllegalStateException();
```
laurit:
```suggestion
        throw new IllegalStateException("null response body not expected");
```

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/RetryAfterParser.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.Locale;
+import java.util.Optional;
+import java.util.regex.Pattern;
+
+public final class RetryAfterParser {
+  private final SystemTime systemTime;
+  public static final Pattern SECONDS_PATTERN = Pattern.compile("\\d+");
+  public static final Pattern DATE_PATTERN =
+      Pattern.compile(
+          "^([A-Za-z]{3}, [0-3][0-9] [A-Za-z]{3} [0-9]{4} [0-2][0-9]:[0-5][0-9]:[0-5][0-9] GMT)$");
```
laurit: Only the `DATE_PATTERN` is surrounded with `^$` which is a bit strange as it is matched the same way as `SECONDS_PATTERN`. For consistency both regexes should either have or not have the line start/end anchors.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/OkHttpSender.java
LikeTheSalad: Good point. Thanks, I've just updated it.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/RetryAfterParser.java
LikeTheSalad: Makes sense. I've added the anchors to the seconds pattern.

laurit: I would have guessed that since `matches()` matches the entire input you don't really need the start/end markers, but really up to you to decide.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
laurit: I think you could also consider whether you need the `start` method at all. Perhaps it could be combined with create? Do you ever need to have a sender in unstarted state? If you decide to work on this it could be done in a subsequent PR.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
```diff
@@ -0,0 +1,225 @@
...
+      this.data = data;
+    }
+
+    @SuppressWarnings("ThrowSpecificExceptions")
+    @Override
+    public void accept(OutputStream outputStream) {
+      try {
+        outputStream.write(data);
+      } catch (IOException e) {
+        throw new RuntimeException(e);
```
laurit: You could use another exception here to avoid the need to suppress the warning. Though you could also consider whether `Consumer<OutputStream>` is what you want to use. Since `HttpSender` is an internal api and the only user of it provides `byte[]` then perhaps that is what would be the easiest, or you could create a custom `ThrowingConsumer` class that would let you avoid handling the exception here.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
```diff
@@ -0,0 +1,225 @@
...
+
+  @Override
+  public void stop() {
+    if (isRunning.compareAndSet(true, false)) {
+      executor.executeNow();
+      executor.stop();
+    }
+  }
+
+  private void enableRetryMode(@Nullable Duration suggestedDelay) {
```
laurit: Imo calling it `enableRetryMode` is misleading. The code never retries, it just sends the same request all the time (or you could say it always retries no matter whether the request succeeds or not). The only thing done here is updating the delay for the next request.
I need to think about this more, perhaps this can be simplified.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/RetryAfterParser.java
LikeTheSalad: I'm fine either way tbh, although I guess there's something to say in favor of having slightly smaller patterns for simplicity. So I've removed the unnecessary bits.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
LikeTheSalad: I'm thinking about making HttpSender non-internal for whenever this lib becomes stable, because I'm guessing some people would want to create their own implementations that won't depend on okhttp (as I'm aware of some issues we've had in core with people not linking the kotlin transitive dependency that it brings), which is why I'd like to keep it as generic as possible. So I've created an interface named `BodyWriter` that is more explicit about what it does and also throws an exception so that it's handled on the caller side.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/RequestService.java
```diff
@@ -0,0 +1,76 @@
...
+     */
+    void onConnectionSuccess();
+
+    /**
+     * For WebSocket implementations, this is called when the connection cannot be made or is lost.
+     * For HTTP implementations, this is called on every HTTP request that fails.
+     *
+     * @param throwable The detailed error.
+     */
+    void onConnectionFailed(Throwable throwable);
```
laurit: currently these are never called, do you plan to add using them in a follow up PR?

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
```diff
@@ -0,0 +1,219 @@
...
+  }
+
+  private void doSendRequest() {
+    try {
+      AgentToServer agentToServer =
+          Objects.requireNonNull(requestSupplier).get().getAgentToServer();
+
+      byte[] data = agentToServer.encodeByteString().toByteArray();
+      try (HttpSender.Response response =
+          requestSender.send(new ByteArrayWriter(data), data.length).get()) {
```
laurit: you could just use `outputStream -> outputStream.write(data)` here

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
LikeTheSalad: This is a good point, and I'd like to get your input on this because I just recalled that this is something I struggled with a lot when defining this API (and also the opampclient one, which also has a `start` method) so I'd like to get a second opinion. 

The thing is that I wanted to _somehow_ enforce the use of the callback by both the implementation and the users. And usually with a regular class, I'd just make it a constructor param, but that won't work here. So I figured that, by having to call the start method (because otherwise the whole thing wouldn't do anything), you'd be forced to follow the API design without having to first read the docs on "how to properly use it" or "how to properly implement this interface" because essentially using it properly will be required for the project to compile. I also thought about adding setters, but then people can forget about calling them... I guess I just try to make APIs in a way that assumes that people won't read the docs and that they can forget things as long as their project compiles.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/RequestService.java
LikeTheSalad: This doesn't sound right, I think they should be called right now. Let me check if I missed something when copying the code over here.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
laurit: I think one way to simplify the logic would be to have periodic requests and manually triggered requests be independent of each other. So that periodic events will fire as scheduled even when there were manual requests. Then we could have `doSendRequest` report back the next scheduled time it parsed from the response. Would that be ok? Untested code follows
```java
  @Override
  public void start(Callback callback, Supplier<Request> requestSupplier) {
    if (!isRunning.compareAndSet(false, true)) {
      throw new IllegalStateException("HttpRequestService is already running");
    }
    this.callback = callback;
    this.requestSupplier = requestSupplier;

    executorService = Executors.newSingleThreadScheduledExecutor(new DaemonThreadFactory());
    // schedule the first execution
    executorService.schedule(
        this::periodicSend, periodicRequestDelay.getNextDelay().toNanos(), TimeUnit.NANOSECONDS);
  }

  private void periodicSend() {
    AtomicReference<Duration> durationReference =
        new AtomicReference<>(periodicRequestDelay.getNextDelay());
    doSendRequest(durationReference::set);
    // schedule the next execution
    executorService.schedule(
        this::periodicSend, durationReference.get().toNanos(), TimeUnit.NANOSECONDS);
  }

  @Override
  public void sendRequest() {
    if (!isRunning.get()) {
      throw new IllegalStateException("HttpRequestService is not running");
    }

    doSendRequest(duration -> {});
  }

  private void doSendRequest(Consumer<Duration> nextExecution) {
    AgentToServer agentToServer = requestSupplier.get().getAgentToServer();

    byte[] data = agentToServer.encodeByteString().toByteArray();
    CompletableFuture<HttpSender.Response> future =
        requestSender.send(outputStream -> outputStream.write(data), data.length);
    try (HttpSender.Response response = future.get(30, TimeUnit.SECONDS)) {
      if (isSuccessful(response)) {
        handleSuccessResponse(
            Response.create(ServerToAgent.ADAPTER.decode(response.bodyInputStream())),
            nextExecution);
      } else {
        handleHttpError(response, nextExecution);
      }
    } catch (ExecutionException e) {
      if (e.getCause() != null) {
        callback.onRequestFailed(e.getCause());
      } else {
        callback.onRequestFailed(e);
      }
    } catch (Exception exception) {
      callback.onRequestFailed(exception);
    }
  }

  private void handleHttpError(HttpSender.Response response, Consumer<Duration> nextExecution) {
    int errorCode = response.statusCode();
    callback.onRequestFailed(new HttpErrorException(errorCode, response.statusMessage()));

    if (errorCode == 503 || errorCode == 429) {
      String retryAfterHeader = response.getHeader("Retry-After");
      Duration retryAfter = null;
      if (retryAfterHeader != null) {
        Optional<Duration> duration = retryAfterParser.tryParse(retryAfterHeader);
        if (duration.isPresent()) {
          retryAfter = duration.get();
        }
      }
      nextExecution.accept(retryAfter);
    }
  }

  private static boolean isSuccessful(HttpSender.Response response) {
    return response.statusCode() >= 200 && response.statusCode() < 300;
  }

  private void handleSuccessResponse(Response response, Consumer<Duration> nextExecution) {
    ServerToAgent serverToAgent = response.getServerToAgent();

    if (serverToAgent.error_response != null) {
      handleErrorResponse(serverToAgent.error_response, nextExecution);
    }

    callback.onRequestSuccess(response);
  }

  private void handleErrorResponse(
      ServerErrorResponse errorResponse, Consumer<Duration> nextExecution) {
    if (errorResponse.type.equals(ServerErrorResponseType.ServerErrorResponseType_Unavailable)) {
      Duration retryAfter = null;
      if (errorResponse.retry_info != null) {
        retryAfter = Duration.ofNanos(errorResponse.retry_info.retry_after_nanoseconds);
      }
      nextExecution.accept(retryAfter);
    }
  }
```

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
```diff
@@ -0,0 +1,219 @@
...
+          handleSuccessResponse(
+              Response.create(ServerToAgent.ADAPTER.decode(response.bodyInputStream())));
+        } else {
+          handleHttpError(response);
+        }
+      } catch (IOException e) {
+        getCallback().onRequestFailed(e);
+      }
+
+    } catch (InterruptedException e) {
```
laurit: i suspect that you might not need the nested try blocks and could add the exceptions to the inner try

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
laurit: perhaps we should also add a timeout to the `get`?

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
```diff
@@ -0,0 +1,219 @@
...
+      this.requestSupplier = requestSupplier;
+      executor.start(this);
+    } else {
+      throw new IllegalStateException("RequestDispatcher is already running");
+    }
+  }
+
+  @Override
+  public void stop() {
+    if (isRunning.compareAndSet(true, false)) {
```
laurit, line 109: This way it is possible to call `start` after `calling` stop, though it will probably not work because the underlying executor was shut down.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
laurit: Currently there is only the `OpampClient` interface but no implementation or a way to create an instance of `OpampClient`. You could require user to provide the `Callbacks` when the client is created.
I wouldn't worry too much about the implementors, there aren't going to be many of them anyway.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
LikeTheSalad: > So that periodic events will fire as scheduled even when there were manual requests

I think that's ok but I'm a bit confused because that should be the way `PeriodicTaskExecutor` works right now. When you call `PeriodicTaskExecutor.executeNow()`, it sends the task directly to the executor service without cancelling the periodic schedule nor creating a new periodic one. Unless I'm missing something?

On top of that, `PeriodicTaskExecutor` allows to swap the `PeriodicDelay` implementation, which helps with having a different schedule when retrying, for example we could create a PeriodicDelay impl with an exponential backoff strategy, as mentioned in the spec, and use it until we get a successful response. Though I don't see that behavior available in the code changes you mentioned.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
LikeTheSalad: Fair enough. I've added this to the [same issue](https://github.com/open-telemetry/opentelemetry-java-contrib/issues/1902) to address both this interface and the client one in a single PR.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/RequestService.java
LikeTheSalad: You're right. It wasn't called anywhere, and it was a miss from my side. I've updated it so that onConnectionFailed is called when something interrupts the HTTP request before we can get a response from the server, and onConnectionSuccess is called when we do (regardless of the HTTP response code).

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
LikeTheSalad: Thanks. I've updated it.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
LikeTheSalad: Sure, I've added a timeout.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
LikeTheSalad: Thanks, I've addressed it temporarily because no validation will be needed when the `start` method is removed as part of [this issue](https://github.com/open-telemetry/opentelemetry-java-contrib/issues/1902).

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
laurit: I just found the `PeriodicTaskExecutor` and interactions between `enableRetryMode` and `disableRetryMode` hard to follow. I passed in `Consumer<Duration>` to allow request sending to give feedback on the next scheduled time. This could easily be replaced with something smarter that can also track successes and fails so it can do exponential backoff.
This was just intended as an idea to allow you to simplify the code.

LikeTheSalad: Ah, got it. I thought it was meant as a full replacement. Ok let me take a deeper look to see how we can integrate both ideas such that we get the same behavior but in a clearer way (it's almost eod where I live so I'll probably add the code changes tomorrow morning).

LikeTheSalad: I've just updated it and removed `PeriodicTaskExecutor`. Did some small changes to what you've suggested to keep the tests passing, and also ensured that all requests run via the executor service, even the immediate ones, as I wouldn't like to risk running those in the main thread, which would cause a crash in android. Though overall it should be better now 👍

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
laurit: Looking at the go version in https://github.com/open-telemetry/opamp-go/blob/main/client/client.go I get the impression that start is there so that you could call `SetAgentDescription` etc. before requests are sent to the server.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
laurit: I looked a bit at the go implementation to get a better understanding what this is supposed to do at all. Although I don't know any go I feel it helped a lot.
I still think that you could improve the code.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
```diff
@@ -0,0 +1,254 @@
...
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import opamp.proto.AgentToServer;
+import opamp.proto.ServerErrorResponse;
+import opamp.proto.ServerErrorResponseType;
+import opamp.proto.ServerToAgent;
+
+public final class HttpRequestService implements RequestService {
+  private final HttpSender requestSender;
+  private final ScheduledExecutorService executorService;
```
laurit, line 42: perhaps there should be a comment that this must be a single threaded executor and that requests shouldn't be sent concurrently

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
```diff
@@ -0,0 +1,254 @@
...
+  @Override
+  public void stop() {
+    if (isRunning.compareAndSet(true, false)) {
+      hasStopped.set(true);
+      executorService.shutdown();
+    }
+  }
+
+  @Override
+  public void sendRequest() {
```
laurit, line 116: If I understand it correctly go version calls this `ScheduleSend` https://github.com/open-telemetry/opamp-go/blob/main/client/internal/sender.go#L54 The comment in that method says that message won't be scheduled when it was already sent. I guess this is also a future improvement to figure out whether the request actually needs to be sent.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
```diff
@@ -0,0 +1,254 @@
...
+    if (!isRunning.get()) {
+      throw new IllegalStateException("HttpRequestService is not running");
+    }
+
+    sendOnce();
+  }
+
+  private void doSendRequest() {
+    AgentToServer agentToServer = Objects.requireNonNull(requestSupplier).get().getAgentToServer();
```
laurit, line 147: The retry logic in https://github.com/open-telemetry/opamp-go/blob/b9a8e1cdee7cf33db29607ffc049521e2d6a54d3/client/internal/httpsender.go#L176 retries the original message. In go version https://github.com/open-telemetry/opamp-go/blob/b9a8e1cdee7cf33db29607ffc049521e2d6a54d3/client/internal/nextmessage.go#L49 as far as I can tell some message fields are reset once it is selected for sending. If you are pulling a new message for retrying you need to somehow guarantee that you don't clear the fields that failed sending. This is just something I noticed, idk whether it is relevant since the request supplier part is not included in this PR.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
```diff
@@ -0,0 +1,254 @@
...
+      if (retryAfter != null && periodicRetryDelay instanceof AcceptsDelaySuggestion) {
+        ((AcceptsDelaySuggestion) periodicRetryDelay).suggestDelay(retryAfter);
+      }
+    }
+  }
+
+  private void cancelCurrentTask() {
+    ScheduledFuture<?> future = currentTask.get();
+    if (future != null) {
+      future.cancel(false);
```
laurit: So this is called when transitioning from regular to retry and vice versa. If the transition happens during a periodic event then canceling the future has no effect, the same task is already running. `periodicSend` will schedule a new task. If the transition happens during a manual send you'll end up canceling the periodic event which won't get rescheduled. Is this correct?

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
LikeTheSalad: Probably. I plan to have it set in the builder, so it's still feasible to remove the start method for Java, as we talked earlier.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
LikeTheSalad: I think an analogous tool to "RequestService" doesn't exist in the Go implementation. Therefore, they have to create 2 client implementations, one per transport protocol, which I believe they'd like to avoid in the future. However, if there's something from the HTTP client implementation that you think we're missing here, please let me know.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
LikeTheSalad: > idk whether it is relevant since the request supplier part is not included in this PR.

It's not relevant for this layer, at least not in this implementation. This layer is all about sending requests, which are triggered by the client implementation. The client is also the request provider and response processor, so that's where I'm planning to make sure we send the same fields for retries, as well as only updating some state (such as the sequence number) when there's a successful response. You can take a look at how it should look like, in case you're interested, by checking the client impl from the [Elastic repo](https://github.com/elastic/opamp-java/blob/6ebcb01b57e979a7e800eefde5d21f294045c6c5/client/src/main/java/co/elastic/opamp/client/internal/OpampClientImpl.java#L49).

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
LikeTheSalad: I think it depends on the transport protocol, because for HTTP, the client must send something even if it has nothing to say, in order to get data from the server. However, I'll create an issue so that we can revisit it when working on the client implementation, as I think this can also be interpreted as "not sending the same message twice" which in that case it's probably a client responsibility.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
LikeTheSalad: I see what you mean. This seems related to other comments in which there's a potential risk of concurrent issues between the periodic send and the manual send. I don't remember if that was addressed somewhere else, in the original impl, but let me add some safeguards here to try and avoid those 2 to overlap.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
LikeTheSalad: Let me see if I can come up with a better solution, as I mentioned [here](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1928#discussion_r2138134964).

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
laurit: I don't think it is a concurrency issue, just code complexity. The original elastic code is also hard to reason about. There `setPeriodicDelay` cancels the current scheduled task and schedules a new task.
- If the call is from scheduled task then cancel does nothing as task is already running and schedules a new task, when the running scheduled task completes it will also schedule a new task so you'll have 2 scheduled task.
- If the call is from manual send then I think it will work correctly.

Could be wrong, too complicated for me to follow. That is why it would be nice if you could simplify it. If you so wish I can also give a try at reworking it a bit.

LikeTheSalad: Got it. I think you might be right, the thing is that it's been a long time since I worked on this, so I don't remember all the details. But what you explained about cancelling scheduled tasks while that same task is running definitely sounds wrong, and it can be something that I missed. I want it to be simple to read because that way it's also going to be easier for me to maintain it, so I agree we should simplify it, and thank you for volunteering. Let me give it one more try today, now that I think I properly understood what the complicated part is, and if afterwards you think it could be further simplified, then I would appreciate it if you could give it a try at reworking it.

LikeTheSalad: I took a deeper look and noticed that the logic to cancel a current task only makes sense when an on-demand request changes the delay strategy. In this scenario, cancelling a previously scheduled task is necessary to avoid sending the next request earlier than it should be sent (based on the retry logic, which could either come from the server or the retry delay implementation). Based on that, it's not necessary to cancel the next task when running periodically, because, as you mentioned, the task is already running and the next one hasn't gotten scheduled yet.

Now, I didn't take into account, when I first created this logic, what would happen should an on-demand and a periodic execution run in parallel. And it seems like there could be some cases in which the periodic event is cancelled forever. Then I tried to come up with some validations to avoid it, but it ended up becoming more complicated and possibly causing more than one periodic event to get scheduled, which is also bad.

So I ended up ensuring that only one request can happen at a time, and that in both on-demand and periodic executions there's always a new event scheduled at the end. So that means that for on-demand executions, the pending periodic request is cancelled and rescheduled at the end, even if the delay logic didn't change, which I don't think it's an issue, it's probably even better than getting into a scenario where an on-demand request finishes and right away the periodic one starts.

A simpler alternative could be allowing the old periodic scheduled request to run, even after the delay time has changed in an on-demand request, but I would be concerned about being a bad citizen with this approach, in the event that the server has requested a retry because it can't handle any more requests at that moment.

LikeTheSalad: Btw I'll be off for the rest of the week, just so you know, in case I take long to respond to further comments. Though feel free to add changes if you think we can further simplify this logic, as discussed earlier. Thank you.

laurit: > So I ended up ensuring that only one request can happen at a time, and that in both on-demand and periodic executions there's always a new event scheduled at the end.

I think this is a reasonable approach. Since we know that only one request is executing at time we could also remove all the "fake" concurrency code. No need to lock when there is no concurrency.

```diff
diff --git a/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java b/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
index f0f459c6..32d5a60f 100644
--- a/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
+++ b/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
@@ -27,8 +27,6 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
-import java.util.concurrent.locks.Lock;
-import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Supplier;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
@@ -39,14 +37,13 @@ import opamp.proto.ServerToAgent;

 public final class HttpRequestService implements RequestService {
   private final HttpSender requestSender;
+  // must be a single threaded executor, the code in this class relies on requests being processed
+  // serially
   private final ScheduledExecutorService executorService;
-  private final PeriodicDelay periodicRequestDelay;
-  private final PeriodicDelay periodicRetryDelay;
   private final AtomicBoolean isRunning = new AtomicBoolean(false);
   private final AtomicBoolean hasStopped = new AtomicBoolean(false);
-  private final AtomicReference<PeriodicDelay> currentDelay;
+  private final ConnectionStatus connectionStatus;
   private final AtomicReference<ScheduledFuture<?>> scheduledTask = new AtomicReference<>();
-  private final Lock sendLock = new ReentrantLock();
   private final RetryAfterParser retryAfterParser;
   @Nullable private Callback callback;
   @Nullable private Supplier<Request> requestSupplier;
@@ -89,10 +86,8 @@ public final class HttpRequestService implements RequestService {
       RetryAfterParser retryAfterParser) {
     this.requestSender = requestSender;
     this.executorService = executorService;
-    this.periodicRequestDelay = periodicRequestDelay;
-    this.periodicRetryDelay = periodicRetryDelay;
     this.retryAfterParser = retryAfterParser;
-    currentDelay = new AtomicReference<>(periodicRequestDelay);
+    this.connectionStatus = new ConnectionStatus(periodicRequestDelay, periodicRetryDelay);
   }

   @Override
@@ -109,10 +104,6 @@ public final class HttpRequestService implements RequestService {
     }
   }

-  private Duration getNextDelay() {
-    return Objects.requireNonNull(currentDelay.get()).getNextDelay();
-  }
-
   @Override
   public void stop() {
     if (isRunning.compareAndSet(true, false)) {
@@ -127,32 +118,16 @@ public final class HttpRequestService implements RequestService {
       throw new IllegalStateException("HttpRequestService is not running");
     }

-    executorService.execute(this::requestOnDemand);
-  }
-
-  private void requestOnDemand() {
-    if (sendLock.tryLock()) {
-      try {
-        ScheduledFuture<?> scheduledFuture = scheduledTask.get();
-        if (scheduledFuture != null) {
-          // Cancel future task
-          scheduledFuture.cancel(false);
-        }
-        sendAndScheduleNext();
-      } finally {
-        sendLock.unlock();
-      }
-    }
-  }
-
-  private void periodicRequest() {
-    if (sendLock.tryLock()) {
-      try {
-        sendAndScheduleNext();
-      } finally {
-        sendLock.unlock();
-      }
-    }
+    executorService.execute(
+        () -> {
+          // cancel the already scheduled task, a new one is created after current request is
+          // processed
+          ScheduledFuture<?> scheduledFuture = scheduledTask.get();
+          if (scheduledFuture != null) {
+            scheduledFuture.cancel(false);
+          }
+          sendAndScheduleNext();
+        });
   }

   private void sendAndScheduleNext() {
@@ -163,7 +138,9 @@ public final class HttpRequestService implements RequestService {
   private void scheduleNextExecution() {
     scheduledTask.set(
         executorService.schedule(
-            this::periodicRequest, getNextDelay().toNanos(), TimeUnit.NANOSECONDS));
+            this::sendAndScheduleNext,
+            connectionStatus.getNextDelay().toNanos(),
+            TimeUnit.NANOSECONDS));
   }

   private void doSendRequest() {
@@ -204,7 +181,7 @@ public final class HttpRequestService implements RequestService {
           retryAfter = duration.get();
         }
       }
-      useRetryDelay(retryAfter);
+      connectionStatus.retryAfter(retryAfter);
     }
   }

@@ -213,7 +190,7 @@ public final class HttpRequestService implements RequestService {
   }

   private void handleHttpSuccess(Response response) {
-    useRegularDelay();
+    connectionStatus.success();
     ServerToAgent serverToAgent = response.getServerToAgent();

     if (serverToAgent.error_response != null) {
@@ -229,28 +206,54 @@ public final class HttpRequestService implements RequestService {
       if (errorResponse.retry_info != null) {
         retryAfter = Duration.ofNanos(errorResponse.retry_info.retry_after_nanoseconds);
       }
-      useRetryDelay(retryAfter);
+      connectionStatus.retryAfter(retryAfter);
     }
     getCallback().onRequestFailed(new OpampServerResponseError(errorResponse.error_message));
   }

-  private void useRegularDelay() {
-    if (currentDelay.compareAndSet(periodicRetryDelay, periodicRequestDelay)) {
-      periodicRequestDelay.reset();
-    }
+  private Callback getCallback() {
+    return Objects.requireNonNull(callback);
   }

-  private void useRetryDelay(@Nullable Duration retryAfter) {
-    if (currentDelay.compareAndSet(periodicRequestDelay, periodicRetryDelay)) {
-      periodicRetryDelay.reset();
-      if (retryAfter != null && periodicRetryDelay instanceof AcceptsDelaySuggestion) {
-        ((AcceptsDelaySuggestion) periodicRetryDelay).suggestDelay(retryAfter);
+  // this class is only used from a single threaded ScheduledExecutorService, hence no
+  // synchronization is needed
+  private static class ConnectionStatus {
+    private final PeriodicDelay periodicRequestDelay;
+    private final PeriodicDelay periodicRetryDelay;
+
+    private boolean retrying;
+    private PeriodicDelay currentDelay;
+
+    ConnectionStatus(PeriodicDelay periodicRequestDelay, PeriodicDelay periodicRetryDelay) {
+      this.periodicRequestDelay = periodicRequestDelay;
+      this.periodicRetryDelay = periodicRetryDelay;
+      currentDelay = periodicRequestDelay;
+    }
+
+    void success() {
+      // after successful request transition from retry to regular delay
+      if (retrying) {
+        retrying = false;
+        periodicRequestDelay.reset();
+        currentDelay = periodicRequestDelay;
       }
     }
-  }

-  private Callback getCallback() {
-    return Objects.requireNonNull(callback);
+    void retryAfter(@Nullable Duration retryAfter) {
+      // after failed request transition from regular to retry delay
+      if (!retrying) {
+        retrying = true;
+        periodicRetryDelay.reset();
+        currentDelay = periodicRetryDelay;
+        if (retryAfter != null && periodicRetryDelay instanceof AcceptsDelaySuggestion) {
+          ((AcceptsDelaySuggestion) periodicRetryDelay).suggestDelay(retryAfter);
+        }
+      }
+    }
+
+    Duration getNextDelay() {
+      return currentDelay.getNextDelay();
+    }
   }

   private static class DaemonThreadFactory implements ThreadFactory {
```

LikeTheSalad: Cheers! I've applied your suggestions.

PR #1934, opened 2025-06-02
add test for Mac OS and Windows

zeitlinger:
Relates https://github.com/open-telemetry/opentelemetry-java-contrib/issues/1907

Using the same config as https://github.com/open-telemetry/opentelemetry-java/blob/main/.github/workflows/build.yml

### DIFF
--- a/opamp-client/build.gradle.kts
+++ b/opamp-client/build.gradle.kts
@@ -53,12 +53,8 @@ abstract class DownloadOpampProtos @Inject constructor(
     val redirectLocation = connection.getHeaderField("Location")
     connection.disconnect()
 
-    val latestTag = if (redirectLocation != null && redirectLocation.contains("/releases/tag/")) {
-      // Extract tag from URL like: https://github.com/open-telemetry/opamp-spec/releases/tag/v0.12.0
-      redirectLocation.substringAfterLast("/")
-    } else {
-      throw RuntimeException("Could not determine latest release tag from redirect. Redirect location: $redirectLocation")
-    }
+    // Extract tag from URL like: https://github.com/open-telemetry/opamp-spec/releases/tag/v0.12.0
+    val latestTag = redirectLocation.substringAfterLast("/")
     // Download the source code for the latest release
     val zipUrl = "https://github.com/open-telemetry/opamp-spec/zipball/$latestTag"
 


inferred-spans/src/test/java/io/opentelemetry/contrib/inferredspans/InferredSpansAutoConfigTest.java
```diff
@@ -31,9 +33,10 @@
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.condition.DisabledOnOs;
-import org.junit.jupiter.api.condition.OS;
+import org.junit.jupiter.api.condition.EnabledOnOs;
 import org.junit.jupiter.api.io.TempDir;
 
+@EnabledOnOs(LINUX) // Uses async-profiler, which is only supported on Linux
```
laurit: async profiler works on mac, see https://github.com/async-profiler/async-profiler

.github/workflows/build.yml
```diff
@@ -6,16 +6,15 @@ on:
       - main
       - release/*
   pull_request:
-  merge_group:
```
trask: we are using merge queue in this repo

.github/workflows/build.yml
```diff
@@ -6,16 +6,15 @@ on:
...
-permissions:
-  contents: read
-
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
   cancel-in-progress: true
 
+permissions:
+  contents: read
+
```
trask: do you mind reverting this change? I think permissions are more important than concurrency, so prefer it to be at the top

.github/workflows/build.yml
```diff
@@ -33,32 +32,57 @@ jobs:
         with:
           cache-read-only: ${{ github.event_name == 'pull_request' }}
       - name: Gradle build and test
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
trask: was this needed?
```suggestion
```

.github/workflows/build.yml
```diff
@@ -69,11 +93,13 @@ jobs:
         with:
           cache-read-only: ${{ github.event_name == 'pull_request' }}
       - name: Gradle test
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
trask: same
```suggestion
```

.github/workflows/build.yml
zeitlinger: the rate limit kicked in due to the parallel builds here: https://github.com/zeitlinger/opentelemetry-java-contrib/blob/c4b126be51817af8c2a95b6d54ba8ab9343077e7/opamp-client/build.gradle.kts#L21

trask: interesting, the problem is that secrets aren't available on PR builds, so I don't think this will help (we have similar problem on link checking workflows)

zeitlinger: it did fix this PR...

trask: can you explain?

zeitlinger: - at first, I only added the matrix test
- this resulted it "rate limit exceeded" errors
- then I added the GH_TOKEN change
- this made the PR pass

trask: looks like you're right: https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication

let me do some testing, I'm not sure why this didn't help with the link check workflows...

trask: finally got to the bottom of why I didn't think it worked before (was because of a bug) 😅: https://github.com/open-telemetry/semantic-conventions/pull/2352#discussion_r2136728568

zeitlinger: great - is this PR ready then?

zeitlinger: @trask can you take a look?

.github/workflows/build.yml
```diff
@@ -36,29 +36,52 @@ jobs:
...
           - 17
           - 21
           - 23
-      fail-fast: false
+        # Collect coverage on latest LTS
+        include:
+          - os: ubuntu-latest
+            test-java-version: 21
+            coverage: true
+            jmh-based-tests: true
```
trask: are these being used?

.github/workflows/build.yml
```diff
@@ -36,29 +36,52 @@ jobs:
...
+          - os: macos-13
+            test-java-version: 21
+          - os: macos-13
+            test-java-version: 23
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
 
-      - id: setup-test-java
-        name: Set up JDK ${{ matrix.test-java-version }} for running tests
+      - id: setup-java-test
```
trask, line 70: `id` is renamed here but not its usage below

.github/workflows/build.yml
trask: I'm still not following how hitting only a single url would require GH_TOKEN, it's not like link checking where we are hitting 100s or 1000s of urls, so I reverted this part to investigate, but not seeing the failure that you were seeing earlier

.github/workflows/build.yml
zeitlinger: removed - that was just copied from the other setup

zeitlinger:
@trask please check again

trask (approved):

Issue #1938, opened 2025-06-06
[disk-buffering] LogRecord EventName is not serialized


Doohl:
### Component(s)

disk-buffering

## Description

**EventName** was introduced recently in opentelemetry-specification (https://github.com/open-telemetry/opentelemetry-specification/issues/4260) to Log and Event record models.

However, EventName hasn't yet been added to the [LogRecordDataMapper](https://github.com/open-telemetry/opentelemetry-java-contrib/blob/main/disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/serialization/mapping/logs/LogRecordDataMapper.java) logic.

## Steps to Reproduce

I encountered this issue testing [opentelemetry-android](https://github.com/open-telemetry/opentelemetry-android). You can repro this issue by:

1. Checking out opentelemetry-android 0.11.0
2. Running the demo-app
3. Observing the cache log files (or with OTEL collector) a lack of EventName in Log Records

## Expected Result

The raw delimited protobuf items should contain EventName context.

## Actual Result

The raw delimited protobuf items do not contain EventName context.

### Component version

1.46.0

### Log output

```
==== RAW LOG RECORD PROTOBUF ====



�
�
�

device.manufacturer
Google
/
device.model.identifier
sdk_gphone64_arm64
)
device.model.name
sdk_gphone64_arm64
K
os.description9
7Android Version 16 (Build BP22.250221.010 API level 36)

os.name	
Android

os.type
linux


os.version
16
*
rum.sdk.version
0.11.0-alpha-SNAPSHOT
&
service.name
OpenTelemetryDemoApp
 
telemetry.sdk.language
java
%
telemetry.sdk.name

opentelemetry
!
telemetry.sdk.version
1.49.0�

otel.demo.app.navigation�YE�=�E2
network.carrier.icc
us2
network.carrier.mcc
3102
network.carrier.mnc
2602'
network.carrier.name

T-Mobile - US2!
network.connection.type
wifi2

product.id

1YMWWN1N4O2&
screen.name
AstronomyShopActivity20

session.id"
 4ef5b0e85f023744f242bc914bb74c332
toolkit
jetpack composeJ 00000000000000000000000000000000R0000000000000000�
�
�

device.manufacturer
Google
/
device.model.identifier
sdk_gphone64_arm64
)
device.model.name
sdk_gphone64_arm64
K
os.description9
7Android Version 16 (Build BP22.250221.010 API level 36)

[...]

PR #1950, opened 2025-06-09
Add missing EventName to disk-buffering LogRecordDataMapper

Doohl:
**Description:**

Updates disk-buffering to use the new [EventName spec](https://opentelemetry.io/docs/specs/otel/logs/data-model/#field-eventname) for LogRecords for both serialization to protobuf and de-serialization to SDK LogRecords.

Existing behavior is preserved; EventName will be excluded if it's not present in the data model.

**Existing Issue(s):**

Fixes https://github.com/open-telemetry/opentelemetry-java-contrib/issues/1938

**Testing:**

- Added two new unit tests to validate the new serialization + de-serialization behavior
- Manually tested the disk-buffering extension against telemetry with and without the EventName LogRecord field

**Documentation:**

No documentation updates needed. We aren't introducing new behavior, but rather preserving the expected behavior.


disk-buffering/src/test/java/io/opentelemetry/contrib/disk/buffering/internal/serialization/mapping/logs/ProtoLogsDataMapperTest.java
```diff
@@ -37,6 +37,7 @@ class ProtoLogsDataMapperTest {
           .setTimestampEpochNanos(100L)
           .setObservedTimestampEpochNanos(200L)
           .setTotalAttributeCount(3)
+          .setEventName("")
```
Doohl, line 40: note: We have to explicitly add `setEventName` to these tests because `io.opentelemetry.proto.logs.v1.LogRecord.Builder` has a default EventName value of `""`. Not adding some string here will cause the tests to fail.

breedx-splk: That feels wrong to me -- the event name should be nullable. Can. you point me to where this is defaulted?

breedx-splk: Oh, I see it now (it's in the build output, generated from the protos). I wonder if we can fix that to allow nullables.... 🤔

breedx-splk (approved):
This looks good to me. Thanks for the contribution!

disk-buffering/src/test/java/io/opentelemetry/contrib/disk/buffering/internal/serialization/mapping/logs/ProtoLogsDataMapperTest.java
LikeTheSalad: > Not adding some string here will cause the tests to fail.

Do you know if the reason would be because of the proto builder, or our LogRecordDataImpl builder? - I'm asking because I recall AutoValue complaining if one non-nullable setter wasn't called. So if that's the case, maybe annotating the parameter [here](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1950/files#diff-0900bb91f1dc1b1573756d7a9a750a6157c167951e0d763fdfdcefda53b7cb30R75) as nullable could help?

LikeTheSalad: I just checked and the [proto definition](https://github.com/open-telemetry/opentelemetry-proto/blob/68f9c6329ca91f96333ee5b240e305945b122b70/opentelemetry/proto/logs/v1/logs.proto#L224) isn't marked as `optional`. So maybe the way to go about this is to check for its default value to either call the setters or not, in both serialization and deserialization cases.

Doohl: > I just checked and the [proto definition](https://github.com/open-telemetry/opentelemetry-proto/blob/68f9c6329ca91f96333ee5b240e305945b122b70/opentelemetry/proto/logs/v1/logs.proto#L224) isn't marked as `optional`. So maybe the way to go about this is to check for its default value to either call the setters or not, in both serialization and deserialization cases.

I considered disabling the setters on `event_name == ""`, but that logic smells off to me. Is there a case where some user may *want* to set event_name to an empty string? It makes some assumptions about use cases here that I can't answer myself.

I'm ok with making that change if we're confident that nobody's going to be trying to explicitly set `event_name` to `""` (empty string) so that we can null the value out and save some bytes for serialization.

LikeTheSalad: Discussed in the Android SIG meeting, we can move forward with it without further changes. Thanks!

PR #1957, opened 2025-06-11
Common Expression Language (CEL) sampler

dol:
This pull request introduces a new `CelBasedSampler` to the OpenTelemetry Java SDK, enabling advanced sampling rules using the [Common Expression Language (CEL)](https://cel.dev/). It also includes updates to the existing `RuleBasedRoutingSampler` for consistency and clarity. Below is a summary of the most important changes grouped by theme:

### New `CelBasedSampler` Implementation:
* Added the `CelBasedSampler` class, which evaluates CEL expressions to make sampling decisions based on span attributes. It includes a fallback sampler and supports multiple expressions. (`samplers/src/main/java/io/opentelemetry/contrib/sampler/CelBasedSampler.java`)
* Introduced the `CelBasedSamplerBuilder` to construct `CelBasedSampler` instances with methods for adding expressions and specifying actions (e.g., `DROP` or `RECORD_AND_SAMPLE`). (`samplers/src/main/java/io/opentelemetry/contrib/sampler/CelBasedSamplerBuilder.java`)
* Added the `CelBasedSamplingExpression` class to encapsulate individual CEL expressions and their associated samplers. (`samplers/src/main/java/io/opentelemetry/contrib/sampler/CelBasedSamplingExpression.java`)
* Implemented a declarative configuration provider for `CelBasedSampler`, enabling configuration through YAML files. (`samplers/src/main/java/io/opentelemetry/contrib/sampler/internal/CelBasedSamplerComponentProvider.java`)

```yaml
# The fallback sampler to use if no expressions match.
fallback_sampler:
  always_on:
# List of CEL expressions to evaluate. Expressions are evaluated in order.
expressions:
  # The action to take when the expression evaluates to true. Must be one of: DROP, RECORD_AND_SAMPLE.
  - action: DROP
    # The CEL expression to evaluate. Must return a boolean.
    expression: attribute['url.path'].startsWith('/actuator')
  - action: RECORD_AND_SAMPLE
    expression: attribute['http.method'] == 'GET' && attribute['http.status_code'] < 400
```

### Documentation Updates:
* Updated `samplers/README.md` to document the new `CelBasedSampler`, including its usage, schema, and example configurations. (`samplers/README.md`) [[1]](diffhunk://#diff-04f8ddc21bc0d07866fd15df18a227ab8a2ff3077a49237b291417b30f603350R8-R16) [[2]](diffhunk://#diff-04f8ddc21bc0d07866fd15df18a227ab8a2ff3077a49237b291417b30f603350R63-R122)

### Dependency Additions:
* Added a dependency on the `dev.cel:cel:0.9.0` library to enable CEL expression evaluation. (`samplers/build.gradle.kts`)

### Updates to `RuleBasedRoutingSampler`:
* Renamed `SamplingRule` to `RuleBasedRoutingSamplingRule` for clarity and updated all related references in the `RuleBasedRoutingSampler` and its builder. (`samplers/src/main/java/io/opentelemetry/contrib/sampler/RuleBasedRoutingSampler.java`, `samplers/src/main/java/io/opentelemetry/contrib/sampler/RuleBasedRoutingSamplerBuilder.java`, `samplers/src/main/java/io/opentelemetry/contrib/sampler/RuleBasedRoutingSamplingRule.java`) [[1]](diffhunk://#diff-8d5c9e7a1055658bca746fc511be27d559d5a05784bee429b66f8d4828b181d8L20-R23) [[2]](diffhunk://#diff-8d5c9e7a1055658bca746fc511be27d559d5a05784bee429b66f8d4828b181d8L40-R45) [[3]](diffhunk://#diff-8d5c9e7a1055658bca746fc511be27d559d5a05784bee429b66f8d4828b181d8L67-R79) [[4]](diffhunk://#diff-6d709cd7757cda0e51ba7d356b3c29290dda9d04b43e4d6edc2919c8fd7f0dd7L18-R18) [[5]](diffhunk://#diff-6d709cd7757cda0e51ba7d356b3c29290dda9d04b43e4d6edc2919c8fd7f0dd7L44-R44) [[6]](diffhunk://#diff-6afb6bd80bcafb5070370afc04daff9a0b0641e85b158cd0c013ab32595a33e4L14-R28) [[7]](diffhunk://#diff-6afb6bd80bcafb5070370afc04daff9a0b0641e85b158cd0c013ab32595a33e4L42-R46)

These changes collectively enhance the SDK's sampling capabilities, allowing users to define sophisticated sampling rules using CEL while maintaining compatibility with existing samplers.

### DIFF
--- a/samplers/README.md
+++ b/samplers/README.md
@@ -5,14 +5,15 @@
 The following samplers support [declarative configuration](https://opentelemetry.io/docs/languages/java/configuration/#declarative-configuration):
 
 * `RuleBasedRoutingSampler`
+* `CelBasedSampler`
 
 To use:
 
 * Add a dependency on `io.opentelemetry:opentelemetry-sdk-extension-incubator:<version>`
 * Follow the [instructions](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk-extensions/incubator/README.md#file-configuration) to configure OpenTelemetry with declarative configuration.
-* Configure the `.tracer_provider.sampler` to include the `rule_based_routing` sampler.
+* Configure the `.tracer_provider.sampler` to include the `rule_based_routing` or `cel_based` sampler.
 
-NOTE: Not yet available for use with the OTEL java agent, but should be in the near future. Please check back for updates.
+## RuleBasedRoutingSampler
 
 Schema for `rule_based_routing` sampler:
 
@@ -57,6 +58,66 @@ tracer_provider:
             - action: DROP
               attribute: url.path
               pattern: /actuator.*
+```
+
+## CelBasedSampler
+
+The `CelBasedSampler` uses [Common Expression Language (CEL)](https://github.com/google/cel-spec) to create advanced sampling rules based on span attributes. CEL provides a powerful, yet simple expression language that allows you to create complex matching conditions.
+
+To use:
+
+* Add a dependency on `io.opentelemetry:opentelemetry-sdk-extension-incubator:<version>`
+* Follow the [instructions](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk-extensions/incubator/README.md#file-configuration) to configure OpenTelemetry with declarative configuration.
+* Configure the `.tracer_provider.sampler` to include the `cel_based` sampler.
+
+Schema for `cel_based` sampler:
+
+```yaml
+# The fallback sampler to use if no expressions match.
+fallback_sampler:
+  always_on:
+# List of CEL expressions to evaluate. Expressions are evaluated in order.
+expressions:
+  # The action to take when the expression evaluates to true. Must be one of: DROP, RECORD_AND_SAMPLE.
+  - action: DROP
+    # The CEL expression to evaluate. Must return a boolean.
+    expression: attribute['url.path'].startsWith('/actuator')
+  - action: RECORD_AND_SAMPLE
+    expression: attribute['http.method'] == 'GET' && attribute['http.status_code'] < 400
+```
+
+Available variables in CEL expressions:
+
+* `name` (string): The span name
+* `spanKind` (string): The span kind (e.g., "SERVER", "CLIENT")
+* `attribute` (map): A map of span attributes
+
+Example of using `cel_based` sampler as the root sampler in `parent_based` sampler configuration:
+
+```yaml
+tracer_provider:
+  sampler:
+    parent_based:
+      root:
+        cel_based:
+          fallback_sampler:
+            always_on:
+          expressions:
+            # Drop health check endpoints
+            - action: DROP
+              expression: spanKind == 'SERVER' && attribute['url.path'].startsWith('/health')
+            # Drop actuator endpoints
+            - action: DROP
+              expression: spanKind == 'SERVER' && attribute['url.path'].startsWith('/actuator')
+            # Sample only HTTP GET requests with successful responses
+            - action: RECORD_AND_SAMPLE
+              expression: spanKind == 'SERVER' && attribute['http.method'] == 'GET' && attribute['http.status_code'] < 400
+            # Selectively sample based on span name
+            - action: RECORD_AND_SAMPLE
+              expression: name.contains('checkout') || name.contains('payment')
+            # Drop spans with specific name patterns
+            - action: DROP
+              expression: name.matches('.*internal.*') && spanKind == 'INTERNAL'
 ```
 
 ## Component owners
--- a/samplers/src/main/java/io/opentelemetry/contrib/sampler/CelBasedSamplingExpression.java
+++ b/samplers/src/main/java/io/opentelemetry/contrib/sampler/CelBasedSamplingExpression.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.contrib.sampler;
+
+import dev.cel.common.CelAbstractSyntaxTree;
+import io.opentelemetry.sdk.trace.samplers.Sampler;
+import java.util.Objects;
+import javax.annotation.Nullable;
+
+public class CelBasedSamplingExpression {
+  final Sampler delegate;
+  final String expression;
+  final CelAbstractSyntaxTree abstractSyntaxTree;
+
+  CelBasedSamplingExpression(
+      Sampler delegate, String expression, CelAbstractSyntaxTree abstractSyntaxTree) {
+    this.delegate = delegate;
+    this.expression = expression;
+    this.abstractSyntaxTree = abstractSyntaxTree;
+  }
+
+  @Override
+  public String toString() {
+    return "CelBasedSamplingExpression{"
+        + "delegate="
+        + delegate
+        + ", expression='"
+        + expression
+        + '}';
+  }
+
+  @Override
+  public boolean equals(@Nullable Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (!(o instanceof CelBasedSamplingExpression)) {
+      return false;
+    }
+    CelBasedSamplingExpression that = (CelBasedSamplingExpression) o;
+    return Objects.equals(delegate, that.delegate)
+        && Objects.equals(expression, that.expression)
+        && Objects.equals(abstractSyntaxTree, that.abstractSyntaxTree);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(delegate, expression, abstractSyntaxTree);
+  }
+}
--- a/samplers/src/main/java/io/opentelemetry/contrib/sampler/internal/CelBasedSamplerComponentProvider.java
+++ b/samplers/src/main/java/io/opentelemetry/contrib/sampler/internal/CelBasedSamplerComponentProvider.java
@@ -1,0 +1,91 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.contrib.sampler.internal;
+
+import dev.cel.common.CelValidationException;
+import io.opentelemetry.api.incubator.config.DeclarativeConfigException;
+import io.opentelemetry.api.incubator.config.DeclarativeConfigProperties;
+import io.opentelemetry.contrib.sampler.CelBasedSampler;
+import io.opentelemetry.contrib.sampler.CelBasedSamplerBuilder;
+import io.opentelemetry.sdk.autoconfigure.spi.internal.ComponentProvider;
+import io.opentelemetry.sdk.extension.incubator.fileconfig.DeclarativeConfiguration;
+import io.opentelemetry.sdk.trace.samplers.Sampler;
+import java.util.List;
+
+/**
+ * Declarative configuration SPI implementation for {@link CelBasedSampler}.
+ *
+ * <p>This class is internal and is hence not for public use. Its APIs are unstable and can change
+ * at any time.
+ */
+public class CelBasedSamplerComponentProvider implements ComponentProvider<Sampler> {
+
+  private static final String ACTION_RECORD_AND_SAMPLE = "RECORD_AND_SAMPLE";
+  private static final String ACTION_DROP = "DROP";
+
+  @Override
+  public Class<Sampler> getType() {
+    return Sampler.class;
+  }
+
+  @Override
+  public String getName() {
+    return "cel_based";
+  }
+
+  @Override
+  public Sampler create(DeclarativeConfigProperties config) {
+    DeclarativeConfigProperties fallbackModel = config.getStructured("fallback_sampler");
+    if (fallbackModel == null) {
+      throw new DeclarativeConfigException(
+          "cel_based sampler .fallback_sampler is required but is null");
+    }
+    Sampler fallbackSampler;
+    try {
+      fallbackSampler = DeclarativeConfiguration.createSampler(fallbackModel);
+    } catch (DeclarativeConfigException e) {
+      throw new DeclarativeConfigException(
+          "cel_based sampler failed to create .fallback_sampler sampler", e);
+    }
+
+    List<DeclarativeConfigProperties> expressions = config.getStructuredList("expressions");
+    if (expressions == null || expressions.isEmpty()) {
+      throw new DeclarativeConfigException("cel_based sampler .expressions is required");
+    }
+
+    CelBasedSamplerBuilder builder = CelBasedSampler.builder(fallbackSampler);
+
+    for (DeclarativeConfigProperties expressionConfig : expressions) {
+      String expression = expressionConfig.getString("expression");
+      if (expression == null) {
+        throw new DeclarativeConfigException(
+            "cel_based sampler .expressions[].expression is required");
+      }
+
+      String action = expressionConfig.getString("action");
+      if (action == null) {
+        throw new DeclarativeConfigException("cel_based sampler .expressions[].action is required");
+      }
+
+      try {
+        if (action.equals(ACTION_RECORD_AND_SAMPLE)) {
+          builder.recordAndSample(expression);
+        } else if (action.equals(ACTION_DROP)) {
+          builder.drop(expression);
+        } else {
+          throw new DeclarativeConfigException(
+              "cel_based sampler .expressions[].action must be "
+                  + ACTION_RECORD_AND_SAMPLE
+                  + " or "
+                  + ACTION_DROP);
+        }
+      } catch (CelValidationException e) {
+        throw new DeclarativeConfigException("Failed to compile CEL expression: " + expression, e);
+      }
+    }
+    return builder.build();
+  }
+}
--- a/samplers/src/test/java/internal/CelBasedSamplerComponentProviderTest.java
+++ b/samplers/src/test/java/internal/CelBasedSamplerComponentProviderTest.java
@@ -1,0 +1,218 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package internal;
+
+import static io.opentelemetry.sdk.trace.samplers.SamplingResult.recordAndSample;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+import dev.cel.common.CelValidationException;
+import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.api.incubator.config.DeclarativeConfigException;
+import io.opentelemetry.api.incubator.config.DeclarativeConfigProperties;
+import io.opentelemetry.api.trace.SpanKind;
+import io.opentelemetry.context.Context;
+import io.opentelemetry.contrib.sampler.CelBasedSampler;
+import io.opentelemetry.contrib.sampler.CelBasedSamplerBuilder;
+import io.opentelemetry.contrib.sampler.internal.CelBasedSamplerComponentProvider;
+import io.opentelemetry.sdk.OpenTelemetrySdk;
+import io.opentelemetry.sdk.extension.incubator.fileconfig.DeclarativeConfiguration;
+import io.opentelemetry.sdk.trace.IdGenerator;
+import io.opentelemetry.sdk.trace.samplers.Sampler;
+import io.opentelemetry.sdk.trace.samplers.SamplingResult;
+import java.io.ByteArrayInputStream;
+import java.nio.charset.StandardCharsets;
+import java.util.Collections;
+import java.util.stream.Stream;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+class CelBasedSamplerComponentProviderTest {
+
+  private static final CelBasedSamplerComponentProvider PROVIDER =
+      new CelBasedSamplerComponentProvider();
+
+  @Test
+  void endToEnd() {
+    String yaml =
+        "file_format: 0.4\n"
+            + "tracer_provider:\n"
+            + "  sampler:\n"
+            + "    parent_based:\n"
+            + "      root:\n"
+            + "        cel_based:\n"
+            + "          fallback_sampler:\n"
+            + "            always_on:\n"
+            + "          expressions:\n"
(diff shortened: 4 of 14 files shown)


trask:
@dol this looks nice!

@johnbley what do you think of CEL vs JEXL in the context of https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/13590?

dependencyManagement/build.gradle.kts
```diff
@@ -2,7 +2,7 @@ plugins {
   `java-platform`
 }
 
-val otelInstrumentationVersion = "2.17.1-alpha"
+val otelInstrumentationVersion = "2.18.0-alpha"
```
dol: Can be removed once the renovate bot bumped the version. The new 2.18.0 version includes the https://github.com/open-telemetry/opentelemetry-java/releases/tag/v1.52.0 release that fixes the double/single quote issues.
open-telemetry/opentelemetry-java#7433

dol: Resolved by #2042

samplers/README.md
```diff
@@ -59,6 +60,66 @@ tracer_provider:
...
+* Add a dependency on `io.opentelemetry:opentelemetry-sdk-extension-incubator:<version>`
+* Follow the [instructions](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk-extensions/incubator/README.md#file-configuration) to configure OpenTelemetry with declarative configuration.
+* Configure the `.tracer_provider.sampler` to include the `cel_based` sampler.
+
+Schema for `cel_based` sampler:
+
+```yaml
+# The fallback sampler to use if no expressions match.
+fallback_sampler:
+  always_on:
```
breedx-splk: If this required or does it default to `always_on` if not present? Might be nice to mention that here?

samplers/src/main/java/io/opentelemetry/contrib/sampler/CelBasedSamplingExpression.java
```diff
@@ -0,0 +1,54 @@
...
+
+import dev.cel.common.CelAbstractSyntaxTree;
+import io.opentelemetry.sdk.trace.samplers.Sampler;
+import java.util.Objects;
+import javax.annotation.Nullable;
+
+public class CelBasedSamplingExpression {
+  final CelAbstractSyntaxTree abstractSyntaxTree;
+  final String expression;
+  final Sampler delegate;
```
breedx-splk: I believe our convention is to make these private and use explicit `get`ters, or use `@AutoValue` with a `create()` method.

samplers/src/main/java/io/opentelemetry/contrib/sampler/internal/CelBasedSamplerComponentProvider.java
```diff
@@ -0,0 +1,91 @@
...
+      throw new DeclarativeConfigException(
+          "cel_based sampler .fallback_sampler is required but is null");
+    }
+    Sampler fallbackSampler;
+    try {
+      fallbackSampler = DeclarativeConfiguration.createSampler(fallbackModel);
+    } catch (DeclarativeConfigException e) {
+      throw new DeclarativeConfigException(
+          "cel_based sampler failed to create .fallback_sampler sampler", e);
+    }
```
breedx-splk: Factoring this out to a method called `getFallbackModel()` could probably help to shrink this long method and improve readability. 👍🏻

samplers/README.md
```diff
@@ -59,6 +60,66 @@ tracer_provider:
...
+# List of CEL expressions to evaluate. Expressions are evaluated in order.
```
breedx-splk: I think it would help to be very explicit. Calling out that it's in-order is great, but is it that the first expression to match is applied and the remaining do not get evaluated, or is it the last one to match "wins"?

samplers/src/test/java/internal/CelBasedSamplerComponentProviderTest.java
```diff
@@ -0,0 +1,246 @@
...
+            + "              expression: '\"example.com\" in attribute[\"http.response.header.host\"] && attribute[\"http.response.status_code\"] == 200'\n"
+            + "            - action: DROP\n"
+            + "              expression: spanKind == \"SERVER\" && attribute[\"url.path\"].matches(\"/actuator.*\")\n"
+            + "            - action: RECORD_AND_SAMPLE\n"
+            + "              expression: spanKind == \"SERVER\" && attribute[\"url.path\"].matches(\"/actuator.*\")\n";
+    OpenTelemetrySdk openTelemetrySdk =
+        DeclarativeConfiguration.parseAndCreate(
+            new ByteArrayInputStream(yaml.getBytes(StandardCharsets.UTF_8)));
+    Sampler sampler = openTelemetrySdk.getSdkTracerProvider().getSampler();
+    assertThat(sampler.toString())
```
breedx-splk: I think it's fine for this first pass (shouldn't gate this PR), but testing through the `toString()` is a bit of an antipattern. We can look at ways of making that more testable later tho.

samplers/src/test/java/internal/CelBasedSamplerComponentProviderTest.java
```diff
@@ -0,0 +1,246 @@
...
+import org.junit.jupiter.params.provider.MethodSource;
+
+class CelBasedSamplerComponentProviderTest {
+
+  private static final CelBasedSamplerComponentProvider PROVIDER =
+      new CelBasedSamplerComponentProvider();
+
+  @Test
+  void endToEnd() {
+    String yaml =
```
breedx-splk: This is _just_ long enough that I think maybe a resource file could be easier to maintain, but definitely in this first pass this is fine.

samplers/src/test/java/internal/CelBasedSamplerComponentProviderTest.java
```diff
@@ -0,0 +1,246 @@
...
+        .isEqualTo(recordAndSample());
+  }
+
+  static Sampler dropSampler(Sampler fallback, String... expressions) {
+    CelBasedSamplerBuilder builder = CelBasedSampler.builder(fallback);
+    for (String expression : expressions) {
+      try {
+        builder.drop(expression);
+      } catch (CelValidationException e) {
+        // Delegate to the provider to handle the exception
```
breedx-splk: would adding a `fail()` call in the `catch()` make this slightly more robust? If not, then we expect this to be hit sometime and should probably verify that. Otherwise, just let this method throw and the test can fail normally.

samplers/src/test/java/internal/CelBasedSamplerComponentProviderTest.java
```diff
@@ -0,0 +1,246 @@
...
+    return builder.build();
+  }
+
+  static Sampler recordAndSampleSampler(Sampler fallback, String... expressions) {
+    CelBasedSamplerBuilder builder = CelBasedSampler.builder(fallback);
+    for (String expression : expressions) {
+      try {
+        builder.recordAndSample(expression);
+      } catch (CelValidationException e) {
+        // Delegate to the provider to handle the exception
```
breedx-splk: Same comment as above about fail/catch.

samplers/src/test/java/internal/CelBasedSamplerComponentProviderTest.java
```diff
@@ -0,0 +1,246 @@
...
+                + "    expression: 'spanKind == \"CLIENT\"'\n",
+            "cel_based sampler .expressions[].action must be RECORD_AND_SAMPLE or DROP"),
+        Arguments.of(
+            "fallback_sampler:\n"
+                + "  always_on: {}\n"
+                + "expressions:\n"
+                + "  - action: DROP\n"
+                + "    expression: 'invalid cel expression!'\n",
+            "Failed to compile CEL expression: invalid cel expression!"));
+  }
```
breedx-splk: Maybe I missed it, but it would be cool to add coverage for the case where `expressions` exists but is empty.

samplers/src/test/java/io/opentelemetry/contrib/sampler/CelBasedSamplerTest.java
```diff
@@ -0,0 +1,275 @@
...
+  public void setup() throws CelValidationException {
+    when(delegate.shouldSample(any(), any(), any(), any(), any(), any()))
+        .thenReturn(SamplingResult.create(SamplingDecision.RECORD_AND_SAMPLE));
+
+    expressions.add(
+        new CelBasedSamplingExpression(
+            CelBasedSampler.celCompiler
+                .compile(
+                    "spanKind == 'SERVER' && attribute[\""
+                        + URL_FULL
```
breedx-splk: This works because `AttributeKey.toString()` merely returns the key (String), but if the `toString()` one day had type information in it or something, this could fail. It can be made slightly more explicit/robust by using `URL_FULL.getKey()`.

samplers/src/test/java/io/opentelemetry/contrib/sampler/CelBasedSamplerTest.java
```diff
@@ -0,0 +1,275 @@
...
+                        + "\"].matches(\"/actuator\")")
+                .getAst(),
+            Sampler.alwaysOff()));
+    expressions.add(
+        new CelBasedSamplingExpression(
+            CelBasedSampler.celCompiler
+                .compile(
+                    "spanKind == 'SERVER' && attribute[\""
+                        + URL_FULL
```
breedx-splk: same comment as above

samplers/src/test/java/io/opentelemetry/contrib/sampler/CelBasedSamplerTest.java
```diff
@@ -0,0 +1,275 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+@ExtendWith(MockitoExtension.class)
+class CelBasedSamplerTest {
+  private static final String SPAN_NAME = "MySpanName";
+  private static final SpanKind SPAN_KIND = SpanKind.SERVER;
```
breedx-splk: In some of the tests at least, this additional redirection of an existing constant makes the tests slightly harder to read/understand. I'd just inline `SpanKind.SERVER` instead of having this indirection.

breedx-splk (commented):
Thanks! I think this is an awesome addition that users will start trying immediately. I left a few comments/questions that I'd like to see addressed...but this is looking solid.

samplers/README.md
dol: Added clarification to the readme. I mostly aligned the look and feel to the `RuleBasedRoutingSampler`.

samplers/src/main/java/io/opentelemetry/contrib/sampler/CelBasedSamplingExpression.java
dol: The code now should adhere to the contributor style guides. Thank you for pointing this out. This was mainly due to a copy/paste from the `RuleBasedRoutingSampler` files.

samplers/src/main/java/io/opentelemetry/contrib/sampler/internal/CelBasedSamplerComponentProvider.java
dol: Make sense. It's implemented.

samplers/src/test/java/internal/CelBasedSamplerComponentProviderTest.java
dol: I did a total rework of this to replace the `toString()`.

samplers/src/test/java/internal/CelBasedSamplerComponentProviderTest.java
dol: All the config parts are now move to test resources and are loaded from disk relative to the class.

samplers/src/test/java/internal/CelBasedSamplerComponentProviderTest.java
dol: Good catch. Was an easy fix.

samplers/src/test/java/internal/CelBasedSamplerComponentProviderTest.java
dol: A new test case was added for this case. It will result in a similar error as the CEL expression is invalid. I opted for a compile error from the CEL engine rather than an is empty check.

samplers/src/test/java/io/opentelemetry/contrib/sampler/CelBasedSamplerTest.java
dol: Good catch. Is fixed now.

samplers/src/test/java/io/opentelemetry/contrib/sampler/CelBasedSamplerTest.java
dol: Fully agree. Inlined all the const for better readability.

dol:
@breedx-splk Thank you for the valuable review. I addressed all the remarks.
The newest versions returns the detailed message if the CEL compiles is unable to parse the expression.

samplers/README.md
dol: Added a hopefully clarifying information.

samplers/README.md
trask, line 65: @dol can we add you as a component owner here to help us?

you'd need to submit a request to join the @open-telemetry github org: https://github.com/open-telemetry/community/issues/new?template=membership.md

you can use myself and @breedx-splk as your sponsors

dol: @trask Request for a new membership was initiated. Thank you for the sponsorship and the support.

samplers/README.md
```diff
@@ -59,9 +60,70 @@ tracer_provider:
...
+            # Drop spans with specific name patterns
+            - action: DROP
+              expression: name.matches('.*internal.*') && spanKind == 'INTERNAL'
+```
+
 ## Component owners
 
-- [Jack Berg](https://github.com/jack-berg), New Relic
-- [Trask Stalnaker](https://github.com/trask), Microsoft
+* [Jack Berg](https://github.com/jack-berg), New Relic
```
dol: In case someone wondering why the list style changed. markdownlint was complaining about [MD004 - Unordered list style](https://github.com/DavidAnson/markdownlint/blob/v0.38.0/doc/md004.md)

samplers/README.md
```diff
@@ -59,9 +60,70 @@ tracer_provider:
...
+* [Trask Stalnaker](https://github.com/trask), Microsoft
+* [Dominic Lüchinger](https://github.com/dol), SIX Group
```
trask: go ahead and alphabetize

and go ahead and add yourself to https://github.com/open-telemetry/opentelemetry-java-contrib/blob/2c41e66c48fa5015ee96e00f8ef7ecd6c925bdf8/.github/component_owners.yml#L76-L78 (where clearly we don't alphabetize 😅)

breedx-splk:
@dol looks like there are some merge conflicts now...

trask:
I've added this PR as a topic for tomorrow's Java SIG meeting to get a few more eyes on it, in particular since this module is automatically pulled in and available in the OpenTelemetry Java agent (@dol you're of course [welcome to join](https://github.com/open-telemetry/community?tab=readme-ov-file#sig-java) if you'd like!)

samplers/src/test/java/internal/CelBasedSamplerComponentProviderTest.java
```diff
@@ -0,0 +1,244 @@
...
+                + "delegate=AlwaysOnSampler"
+                + "}"
+                + "]},"
+                + "remoteParentSampled:AlwaysOnSampler,"
+                + "remoteParentNotSampled:AlwaysOffSampler,"
+                + "localParentSampled:AlwaysOnSampler,"
+                + "localParentNotSampled:AlwaysOffSampler"
+                + "}");
+
+    // SERVER span to other path should be recorded and sampled
```
jaydeluca: I think this one matches the first expression?
```suggestion
    // SERVER span with host header and 200 response is dropped due to matching with first expression.
```

johnbley:
I'm not against putting this into `contrib` at all but wonder if a better long-term approach would be to have a Java implementation of an `ottl` engine (or a large part of one) for this purpose?  For instance, the collector's `tailsamplingprocessor` reuses ottl for its `ottl_condition`s.  If the same syntax, functions, data model, and scoping semconv could be applied at both the language-level and the collector it would be a big win for coherency.

breedx-splk (changes requested):
@dol thank you for your patience and understanding with this PR.

We definitely think there is room for this sampler to exist, and many of us really appreciate the flexibility that the CEL brings here. We also think that @johnbley 's idea about OTTL is also interesting, but no implementation of OTTL yet exists in java...but maybe one day! 🤞🏻 

So as @trask suggested above, we did indeed talk about this in the SIG meeting this week. The conclusion was that we think it would be best to introduce this as a new contrib module instead of adding it to the existing samplers module. Doing that allows the existing stable java instrumentation agent to continue depending on the existing samplers module, without introducing new dependencies or opening up any potential expression exploit/vulnerability that might be dreamed up.

Based on this, I'm going to request changes on this PR in hopes that you can either adjust this one or make a new one that creates a new gradle module. I hope you understand and can work with us to get this great work merged. 

Thanks again for the contribution and looking forward to seeing it land as a new module here.

dol:
@breedx-splk Thanks for the clarification on how to go forward. I'll go ahead and perform the requested changes to outsource the CEL sampler to it's own gradle module.

I was not aware that the OTEL collector already has a similar approeach in complex filtering with it's own DSL, the `OpenTelemetry Transformation Language (OTTL)`.
I agree that I would make sense to have the one common syntax like the OTTL for processing and sampling.

.github/component_owners.yml
```diff
@@ -25,6 +25,8 @@ components:
   baggage-processor:
     - mikegoldsmith
     - zeitlinger
+  cel-sampler:
+    - dol
```
breedx-splk, line 29: Let's also add the existing samplers component owners as well, so it's not just you alone:

```suggestion
    - dol
    - trask
    - jackberg
    - breedx-splk    
```

Since I've been reviewing I'm happy to help out.

cel-sampler/README.md
```diff
@@ -0,0 +1,80 @@
...
+              expression: name.contains('checkout') || name.contains('payment')
+            # Drop spans with specific name patterns
+            - action: DROP
+              expression: name.matches('.*internal.*') && spanKind == 'INTERNAL'
+```
+
+## Component owners
+
+* [Dominic Lüchinger](https://github.com/dol), SIX Group
+* TBD
```
breedx-splk: Let's add @trask and @jack-berg who are also owners of the other sampling module. Add @breedx-splk too since I'm helping out.

cel-sampler/src/main/java/io/opentelemetry/contrib/sampler/cel/CelBasedSamplingExpression.java
```diff
@@ -0,0 +1,95 @@
...
+   *
+   * @param abstractSyntaxTree The compiled CEL abstract syntax tree
+   * @param delegate The sampler to use when this expression evaluates to true
+   */
+  CelBasedSamplingExpression(CelAbstractSyntaxTree abstractSyntaxTree, Sampler delegate) {
+    this.abstractSyntaxTree =
+        requireNonNull(abstractSyntaxTree, "abstractSyntaxTree must not be null");
+    this.expression = abstractSyntaxTree.getSource().getContent().toString();
+    this.delegate = requireNonNull(delegate, "delegate must not be null");
+  }
```
breedx-splk, line 36: nit: This is a slight violation of dependency injection principles because it does a little work in the constructor to obtain the true dependency -- which is the string version of the content from the source of the AST. 

Now that I'm looking at it tho, it's not even a real dependency...so I think it would be better to remove the `expression` field entirely and just do the getter chain in `getExpression()`.

cel-sampler/src/main/java/io/opentelemetry/contrib/sampler/cel/CelBasedSampler.java
```diff
@@ -0,0 +1,157 @@
...
+   */
+  public CelBasedSampler(List<CelBasedSamplingExpression> expressions, Sampler fallback) {
+    this.expressions = requireNonNull(expressions, "expressions must not be null");
+    this.expressions.forEach(
+        expr -> {
+          if (!expr.getAbstractSyntaxTree().isChecked()) {
+            throw new IllegalArgumentException(
+                "Expression and its AST is not checked: " + expr.getExpression());
+          }
+        });
```
breedx-splk: nit: Prefer to copy list into new private field to ensure deep immutability for the life of this object.

cel-sampler/src/main/java/io/opentelemetry/contrib/sampler/cel/CelBasedSampler.java
```diff
@@ -0,0 +1,157 @@
...
+    evaluationContext.put("traceId", traceId);
+    evaluationContext.put("spanKind", spanKind.name());
+    evaluationContext.put("attribute", convertAttributesToMap(attributes));
+
+    for (CelBasedSamplingExpression expression : expressions) {
+      try {
+        CelRuntime.Program program = celRuntime.createProgram(expression.getAbstractSyntaxTree());
+        Object result = program.eval(evaluationContext);
+        // Happy path: Perform sampling based on the boolean result
+        if (result instanceof Boolean && ((Boolean) result)) {
```
breedx-splk:
```suggestion
        if (Boolean.TRUE.equals(result)) {
```

breedx-splk (commented):
This is nearly there! I had a few small comments, and I do think it's helpful to expand the component owners list. Please also add this project to the topmost `README.md`  in the repo.

Thanks again!

.github/component_owners.yml
dol: Will do. Makes sense. Will fix `s/jackberg/jack-berg/`

cel-sampler/src/main/java/io/opentelemetry/contrib/sampler/cel/CelBasedSampler.java
dol: Thanks for the hint. It's now copied to an immutable list.

.github/component_owners.yml
breedx-splk: Whoops, thanks, good catch. :)

breedx-splk (approved):

PR #1959, opened 2025-06-12
[aws-xray] Update SamplerRulesApplier to recognize new HTTP/URL semconv. 

AsakerMohd:
**Description:**
The HTTP/URL semantic conventions were updated making the X-Ray Sampler outdated since it only recognizes the older HTTP semantic conventions. I've updated the SamplingRuleApplier to look at both the old and new conventions for backwards compatibility. 

**Testing:**
Updated unit tests and are now passing. 

**Manual Testing:**

For the manual testing. I created three sampling rules in the X-Ray console
1. Default one with zero sampling
6. Second one with path = `/aws-sdk-call` with 100% sampling
7. Third one with path = `/outgoing-http-call` with 100% sampling

I used a sample app with two APIs:
1. /outgoing-http-call
2. /aws-sdk-call

Before the change, calling any of the above APIs wasn't being matched when it should. After the change, the rules were being matched and each rule showed the correct stats. 

**Before the change:**
Screenshot 2025-06-01 at 11 34 56 PM

**After the change:**
Screenshot 2025-06-01 at 11 29 54 PM


aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/SamplingRuleApplier.java
```diff
@@ -181,11 +183,12 @@ boolean matches(Attributes attributes, Resource resource) {
     String host = null;
 
     for (Map.Entry<AttributeKey<?>, Object> entry : attributes.asMap().entrySet()) {
-      if (entry.getKey().equals(HTTP_TARGET)) {
+      if (entry.getKey().equals(HTTP_TARGET) || entry.getKey().equals(UrlAttributes.URL_PATH)) {
         httpTarget = (String) entry.getValue();
-      } else if (entry.getKey().equals(HTTP_URL)) {
+      } else if (entry.getKey().equals(HTTP_URL) || entry.getKey().equals(UrlAttributes.URL_FULL)) {
```
laurit: server instrumentations don't fill `url.full`, would that be an issue?

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/SamplingRuleApplier.java
```diff
@@ -181,11 +183,12 @@ boolean matches(Attributes attributes, Resource resource) {
...
         httpUrl = (String) entry.getValue();
-      } else if (entry.getKey().equals(HTTP_METHOD)) {
+      } else if (entry.getKey().equals(HTTP_METHOD)
+          || entry.getKey().equals(HttpAttributes.HTTP_REQUEST_METHOD)) {
```
laurit: there is also `http.request.method_original` that could be used to get the request method when `http.request.method` is  `_OTHER`

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/SamplingRuleApplier.java
```diff
@@ -181,11 +183,12 @@ boolean matches(Attributes attributes, Resource resource) {
...
         httpMethod = (String) entry.getValue();
       } else if (entry.getKey().equals(NET_HOST_NAME)) {
```
laurit: `net.host.name` is not filled any more, try `server.address` instead, `http.host` is also not filled (also replaced with `server.address`)

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/SamplingRuleApplier.java
AsakerMohd: That should be fine. It's there incase URL_PATH isn't being populated and the idea here is to then extract the path from URL_FULL.

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/SamplingRuleApplier.java
AsakerMohd: Let me take a look.

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/SamplingRuleApplier.java
AsakerMohd: Thanks for pointing it out. Let me take a look as well.

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/SamplingRuleApplier.java
AsakerMohd: Added logic to check for _OTHER and to use `http.request.method_original` in that case.

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/SamplingRuleApplier.java
AsakerMohd: Updated.

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/SamplingRuleApplier.java
```diff
@@ -181,15 +188,24 @@ boolean matches(Attributes attributes, Resource resource) {
     String host = null;
 
     for (Map.Entry<AttributeKey<?>, Object> entry : attributes.asMap().entrySet()) {
-      if (entry.getKey().equals(HTTP_TARGET)) {
+      if (entry.getKey().equals(HTTP_TARGET) || entry.getKey().equals(UrlAttributes.URL_PATH)) {
```
laurit: Now that there are multiple keys for the same value it might be better to move these out of the loop.
```java
    String httpTarget = attributes.get(URL_PATH);
    if (httpTarget == null) {
      httpTarget = attributes.get(HTTP_TARGET);
    }
    ...
    for (Map.Entry<AttributeKey<?>, Object> entry : attributes.asMap().entrySet()) {
```
If the attributes set contains both attributes then it would be predictable which is used.

AsakerMohd: Good idea. done!

srprash (approved):

aws-xray/src/test/java/io/opentelemetry/contrib/awsxray/SamplingRuleApplierTest.java
```diff
@@ -72,6 +74,15 @@ class ExactMatch {
             .put(AttributeKey.longKey("speed"), 10)
             .build();
 
+    private final Attributes newSemCovAttributes =
```
laurit: we usually call it stable semconv, semantic conventions is usually abbreviated as semconv not semcov

aws-xray/src/test/java/io/opentelemetry/contrib/awsxray/SamplingRuleApplierTest.java
```diff
@@ -72,6 +74,15 @@ class ExactMatch {
...
+        Attributes.builder()
+            .put(HttpAttributes.HTTP_REQUEST_METHOD, "GET")
+            .put(NET_HOST_NAME, "opentelemetry.io")
```
laurit: this is not part of the stable semconv, perhaps it would be better to replace it with `server.address`?

aws-xray/src/test/java/io/opentelemetry/contrib/awsxray/SamplingRuleApplierTest.java
```diff
@@ -120,6 +131,21 @@ void matches() {
           .isTrue();
     }
 
+    @Test
+    void matchesURLFullNewSemCov() {
+      assertThat(applier.matches(newSemCovAttributes, resource)).isTrue();
+
+      // http.url works too
```
laurit: looks like this comment was copy-pasted from another test, it isn't relevant here

AsakerMohd: updated

aws-xray/src/test/java/io/opentelemetry/contrib/awsxray/SamplingRuleApplierTest.java
AsakerMohd: Good catch. Replaced it with server.address and added new unit tests for it.

aws-xray/src/test/java/io/opentelemetry/contrib/awsxray/SamplingRuleApplierTest.java
AsakerMohd: changed abrv to `semconv` instead of `semcov`

laurit (approved):

Issue #1941, opened 2025-06-06
New component: ibm-mq monitoring


breedx-splk:
[IBM MQ](https://en.wikipedia.org/wiki/IBM_MQ) (formerly called WebSphere MQ) is a messaging system which currently lacks OpenTelemetry instrumentation. The platform itself is capable of receiving diagnostic request messages and will respond to them with messages that contain telemetry.

In order to surface this telemetry as metrics, Splunk has began building a system that polls IBM MQ periodically, parses the response messages, and generates synchronous metrics. The repository is here:  https://github.com/signalfx/opentelemetry-ibm-mq-monitoring and the current list of metrics is here https://github.com/signalfx/opentelemetry-ibm-mq-monitoring/blob/main/docs/metrics.md.

The project uses a metrics definition (in yaml format) that is munged by the otel [weaver](https://github.com/open-telemetry/weaver) tool in order to generate documentation (markdown) and code (java). The repository is somewhat young, but it contains a set of integration tests as well as an "end-to-end" gold standard test that verifies that expected metrics are generated.

Splunk would like to add this utility as a new component in `opentelemetry-java-contrib` and will offer to be the component owners for it. We discussed this in the java SIG meeting on June 5th, 2025 and the response seemed favorable. Creating this issue now for tracking the work of standing up the new module. Thanks! 👍🏻

atoulme:
Closing as complete.

PR #1960, opened 2025-06-13
Add ibm-mq-metrics

breedx-splk:
This is the first part of #1941. It adds a new component called `ibm-mq-metrics` which provides a runnable utility that can fetch metrics from IBM MQ series and emit them as otel metrics. It is based on https://github.com/signalfx/opentelemetry-ibm-mq-monitoring. This code is alpha quality.

There are a few pieces that are still missing from this PR:
* e2e tests. We have a set of end-to-end "golden" tests that validate data using a docker-compose environment. These have not yet been added, due in part to the `golden` image only being available in a splunk registry (see [#40721 for more info](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/40721))
* ability to run the `integrationTest` target. This requires running the ibm mq server as a docker container, which is not yet included.
* workflows to run the above


ibm-mq-metrics/src/main/java/io/opentelemetry/ibm/mq/metrics/MetricsConfig.java
```diff
@@ -0,0 +1,213 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.ibm.mq.metrics;
```
mcmho, line 6: @breedx-splk   I noticed that we have package io.opentelemetry.ibm.mq.config;    I wonder what is the reason to package this in a different location?

ibm-mq-metrics/src/main/java/io/opentelemetry/ibm/mq/opentelemetry/Config.java
```diff
@@ -0,0 +1,83 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.ibm.mq.opentelemetry;
```
mcmho, line 6: i also wonder if this should be under package io.opentelemetry.ibm.mq.config as well.

ibm-mq-metrics/model/metrics.yaml
```diff
@@ -0,0 +1,611 @@
+groups:
+  - id: mq.message.retry.count
```
trask: this is new-ish semconv convention for top-level namespacing (https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/naming.md#choosing-a-system-name)

```suggestion
  - id: ibm.mq.message.retry.count
```

ibm-mq-metrics/model/attributes.yaml
```diff
@@ -0,0 +1,71 @@
+groups:
+  - id: shared.attributes
+    type: attribute_group
+    brief: Attributes of metrics.
+    attributes:
+      - id: queue.manager
```
trask: or we can see if there's a general attribute under `messaging.*` already
```suggestion
      - id: ibm.mq.queue.manager
```

ibm-mq-metrics/src/main/java/io/opentelemetry/ibm/mq/metrics/MetricsConfig.java
breedx-splk: Both `MetricsConfig` and its neighbor `Config` are code generated with weaver. I didn't pick the original package name, so we _could_ change it, but it does simplify things to have these output to the same location. We could consider a different package name, like `codegen` or something, or we could bite the bullet and write them to different places, but I like keeping them grouped.

ibm-mq-metrics/src/main/java/io/opentelemetry/ibm/mq/opentelemetry/Config.java
breedx-splk: I think it was put here to simply group the codegen/weaver output in one place.

ibm-mq-metrics/model/metrics.yaml
breedx-splk: Thanks, ok, I think that makes sense. I mean, to the extent that having the semconv here at all makes sense. :)

Do you think we should just run down the entire list here and prefix them all?

trask: 👍

ibm-mq-metrics/src/main/java/io/opentelemetry/ibm/mq/metrics/MetricsConfig.java
mcmho: ok, that's fine for now i guess.

atoulme: We could also move this code under a different source folder

ibm-mq-metrics/model/metrics.yaml
```diff
@@ -0,0 +1,611 @@
+groups:
+  - id: ibm.mq.message.retry.count
```
mcmho, line 2: looks good 👍🏻

ibm-mq-metrics/model/attributes.yaml
mcmho: ibm.mq.queue.manager     @breedx-splk?

ibm-mq-metrics/src/integrationTest/resources/conf/test-queuemgr-config.yml
```diff
@@ -0,0 +1,170 @@
...
+        # https://www.ibm.com/support/knowledgecenter/en/SSFKSJ_7.5.0/com.ibm.mq.pla.doc/q005020_.htm
+        include: ["#"]
+        exclude:
+             #type value: STARTSWITH, EQUALS, ENDSWITH, CONTAINS
+           - type: "STARTSWITH"
+             #The name of the queue or queue name pattern as per queue filter, comma separated values
+             values: ["SYSTEM","$SYS"]
+
+metrics:
+  "mq.message.retry.count":  # Number of message retries
```
mcmho: do these need to have the 'ibm' prefix?

breedx-splk: yes, should be updated now.

breedx-splk:
Ok, so I've done an update that does the following:
* updates most metric names to include the `ibm.mq` namespace prefix
* updates most attribute names to include the `ibm.mq` prefix
* creates a new attributes constant class, to reduce the duplication of string constants, which is a maintainability problem.

Thanks for looking.

mcmho (approved):

breedx-splk:
> thank you @breedx-splk for all the update! So we are waiting for @trask to give the final blessing?

Not necessarily @trask, but any of the approvers/maintainers of this repo.

ibm-mq-metrics/README.md
```diff
@@ -0,0 +1,225 @@
...
+* `SYSTEM.ADMIN.CONFIG.EVENT`: Configuration events
+
+Please turn on those events to take advantage of this monitoring.
+
+## Build
+
+Build the package with:
+
+```shell
+./gradlew shadowJar
```
laurit: maybe just building this module with `./gradlew :ibm-mq-metrics:shadowJar` would be better

ibm-mq-metrics/README.md
```diff
@@ -0,0 +1,225 @@
...
+```
+
+Note: Due to restrictive licensing, this uber-jar (fat-jar) does not include the IBM client jar f
```
laurit: `jar f` looks like a typo

ibm-mq-metrics/README.md
```diff
@@ -0,0 +1,225 @@
...
+
+Run the standalone jar alongside the IBM jar:
+
+```shell
+java \
+   -Djavax.net.ssl.keyStore=key.jks \
+   -Djavax.net.ssl.keyStorePassword=<password> \
+   -Djavax.net.ssl.trustStore=key.jks \
+   -Djavax.net.ssl.trustStorePassword=<password> \
+   -cp target/ibm-mq-monitoring-<version>-all.jar:lib/com.ibm.mq.allclient.jar \
```
laurit: should be `ibm-mq-metrics/build/libs/opentelemetry-ibm-mq-metrics-<version>-all.jar`

ibm-mq-metrics/README.md
```diff
@@ -0,0 +1,225 @@
...
+The user connecting to the queueManager should have the inquire, get, put (since PCF responses
+cause dynamic queues to be created) permissions. For metrics that execute MQCMD_RESET_Q_STATS
+command, chg permission is needed.
+
+### SSL Support
+
+1. Configure the IBM SSL Cipher Suite in the config.yml.
+   Note that, to use some CipherSuites the unrestricted policy needs to be configured in JRE.
+   Please visit [this link](http://www.ibm.com/support/knowledgecenter/SSYKE2_8.0.0/com.ibm.java.security.component.80.doc/security-component/sdkpolicyfiles.html)
+   for more details. For Oracle JRE, please update with [JCE Unlimited Strength Jurisdiction Policy](http://www.oracle.com/technetwork/java/javase/downloads/jce8-download-2133166.html).
```
laurit, line 117: I think it would be a good idea to mention that this is not needed on oracle 9 and later and also in 8 since u161 https://www.oracle.com/java/technologies/javase/8u161-relnotes.html#JDK-8170157

ibm-mq-metrics/README.md
```diff
@@ -0,0 +1,225 @@
...
+`modelQueueName` and `replyQueuePrefix` fields in the config.yml.
+[More details mentioned here](https://www.ibm.com/support/knowledgecenter/SSFKSJ_7.5.0/com.ibm.mq.ref.adm.doc/q083240_.htm)
+
+## Metrics
+
+See [docs/metrics.md](docs/metrics.md).
+
+## Troubleshooting
+
+1. Please follow the steps listed in this [troubleshooting-document](https://community.appdynamics.com/t5/Knowledge-Base/How-to-troubleshoot-missing-custom-metrics-or-extensions-metrics/ta-p/28695) in order to troubleshoot your issue. These are a set of common issues that customers might have faced during the installation of the extension.
```
laurit: this seems to be for appd, is it relevant here?

ibm-mq-metrics/README.md
```diff
@@ -0,0 +1,225 @@
...
+4. `MQJE001: Completion Code '2', Reason '2195'`
+   This could happen in **Client** mode. Please make sure that the IBM MQ dependency jars are correctly referenced in classpath of monitor.xml
+
+5. `MQJE001: Completion Code '2', Reason '2400'`
+   This could happen if unsupported cipherSuite is provided or JRE not having/enabled unlimited jurisdiction policy files. Please check SSL Support section.
+
+6. If you are seeing "NoClassDefFoundError" or "ClassNotFound" error for any of the MQ dependency even after providing correct path in monitor.xml, then you can also try copying all the required jars in WMQMonitor (MAHome/monitors/WMQMonitor) folder and provide classpath in monitor.xml like below
+
+   ```
+    <classpath>ibm-mq-monitoring-<version>-all.jar;com.ibm.mq.allclient.jar</classpath>
```
laurit: should be `opentelemetry-ibm-mq-metrics-<version>-all.jar`

ibm-mq-metrics/README.md
```diff
@@ -0,0 +1,225 @@
...
+   traffics from other ips, you need to contact admin to provide you access to the queue manager.
+   For Bindings mode, please make sure that the MA is owned by a mqm user.
+
+4. `MQJE001: Completion Code '2', Reason '2195'`
+   This could happen in **Client** mode. Please make sure that the IBM MQ dependency jars are correctly referenced in classpath of monitor.xml
+
+5. `MQJE001: Completion Code '2', Reason '2400'`
+   This could happen if unsupported cipherSuite is provided or JRE not having/enabled unlimited jurisdiction policy files. Please check SSL Support section.
+
+6. If you are seeing "NoClassDefFoundError" or "ClassNotFound" error for any of the MQ dependency even after providing correct path in monitor.xml, then you can also try copying all the required jars in WMQMonitor (MAHome/monitors/WMQMonitor) folder and provide classpath in monitor.xml like below
```
laurit: should be `ClassNotFoundException` instead of `ClassNotFound`

ibm-mq-metrics/README.md
```diff
@@ -0,0 +1,225 @@
...
+
+   ```
+    <classpath>ibm-mq-monitoring-<version>-all.jar;com.ibm.mq.allclient.jar</classpath>
+   ```
+
+   OR
+
+   ```
+    <classpath>ibm-mq-monitoring-<version>-all.jar;com.ibm.mq.jar;com.ibm.mq.jmqi.jar;com.ibm.mq.commonservices.jar;com.ibm.mq.headers.jar;com.ibm.mq.pcf.jar;connector.jar;dhbcore.jar</classpath>
```
laurit: should be `opentelemetry-ibm-mq-metrics-<version>-all.jar`

ibm-mq-metrics/build.gradle.kts
```diff
@@ -0,0 +1,79 @@
+plugins {
+  application
+  id("com.github.johnrengelman.shadow")
+  id("otel.java-conventions")
+  id("otel.publish-conventions")
+}
+
+description = "IBM-MQ metrics"
+otelJava.moduleName.set("io.opentelemetry.contrib.jmxscraper")
+application.mainClass.set("io.opentelemetry.contrib.jmxscraper.JmxScraper")
```
laurit: I believe these aren't correct

ibm-mq-metrics/build.gradle.kts
```diff
@@ -0,0 +1,79 @@
...
+  testImplementation("com.google.guava:guava")
+  testImplementation("io.opentelemetry:opentelemetry-sdk-testing:1.50.0")
+  integrationTestImplementation("org.assertj:assertj-core:3.27.3")
+  integrationTestImplementation("org.junit.jupiter:junit-jupiter-api:5.12.2")
+  integrationTestImplementation("io.opentelemetry:opentelemetry-sdk-testing:1.50.0")
+  integrationTestImplementation("com.ibm.mq:com.ibm.mq.jakarta.client:9.4.2.0")
+  integrationTestImplementation("jakarta.jms:jakarta.jms-api:3.1.0")
+  integrationTestImplementation("org.junit.jupiter:junit-jupiter-engine:5.12.2")
+  integrationTestRuntimeOnly("org.junit.platform:junit-platform-launcher:1.13.0")
+  ibmClientJar("com.ibm.mq:com.ibm.mq.allclient:9.4.2.1") {
```
laurit, line 51: this is also present as `api` dependency

ibm-mq-metrics/docs/metrics.md
```diff
@@ -0,0 +1,801 @@
...
+| `ibm.mq.job.name` | string | The job name | `0000074900000003` | `Required` | ![Development](https://img.shields.io/badge/-development-blue) |
+| `ibm.mq.queue.manager` | string | The name of the IBM queue manager | `MQ1` | `Required` | ![Development](https://img.shields.io/badge/-development-blue) |
+
+
+
+## Metric `ibm.mq.byte.received`
+
+| Name     | Instrument Type | Unit (UCUM) | Description    | Stability |
+| -------- | --------------- | ----------- | -------------- | --------- |
+| `ibm.mq.byte.received` | Gauge | `{bytes}` | Number of bytes received | ![Development](https://img.shields.io/badge/-development-blue) |
```
laurit: Jvm metrics use `By` as unit.
should be a counter?

ibm-mq-metrics/docs/metrics.md
```diff
@@ -0,0 +1,801 @@
+# Produced Metrics
+
+
+## Metric `ibm.mq.message.retry.count`
+
+| Name     | Instrument Type | Unit (UCUM) | Description    | Stability |
+| -------- | --------------- | ----------- | -------------- | --------- |
+| `ibm.mq.message.retry.count` | Gauge | `{messages}` | Number of message retries | ![Development](https://img.shields.io/badge/-development-blue) |
```
laurit: should be singular `{message}`
should be a counter?

ibm-mq-metrics/docs/metrics.md
```diff
@@ -0,0 +1,801 @@
...
+
+**[9] `messaging.destination.name`:** This is duplicated from otel semantic-conventions.
+
+
+
+## Metric `ibm.mq.current.max.queue.filesize`
+
+| Name     | Instrument Type | Unit (UCUM) | Description    | Stability |
+| -------- | --------------- | ----------- | -------------- | --------- |
+| `ibm.mq.current.max.queue.filesize` | Gauge | `mib` | Current maximum queue file size | ![Development](https://img.shields.io/badge/-development-blue) |
```
laurit: https://github.com/open-telemetry/semantic-conventions/blob/69f40e1ccdb32e28c6ca63a44d7348a07760aeef/docs/general/metrics.md?plain=1#L114 use `MiBy` as the unit. It also suggest that instrumentations should use non-prefixed units

ibm-mq-metrics/docs/metrics.md
```diff
@@ -0,0 +1,801 @@
...
+| `ibm.mq.job.name` | string | The job name | `0000074900000003` | `Required` | ![Development](https://img.shields.io/badge/-development-blue) |
+| `ibm.mq.queue.manager` | string | The name of the IBM queue manager | `MQ1` | `Required` | ![Development](https://img.shields.io/badge/-development-blue) |
+
+
+
+## Metric `ibm.mq.max.sharing.conversations`
+
+| Name     | Instrument Type | Unit (UCUM) | Description    | Stability |
+| -------- | --------------- | ----------- | -------------- | --------- |
+| `ibm.mq.max.sharing.conversations` | Gauge | `{conversations}` | Maximum number of conversations permitted on this channel instance. | ![Development](https://img.shields.io/badge/-development-blue) |
```
laurit: should have a `.count` suffix?
should use singular `conversation` instead of `conversations` in metric name?
should be an up down counter?
@PeterF778 you are more knowledgeable about this, could you check whether the instrument names, units and types are correct here.

ibm-mq-metrics/src/integrationTest/resources/log4j2.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Configuration status="WARN" monitorInterval="30">
+
+    <Appenders>
+        <!-- Console appender: nice for quick output during Maven run -->
```
laurit: not using maven

ibm-mq-metrics/src/integrationTest/resources/log4j2.xml
```diff
@@ -0,0 +1,29 @@
...
+        <Console name="Console" target="SYSTEM_OUT">
+            <PatternLayout pattern="%d{yyyy-MM-dd HH:mm:ss.SSS} [%t] %-5level %logger{36} - %msg%n" />
+        </Console>
+
+        <!-- File appender: capture full logs for deeper analysis -->
+        <File name="IntegrationTestFile" fileName="target/integration-test.log" immediateFlush="true">
```
laurit: could write the log to `build` instead of `target`

ibm-mq-metrics/src/main/java/io/opentelemetry/ibm/mq/WmqMonitor.java
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    // Disconnect queue manager
+    if (ibmQueueManager != null) {
+      String name = "";
+      try {
+        name = ibmQueueManager.getName();
+        ibmQueueManager.disconnect();
+        // logger.debug("Connection disconnected for queue manager {} in thread {}",
+        // ibmQueueManager.getName(), Thread.currentThread().getName());
```
laurit: commented out code

ibm-mq-metrics/src/test/resources/log4j.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
laurit: as far as I can tell tests don't include log4j

ibm-mq-metrics/src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,59 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
laurit: as far as I can tell log4j isn't included in the fat jar

ibm-mq-metrics/src/main/java/io/opentelemetry/ibm/mq/opentelemetry/Main.java
```diff
@@ -0,0 +1,91 @@
...
+
+    run(config, service, otel);
+  }
+
+  @VisibleForTesting
+  public static void run(
+      ConfigWrapper config, ScheduledExecutorService service, OpenTelemetry otel) {
+    MeterProvider meterProvider = otel.getMeterProvider();
+
+    Runtime.getRuntime().addShutdownHook(new Thread(service::shutdown));
```
laurit: I suspect that this doesn't do anything. Since the `ScheduledExecutorService` used here doesn't use daemon threads (and if it did then app would terminate when the main thread exits).

ibm-mq-metrics/src/main/java/io/opentelemetry/ibm/mq/opentelemetry/Main.java
```diff
@@ -0,0 +1,91 @@
...
+  private Main() {}
+
+  public static void main(String[] args) throws Exception {
+    if (args.length == 0) {
+      System.err.println("Usage: Main <config-file>");
+      System.exit(1);
+    }
+
+    try {
+      Main.class.getClassLoader().loadClass("com.ibm.mq.headers.MQDataException");
```
laurit: usually `Class.forName` is used for class presence checks

ibm-mq-metrics/README.md
breedx-splk: Yeah, this was fallout from when it was a stand-alone project.  Another option is to assume the current directory is this module, so like
```
cd ibm-mq-metrics
../gradlew shadowJar
```
Do you have a preference?

ibm-mq-metrics/README.md
breedx-splk: Thanks, yeah good catch from when this was a standalone maven project. 👍🏻

ibm-mq-metrics/README.md
breedx-splk: Seems then like it would be safe to say just "java 8 below 8u161" then right?

ibm-mq-metrics/README.md
breedx-splk: Nah, shouldn't be here. Thanks, will remove.

ibm-mq-metrics/build.gradle.kts
breedx-splk: Copypasta!? WHAT COPYPASTA!? :) Thanks.

ibm-mq-metrics/build.gradle.kts
breedx-splk: true. isn't this a different usage?

ibm-mq-metrics/docs/metrics.md
breedx-splk: In concept, sure, this seems like it should be a counter. But the values received from mq are absolute, cumulative values, and the `LongCounter` api has no way to set the value directly (only add). So using a counter I think would break accuracy.

ibm-mq-metrics/docs/metrics.md
breedx-splk: Agreed about the singular, but same challenge with receiving absolute values from mq.

ibm-mq-metrics/README.md
breedx-splk: I chose the two-line current directory version. LMK if you really hate it (I don't feel strongly).

ibm-mq-metrics/docs/metrics.md
breedx-splk: Switched to By and did the unit conversion in code.

ibm-mq-metrics/docs/metrics.md
PeterF778: I think the semantic conventions for metrics suggest something like `ibm.mq.sharing.conversation.limit` for this case.

ibm-mq-metrics/docs/metrics.md
PeterF778: Yes, most of these metrics should be Counters. If you used asynchronous measurements (LongCounterBuilder), you could set the counter values directly. Perhaps it is worth considering.

ibm-mq-metrics/docs/metrics.md
PeterF778: With respect to the ".count" suffix, I cannot find the relevant page, but I believe it is supposed to be used only if the preceding part of the name can be used as metric namespace. This is to avoid having metric names identical to metric namespaces, which is not forbidden, but discouraged.

ibm-mq-metrics/docs/metrics.md
laurit: In jvm metrics counters use async measurements (`buildWithCallback` instead of `build`)  https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/1115cda7a2f4aefa0bbc25aa64f9b35f694a2df4/instrumentation/runtime-telemetry/runtime-telemetry-java8/library/src/main/java/io/opentelemetry/instrumentation/runtimemetrics/java8/Classes.java#L53 You cold keep the absolute value in an `AtomicLong` and have the callback read it. Another option would be to use a `BatchCallback`, like in db connection pool metrics, but I think here it could require a lot of work.

ibm-mq-metrics/src/main/resources/log4j2.xml
breedx-splk: Removed the configs....I agree they don't seem to be needed.

ibm-mq-metrics/src/test/resources/log4j.xml
breedx-splk: removed

ibm-mq-metrics/docs/metrics.md
breedx-splk: This all makes sense, and I even made an attempt to see how complicated it might be....and it's considerably more challenging than I had expected. This is because some of the collector classes ([like this one](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1960/files#diff-5e467f091480fa027351585744ebd6c838f700b8546f83fac5f2ca56651171f4R94)) are able to be used across multiple queue managers and multiple queues. 

The queue name and queue manager name are metric Attributes that are built dynamically at the time the value is recorded! The async counter doesn't have a way to specify attributes at record time, so we would need to know, at creation time, the permutation of all the queues being monitored for each queue manager, and we'd have to precompute the attributes in advance of creating the counter. 

From what I can tell, the queue names are NOT known in advance (there's just a filter for the names, and the filter is applied when reading responses).

So yeah, given all that, can we please keep these as inappropriate gauges for now?

PeterF778: Yes, it looks like the OTel API/SDK does not have a good solution for this case. I agree that we should not block this PR, and let's lobby for a better support for "synchronous" Counters.

PeterF778 (approved):

ibm-mq-metrics/docs/metrics.md
laurit: @jack-berg ^^^ critique on usability (or lack of it) of otel metrics api. @breedx-splk I totally get your frustration. For synchronous counters otel spec only defines the add operation. https://opentelemetry.io/docs/specs/otel/metrics/supplementary-guidelines/ suggest using an asynchronous counter for monotonically increasing additive values. Usually it isn't too hard, you can record your values into an `AtomicLong` and have the counter callback read them. If you need record multiple values with different attributes I believe you could write them into a map with attributes as keys. In the callback you can then record all the map elements. This isn't super easy as you have to consider the concurrency and make sure you properly synchronize the map access. Another option would be to use the batch callback, but that might be hard to use here.

ibm-mq-metrics/model/metrics.yaml
```diff
@@ -0,0 +1,611 @@
...
+    stability: development
+    brief: "The number of full queue events"
+    instrument: counter
+    unit: "{event}"
+    attributes:
+      - ref: ibm.mq.queue.manager
+        requirement_level: required
+      - ref: messaging.destination.name
+        requirement_level: required
+  - id: ibm.mq..queue.depth.high.event
```
trask: how did weaver not catch this?
```suggestion
  - id: ibm.mq.queue.depth.high.event
```

ibm-mq-metrics/model/metrics.yaml
```diff
@@ -0,0 +1,611 @@
...
+    attributes:
+      - ref: ibm.mq.queue.manager
+        requirement_level: required
+      - ref: messaging.destination.name
+        requirement_level: required
+      - ref: ibm.mq.queue.type
+        requirement_level: required
+  - id: ibm.mq.onqtime.1
+    type: metric
+    metric_name: ibm.mq.onqtime.1
```
trask: should this be a semconv rule for namespaces not to start with numbers?

```suggestion
    metric_name: ibm.mq.onqtime.short_window/period/window_1
```

PR #1969, opened 2025-06-20
OpAMP WebSocket service

LikeTheSalad:
**Description:**

This is a continuation of the OpAMP implementation work. The changes here are focused on adding the [RequestService's](https://github.com/open-telemetry/opentelemetry-java-contrib/blob/37f0278c849184f3b96622847634abedd89c578e/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/RequestService.java#L20) WebSocket implementation.

### DIFF
--- a/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/WebSocketRequestService.java
+++ b/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/WebSocketRequestService.java
@@ -29,15 +29,14 @@ import opamp.proto.ServerErrorResponseType;
 import opamp.proto.ServerToAgent;
 
 public final class WebSocketRequestService implements RequestService, WebSocket.Listener {
+  private static final PeriodicDelay DEFAULT_DELAY_BETWEEN_RETRIES =
+      PeriodicDelay.ofFixedDuration(Duration.ofSeconds(30));
+
   private final WebSocket webSocket;
-  private final PeriodicDelay periodicRetryDelay;
-  private final AtomicBoolean retryingConnection = new AtomicBoolean(false);
-  private final AtomicBoolean nextRetryScheduled = new AtomicBoolean(false);
   private final AtomicBoolean isRunning = new AtomicBoolean(false);
   private final AtomicBoolean hasStopped = new AtomicBoolean(false);
+  private final ConnectionStatus connectionStatus;
   private final ScheduledExecutorService executorService;
-  public static final PeriodicDelay DEFAULT_DELAY_BETWEEN_RETRIES =
-      PeriodicDelay.ofFixedDuration(Duration.ofSeconds(30));
 
   /** Defined <a href="https://datatracker.ietf.org/doc/html/rfc6455#section-7.4.1">here</a>. */
   private static final int WEBSOCKET_NORMAL_CLOSURE_CODE = 1000;
@@ -77,8 +76,8 @@ public final class WebSocketRequestService implements RequestService, WebSocket.
       PeriodicDelay periodicRetryDelay,
       ScheduledExecutorService executorService) {
     this.webSocket = webSocket;
-    this.periodicRetryDelay = periodicRetryDelay;
     this.executorService = executorService;
+    this.connectionStatus = new ConnectionStatus(periodicRetryDelay);
   }
 
   @Override
@@ -154,7 +153,7 @@ public final class WebSocketRequestService implements RequestService, WebSocket.
 
   @Override
   public void onOpen() {
-    retryingConnection.set(false);
+    connectionStatus.success();
     getCallback().onConnectionSuccess();
     synchronized (hasPendingRequestLock) {
       if (hasPendingRequest) {
@@ -200,34 +199,12 @@ public final class WebSocketRequestService implements RequestService, WebSocket.
       }
 
       webSocket.close(WEBSOCKET_NORMAL_CLOSURE_CODE, null);
-      scheduleConnectionRetry(retryAfter);
+      connectionStatus.retryAfter(retryAfter);
     }
   }
 
   private static boolean serverIsUnavailable(ServerErrorResponse errorResponse) {
     return errorResponse.type.equals(ServerErrorResponseType.ServerErrorResponseType_Unavailable);
-  }
-
-  @SuppressWarnings("FutureReturnValueIgnored")
-  private void scheduleConnectionRetry(@Nullable Duration retryAfter) {
-    if (hasStopped.get()) {
-      return;
-    }
-    if (retryingConnection.compareAndSet(false, true)) {
-      periodicRetryDelay.reset();
-      if (retryAfter != null && periodicRetryDelay instanceof AcceptsDelaySuggestion) {
-        ((AcceptsDelaySuggestion) periodicRetryDelay).suggestDelay(retryAfter);
-      }
-    }
-    if (nextRetryScheduled.compareAndSet(false, true)) {
-      executorService.schedule(
-          this::retryConnection, periodicRetryDelay.getNextDelay().toNanos(), TimeUnit.NANOSECONDS);
-    }
-  }
-
-  private void retryConnection() {
-    nextRetryScheduled.set(false);
-    startConnection();
   }
 
   @Override
@@ -238,17 +215,57 @@ public final class WebSocketRequestService implements RequestService, WebSocket.
   @Override
   public void onClosed() {
     // If this service isn't stopped, we should retry connecting.
-    scheduleConnectionRetry(null);
+    connectionStatus.retryAfter(null);
   }
 
   @Override
   public void onFailure(Throwable t) {
     getCallback().onConnectionFailed(t);
-    scheduleConnectionRetry(null);
+    connectionStatus.retryAfter(null);
   }
 
   @Nonnull
   private Callback getCallback() {
     return Objects.requireNonNull(callback);
+  }
+
+  private class ConnectionStatus {
+    private final PeriodicDelay periodicRetryDelay;
+    private final AtomicBoolean retryingConnection = new AtomicBoolean(false);
+    private final AtomicBoolean nextRetryScheduled = new AtomicBoolean(false);
+
+    ConnectionStatus(PeriodicDelay periodicRetryDelay) {
+      this.periodicRetryDelay = periodicRetryDelay;
+    }
+
+    void success() {
+      retryingConnection.set(false);
+    }
+
+    @SuppressWarnings("FutureReturnValueIgnored")
+    void retryAfter(@Nullable Duration retryAfter) {
+      if (hasStopped.get()) {
+        return;
+      }
+
+      if (retryingConnection.compareAndSet(false, true)) {
+        periodicRetryDelay.reset();
+        if (retryAfter != null && periodicRetryDelay instanceof AcceptsDelaySuggestion) {
+          ((AcceptsDelaySuggestion) periodicRetryDelay).suggestDelay(retryAfter);
+        }
+      }
+
+      if (nextRetryScheduled.compareAndSet(false, true)) {
+        executorService.schedule(
+            this::retryConnection,
+            periodicRetryDelay.getNextDelay().toNanos(),
+            TimeUnit.NANOSECONDS);
+      }
+    }
+
+    private void retryConnection() {
+      nextRetryScheduled.set(false);
+      startConnection();
+    }
   }
 }


LikeTheSalad:
cc @tigrannajaryan

dependencyManagement/build.gradle.kts
```diff
@@ -57,5 +57,6 @@ dependencies {
     api("tools.profiler:async-profiler:4.0")
     api("com.blogspot.mydailyjava:weak-lock-free:0.18")
     api("org.agrona:agrona:1.22.0")
+    api("com.google.protobuf:protobuf-java-util:4.29.1")
```
laurit: could apply the bom instead
```
val protobufVersion = "4.31.1"
api(enforcedPlatform("com.google.protobuf:protobuf-bom:$protobufVersion"))
```

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/websocket/OkHttpWebSocket.java
```diff
@@ -0,0 +1,116 @@
...
+import okhttp3.Response;
+import okhttp3.WebSocketListener;
+import okio.ByteString;
+
+public class OkHttpWebSocket implements WebSocket {
+  private final String url;
+  private final OkHttpClient client;
+  private final AtomicBoolean starting = new AtomicBoolean(false);
+  private final AtomicBoolean closing = new AtomicBoolean(false);
+  private final AtomicBoolean running = new AtomicBoolean(false);
```
laurit: Perhaps it would be better to express the possible lifecycle states as an enum?

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/websocket/OkHttpWebSocket.java
```diff
@@ -0,0 +1,116 @@
...
+    public void onOpen(@Nonnull okhttp3.WebSocket webSocket, @Nonnull Response response) {
+      running.set(true);
+      starting.set(false);
+      delegate.onOpen();
+    }
+
+    @Override
+    public void onClosing(@Nonnull okhttp3.WebSocket webSocket, int code, @Nonnull String reason) {
+      running.set(false);
+      closing.set(true);
```
laurit: other events are passed on to the delegate

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/websocket/OkHttpWebSocket.java
```diff
@@ -0,0 +1,116 @@
...
+  @Override
+  public void close(int code, @Nullable String reason) {
+    if (!running.get()) {
+      return;
+    }
+    if (closing.compareAndSet(false, true)) {
+      if (!getWebSocket().close(code, reason)) {
+        closing.set(false);
+        running.set(false);
+      }
```
laurit: Lets say there is a failure at the same time when `close` is called. Then it is probably possible to reach a situation where the closing flag remains set to `true` after the close completes.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/websocket/WebSocket.java
```diff
@@ -0,0 +1,66 @@
...
+   *     href="https://datatracker.ietf.org/doc/html/rfc6455#section-5.5.1">Section 5.5.1 of RFC
+   *     6455</a>
+   */
+  void close(int code, @Nullable String reason);
+
+  /**
+   * Sends a message via the websocket connection.
+   *
+   * @param request The message payload.
+   * @return {@link Boolean#FALSE} If the message can't get dispatched for any reason, whether the
```
laurit: I'd use `@return false` or `{@code false}`
```suggestion
   * @return {@link Boolean#FALSE} If the message can't be dispatched for any reason, whether the
```

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/websocket/WebSocket.java
```diff
@@ -0,0 +1,66 @@
...
+    /**
+     * Called when receiving a message from the remote peer.
+     *
+     * @param data The payload sent by the remote peer.
+     */
+    void onMessage(byte[] data);
+
+    /**
+     * Called when the connection is closed or cannot be established due to an error. No messages
+     * can get transmitted after this method is called.
```
laurit:
```suggestion
     * can be transmitted after this method is called.
```

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/websocket/WebSocket.java
```diff
@@ -0,0 +1,66 @@
...
+
+  interface Listener {
+
+    /**
+     * Called when the websocket connection is successfully established with the remote peer and may
+     * start sending messages.
+     */
+    void onOpen();
+
+    /** Called when the connection is terminated and no further messages can get transmitted. */
```
laurit:
```suggestion
    /** Called when the connection is terminated and no further messages can be transmitted. */
```

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/websocket/WebSocket.java
```diff
@@ -0,0 +1,66 @@
...
+   *     Boolean#TRUE} if the message can get sent. Returning {@link Boolean#TRUE} doesn't guarantee
+   *     that the message will arrive at the remote peer.
+   */
+  boolean send(byte[] request);
+
+  interface Listener {
+
+    /**
+     * Called when the websocket connection is successfully established with the remote peer and may
+     * start sending messages.
```
laurit: It feels to me that there is a word missing before `may start sending messages.`

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/WebSocketRequestService.java
```diff
@@ -0,0 +1,228 @@
...
+
+  private void handleServerError(ServerErrorResponse errorResponse) {
+    if (serverIsUnavailable(errorResponse)) {
+      Duration retryAfter = null;
+
+      if (errorResponse.retry_info != null) {
+        retryAfter = Duration.ofNanos(errorResponse.retry_info.retry_after_nanoseconds);
+      }
+
+      webSocket.close(1000, null);
```
laurit: will the call to `close` here trigger a `onClosed` event that will also call `scheduleConnectionRetry`?

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/WebSocketRequestService.java
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    doSendRequest();
+  }
+
+  private void doSendRequest() {
+    try {
+      if (!trySendRequest()) {
+        hasPendingRequest.set(true);
+      }
```
laurit: I guess this is to handle when request is sent before websocket is ready. Imo there is race where `trySendRequest` returns `false` but `onOpen` completes before the `hasPendingRequest` flag is set.

dependencyManagement/build.gradle.kts
LikeTheSalad: Thanks, I've added the changes.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/websocket/OkHttpWebSocket.java
LikeTheSalad: Thanks. I added an enum instead. I noticed it's easier to read with the enums.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/websocket/OkHttpWebSocket.java
LikeTheSalad: It wasn't there because I didn't have a use case for this event in the implementation right now. But I just added it as it's part of the RFC anyway.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/websocket/OkHttpWebSocket.java
LikeTheSalad: Thanks. I missed the "IllegalArgumentException" warning from the okhttp method. I've just updated it to avoid this issue.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/websocket/WebSocket.java
LikeTheSalad: I've rephrased it a bit, let me know what you think.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/WebSocketRequestService.java
LikeTheSalad: Yes, in theory. It can also happen in an `onFailure` event. That's why I made sure only one retry task can be scheduled, no matter how many times `scheduleConnectionRetry` is called.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/WebSocketRequestService.java
LikeTheSalad: Thanks, it seems possible indeed. I've just updated it to make sure it won't happen.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/websocket/OkHttpWebSocket.java
```diff
@@ -0,0 +1,114 @@
...
+  public void close(int code, @Nullable String reason) {
+    if (status.compareAndSet(Status.RUNNING, Status.CLOSING)) {
+      try {
+        if (!getWebSocket().close(code, reason)) {
+          status.set(Status.NOT_RUNNING);
+        }
+      } catch (IllegalArgumentException e) {
+        status.set(Status.RUNNING);
+        throw e;
+      }
```
laurit: If you want to keep it then there should be a comment explaining why this is needed. I think that we don't really need to handle this since using an invalid code or reason that is too long is a caller error.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/websocket/OkHttpWebSocket.java
```diff
@@ -0,0 +1,114 @@
...
+
+    @Override
+    public void onFailure(
+        @Nonnull okhttp3.WebSocket webSocket, @Nonnull Throwable t, @Nullable Response response) {
+      status.set(Status.NOT_RUNNING);
+      delegate.onFailure(t);
+    }
+
+    @Override
+    public void onMessage(@Nonnull okhttp3.WebSocket webSocket, @Nonnull ByteString bytes) {
```
laurit, line 104: there is also `fun onMessage(webSocket: WebSocket, text: String)` is it fine to leave that unhandled? I guess it is since it isn't used?

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/WebSocketRequestService.java
```diff
@@ -0,0 +1,243 @@
...
+  /**
+   * Creates an {@link WebSocketRequestService}.
+   *
+   * @param webSocket The WebSocket implementation.
+   * @param periodicRetryDelay The time to wait between retries.
+   */
+  public static WebSocketRequestService create(
+      WebSocket webSocket, PeriodicDelay periodicRetryDelay) {
+    return new WebSocketRequestService(
+        webSocket, periodicRetryDelay, Executors.newSingleThreadScheduledExecutor());
```
laurit: could use daemon thread here like in the http service

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/websocket/OkHttpWebSocket.java
LikeTheSalad: Thanks. I've added a comment.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/websocket/OkHttpWebSocket.java
LikeTheSalad: Binary format should be the only one we need, so yeah I think it's fine to leave the utf-8 message unhandled.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/WebSocketRequestService.java
LikeTheSalad: Thanks, I overlooked it. It's updated now.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/WebSocketRequestService.java
```diff
@@ -0,0 +1,245 @@
...
+
+  @Nonnull
+  private Request getRequest() {
+    return Objects.requireNonNull(requestSupplier).get();
+  }
+
+  @Override
+  public void stop() {
+    if (hasStopped.compareAndSet(false, true)) {
+      doSendRequest();
```
laurit, line 148: I assume that sending the request is here because of https://opentelemetry.io/docs/specs/opamp/#websocket-transport-opamp-client-initiated The `agent_disconnect` flag must be set on the request by the caller. I think this could use a comment explaining why the request is sent before closing.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/WebSocketRequestService.java
```diff
@@ -0,0 +1,245 @@
...
+      webSocket.close(1000, null);
```
laurit: could use a a comment describing what 1000 is. Maybe link to https://datatracker.ietf.org/doc/html/rfc6455#section-7.4.1

> 1000 indicates a normal closure, meaning that the purpose for
      which the connection was established has been fulfilled.

LikeTheSalad: Thanks. I've just updated it.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/WebSocketRequestService.java
LikeTheSalad: Thanks, I've just added a comment to explain it.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/WebSocketRequestService.java
laurit: Since there are only tests with mocks it is hard to try this out. Looking at the code if `onClosed` gets called as a result of calling `close` here then the following could happen 
- `scheduleConnectionRetry` sets `nextRetryScheduled` to `true` and schedules retry
- `retryConnection` runs and clears `nextRetryScheduled` and starts reconnecting
- `onClosed` is called which calls `scheduleConnectionRetry` which ends up calling `WebSocket.open` again
Firstly for this to be possible at all the delay used in first call to `scheduleConnectionRetry` would need to be really small. Secondly it would be fine since apparently calling `WebSocket.open` on an already open doesn't do anything. Just a bit confusing.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/WebSocketRequestService.java
```diff
@@ -0,0 +1,254 @@
...
+  }
+
+  @SuppressWarnings("FutureReturnValueIgnored")
+  private void scheduleConnectionRetry(@Nullable Duration retryAfter) {
+    if (hasStopped.get()) {
+      return;
+    }
+    if (retryingConnection.compareAndSet(false, true)) {
+      periodicRetryDelay.reset();
+      if (retryAfter != null && periodicRetryDelay instanceof AcceptsDelaySuggestion) {
```
laurit: Here we are only ever using the retry suggested from the first retry attempt which is fine since subsequent failed attempts would use `null` for retry anyway. In http service there is the same code but there subsequent retry attempts could suggest the delay which we ignore. Just something I found a bit confusing.

laurit:
Maybe we should also move the retry handling into a nested class like in the http service?
```diff
diff --git a/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/WebSocketRequestService.java b/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/WebSocketRequestService.java
index 3450efa4..3010d00f 100644
--- a/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/WebSocketRequestService.java
+++ b/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/WebSocketRequestService.java
@@ -29,15 +29,14 @@ import opamp.proto.ServerErrorResponseType;
 import opamp.proto.ServerToAgent;
 
 public final class WebSocketRequestService implements RequestService, WebSocket.Listener {
+  private static final PeriodicDelay DEFAULT_DELAY_BETWEEN_RETRIES =
+      PeriodicDelay.ofFixedDuration(Duration.ofSeconds(30));
+
   private final WebSocket webSocket;
-  private final PeriodicDelay periodicRetryDelay;
-  private final AtomicBoolean retryingConnection = new AtomicBoolean(false);
-  private final AtomicBoolean nextRetryScheduled = new AtomicBoolean(false);
   private final AtomicBoolean isRunning = new AtomicBoolean(false);
   private final AtomicBoolean hasStopped = new AtomicBoolean(false);
+  private final ConnectionStatus connectionStatus;
   private final ScheduledExecutorService executorService;
-  public static final PeriodicDelay DEFAULT_DELAY_BETWEEN_RETRIES =
-      PeriodicDelay.ofFixedDuration(Duration.ofSeconds(30));
 
   @GuardedBy("hasPendingRequestLock")
   private boolean hasPendingRequest = false;
@@ -74,8 +73,8 @@ public final class WebSocketRequestService implements RequestService, WebSocket.
       PeriodicDelay periodicRetryDelay,
       ScheduledExecutorService executorService) {
     this.webSocket = webSocket;
-    this.periodicRetryDelay = periodicRetryDelay;
     this.executorService = executorService;
+    this.connectionStatus = new ConnectionStatus(periodicRetryDelay);
   }
 
   @Override
@@ -145,7 +144,7 @@ public final class WebSocketRequestService implements RequestService, WebSocket.
 
   @Override
   public void onOpen() {
-    retryingConnection.set(false);
+    connectionStatus.success();
     getCallback().onConnectionSuccess();
     synchronized (hasPendingRequestLock) {
       if (hasPendingRequest) {
@@ -191,7 +190,7 @@ public final class WebSocketRequestService implements RequestService, WebSocket.
       }
 
       webSocket.close(1000, null);
-      scheduleConnectionRetry(retryAfter);
+      connectionStatus.retryAfter(retryAfter);
     }
   }
 
@@ -199,28 +198,6 @@ public final class WebSocketRequestService implements RequestService, WebSocket.
     return errorResponse.type.equals(ServerErrorResponseType.ServerErrorResponseType_Unavailable);
   }
 
-  @SuppressWarnings("FutureReturnValueIgnored")
-  private void scheduleConnectionRetry(@Nullable Duration retryAfter) {
-    if (hasStopped.get()) {
-      return;
-    }
-    if (retryingConnection.compareAndSet(false, true)) {
-      periodicRetryDelay.reset();
-      if (retryAfter != null && periodicRetryDelay instanceof AcceptsDelaySuggestion) {
-        ((AcceptsDelaySuggestion) periodicRetryDelay).suggestDelay(retryAfter);
-      }
-    }
-    if (nextRetryScheduled.compareAndSet(false, true)) {
-      executorService.schedule(
-          this::retryConnection, periodicRetryDelay.getNextDelay().toNanos(), TimeUnit.NANOSECONDS);
-    }
-  }
-
-  private void retryConnection() {
-    nextRetryScheduled.set(false);
-    startConnection();
-  }
-
   @Override
   public void onClosing() {
     // Noop
@@ -229,17 +206,57 @@ public final class WebSocketRequestService implements RequestService, WebSocket.
   @Override
   public void onClosed() {
     // If this service isn't stopped, we should retry connecting.
-    scheduleConnectionRetry(null);
+    connectionStatus.retryAfter(null);
   }
 
   @Override
   public void onFailure(Throwable t) {
     getCallback().onConnectionFailed(t);
-    scheduleConnectionRetry(null);
+    connectionStatus.retryAfter(null);
   }
 
   @Nonnull
   private Callback getCallback() {
     return Objects.requireNonNull(callback);
   }
+
+  private class ConnectionStatus {
+    private final PeriodicDelay periodicRetryDelay;
+    private final AtomicBoolean retryingConnection = new AtomicBoolean(false);
+    private final AtomicBoolean nextRetryScheduled = new AtomicBoolean(false);
+
+    ConnectionStatus(PeriodicDelay periodicRetryDelay) {
+      this.periodicRetryDelay = periodicRetryDelay;
+    }
+
+    void success() {
+      retryingConnection.set(false);
+    }
+
+    @SuppressWarnings("FutureReturnValueIgnored")
+    void retryAfter(@Nullable Duration retryAfter) {
+      if (hasStopped.get()) {
+        return;
+      }
+
+      if (retryingConnection.compareAndSet(false, true)) {
+        periodicRetryDelay.reset();
+        if (retryAfter != null && periodicRetryDelay instanceof AcceptsDelaySuggestion) {
+          ((AcceptsDelaySuggestion) periodicRetryDelay).suggestDelay(retryAfter);
+        }
+      }
+
+      if (nextRetryScheduled.compareAndSet(false, true)) {
+        executorService.schedule(
+            this::retryConnection,
+            periodicRetryDelay.getNextDelay().toNanos(),
+            TimeUnit.NANOSECONDS);
+      }
+    }
+
+    private void retryConnection() {
+      nextRetryScheduled.set(false);
+      startConnection();
+    }
+  }
 }
```

laurit (approved):

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/WebSocketRequestService.java
LikeTheSalad: I see what you mean. As you mentioned, it should be fine to call `websocket.open` multiple times, although it does seem confusing in the event that the delay time is really short, which I would hope would never be the case. Trying to think of ways to avoid this, probably we could store the "suggested delay" in memory and use it the next time `scheduleConnectionRetry` is called, since that should happen anyway after calling `websocket.close` here (even if the connection is interrupted with an error, because `scheduleConnectionRetry` would be called from `onFailure` in that case). It would bring other things to look out for though, as the suggested delay would become a state for this class, so it might make things confusing in a different way? - If you think it's worth it, I can look for ways to make it bullet-proof, but for now, I left it as is because it seemed unlikely that this scenario could happen in practice.

LikeTheSalad:
> Maybe we should also move the retry handling into a nested class like in the http service?

Sure, I'll add the changes.

LikeTheSalad:
> > Maybe we should also move the retry handling into a nested class like in the http service?
> 
> Sure, I'll add the changes.

Thanks, it's done now.

PR #1986, opened 2025-06-30
Make link check consistent

breedx-splk:
This aligns the way we do link checking in contrib with [how we do it in the instrumenation repo](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/.github/workflows/reusable-markdown-link-check.yml#L18).

One exception is the default `github.token` but I will align that by removing it in instrumentation as well.


breedx-splk:
There's a comment in https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/14163 about waiting for https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/14155#pullrequestreview-2972199235, and I'm cool with that. We can rework this one to align with the new approach in instrumentation.

lychee.toml
```diff
@@ -0,0 +1,19 @@
+timeout = 30
```
laurit: could you copy the changes from https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/14155/files#diff-914f879a5815b9d9802142d70a04f4f511049003fd6e3c30b58b3c3bbe4e74bb

breedx-splk: on it now

PR #1989, opened 2025-07-01
code stable semconv

SylvainJuge:
**Description:**

Removes usage of incubating code semconv attributes that have been made stable in semconv 1.33.0: https://github.com/open-telemetry/semantic-conventions/issues/1377


breedx-splk (approved):
Looks good. Let's get this rebased so the checks will pass and then we're good to merge. Thanks!

PR #1991, opened 2025-07-02
fix tomcat integration test

breedx-splk:
Tomcat integration test breaks when updating to the new instrumentation. This fixes it. Would love to have a look from @SylvainJuge and @robsunday. Thanks!


breedx-splk:
Note: This is PRing into the renovate branch in #1983.

breedx-splk:
Thanks @SylvainJuge for the detailed explanation. I'll roll those test changes back and see how it looks with  `otel.jmx.target.source=legacy`. Appreciate it. 🙏🏻

trask (approved):

PR #1997, opened 2025-07-04
Recreate #1994

breedx-splk:
#1994 seems to have a merge conflict in the changelog that cannot be resolved there...so I'm making a new PR here to supersede that. This also includes the version bump, which is required to do the release build/action.


breedx-splk:
I managed to get 1994 merged. Phew.

PR #1998, opened 2025-07-04
remove unreleased from main

breedx-splk:
Another shot at unblocking the merge conflict in #1994, to unblock the release build.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,5 @@
 # Changelog
 
-## Unreleased
-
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,7 +1,5 @@
 # Changelog
 
-## Unreleased
```
laurit, line 3: As far as I know release scripts use the `## Unreleased` marker. It might be best to add it back before the next release to avoid surprises. cc @trask

PR #2002, opened 2025-07-07
Stop lychee from causing so many PR failures

trask:


.github/workflows/build.yml
```diff
@@ -8,6 +8,9 @@ on:
   pull_request:
   merge_group:
   workflow_dispatch:
+  schedule:
+    # Run daily at 7:30 AM UTC
+    - cron: '30 7 * * *'
```
breedx-splk, line 13: I'm honestly a little confused as to why this PR about lychee is adding a daily scheduled build?

breedx-splk (approved):
Thanks!

.github/workflows/build.yml
trask: oh, ya, it's to make sure we get some notification when an external link disappears (since we'll no longer be checking all external links in PRs anymore and so may not notice)

PR #2004, opened 2025-07-07
Add unreleased section

trask:
The release automation expects this section to be present (even if empty)


trask:
no worries, thanks for running with it last week ❤️

PR #2021, opened 2025-07-11
Opamp client implementation

LikeTheSalad:
This is a continuation of the OpAMP implementation work. The OpampClient implementation is created here, along with its builder. It brings together what has been done in previous PRs.


LikeTheSalad:
cc @tigrannajaryan

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/OpampClient.java
```diff
@@ -62,7 +67,7 @@ interface Callbacks {
      *
      * @param client The relevant {@link OpampClient} instance.
      */
-    void onConnect(OpampClient client);
+    void onConnect(@Nonnull OpampClient client);
```
laurit: I think usually we expect parameters to be non-null https://github.com/open-telemetry/opentelemetry-java-contrib/blob/5d702ed0d3cb9511b488fa2c5fffe29da2a7badf/aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/package-info.java#L7 and use `@Nullable` when they can be null.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/OpampClientBuilder.java
```diff
@@ -0,0 +1,490 @@
...
+import opamp.proto.AgentCapabilities;
+import opamp.proto.AgentDescription;
+import opamp.proto.AnyValue;
+import opamp.proto.ArrayValue;
+import opamp.proto.KeyValue;
+import opamp.proto.RemoteConfigStatus;
+
+/**
+ * This class is internal and is hence not for public use. Its APIs are unstable and can change at
+ * any time.
```
laurit, line 33: is this a typo? I would expect this class to be part of the public api since it is needed to build the client.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/OpampClient.java
```diff
@@ -6,12 +6,17 @@
 package io.opentelemetry.opamp.client.internal;
 
 import io.opentelemetry.opamp.client.internal.response.MessageData;
+import javax.annotation.Nonnull;
 import opamp.proto.AgentDescription;
 import opamp.proto.RemoteConfigStatus;
 import opamp.proto.ServerErrorResponse;
 
 public interface OpampClient {
```
laurit, line 14: should this class be in the internal package?

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/impl/OpampClientImpl.java
```diff
@@ -0,0 +1,268 @@
...
+  private final RequestService requestService;
+  private final AgentToServerAppenders appenders;
+  private final OpampClientState state;
+  private final RecipeManager recipeManager;
+  private final AtomicBoolean isRunning = new AtomicBoolean(false);
+  private final AtomicBoolean hasStopped = new AtomicBoolean(false);
+  @Nullable private Callbacks callbacks;
+
+  /** Fields that must always be sent. */
+  private static final List<Field> CONSTANT_FIELDS;
```
laurit: Maybe it would be better to call these required fields.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/state/InMemoryState.java
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  public void set(@Nonnull T value) {
+    state.set(value);
+  }
+
+  @Nonnull
+  @Override
+  public T get() {
+    return Objects.requireNonNull(state.get());
```
laurit, line 36: from debugging perspective would probably be better to verify that `null` is not passed as initial state or through the setter, though they are already annotated with `@Nonnull`

opamp-client/src/test/java/io/opentelemetry/opamp/client/internal/impl/OpampClientImplTest.java
```diff
@@ -0,0 +1,382 @@
...
+import opamp.proto.ServerToAgent;
+import opamp.proto.ServerToAgentFlags;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+@ExtendWith(MockitoExtension.class)
+class OpampClientImplTest {
```
laurit, line 63: I'm not a fan of testing only with mocks. It would be nice if there was a tests tested agains a server. The server could respond with predefined responses.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/OpampClientBuilder.java
```diff
@@ -0,0 +1,490 @@
...
+ *
+ * <p>Builds an {@link OpampClient} instance.
+ */
+public final class OpampClientBuilder {
+  private final Map<String, AnyValue> identifyingAttributes = new HashMap<>();
+  private final Map<String, AnyValue> nonIdentifyingAttributes = new HashMap<>();
+  private long capabilities = 0;
+  @Nullable private byte[] instanceUid;
+  @Nullable private State.EffectiveConfig effectiveConfigState;
+  @Nullable private RequestService service;
```
laurit: perhaps it would make sense to default to http?

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/OpampClient.java
LikeTheSalad: Got it, thanks. I'll add the package annotation and remove these.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/OpampClientBuilder.java
LikeTheSalad: I couldn't find where we discussed this, though I remember talking about we (the OTel Java community) not being ready to have a standalone opamp client repo and committing to provide support for it right now, so my understanding was that this contrib module would be a sort of incubating place for us to use and refine its API in the meantime without fearing of affecting users with breaking changes, and so that's why everything here is in an internal package right now.

Now, I'm fine with having a non-internal API so long as we can still add breaking changes in the future without issues, which I _think_ it's the case for all the contrib modules in `alpha` status, so if that's the case, I think it's fine to move this (and other types as needed) out of the internal package, because right now, even though I remember that we wanted to use this internally for now, I don't recall having everything in an `internal` package to be a requirement.

I'd like to hear your thoughts on it.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/OpampClient.java
LikeTheSalad: [This reply](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2021#discussion_r2209538191) applies here too.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/OpampClientBuilder.java
laurit: I think it would be better to move the api out of the internal package as then we could get a better feel of what apis we have and what classes they need. We'll have to do this eventually anyway. If needed we can make parts of the api experimental (or should I say more experimental than the rest) and have them in the internal package. In the agent repository we have done this for example in https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/internal/Experimental.java @trask do you have any suggestions

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/OpampClient.java
LikeTheSalad: It's updated now.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/impl/OpampClientImpl.java
LikeTheSalad: Thanks. This is done now.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/state/InMemoryState.java
LikeTheSalad: Thanks. This is done now.

opamp-client/src/test/java/io/opentelemetry/opamp/client/internal/impl/OpampClientImplTest.java
LikeTheSalad: Thanks, it's updated now. This change helped me find an issue with OkHttpSender 👍

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/OpampClientBuilder.java
LikeTheSalad: I've added http as the default. My only concern is that it ties the builder with the okhttpsender impl, though that's not a problem for now.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/OkHttpSender.java
```diff
@@ -51,17 +51,12 @@ public CompletableFuture<Response> send(BodyWriter writer, int contentLength) {
...
-                  future.completeExceptionally(
-                      new HttpErrorException(response.code(), response.message()));
-                }
+              public void onResponse(@Nonnull Call call, @Nonnull okhttp3.Response response) {
+                future.complete(new OkHttpResponse(response));
               }
 
               @Override
-              public void onFailure(@NotNull Call call, @NotNull IOException e) {
+              public void onFailure(@Nonnull Call call, @Nonnull IOException e) {
```
laurit: could add a package-info with `@ParametersAreNonnullByDefault` into this package too

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/OpampClientBuilder.java
laurit: Yes, this will definitely require some more thought. I guess the whole point in having a custom `HttpSender` would be to use something else besides okhttp. From that perspective all of the okhttp specific code should be in a separate module. On the other hand most users are going to be fine with okhttp and perhaps only wish to provide the url, for them it would be nice to have s simpler option than `builder.setRequestService(HttpRequestService.create(OkHttpSender.create("http://localhost:4320/v1/opamp")))`. We could build a spi similar to `HttpSenderProvider` in the sdk that could default to okhttp when it is available and when some other provider is available use that. The tricky part would be when there are multiple providers. Sdk resolves that through configuration options, but that probably is not going to work here. Anyway this is not for the current PR.

laurit (approved):

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/OkHttpSender.java
trask: we typically only annotate @Nullable params, see https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/docs/contributing/style-guideline.md#nullable-annotation-usage

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/OpampClient.java
```diff
@@ -72,7 +77,7 @@ interface Callbacks {
      * @param client The relevant {@link OpampClient} instance.
      * @param throwable The exception.
      */
-    void onConnectFailed(OpampClient client, Throwable throwable);
+    void onConnectFailed(OpampClient client, @Nullable Throwable throwable);
```
trask: why do the callbacks pass in OpampClient? it seems that is available if needed already when registering the callback?

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/OpampClientBuilder.java
trask: It could depend on whether we're planning to publish a general purpose OpAmp client library, or whether this is eventually only going to be an implementation detail of a higher level OpAmp-based remote configuration solution.

I don't mind either approach for now.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/OkHttpSender.java
LikeTheSalad: Thanks. I've just added the changes. I was confused at first because I thought that the package-info file was also applied to the subpackages automatically, though the IDE kept complaining until I added it to this specific one.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/OpampClientBuilder.java
LikeTheSalad: I'm fine with either approach as long as it's ok to add breaking changes without worrying about affecting users, as my understanding is that this module should serve as an incubator to iteratively define the client API that better works for Java. In any case, since this PR is approved, and adding this change might make the PR quite bigger, I've created [this issue](https://github.com/open-telemetry/opentelemetry-java-contrib/issues/2046) to keep track of this discussion.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/OpampClient.java
LikeTheSalad: This is a good point, and tbh I don't have a good answer for it 😅 I just saw that in the opamp-go callback they pass a `Context` as first param, which is a go-specific type, and it just felt like it might be helpful to pass the client impl here instead, as a way to provide some sort of context of where was this callback called from, in case that there might be multiple clients using a single callback? It's a bit of a stretch, and tbh, after thinking about it again now, I couldn't find a use case for it, so I just removed it following the "when in doubt, leave it out" mantra. In any case, since this is an incutating API, we could add it back later if needed.

Issue #2007, opened 2025-07-08
ConsistentRateLimitingSampler can fail if used in combination with legacy samplers


PeterF778:
### Component(s)

samplers

## Description

When ConsistentRateLimitingSampler is using ConsistentParentBasedSampler as the delegate, and when the parent span has no threshold value, the calculated suggested threshold is correct only under assumption that the randomness value is uniformly distributed in the interval [0, 2^56). This assumption may be incorrect.

Please also see this [issue](https://github.com/open-telemetry/opentelemetry-specification/issues/4566).

## Steps to Reproduce

Let's assume 100 traces per second. Now,
1. Use the legacy AlwayOn sampler for the parent span
2. For child spans, use ConsistentRateLimitingSampler using ConsistentParentBasedSampler, and specify max rate as 50 spans/s
3. For grandchild spans, use ConsistentRateLimitingSampler using ConsistentParentBasedSampler, and specify max rate as 25 spans/s

## Expected Result

The grandchild spans are sampled at the rate of about 25/s.

## Actual Result

The grandchild spans are sampled at the rate of about 50/s.

### Component version

1.47.0

PR #2022, opened 2025-07-11
2007_ConsistentRateLimitingSampler

PeterF778:
Fixes https://github.com/open-telemetry/opentelemetry-java-contrib/issues/2007.

**Description:**

Using a new randomness value whenever making sampling decision with unreliable adjusted count.

**Existing Issue(s):**

https://github.com/open-telemetry/opentelemetry-java-contrib/issues/2007

**Testing:**

Added unit test for this scenario.

### DIFF
--- a/consistent-sampling/src/main/java/io/opentelemetry/contrib/sampler/consistent56/ConsistentSampler.java
+++ b/consistent-sampling/src/main/java/io/opentelemetry/contrib/sampler/consistent56/ConsistentSampler.java
@@ -186,10 +186,20 @@ public abstract class ConsistentSampler implements Sampler, Composable {
     boolean isSampled;
     boolean isAdjustedCountCorrect;
     if (isValidThreshold(threshold)) {
-      long randomness = getRandomness(otelTraceState, traceId);
-      isSampled = threshold <= randomness;
       isAdjustedCountCorrect = intent.isAdjustedCountReliable();
-    } else { // DROP
+      // determine the randomness value to use
+      long randomness;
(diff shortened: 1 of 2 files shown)


consistent-sampling/src/main/java/io/opentelemetry/contrib/sampler/consistent56/ConsistentSampler.java
```diff
@@ -186,10 +186,20 @@ public final SamplingResult shouldSample(
     boolean isSampled;
     boolean isAdjustedCountCorrect;
     if (isValidThreshold(threshold)) {
-      long randomness = getRandomness(otelTraceState, traceId);
-      isSampled = threshold <= randomness;
       isAdjustedCountCorrect = intent.isAdjustedCountReliable();
-    } else { // DROP
+      // determine the randomness value to use
+      long randomness;
+      randomness = getRandomness(otelTraceState, traceId);
```
breedx-splk: The value is set in the `if` and the `else`. I think either don't default it here, or remove the `if` block by inverting and keeping the `else`.
 
```suggestion
      long randomness;
```

breedx-splk (approved):
This is getting tricky, but I think I've followed. Thanks!

consistent-sampling/src/main/java/io/opentelemetry/contrib/sampler/consistent56/ConsistentSampler.java
PeterF778: Good catch, thanks!

PR #2027, opened 2025-07-14
use mise for easier local link checking

zeitlinger:


.github/workflows/reusable-link-check.yml
```diff
@@ -13,12 +13,12 @@ jobs:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           fetch-depth: 0 # needed for merge-base below
-
+      - uses: jdx/mise-action@5cb1df66ed5e1fb3c670ea0b62fd17a76979826a # v2.3.1
```
breedx-splk: At first glance, this project seems mature...but it does increase our overall attack surface doesn't it?

breedx-splk: Maybe it's no change, since this also drops the lychee docker dep. 🤷🏻

zeitlinger: Well, this pins the version of the mise action, but takes the latest mise.

We could also pin the mise version (and add a sha256) - but we'd have to come up with a script to update that, because renovate can't do that yet.

See https://github.com/jdx/mise-action

zeitlinger: I found a way now to pin mise

zeitlinger: done

.github/workflows/reusable-link-check.yml
```diff
@@ -13,12 +13,15 @@ jobs:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
         with:
           fetch-depth: 0 # needed for merge-base below
-
+      - uses: jdx/mise-action@bfb9fa0b029db830a8c570757cee683df207a6c5 # v2.4.0
+        with:
+          version: v2025.7.11
+          sha256: fe2d4c5c681c942b2f52bf0c71d04429ba4a5e090973514bde466a411190cd00
```
trask: is this needed? or is it good enough without? and by removing this would it allow us to remove the custom renovate manager?

```suggestion
```

zeitlinger: it failed for me a couple of times, so I'd suggest to keep it

trask: looks like it was an old version of the action, I've pushed an update, I like to try to limit the renovate custom managers when possible b/c they're pretty confusing

trask (approved):
I like it, thanks!

PR #2029, opened 2025-07-14
declarative config: Inferred spans

zeitlinger:
Based on https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2020

Fixes #2011

Issue #2011, opened 2025-07-09
Declarative config: support inferred spans


zeitlinger:
port InferredSpansAutoConfig to declarative config

PR #2030, opened 2025-07-15
declarative config: support Inferred spans

zeitlinger:
The declarative config bridge has been copied from the agent - eventually, we can use only one, but that's a later step.

Fixes [#2029 ](https://github.com/open-telemetry/opentelemetry-java-contrib/issues/2011)

### DIFF
--- a/inferred-spans/src/test/java/io/opentelemetry/contrib/inferredspans/internal/SamplingProfilerTest.java
+++ b/inferred-spans/src/test/java/io/opentelemetry/contrib/inferredspans/internal/SamplingProfilerTest.java
@@ -57,10 +57,12 @@ class SamplingProfilerTest {
   }
 
   @Test
-  void shouldLazilyCreateTempFilesAndCleanThem() throws Exception {
-
-    List<Path> tempFiles = getProfilerTempFiles();
-    assertThat(tempFiles).isEmpty();
+  void shouldLazilyCreateTempFilesAndCleanThem() {
+    for (Path file : getProfilerTempFiles()) {
+      if (!file.toFile().delete()) {
+        throw new IllegalStateException("Could not delete temp file: " + file);
+      }
+    }
 
     // temporary files should be created on-demand, and properly deleted afterwards
     setupProfiler(false);


zeitlinger:
@JonasKunz please have a look

declarative-config-bridge/build.gradle.kts
```diff
@@ -0,0 +1,18 @@
+plugins {
+  id("otel.java-conventions")
+  id("otel.publish-conventions")
+}
+
+description = "OpenTelemetry extension that provides a bridge for declarative configuration."
+otelJava.moduleName.set("io.opentelemetry.contrib.sdk.declarative.config.bridge")
```
jaydeluca: the package used in the code of this module is `io.opentelemetry.contrib.sdk.autoconfigure`, should they match this moduleName here?

declarative-config-bridge/src/test/resources/config.yaml
```diff
@@ -0,0 +1,27 @@
+file_format: 0.4
```
jaydeluca: is there a reason we couldn't use the latest `1.0-rc.1` here?

inferred-spans/src/test/java/io/opentelemetry/contrib/inferredspans/InferredSpansCustomizerProviderTest.java
```diff
@@ -0,0 +1,58 @@
...
+import io.opentelemetry.sdk.extension.incubator.fileconfig.DeclarativeConfiguration;
+import java.io.ByteArrayInputStream;
+import java.nio.charset.StandardCharsets;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.DisabledOnOs;
+import org.junit.jupiter.api.condition.OS;
+
+@DisabledOnOs(OS.WINDOWS)
```
jaydeluca: why do we disable this for windows?

inferred-spans/src/test/java/io/opentelemetry/contrib/inferredspans/InferredSpansCustomizerProviderTest.java
```diff
@@ -0,0 +1,58 @@
...
+  @AfterEach
+  public void resetGlobalOtel() {
+    ProfilingActivationListener.ensureInitialized();
+    OtelReflectionUtils.shutdownAndResetGlobalOtel();
+  }
+
+  @Test
+  void declarativeConfig() {
+    String yaml =
+        "file_format: 0.4\n"
```
jaydeluca: Same as other comment, is there a reason we can't use a later version? is it because the latest is an RC?
```suggestion
        "file_format: 1.0-rc.1\n"
```

inferred-spans/src/test/java/io/opentelemetry/contrib/inferredspans/InferredSpansCustomizerProviderTest.java
jaydeluca: actually, just noticed on the other test there is a comment explaining it, so maybe we could add that same comment to this test as well for clarity?

```suggestion
// async-profiler doesn't work on Windows
@DisabledOnOs(OS.WINDOWS)
```

declarative-config-bridge/build.gradle.kts
zeitlinger: yes, it should!

declarative-config-bridge/src/test/resources/config.yaml
zeitlinger: consistency in the repo - but I'll update all

declarative-config-bridge/build.gradle.kts
```diff
@@ -0,0 +1,18 @@
+plugins {
+  id("otel.java-conventions")
+  id("otel.publish-conventions")
+}
+
+description = "OpenTelemetry extension that provides a bridge for declarative configuration."
+otelJava.moduleName.set("io.opentelemetry.contrib.sdk.config.bridge")
```
trask: can you add a README for this module, and describe why it's needed? thanks

zeitlinger: done - as agreed, I've also added @jaydeluca as co-component owner

trask: thanks, makes sense now

would you mind doing this in two steps, first adding declarative config support without creating a reusable bridge component, and then we can circle back to the reusable bridge concept afterwards? if we are going to create a reusable bridge, it might make sense for it to be in core, but that will require more consideration and discussions

zeitlinger: > without creating a reusable bridge component

- it's actually copied from the agent (and we can re-use it in the agent later) (@laurit reviewed it there)
- not sure what you mean by "without" - do you mean to copy-paste the entire config related logic? or simply move the classes to the inferred spans project?

> If we are going to create a reusable bridge, it might make sense for it to be in core, but that will require more consideration and discussions

- core only implements things that are specified - but I'm open to doing that too
- I would suggest doing the move as a follow-up task

jackshirazi:
per the updated schedule from the SIG, will this be closed for a later new PR to be opened next year?

zeitlinger:
@jackshirazi @JonasKunz please have a look

inferred-spans/src/test/java/io/opentelemetry/contrib/inferredspans/InferredSpansCustomizerProviderTest.java
```diff
@@ -0,0 +1,58 @@
...
+  void declarativeConfig() {
+    String yaml =
+        "file_format: 1.0-rc.1\n"
+            + "tracer_provider:\n"
+            + "  processors:\n"
+            + "    - experimental_inferred_spans:\n"
+            + "        enabled: false\n"
+            + "        backup:\n"
+            + "          diagnostic:\n"
+            + "            files: true\n";
```
SylvainJuge: @jackshirazi I think we should keep `backup_diagnostric_files` as a single attribute, WDYT ?

zeitlinger: please review all properties https://github.com/open-telemetry/opentelemetry-java-contrib/blob/168161c08558ed243b3440f63bc880e0a0f83345/inferred-spans/README.md#autoconfiguration - where `.` should become `_` instead of a new nesting level

inferred-spans/README.md
```diff
@@ -43,6 +43,31 @@ So if you are using an autoconfigured OpenTelemetry SDK, you'll only need to add
...
+### Usage with declarative configuration
+
+You can configure the inferred spans processor using declarative YAML configuration with the
+OpenTelemetry SDK. For example:
+
+```yaml
+file_format: 1.0-rc.1
+tracer_provider:
+  processors:
+    - experimental_inferred_spans:
```
jackshirazi: does this mean if we become stable the name would change and it would break all existing config? Or to avoid breaking we'd need to support `experimental_inferred_spans` as well as the stable name?

inferred-spans/src/main/java/io/opentelemetry/contrib/inferredspans/InferredSpansComponentProvider.java
```diff
@@ -0,0 +1,37 @@
...
+
+import com.google.auto.service.AutoService;
+import io.opentelemetry.api.incubator.config.DeclarativeConfigProperties;
+import io.opentelemetry.instrumentation.config.bridge.DeclarativeConfigPropertiesBridgeBuilder;
+import io.opentelemetry.sdk.autoconfigure.spi.internal.ComponentProvider;
+import io.opentelemetry.sdk.trace.SpanProcessor;
+
+@SuppressWarnings("rawtypes")
+@AutoService(ComponentProvider.class)
+public class InferredSpansComponentProvider implements ComponentProvider<SpanProcessor> {
```
jackshirazi: This is an InferredSpansSpanProcessorProvider since the component is defined

inferred-spans/src/main/java/io/opentelemetry/contrib/inferredspans/InferredSpansConfig.java
```diff
@@ -0,0 +1,120 @@
...
+  static final String MIN_DURATION_OPTION = "otel.inferred.spans.min.duration";
+  static final String INCLUDED_CLASSES_OPTION = "otel.inferred.spans.included.classes";
+  static final String EXCLUDED_CLASSES_OPTION = "otel.inferred.spans.excluded.classes";
+  static final String INTERVAL_OPTION = "otel.inferred.spans.interval";
+  static final String DURATION_OPTION = "otel.inferred.spans.duration";
+  static final String LIB_DIRECTORY_OPTION = "otel.inferred.spans.lib.directory";
+  static final String PARENT_OVERRIDE_HANDLER_OPTION =
+      "otel.inferred.spans.parent.override.handler";
+
+  static SpanProcessor create(ConfigProperties properties) {
```
jackshirazi: this is createSpanProcessor(), since we're not creating an InferredSpansConfig

inferred-spans/src/test/java/io/opentelemetry/contrib/inferredspans/InferredSpansCustomizerProviderTest.java
jackshirazi: I think there should be no nesting for any of the options, all of these should be unnested attributes
logging.enabled
backup.diagnostic.files
safe.mode
post.processing.enabled
sampling.interval
min.duration
included.classes
excluded.classes
interval
duration
lib.directory
parent.override.handler

inferred-spans/README.md
zeitlinger: yes, it would be a breaking change

This is the policy to signal unstable parts in an otherwise stable section

inferred-spans/src/test/java/io/opentelemetry/contrib/inferredspans/InferredSpansCustomizerProviderTest.java
zeitlinger: done

inferred-spans/src/main/java/io/opentelemetry/contrib/inferredspans/InferredSpansSpanProcessorProvider.java
```diff
@@ -0,0 +1,43 @@
...
+
+@SuppressWarnings("rawtypes")
+@AutoService(ComponentProvider.class)
+public class InferredSpansSpanProcessorProvider implements ComponentProvider<SpanProcessor> {
+
+  private static final String PREFIX = "otel.inferred.spans.";
+
+  @Override
+  public String getName() {
+    return "experimental_inferred_spans";
```
jackshirazi: so clearly from the discussion in the README, we'll want to support more than one name. How will that be achieved (specifically, we'll want both `experimental_inferred_spans` and `inferred_spans` to be supported at some point in the future)

inferred-spans/README.md
jackshirazi: moved discussion to the getName() method where this is defined

inferred-spans/src/test/java/io/opentelemetry/contrib/inferredspans/InferredSpansSpanProcessorProviderTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+  @BeforeEach
+  @AfterEach
+  public void resetGlobalOtel() {
+    ProfilingActivationListener.ensureInitialized();
+    OtelReflectionUtils.shutdownAndResetGlobalOtel();
+  }
+
+  @Test
+  void declarativeConfig() {
```
jackshirazi, line 39: at least one equivalent negative test, eg `enabled: 5`

inferred-spans/src/main/java/io/opentelemetry/contrib/inferredspans/InferredSpansSpanProcessorProvider.java
trask: @zeitlinger for naming, what about `inferred_spans/development` to match the `*/development` convention for declarative configuration property names? would be great to get some common guidance from configuration SIG

@jackshirazi maybe a `@Deprecated` version of this class with a different `getName()`?

jackshirazi: So a duplicated class (or would a subclass work?), with either one or the other installed depending on the declaration. I'm good with that, will that work @zeitlinger

zeitlinger: > @zeitlinger for naming, what about `inferred_spans/development` to match the `*/development` convention for declarative configuration property names?

@jack-berg pointed out that we should use the `experimental_*` convention

trask: @zeitlinger thanks, can you try to send a PR to configuration repo to document this? I think that's the only way we're all going to follow the same convention

zeitlinger: @jack-berg please help me out here

inferred-spans/src/test/java/io/opentelemetry/contrib/inferredspans/InferredSpansSpanProcessorProviderTest.java
zeitlinger: I added a test for disabled.
Turned out that enabled wasn't supported at all - so I added it with a default value of `true` - since the user has already added the span processor explicitly.

Using `enabled: 5` results in using the default value and a warning:

"WARNING: Ignoring value for key [enabled] because it is Integer instead of Boolean: 5"

trask:
(testing out the [copilot-instructions.md](https://github.com/open-telemetry/opentelemetry-java-contrib/blob/main/.github/copilot-instructions.md))

inferred-spans/src/main/java/io/opentelemetry/contrib/inferredspans/InferredSpansSpanProcessorProvider.java
zeitlinger: I found the reference: https://github.com/open-telemetry/opentelemetry-configuration/blob/main/README.md#experimental-features - so it should be `inferred_spans/development` as @trask said

zeitlinger: > so clearly from the discussion in the README, we'll want to support more than one name.

let's make this a separate discussion - I've added it to the Java SIG agenda

jackshirazi: The name isn't my focus here, it's being able to support both names indefinitely when this becomes stable

zeitlinger: > The name isn't my focus here, it's being able to support both names indefinitely when this becomes stable

this is why I added it to the meeting agenda

zeitlinger:
@jackshirazi please check again after the Java SIG meeting

Issue #1956, opened 2025-06-11
[opamp] Add retry periodic delay implementation


LikeTheSalad:
This is to create a periodic delay implementation that enables exponential backoff logic, to be used as the default retry periodic delay for opamp requests.

Issue #1954, opened 2025-06-11
[opamp] Add animalsniffer


LikeTheSalad:
This is to ensure the same compatibility guarantees with Android apps as the one from [OTel Java](https://github.com/open-telemetry/opentelemetry-java) core.o

jackshirazi:
@LikeTheSalad I think this is done?

PR #2043, opened 2025-07-22
Update change log for upcoming release

trask:


CHANGELOG.md
```diff
@@ -2,6 +2,18 @@
 
 ## Unreleased
 
+### Consistent sampling
```
laurit, line 25: @SylvainJuge I believe you wised to call out some breaking changes. If so please provide the language to be included in the change log.

CHANGELOG.md
```diff
@@ -2,6 +2,18 @@
...
+
+- Refactor ConsistentFixedThresholdSampler to prepare for dynamic threshold support
+  ([#2018](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2018))
+- ConsistentRateLimitingSampler can fail if used in combination with legacy samplers
+  ([#2022](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2022))
+
+### OpAMP client
+
+- Add OpAMP WebSocket service implementation
+  ([#1969](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1969))
```
laurit: Since currently merged opamp client is not functional imo there isn't much point in adding change log about it. After merging the last open PR it should be functional so we could start adding change log for it in the next release.

laurit: If we are going to merge #2021 in this release we could either add a single entry that opamp client implementation was added, or just add entries for both PRs.

LikeTheSalad: I just wanted to mention that, as @laurit said, in its current state it's not functional, and even though #2021 makes it almost ready to go, I was planning to finish this first iteration with a last PR after it, where we configure the maven publishing in this module, as well as addressing a couple of small issues such as these ones: https://github.com/open-telemetry/opentelemetry-java-contrib/issues/1902, https://github.com/open-telemetry/opentelemetry-java-contrib/issues/1956, and https://github.com/open-telemetry/opentelemetry-java-contrib/issues/1954. Which should fit in a single PR to polish it all before its first release.

Btw, I'm participating in a company event this week so the time I can dedicate to this work is limited. I was planning to address the issues mentioned above next week.

CHANGELOG.md
SylvainJuge: do you mean about the consistent sampling ? (if yes, then maybe asking @jackshirazi might be relevant), or for marking the jmx gatherer as deprecated with https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2034 ? The latter is only a documentation change for now, but if you plan to include it you can add the following:

------
### JMX Gatherer
- JMX Gatherer is now deprecated in favor of JMX Scraper
  ([#2034](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2034))

jackshirazi: I believe https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2022 is a breaking change as in the edge case it covers, sampling will cascade correctly. But it's actually a bugfix. Nothing else that I know about is a notifiable change

CHANGELOG.md
trask: good point, I forgot the module isn't published to maven central yet, so will remove this section entirely

CHANGELOG.md
laurit: I meant https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2033

> However, this should be included in the release notes as users of jmx-scraper with the following configuration will see changes in captured metrics:
otel.jmx.target.source = auto (default)
otel.jmx.target.system with jetty

trask: thanks, I've added #2033

laurit (approved):

PR #2014, opened 2025-07-09
Declarative config: Cloud component providers

zeitlinger:
Part of https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/14081

PR #2047, opened 2025-07-22
add end to end test for declarative config for aws

zeitlinger:
See https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2014#discussion_r2221287484

I'll add the other cloud providers, too, if this is a good pattern.


aws-resources/build.gradle.kts
```diff
@@ -21,11 +21,25 @@ dependencies {
...
+  testImplementation("io.opentelemetry:opentelemetry-sdk-extension-incubator")
   testImplementation("io.opentelemetry:opentelemetry-api-incubator")
   testImplementation("io.opentelemetry:opentelemetry-sdk-testing")
+  testImplementation("io.opentelemetry:opentelemetry-exporter-logging")
 
   testImplementation("com.linecorp.armeria:armeria-junit5")
   testRuntimeOnly("org.bouncycastle:bcpkix-jdk15on")
   testImplementation("com.google.guava:guava")
   testImplementation("org.skyscreamer:jsonassert")
+  testImplementation("org.junit-pioneer:junit-pioneer")
```
jaydeluca: what is this added dependency used for?

aws-resources/src/test/java/io/opentelemetry/contrib/aws/resource/ResourceComponentProviderTest.java
```diff
@@ -19,6 +24,22 @@ void providerIsLoaded() {
     Iterable<ComponentProvider> providers =
         ComponentLoader.forClassLoader(ResourceComponentProviderTest.class.getClassLoader())
             .load(ComponentProvider.class);
-    assertThat(providers).extracting(ComponentProvider::getName).containsExactly("aws");
+    assertThat(providers).extracting(ComponentProvider::getName).contains("aws");
```
jaydeluca: I was curious about the need for changing `containsExactly` to `contains` so i tried to put this back to `containsExactly` to see what would happen, and noticed that in the test failure it lists `console` 3 times, plus an additional `service`:

```
Expecting actual:
  ["aws", "console", "console", "console", "service"]
to contain exactly (and in same order):
  ["aws"]
but some elements were not expected:
  ["console", "console", "console", "service"]
```

I changed the test code to get some insights into what was actually loaded to see. Looks like this is because declarative configuration is automatically adding additional console exporters for metrics and logs plus the service resource detector. 

```
Expecting actual:
  [io.opentelemetry.contrib.aws.resource.internal.AwsResourceDetector,
    io.opentelemetry.exporter.logging.internal.ConsoleMetricExporterComponentProvider,
    io.opentelemetry.exporter.logging.internal.ConsoleSpanExporterComponentProvider,
    io.opentelemetry.exporter.logging.internal.ConsoleLogRecordExporterComponentProvider,
    io.opentelemetry.sdk.extension.incubator.fileconfig.ServiceResourceDetector]
```

This surprised me a bit because I thought if the metric and log exporters weren't explicitly defined, they wouldn't be configured. Is my understanding off here?

zeitlinger: ComponentProvider is only about what providers are available to be loaded if they appear in the yaml

This test seems redundant when the e2e test is there - I'll delete it

aws-resources/build.gradle.kts
zeitlinger: removed

zeitlinger:
> 👍 since these endToEnd tests are all very similar we could probably create a general AbstractResourceProviderTest or utility class and just pass the attributes

there is no "test-common" project in contrib - and I'm not sure this use case justifies it

PR #2051, opened 2025-07-23
Fix flaky aws remote sampler test

laurit:
Sampler update time is set up in a background thread. By the time the removed assertion is run it is usually not set, but sometimes is.


trask:
Closing and re-opening to retrigger component owner reviews since I've just fixed that (hopefully)

PR #2067, opened 2025-07-30
Opamp first iteration completion

LikeTheSalad:
Finishes the first iteration of the `opamp-client` module by addressing these issues: https://github.com/open-telemetry/opentelemetry-java-contrib/issues/1902, https://github.com/open-telemetry/opentelemetry-java-contrib/issues/1956, and https://github.com/open-telemetry/opentelemetry-java-contrib/issues/1954.

> The issue https://github.com/open-telemetry/opentelemetry-java-contrib/issues/1902 is partially done, as it was about updating both the `OpampClient` and `RequestService` implementations to extend Closeable. However, this was only feasible for `OpampClient`, as doing so for `RequestService` required keeping its abstraction within `OpampClientImpl`, leading to having to use factories and extra steps to keep it generic. In the end, it was a lot of code changes that didn't seem feasible for this PR.

The following changes have been added as well:

* Updating the README file to include a usage example.
* Applying the maven publishing conventions.
* Suppressing auto instrumentation of the opamp-related HTTP requests.


buildSrc/src/main/kotlin/otel.animalsniffer-conventions.gradle.kts
```diff
@@ -0,0 +1,24 @@
...
+
+// Always having declared output makes this task properly participate in tasks up-to-date checks
+tasks.withType<AnimalSniffer> {
+  reports.text.required.set(true)
+}
+
+// Attaching animalsniffer check to the compilation process.
+tasks.named("classes").configure {
+  finalizedBy("animalsnifferMain")
+}
```
laurit: opentelemetry-java and opentelemetry-java-instrumentation don't have this

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/OkHttpSender.java
```diff
@@ -45,8 +47,14 @@ public CompletableFuture<Response> send(BodyWriter writer, int contentLength) {
     RequestBody body = new RawRequestBody(writer, contentLength, MEDIA_TYPE);
     builder.post(body);
 
+    InstrumentationUtil.suppressInstrumentation(() -> doSendRequest(builder.build(), future));
```
laurit, line 52: Perhaps there should be a comment explaining that this prevents automatic instrumentation from tracing the underlying okhttp request?

buildSrc/src/main/kotlin/otel.animalsniffer-conventions.gradle.kts
LikeTheSalad: Got it. It's just a way of getting notified of an api compatibility issue during development without having to run all the build checks, though it's not strictly necessary, so I've just removed it.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/OkHttpSender.java
LikeTheSalad: Sure. I've just added a comment.

buildSrc/src/main/kotlin/otel.animalsniffer-conventions.gradle.kts
laurit: I don't have anything particular against it just that it would be preferable to have this aligned with our other repositories.

laurit (approved):

Issue #2077, opened 2025-08-05
Move Disk Buffering into Beta


bidetofevil:
The [disk buffering package](https://github.com/open-telemetry/opentelemetry-java-contrib/blob/main/disk-buffering/README.md) is an important part of any usage of the OTel on mobile, and it's been used in `opentelemetry-android` for a while now. I think it's worth moving it from alpha to beta to encourage usage.

LikeTheSalad:
Thanks for opening this issue, @bidetofevil 

I agree it's mostly ready to go, but there are a couple of API changes that I've been planning to make based on the feedback I've collected so far, which I think should be made before moving it into a more stable status, as those changes will most likely break the current way of using the API. I've been recently busy with the OpAMP development, which is why I haven't been able to address those yet, but given that the last [OpAMP PR](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2067) is already created and ready for review, I was planning to move on to the disk buffering changes next. Still, I think we can move it to beta this month, though (depending on how fast the reviews are).

PR #2084, opened 2025-08-08
Disk buffering api changes

LikeTheSalad:
These changes are meant to address the feedback received since the creation of this module, as well as prepare it to get promoted from alpha status to beta, as mentioned [here](https://github.com/open-telemetry/opentelemetry-java-contrib/issues/2077).

The key takeaways for this new API are the following:

- Abstracts the storage mechanism to allow for custom implementations to store signals using a different approach (unlike the current API that forces storing them into protobuf files as the only possible approach).
- Provide more control over the writing process from disk buffering exporter instances (unlike the current API, which makes opinionated decisions around what to do when a write operation fails and what happens when an exporter's shutdown method is called).
- Provide a cleaner reader API. (The current implementation consists of passing around export functions, with multiple overlapping enum states and making opinionated decisions for non-happy paths).
- Provide a way to clear stored data on-demand. (The current API doesn't have a way to clear data other than reading and exporting it).

The plan is to create a default implementation for the new API that covers the existing API's behavior. Since migrating the existing behavior to the new API will require changing many files, I've decided to leave it for a follow-up PR, so that we can focus on the new API design on this one.


fractalwrench (approved):
LGTM

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/exporters/ExporterCallback.java
```diff
@@ -0,0 +1,38 @@
...
+   * @param error Optional - provides more information of why the operation failed.
+   */
+  void onExportError(SignalType type, @Nullable Throwable error);
+
+  /**
+   * Called when the exporter is closed.
+   *
+   * @param type The type of signal associated to the exporter.
+   */
+  void onShutdown(SignalType type);
```
breedx-splk, line 33: Not a huge deal, but I often find it convenient to provide default empty methods for implementations that may not necessarily care about all of the interface methods. That way, implementers can choose to implement the ones that they are interested in, and it doesn't require many implementations to repeat empty bodies.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/exporters/SignalStorageExporter.java
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  public CompletableResultCode exportToStorage(Collection<T> items) {
+    CompletableFuture<WriteResult> future = storage.write(items);
+    try {
+      WriteResult operation = future.get(writeTimeout.toMillis(), TimeUnit.MILLISECONDS);
+      if (operation.isSuccessful()) {
+        callback.onExportSuccess(type);
+        return CompletableResultCode.ofSuccess();
+      } else {
```
breedx-splk: prefer removing the redundant `else` and unindenting the part below it.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/exporters/SignalStorageExporter.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.concurrent.TimeoutException;
+
+/** Internal utility for common export to disk operations across all exporters. */
+final class SignalStorageExporter<T> {
+  private final SignalStorage<T> storage;
+  private final ExporterCallback callback;
+  private final Duration writeTimeout;
+  private final SignalType type;
+
+  public SignalStorageExporter(
```
breedx-splk, line 26: Would be good to have unit test coverage for this new class.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/FileSpanStorage.java
```diff
@@ -0,0 +1,39 @@
...
+import io.opentelemetry.contrib.disk.buffering.storage.SignalStorage;
+import io.opentelemetry.contrib.disk.buffering.storage.result.WriteResult;
+import io.opentelemetry.sdk.trace.data.SpanData;
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.concurrent.CompletableFuture;
+import javax.annotation.Nonnull;
+
+/** Default storage implementation where items are stored in multiple protobuf files. */
+public final class FileSpanStorage implements SignalStorage.Span {
```
breedx-splk, line 17: Presumably there will be analogs for the other signal types? Why not include those in this PR as well?

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/exporters/ExporterCallback.java
LikeTheSalad: Got it. I think I see what you mean, though considering that this interface is only for callback purposes, do you think that also applies here? And if so, how can we know what callbacks to set a default body to and which ones to enforce? Given that none of them are required to make the exporter work? - Usually, my concern with default impls, is that people might miss available options as the compiler won't complain if they're not implemented, unless users read the docs and take the time to scan through the lib's source code. It's not a big deal if we can provide defaults that work for everyone in case an implementation is used as part of the business logic, but for merely informative ones, such as this one, I'm not sure it applies.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/exporters/SignalStorageExporter.java
LikeTheSalad: Good point, I've added the changes.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/exporters/SignalStorageExporter.java
LikeTheSalad: I've just added unit tests for it.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/FileSpanStorage.java
LikeTheSalad: > Presumably there will be analogs for the other signal types?

Yes, that's the idea.

> Why not include those in this PR as well?

I just wanted to put the new API design to test with this PR, not to fully implement it yet; this is to avoid wasting time in case some design changes were requested during the review. If there are no issues with the design itself, I'm planning to create a follow-up PR with all the implementations and refactorings needed to use the new approach.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/exporters/ExporterCallback.java
zeitlinger: - we already use the pattern: https://github.com/open-telemetry/opentelemetry-java/blob/a0d13cfa80ce40455acbc01306279400e394ea39/sdk-extensions/incubator/src/main/java/io/opentelemetry/sdk/extension/incubator/trace/OnEndSpanProcessor.java#L36
- it uses a second method to indicate which method is actually implemented, such as 
- it doesn't use default methods, but adds an implementation as a convenience

fractalwrench: IMO default method bodies should be avoided in interfaces that form public APIs. It makes discoverability harder, doesn't cleanly separate the interface from the implementation, and the behavior can become part of the public API's contract which can constrain changes in future.

Having said that, an empty body is less of a big deal to me than a body containing actual business logic would be

breedx-splk: I think @fractalwrench makes some good points here, but like I said, I have found them convenient. Just know that interfaces with multiple `onXXX` type methods (like this one) especially when on public APIs will have many implementers with no-op implementations. Check out `SpanProcessor` implementations, for example.

An alternative is to separate the methods into separate interfaces, but that's fairly intrusive to this existing design. It's not that important.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/SignalType.java
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.contrib.disk.buffering;
+
+public enum SignalType {
```
tylerbenson, line 8: @LikeTheSalad Any reason you added this instead of using `io.opentelemetry.contrib.disk.buffering.internal.utils.SignalTypes`?

PR #2090, opened 2025-08-12
Remove unnecessary gradle option

trask:


laurit (approved):
instrumentation repo has the same option

PR #2097, opened 2025-08-12
Auto-update Java version used in tests

trask:
Assuming this works, we'll still need to remember to always keep behind LTS versions.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -54,7 +54,7 @@ jobs:
           - 11
           - 17
           - 21
-          - 23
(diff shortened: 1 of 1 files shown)


.github/renovate.json5
```diff
@@ -170,11 +171,22 @@
       customType: 'regex',
       datasourceTemplate: 'npm',
       managerFilePatterns: [
-        '/^.github/workflows//',
+        '.github/workflows/**'
```
trask, line 174: just a simplification to match what I used below

Issue #1964, opened 2025-06-17
[jmx-scraper] Package jmx scraper as a deb/rpm/tar.gz package


atoulme:
### Component(s)

jmx-scraper

### Is your feature request related to a problem? Please describe.

There is no installation package for the JMX scraper at this time.

### Describe the solution you'd like

Package the JMX scraper tooling as a deb/rpm/tar.gz to allow installation in unix systems.

It should come with the following features:
* Systemd unit file
* Default config file

Optionally, the package should have dependencies such as jre spelled out for installation as part of this installation method.

mcmho:
Hey @atoulme!   I'm trying to get more involve in the community.   I wonder if this issue is still needed to be fixed?

atoulme:
@SylvainJuge anything towards the end goal described in this issue helps.

mcmho:
@SylvainJuge   Here is a proposed change for this issue - https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2099/files   Generally systemd file is pretty universal across all flavors of modern linux distributions.  

Please decide and let me know whether we want to proceed with this commit or not.

trask:
> While I definitely see some benefits to provide a one-liner command to install jmx-scraper, doing so also brings some extra complexity and maintenance challenges. In particular, there are potentially lots of architectures/distributions/variants to test and support, as we can see with the [collector install instructions](https://opentelemetry.io/docs/collector/installation/).

I have the same sentiment.

@atoulme how important is this? are we expecting most users to use it via the collector integration?

breedx-splk:
> > While I definitely see some benefits to provide a one-liner command to install jmx-scraper, doing so also brings some extra complexity and maintenance challenges. In particular, there are potentially lots of architectures/distributions/variants to test and support, as we can see with the [collector install instructions](https://opentelemetry.io/docs/collector/installation/).
> 
> I have the same sentiment.

Adding my +1 here too. I don't think we should do this. I think that providing a systemd file and instructions on how to use it is probably fine, but I don't think we should provide additional packaging or deploy artifacts.

atoulme:
The collector is in go and is statically compiled for each OS and architecture, this is a Java application that is going to work with Java 11+, iirc. I don't think the comparison holds.

> @atoulme how important is this? are we expecting most users to use it via the collector integration?

[...]

PR #2099, opened 2025-08-13
[jmx-scraper] Package jmx scraper as a deb/rpm/tar.gz package #1964

mcmho:
**Description:**
 - added copyJarForPackage for gradle to copy jar for packaging
 - added createTarBall to create a packaged tar.gz
 - added systemd service file

**Existing Issue(s):**
 https://github.com/open-telemetry/opentelemetry-java-contrib/issues/1964
  
**Testing:**
 - 'gradle assemble' produces the opentelemetry-jmx-scraper-1.49.0-alpha.tar.gz
 - tested locally with a docker image with systemd support

**Documentation:**
updated README.MD

**Outstanding items:**
If we are going ahead with this, then let me add the .tar.gz to our Release page as well.

### DIFF
--- a/maven-extension/README.md
+++ b/maven-extension/README.md
@@ -15,13 +15,13 @@ In the code snippets below, replace `OPENTELEMETRY_MAVEN_VERSION` with the [late
 release](https://search.maven.org/search?q=g:io.opentelemetry.contrib%20AND%20a:opentelemetry-maven-extension).
 
 ## Installation
-jmx-scraper is typically run as a standalone JAR file. 
-o use it, you generally need to download the latest JAR from the OpenTelemetry Java Contrib releases page 
-or retrieve it as a dependency from a Maven repository. 
+jmx-scraper is typically run as a standalone JAR file.
+o use it, you generally need to download the latest JAR from the OpenTelemetry Java Contrib releases page
+or retrieve it as a dependency from a Maven repository.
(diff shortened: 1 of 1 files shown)


maven-extension/README.md
```diff
@@ -14,6 +14,17 @@ The Maven OpenTelemetry Extension is configured using environment variables or J
...
+jmx-scraper is typically run as a standalone JAR file.
+o use it, you generally need to download the latest JAR from the OpenTelemetry Java Contrib releases page
+or retrieve it as a dependency from a Maven repository.
+Once obtained, you can run it using the Java command line tool.
+Please refer to [Usage](#usage).
+
+For systemd service, the opentelemetry-jmx-scraper-x.xx.x.tar.gz tar.gz file is available
+with the JAR and a ready to install systemd service file.
+
+
```
breedx-splk, line 27: Whoops, I think this was added to the wrong README.md. Shouldn't be in the maven extension dir I think.

breedx-splk (changes requested):
So I don't think we should bundle/package a separate .tar or .tgz file -- in the java world the `.jar` should be enough, and we started publishing that with the last release.

I think that adding the `systemd` service file is fine, if it comes with docs on how to use it.

breedx-splk:
@mcmho how do you feel about closing this one?

Issue #2102, opened 2025-08-14
Make GOOGLE_CLOUD_PROJECT property not mandatory


gustavovnicius:
### Component(s)

gcp-auth-extension

### Is your feature request related to a problem? Please describe.

Currently the `gcp-auth-extension` has the `GOOGLE_CLOUD_PROJECT` as a required property (either system properties or environment variable).

That, on some setups, is an inconvenience and one needs a few workarounds to supply that property. Which, in a first glance, is also redundant, as the extension itself triggers a metadata request which could be used to infer the project id.

Having the option to provide the project as a configuration option is still useful though, for those who have setups where the reporting and quota are on different projects.

### Describe the solution you'd like

The extension should infer the project from the metadata request and not have it as a required configuration property.

In case it can neither infer the project from the metadata, and a project id isn't provided, then it should fail (same behavior today when the option isn't provided).

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

laurit:
cc @psx95

psx95:
Hi @gustavovnicius, thanks for opening this issue and the related PR.

There was some discussion when deciding this behavior, but basically the reason why this property was made a mandatory field was because of 2 primary reasons: 

1. We use [google-auth-library-java](https://github.com/googleapis/google-auth-library-java/tree/main) to authenticate using ADC, but last I checked there was no way to extract GCP Project ID from GoogleCredentials. I had [opened an issue](https://github.com/googleapis/google-auth-library-java/issues/1614) with the library authors requesting a feature to add a method to retrieve Project IDs from OAuth2 credentials (*Unfortunately, there hasn't been much activity on it, but I will follow through with this issue again).*
[...]

PR #2109, opened 2025-08-15
[gcp-auth-extension]: Try resolving GCP_PROJECT with ServiceOptions if not provided

gustavovnicius:
**Description:**

Enhancement - Try resolving GOOGLE_CLOUD_PROJECT value using google-cloud-sdk.

Having a mandatory `GOOCLE_CLOUD_PROJECT` configuration is unnecessary, given the extension already depends on ADC context existing. It can try to infer the project from it.
Still mandatory to have a project id, but this gives more flexibility in the usage of the extension.

**Existing Issue(s):**
https://github.com/open-telemetry/opentelemetry-java-contrib/issues/2102

**Testing:**

Refactored the project id resolver logic into a separate method, including the more comprehensive `ServiceOptions` resolver. Tested the method for both the current strategy and falling back to `ServiceOptions`, also added a missing test for not being able to find the option and throwing an exception.

**Documentation:**

`GOOGLE_CLOUD_PROJECT` configuration was moved into the optional config section, with a note that even though it's optional, the extension needs it either provided or being able to infer it. 

**Outstanding items:**

N/A


gustavovnicius:
Any outstanding blocker about this one?

trask:
cc @psx95 @jsuereth

gcp-auth-extension/src/test/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProviderTest.java
```diff
@@ -538,6 +541,49 @@ void testTargetSignalsBehavior(TargetSignalBehavior testCase) {
...
+        GcpAuthAutoConfigurationCustomizerProvider.getGoogleProjectId(configProperties));
+  }
+
+  @Test
+  void testResolveGoogleProjectIdFromServiceOptions() {
+    // Clear the system property to ensure it does not interfere with the test
+    System.clearProperty(ConfigurableOption.GOOGLE_CLOUD_PROJECT.getSystemProperty());
+    // Property that the extension expects is google.cloud.project, ServiceOptions works with
+    // GOOGLE_CLOUD_PROJECT, so this will highlight the fallback to ServiceOptions
+    System.setProperty("GOOGLE_CLOUD_PROJECT", DUMMY_GCP_RESOURCE_PROJECT_ID);
```
laurit: consider using junit `SetSystemProperty` and `ClearSystemProperty` annotations. Then you won't need to worry about properties set in one test messing with the next test.

gustavovnicius: Fixed

gustavovnicius:
New version has been adopted and project id is attempted to be resolved based on credentials. @psx95

gcp-auth-extension/README.md
```diff
@@ -34,14 +34,13 @@ The extension can be configured either by environment variables or system proper
...
-#### Required Config
+#### Optional Config
 
 - `GOOGLE_CLOUD_PROJECT`: Environment variable that represents the Google Cloud Project ID to which the telemetry needs to be exported.
 
   - Can also be configured using `google.cloud.project` system property.
-  - This is a required option, the agent configuration will fail if this option is not set.
-
-#### Optional Config
+  - If neither of these options are set, the extension will attempt to infer the project id from the credentials used, however notice that not all credentials implementations will be able to provide a project id, so the inference is only a best-effort attempt.
```
psx95:
```suggestion
  - If neither of these options are set, the extension will attempt to infer the project id from the current credentials as a fallback, however notice that not all credentials implementations will be able to provide a project id, so the inference is only a best-effort attempt.
```

gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
```diff
@@ -228,10 +234,36 @@ private static Map<String, String> getRequiredHeaderMap(
...
+
+  /**
+   * Retrieves the Google Cloud Project ID from the configuration properties, falling back to
+   * google-cloud-core's ServiceOptions project ID resolution if not explicitly set.
+   *
+   * @param configProperties The configuration properties containing the GCP project ID.
+   * @return The Google Cloud Project ID.
+   */
+  @Nonnull
+  static String getGoogleProjectId(
```
psx95: We could do away with this functions since much of this functionality is already available in `ConfigurableOption` class.

See suggestion for `customizeResource` method.

gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
```diff
@@ -228,10 +234,36 @@ private static Map<String, String> getRequiredHeaderMap(
...
-  private static Resource customizeResource(Resource resource, ConfigProperties configProperties) {
-    String gcpProjectId =
-        ConfigurableOption.GOOGLE_CLOUD_PROJECT.getConfiguredValue(configProperties);
-    Resource res = Resource.create(Attributes.of(stringKey(GCP_USER_PROJECT_ID_KEY), gcpProjectId));
+  private static Resource customizeResource(Resource resource, String gcpProjectId) {
+    Resource res =
+        Resource.create(
+            Attributes.of(AttributeKey.stringKey(GCP_USER_PROJECT_ID_KEY), gcpProjectId));
     return resource.merge(res);
   }
```
psx95:
```suggestion
  // Note that credentials can be passed from `customize` function directly
  private static Resource customizeResource(Resource resource, ConfigProperties configProperties, GoogleCredentials credentials) {
    String gcpProjectId;
    try {
      gcpProjectId = ConfigurableOption.GOOGLE_CLOUD_PROJECT.getConfiguredValue(configProperties);
    } catch (GoogleAuthException e) {
      gcpProjectId = credentials.getProjectId();
      if (gcpProjectId == null) {
        // this exception will still contain the accurate message.
        throw e;  
      }      
    }
    Resource res = Resource.create(Attributes.of(stringKey(GCP_USER_PROJECT_ID_KEY), gcpProjectId));
    return resource.merge(res);
  }
```

gcp-auth-extension/src/test/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProviderTest.java
```diff
@@ -538,6 +543,51 @@ void testTargetSignalsBehavior(TargetSignalBehavior testCase) {
...
+    Mockito.verifyNoInteractions(mockedGoogleCredentials);
+  }
+
+  @Test
+  @ClearSystemProperty(key = "google.cloud.project")
+  @ClearSystemProperty(key = "GOOGLE_CLOUD_PROJECT")
+  void testResolveGoogleProjectIdFromCredentials() {
+    DefaultConfigProperties configProperties =
+        DefaultConfigProperties.create(Collections.emptyMap(), Mockito.mock(ComponentLoader.class));
+
```
psx95: nit: extra line

psx95 (commented):
Thanks for the PR! 
Left a couple of suggestions that might be helpful.

gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
gustavovnicius: @psx95, this method was added mostly so we could isolate the project id resolution in the tests.
Any suggestions for that?

psx95: Hi @gustavovnicius, this method is an implementation detail of the class and should have been private. For testing, I think you should test the expected behavior instead of testing the method in isolation. I think `testQuotaProjectBehavior` does something similar. 

 So an idea here would be to mock the `GoogleCredentials` (particularly the `getProjectId` method) and in one of the cases, you return a valid Project ID and see if the extension loads or throws an expected `GoogleAuthException`

Let me know if this makes sense.

gustavovnicius: Hey @psx95 , suggestion applied. Tried to replicate the same test structure as `testQuotaProjectBehavior`

gustavovnicius:
@psx95 ?

gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
```diff
@@ -228,9 +230,19 @@ private static Map<String, String> getRequiredHeaderMap(
...
 
   // Updates the current resource with the attributes required for ingesting OTLP data on GCP.
-  private static Resource customizeResource(Resource resource, ConfigProperties configProperties) {
-    String gcpProjectId =
-        ConfigurableOption.GOOGLE_CLOUD_PROJECT.getConfiguredValue(configProperties);
+  // Note that credentials can be passed from `customize` function directly
```
psx95: Nit: Can remove this comment. 

I added this in the suggestion only to clarify my reasoning.

gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
```diff
@@ -228,9 +230,19 @@ private static Map<String, String> getRequiredHeaderMap(
...
+  private static Resource customizeResource(
+      Resource resource, GoogleCredentials credentials, ConfigProperties configProperties) {
+    String gcpProjectId;
+    try {
+      gcpProjectId = ConfigurableOption.GOOGLE_CLOUD_PROJECT.getConfiguredValue(configProperties);
+    } catch (io.opentelemetry.sdk.autoconfigure.spi.ConfigurationException e) {
```
psx95: Nit: Can replace it with an import statement.

gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
```diff
@@ -228,9 +230,19 @@ private static Map<String, String> getRequiredHeaderMap(
...
+      gcpProjectId = credentials.getProjectId();
+      if (gcpProjectId == null) {
+        // this exception will still contain the accurate message.
```
psx95: Nit: Can remove this comment. 

I added this in the suggestion only to clarify my reasoning.

gcp-auth-extension/src/test/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProviderTest.java
```diff
@@ -680,6 +767,34 @@ private static Stream<Arguments> provideTargetSignalBehaviorTestCases() {
    * indicates that the mocked credentials are configured to provide DUMMY_GCP_QUOTA_PROJECT_ID as
    * the quota project ID.
    */
+  private static Stream<Arguments> provideProjectIdBehaviorTestCases() {
```
psx95, line 762: The javadoc string from line 755 to 769 applies only to `provideQuotaBehaviorTestCases`, please move it so that it sits right above that function. 

`provideProjectIdBehaviorTestCases` would require a *similar*, but *different* doc string.

gcp-auth-extension/src/test/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProviderTest.java
```diff
@@ -680,6 +767,34 @@ private static Stream<Arguments> provideTargetSignalBehaviorTestCases() {
...
+        // If user doesn't specify project ID, fallback to credentials.getProjectId()
+        Arguments.of(
+            ProjectIdTestBehavior.builder()
+                .setUserSpecifiedProjectId(null)
+                .setCredentialsProjectId("credentials-project-id")
+                .setExpectedProjectIdInResource("credentials-project-id")
+                .setExpectedToThrow(false)
+                .build()),
+        // If user doesn't specify and credentials.getProjectId() returns null, throw exception
+        Arguments.of(
```
psx95, line 789: Some edge cases should also be tested here - particularly around the extension's behavior on empty string values (`""`) for `setUserSpecifiedProjectId` and `setCredentialsProjectId`.

gcp-auth-extension/src/test/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProviderTest.java
```diff
@@ -456,6 +457,92 @@ void testQuotaProjectBehavior(QuotaProjectIdTestBehavior testCase) throws IOExce
...
+  @MethodSource("provideProjectIdBehaviorTestCases")
+  @ClearSystemProperty(key = "google.cloud.project")
+  @ClearSystemProperty(key = "google.otel.auth.target.signals")
+  @SuppressWarnings("CannotMockMethod")
+  void testProjectIdBehavior(ProjectIdTestBehavior testCase) throws IOException {
+
+    // configure environment according to test case
+    String userSpecifiedProjectId = testCase.getUserSpecifiedProjectId();
+    if (userSpecifiedProjectId != null && !userSpecifiedProjectId.isEmpty()) {
+      System.setProperty(
```
psx95, line 470: The user can pass empty strings as the projectID, as such the test should account for it.

gcp-auth-extension/src/test/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProviderTest.java
```diff
@@ -456,6 +457,92 @@ void testQuotaProjectBehavior(QuotaProjectIdTestBehavior testCase) throws IOExce
...
+        Mockito.mockStatic(GoogleCredentials.class)) {
+      googleCredentialsMockedStatic
+          .when(GoogleCredentials::getApplicationDefault)
+          .thenReturn(mockedGoogleCredentials);
+
+      if (testCase.getExpectedToThrow()) {
+        // expect exception to be thrown when project ID is not available
+        assertThatThrownBy(
+                () -> {
+                  OpenTelemetrySdk sdk =
```
psx95: If it is expected to throw an exception, on account of improper configuration (User project ID cannot be determined), then it should be thrown only from `buildOpenTelemetrySdkWithExporter`, right ? 

Then there is no need to invoke `generateTestSpan` or `shutdown` on the SDK - we should remove these, because this test will pass even if the exception is coming from `generateTestSpan` or `shutdown`.

psx95 (changes requested):
Hi @gustavovnicius, 
Apologies for the delay on this, and thank you for the changes, they look good to me now, just suggested a few changes in the test, mostly about adding a few edge cases and updating the documentation there.

We should be good to merge after this 👍🏻 \
Thanks again for your contribution and patience!

PR #2121, opened 2025-08-15
Use the generated code to drive the definition of the metric

atoulme:
A no-op change to rely on the generated code to take the definition of the gauge instead of duplicating this code.


ibm-mq-metrics/src/main/java/io/opentelemetry/ibm/mq/WmqMonitor.java
```diff
@@ -66,7 +67,7 @@ public WmqMonitor(ConfigWrapper config, ExecutorService threadPool, Meter meter)
 
     this.metricsConfig = new MetricsConfig(config);
 
-    this.heartbeatGauge = meter.gaugeBuilder("ibm.mq.heartbeat").setUnit("1").ofLongs().build();
+    this.heartbeatGauge = Metrics.createIbmMqHeartbeat(meter);
```
atoulme, line 70: The code of that function:
```
    return meter
        .gaugeBuilder("ibm.mq.heartbeat")
        .ofLongs()
        .setUnit("1")
        .setDescription("Queue manager heartbeat")
        .build();
```

PR #2122, opened 2025-08-15
[ibm-mq-metrics] Connection errors counter

atoulme:
**Description:**
Add a counter that records the number of times the connection was not… established successfully, with an error.code dimension recording the code of the error reported

**Testing:**

Tests the result of a bad connection in an isolated unit test.


ibm-mq-metrics/model/metrics.yaml
```diff
@@ -609,3 +609,15 @@ groups:
     attributes:
       - ref: ibm.mq.queue.manager
         requirement_level: required
+  - id: mq.connection.errors
+    type: metric
+    metric_name: mq.connection.errors
```
breedx-splk:
```suggestion
  - id: ibm.mq.connection.errors
    type: metric
    metric_name: ibm.mq.connection.errors
```

ibm-mq-metrics/src/integrationTest/java/io/opentelemetry/ibm/mq/integration/tests/WMQMonitorIntegrationTest.java
```diff
@@ -281,4 +284,34 @@ void test_otlphttp() throws Exception {
...
+        connectionErrors = (SumData<LongPointData>) metricData.getData();
+      }
+    }
+
+    assertThat(connectionErrors).isNotNull();
+
+    LongPointData metricPoint = connectionErrors.getPoints().iterator().next();
+    String value = metricPoint.getAttributes().get(AttributeKey.stringKey("error.code"));
+
+    assertThat(value).isEqualTo("2538");
```
breedx-splk: Would it be possible to verify the _value_ as well?

breedx-splk:
Oh, test failures and rebase tho. :)

ibm-mq-metrics/src/integrationTest/java/io/opentelemetry/ibm/mq/integration/tests/WMQMonitorIntegrationTest.java
atoulme: the value is 1, we increment a counter, is that _valuable_ ?

atoulme: alright added

PR #2123, opened 2025-08-15
Add back yaml templates used to generate the default config an example

atoulme:
**Description:**

Add back the yaml template used to generate a default config file.


jaydeluca (approved):
having a section in this component's readme that explains the different pieces of code that are maintained via weaver and templates, and how to go about regenerating them if needed etc would be helpful IMO. 

Should [this](https://github.com/open-telemetry/opentelemetry-java-contrib/blob/main/ibm-mq-metrics/README.md?plain=1#L149-L150) be updated to reference the location of the example config file added in this PR? i dont see anything in that `src/main/resources/config.yml` location currently

PR #2135, opened 2025-08-19
Restructure workflows

trask:
This accomplishes 3 things:
- There will no longer be a "skipped" PR check and so the whole section should collapse PRs when green
- Introduces a daily build (with --no-build-cache), which will help catch weird gradle caching issues like we saw recently
- Aligns more with instrumentation repo

Potentially in future could call the common workflow from release workflow too (similar to instrumentation repo).

GitHub diff isn't great, so I split it into 2 commits, the first commit is just a file rename, so you can review the second commit which then is better (after the rename).


.github/workflows/build-common.yml
```diff
@@ -121,16 +118,6 @@ jobs:
           name: integration-test-results
           path: jmx-metrics/build/reports/tests/integrationTest
 
-  link-check:
```
trask: move into the callers so they can choose not to block merging / snapshots if it fails

.github/workflows/build-common.yml
```diff
@@ -139,97 +126,3 @@ jobs:
 
   shell-script-check:
     uses: ./.github/workflows/reusable-shell-script-check.yml
-
-  publish-snapshots:
```
trask: moved to build-daily which is only place it's used

.github/workflows/build-common.yml
```diff
@@ -139,97 +126,3 @@ jobs:
...
-      - name: Build and publish snapshots
-        if: ${{ github.ref_name == 'main' && github.repository == 'open-telemetry/opentelemetry-java-contrib' }}
-        run: ./gradlew assemble publishToSonatype
-        env:
-          SONATYPE_USER: ${{ secrets.SONATYPE_USER }}
-          SONATYPE_KEY: ${{ secrets.SONATYPE_KEY }}
-          GPG_PRIVATE_KEY: ${{ secrets.GPG_PRIVATE_KEY }}
-          GPG_PASSWORD: ${{ secrets.GPG_PASSWORD }}
-
-  required-status-check:
```
trask: moved to build-pull-request which is only place it's used

.github/workflows/build-common.yml
```diff
@@ -139,97 +126,3 @@ jobs:
...
-            !startsWith(github.base_ref, 'release/') &&
-            (
-              needs.markdown-lint-check.result != 'success' ||
-              needs.misspell-check.result != 'success' ||
-              needs.shell-script-check.result != 'success'
-            )
-          )
-        run: exit 1 # fail
-
-  workflow-notification:
```
trask: moved to build-daily which is only place it's used

laurit (approved):

Issue #2016, opened 2025-07-09
Add dynamic updateable consistent threshold sampler


jackshirazi:
### Component(s)

consistent-sampling

### Is your feature request related to a problem? Please describe.

static ratio sampling is too restrictive, I need dynamic updateable ratio

### Describe the solution you'd like

I already have the solution working in https://github.com/jackshirazi/elastic-otel-java/tree/dynamic-sampling , it's basically an additional sampler. But it's cleaner to refactor the existing samplers to make it a subclass 

### Additional context

I'll refactor and add the class in multiple steps since that makes the PRs easier to review unless one big PR is really preferred

PR #2137, opened 2025-08-19
add updateable threshold sampler

jackshirazi:
**Description:**

Refactored the ConsistentFixedThresholdSampler class into a ConsistentThresholdSampler abstract class and ConsistentFixedThresholdSampler subclass, then added ConsistentVariableThresholdSampler class which is mutable

**Existing Issue(s):**

closes https://github.com/open-telemetry/opentelemetry-java-contrib/issues/2016

**Testing:**

Test class added for new updateable sampler.

**Documentation:**

ConsistentSampler additional public method has added documentation

**Outstanding items:**

Nothing obvious

### DIFF
--- a/consistent-sampling/src/main/java/io/opentelemetry/contrib/sampler/consistent56/ConsistentThresholdSampler.java
+++ b/consistent-sampling/src/main/java/io/opentelemetry/contrib/sampler/consistent56/ConsistentThresholdSampler.java
@@ -1,0 +1,61 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.contrib.sampler.consistent56;
+
+import static io.opentelemetry.contrib.sampler.consistent56.ConsistentSamplingUtil.calculateSamplingProbability;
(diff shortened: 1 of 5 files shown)


consistent-sampling/src/main/java/io/opentelemetry/contrib/sampler/consistent56/ConsistentThresholdSampler.java
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  protected static String getThresholdDescription(long threshold) {
+    String thresholdString;
+    if (threshold == getMaxThreshold()) {
+      thresholdString = "max";
+    } else {
+      thresholdString =
+          ConsistentSamplingUtil.appendLast56BitHexEncodedWithoutTrailingZeros(
+                  new StringBuilder(), threshold)
```
PeterF778: Other static methods of ConsistentSamplingUtil are in the "import static", why this is different?

jackshirazi: good point, thanks! Updated

trask (approved):

PR #2138, opened 2025-08-19
Use a more efficient serializer and write bytes directly to disk

tylerbenson:
Instead of an intermediary byte array.

The OTel Java SDK has a hand-coded and optimized serializer implementation that is well tested and used extensively. This PR switches to using that serializer.

It is also more efficient to serialize directly to the file output stream instead of an intermediary byte array, so I used the same marshaling strategy used by the SDK and passed the marshaler down to the WritableFile.

See relevant discussion here:
https://github.com/open-telemetry/opentelemetry-java-contrib/discussions/1911

### DIFF
--- a/disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/exporter/ToDiskExporter.java
+++ b/disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/exporter/ToDiskExporter.java
@@ -38,10 +38,11 @@ public class ToDiskExporter<EXPORT_DATA> {
     return new ToDiskExporterBuilder<>(storage);
   }
 
-  public CompletableResultCode export(Collection<EXPORT_DATA> data) {
+  public synchronized CompletableResultCode export(Collection<EXPORT_DATA> data) {
     logger.log("Intercepting exporter batch.", Level.FINER);
     try {
-      if (storage.write(serializer.serialize(data))) {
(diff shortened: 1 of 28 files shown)


disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/exporter/ToDiskExporter.java
```diff
@@ -38,10 +38,11 @@ public static <T> ToDiskExporterBuilder<T> builder(Storage storage) {
     return new ToDiskExporterBuilder<>(storage);
   }
 
-  public CompletableResultCode export(Collection<EXPORT_DATA> data) {
+  public synchronized CompletableResultCode export(Collection<EXPORT_DATA> data) {
```
zeitlinger, line 41: why sychronized?

zeitlinger (commented):
I would have expected more work - but the integration test works 😄

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/exporter/ToDiskExporter.java
tylerbenson: The serializer is stateful now, plus it seems like a bad idea to allow multiple exports to be writing to disk concurrently anyway.

trask (approved):

PR #2147, opened 2025-08-20
Add AWS X-Ray Adaptive Sampling Support

majanjua-amzn:
**Description:**
Adding support for X-Ray's new Adaptive Sampling feature. The feature will allow users to detect anomalies in their application across distributed services and boost the sampling rate of their root service based on their X-Ray sampling rule configuration and optionally a user-provided SDK level configuration. It will also allow users to optionally provide an error capture configuration, where - if configured - the sampler will send unsampled anomaly spans to be exported directly:
- Configure a max boost rate and boost cooldown in their X-Ray sampling rules that defines how high the sampling rate can go for a given rule based on anomalies detected in the instrumentation and for how long
- Configure anomaly conditions based on error code (RegEx), latency, and span operation - these will be used to check if a given span is an anomaly. By default, spans with a `statusCode > 499` (i.e. 5XX) will be considered anomalies
- Configure a anomaly/error capture rate that allows spans to be sent directly to the configured span exporter.

The changes:
- Provide APIs called `setSpanExporter` and `setAdaptiveSamplingConfig` to set up the feature - if these are not provided any attempt to use the `adaptSampling` API will throw an `IllegalStateException`
- Provide an API called `adaptSampling` that accepts a span and its associated spanData:
  - This API performs the necessary logic to determine whether the span is an anomaly based on the user-provided conditions (or default 5XX) and makes the decision whether to count it towards the boost-related statistics and/or whether to send it directly for export based on the error capture rate
- Update the calls to GetSamplingRules and GetSamplingTargets according to the new API and the collected anomaly statistics
- Propagate sampling information between instrumented services - specifically, the sampling rule in the root service is passed to all downstream services via the trace state, such that statistics can be recorded meaningfully for boost to be triggered for the root service in a distributed system

**Testing:**
Re-testing needs to be done based on latest changes in aws-otel-java-instrumentation once ready, hence the need for the draft revision of the PR
- Manual testing: Application launched on fresh EC2 instance with 3 services
- Performance testing: Revealed no difference between 

**Documentation:**
TBD - hence draft PR

**Outstanding items:**
These changes are the first phase and iteration of AWS X-Ray's adaptive sampling feature. As we get feedback, more changes may be introduced to improve or streamline the experience.

### DIFF
--- a/aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/XrayRulesSampler.java
+++ b/aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/XrayRulesSampler.java
@@ -5,42 +5,79 @@
 
 package io.opentelemetry.contrib.awsxray;
 
+import static io.opentelemetry.semconv.HttpAttributes.HTTP_RESPONSE_STATUS_CODE;
+import static io.opentelemetry.semconv.ServiceAttributes.SERVICE_NAME;
+
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
+import io.opentelemetry.api.common.AttributeKey;
 import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.api.trace.Span;
 import io.opentelemetry.api.trace.SpanKind;
+import io.opentelemetry.api.trace.StatusCode;
 import io.opentelemetry.context.Context;
 import io.opentelemetry.contrib.awsxray.GetSamplingTargetsResponse.SamplingTargetDocument;
 import io.opentelemetry.sdk.common.Clock;
 import io.opentelemetry.sdk.resources.Resource;
+import io.opentelemetry.sdk.trace.ReadableSpan;
 import io.opentelemetry.sdk.trace.data.LinkData;
+import io.opentelemetry.sdk.trace.data.SpanData;
 import io.opentelemetry.sdk.trace.samplers.Sampler;
 import io.opentelemetry.sdk.trace.samplers.SamplingResult;
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.time.Duration;
 import java.util.Arrays;
 import java.util.Comparator;
 import java.util.Date;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
+import java.util.function.Consumer;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
+import javax.annotation.Nullable;
 
 final class XrayRulesSampler implements Sampler {
 
   private static final Logger logger = Logger.getLogger(XrayRulesSampler.class.getName());
+
+  public static final AttributeKey<String> AWS_XRAY_SAMPLING_RULE =
+      AttributeKey.stringKey("aws.xray.sampling_rule");
+
+  // Used for generating operation
+  private static final String UNKNOWN_OPERATION = "UnknownOperation";
+  private static final AttributeKey<String> URL_PATH = AttributeKey.stringKey("url.path");
+  private static final AttributeKey<String> HTTP_TARGET = AttributeKey.stringKey("http.target");
+  private static final AttributeKey<String> HTTP_REQUEST_METHOD =
+      AttributeKey.stringKey("http.request.method");
+  private static final AttributeKey<String> HTTP_METHOD = AttributeKey.stringKey("http.method");
 
   private final String clientId;
   private final Resource resource;
   private final Clock clock;
   private final Sampler fallbackSampler;
   private final SamplingRuleApplier[] ruleAppliers;
+  private final Map<String, String> ruleToHashMap;
+  private final Map<String, String> hashToRuleMap;
+
+  private final boolean adaptiveSamplingRuleExists;
+  private final Cache<String, AwsXrayAdaptiveSamplingConfig.UsageType> traceUsageCache;
+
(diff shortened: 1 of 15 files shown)


aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/XrayRulesSampler.java
```diff
@@ -74,10 +124,30 @@ public SamplingResult shouldSample(
...
+            .getTraceState()
+            .get(AwsSamplingResult.AWS_XRAY_SAMPLING_RULE_TRACE_STATE_KEY);
     for (SamplingRuleApplier applier : ruleAppliers) {
       if (applier.matches(attributes, resource)) {
-        return applier.shouldSample(
-            parentContext, traceId, name, spanKind, attributes, parentLinks);
+        SamplingResult result =
+            applier.shouldSample(parentContext, traceId, name, spanKind, attributes, parentLinks);
+
+        String ruleToPropagate =
```
wangzlei: nit: It is readable to update line 134-150:
1) if there is rule from upstream, return upstream's rule;
2) if adaptiveSamplingConfig, return current rule;

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/XrayRulesSampler.java
```diff
@@ -96,7 +166,144 @@ public String getDescription() {
...
+    // Empty list -> no conditions will apply and we will not do anything
+    if (anomalyConditions != null && !anomalyConditions.isEmpty()) {
+      String operation = spanData.getAttributes().get(AwsAttributeKeys.AWS_LOCAL_OPERATION);
+      if (operation == null) {
+        operation = generateIngressOperation(spanData);
+      }
+      for (AwsXrayAdaptiveSamplingConfig.AnomalyConditions condition : anomalyConditions) {
+        // Check if the operation matches any in the list or if operations list is null (match all)
+        List<String> operations = condition.getOperations();
+        if (!(operations == null || operations.isEmpty() || operations.contains(operation))) {
```
wangzlei: Is operation(s) in condition required?

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/XrayRulesSampler.java
```diff
@@ -96,7 +166,144 @@ public String getDescription() {
...
+              case SAMPLING_BOOST:
+                shouldBoostSampling = true;
+                break;
+              case ERROR_SPAN_CAPTURE:
+                shouldCaptureAnomalySpan = true;
+                break;
+            }
+          }
+        }
+        if (shouldBoostSampling && shouldCaptureAnomalySpan) {
```
wangzlei: For performance, maybe if shouldBoostSampling or shouldCaptureAnomalySpan is true we can skip the remaining BoostSampling or CaptureAnomaly check.

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/XrayRulesSampler.java
```diff
@@ -96,7 +166,144 @@ public String getDescription() {
...
+      shouldBoostSampling = true;
+      shouldCaptureAnomalySpan = true;
+    }
+
+    String traceId = spanData.getTraceId();
+    SpanContext parentContext = spanData.getParentSpanContext();
+    boolean isLocalRootSpan =
+        parentContext == null || !parentContext.isValid() || parentContext.isRemote();
+
+    if (shouldBoostSampling || shouldCaptureAnomalySpan || isLocalRootSpan) {
```
wangzlei: Readability, not sure whether the "if" here is useful since you have 
// Anomaly Capture, Sampling Boost below

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/XrayRulesSampler.java
majanjua-amzn: No

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/XrayRulesSampler.java
majanjua-amzn: Split `// Anomaly Capture` from `// Sampling Boost` section. Still need `shouldBoostSampling || isLocalRootSpan` together since they both trigger the same loop logic that finds the `ruleToReportTo`

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/XrayRulesSampler.java
majanjua-amzn: Done

aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/XrayRulesSampler.java
majanjua-amzn: Added a comment. Splitting the two cases would require some code duplication or turning the logic into a function which doesn't really make sense to me here.

docs/apidiffs/current_vs_latest/opentelemetry-aws-xray.txt
trask, line 1: @majanjua-amzn @wangzlei please review new public API carefully since this module is already marked stable (if you are unsure you can always add it initially under `.internal.` package as an experimental feature), thanks

majanjua-amzn: We've tested and reviewed it extensively more than once as part of a contribution to https://github.com/aws-observability/aws-otel-java-instrumentation, and have also verified there is no impact on existing behaviour for those not using the new APIs.

As such, we're okay with releasing these APIs directly. Thanks for the callout!

Issue #2150, opened 2025-08-21
Sporadic failing test in baggage-processor


laurit:
See https://scans.gradle.com/s/jjq3wmv4qs6fg https://scans.gradle.com/s/gqw4jupazwusy
Failures produced while attempting to merge https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2031 so could be caused by that PR.
```
java.lang.AssertionError: 	
Expecting all elements of:	
  [ExtendedSdkLogRecordData{resource=Resource{schemaUrl=null, attributes={service.instance.id="a6a744b8-2dcb-4a31-a46f-31fc9fd4b6cf", service.name="unknown_service:java", telemetry.sdk.language="java", telemetry.sdk.name="opentelemetry", telemetry.sdk.version="1.53.0"}}, instrumentationScopeInfo=InstrumentationScopeInfo{name=test, version=null, schemaUrl=null, attributes={}}, timestampEpochNanos=0, observedTimestampEpochNanos=1755714644912715000, spanContext=ImmutableSpanContext{traceId=00000000000000000000000000000000, spanId=0000000000000000, traceFlags=00, traceState=ArrayBasedTraceState{entries=[]}, remote=false, valid=false}, severity=UNDEFINED_SEVERITY_NUMBER, severityText=null, totalAttributeCount=0, extendedAttributes={}, bodyValue=ValueString{test}, eventName=null}]	
to satisfy given requirements, but these elements did not:	
ExtendedSdkLogRecordData{resource=Resource{schemaUrl=null, attributes={service.instance.id="a6a744b8-2dcb-4a31-a46f-31fc9fd4b6cf", service.name="unknown_service:java", telemetry.sdk.language="java", telemetry.sdk.name="opentelemetry", telemetry.sdk.version="1.53.0"}}, instrumentationScopeInfo=InstrumentationScopeInfo{name=test, version=null, schemaUrl=null, attributes={}}, timestampEpochNanos=0, observedTimestampEpochNanos=1755714644912715000, spanContext=ImmutableSpanContext{traceId=00000000000000000000000000000000, spanId=0000000000000000, traceFlags=00, traceState=ArrayBasedTraceState{entries=[]}, remote=false, valid=false}, severity=UNDEFINED_SEVERITY_NUMBER, severityText=null, totalAttributeCount=0, extendedAttributes={}, bodyValue=ValueString{test}, eventName=null}	
error: java.lang.AssertionError: 	
[...]

PR #2031, opened 2025-07-16
add declarative config for baggage

zeitlinger:
Fixes #2013

PR #2152, opened 2025-08-21
add baggage before it can be exported

zeitlinger:
Fixes https://github.com/open-telemetry/opentelemetry-java-contrib/issues/2150

### DIFF
--- a/baggage-processor/src/main/java/io/opentelemetry/contrib/baggage/processor/BaggageProcessorCustomizer.java
+++ b/baggage-processor/src/main/java/io/opentelemetry/contrib/baggage/processor/BaggageProcessorCustomizer.java
@@ -57,6 +57,7 @@ public class BaggageProcessorCustomizer implements AutoConfigurationCustomizerPr
       return;
     }
 
+    // need to add before the batch span processor
     sdkLoggerProviderBuilder.addLogRecordProcessorFirst(createBaggageLogRecordProcessor(keys));
   }
 


zeitlinger:
We need to add the baggage processor before the batch span processor (that is usually already configured) - because it can start to export as soon as it's offered the span - see https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java#L177-L182

@trask fyi since we discussed this

baggage-processor/src/main/java/io/opentelemetry/contrib/baggage/processor/BaggageProcessorCustomizer.java
```diff
@@ -37,7 +37,7 @@ private static void addSpanProcessor(
       return;
     }
 
-    sdkTracerProviderBuilder.addSpanProcessor(createBaggageSpanProcessor(keys));
+    sdkTracerProviderBuilder.addSpanProcessorFirst(createBaggageSpanProcessor(keys));
```
trask, line 41:
```suggestion
    // need to add before the batch span processor
    sdkTracerProviderBuilder.addSpanProcessorFirst(createBaggageSpanProcessor(keys));
```

baggage-processor/src/main/java/io/opentelemetry/contrib/baggage/processor/BaggageProcessorCustomizer.java
```diff
@@ -56,7 +56,7 @@ private static void addLogRecordProcessor(
       return;
     }
 
-    sdkLoggerProviderBuilder.addLogRecordProcessor(createBaggageLogRecordProcessor(keys));
+    sdkLoggerProviderBuilder.addLogRecordProcessorFirst(createBaggageLogRecordProcessor(keys));
```
trask, line 61:
```suggestion
    // need to add before the batch span processor
    sdkLoggerProviderBuilder.addLogRecordProcessorFirst(createBaggageLogRecordProcessor(keys));
```

trask (approved):

trask:
> We need to add the baggage processor before the batch span processor (that is usually already configured) - because it can start to export as soon as it's offered the span - see https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java#L177-L182

hm, I feel like this would have been a bigger problem if batch span processor wasn't always added last in this typical situation (and not sporadic then)

zeitlinger:
> hm, I feel like this would have been a bigger problem if batch span processor wasn't always added last in this typical situation (and not sporadic then)

Possible - but I didn't spot anything else.

@laurit do you have an idea?

zeitlinger:
> > We need to add the baggage processor before the batch span processor (that is usually already configured) - because it can start to export as soon as it's offered the span - see https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java#L177-L182
> 
> hm, I feel like this would have been a bigger problem if batch span processor wasn't always added last in this typical situation (and not sporadic then)

maybe it happens in the test due to the short delay: https://github.com/zeitlinger/opentelemetry-java-contrib/blob/00c7b2841f7c16371a6c4a7c3eae5358fd0e3516/baggage-processor/src/test/java/io/opentelemetry/contrib/baggage/processor/BaggageProcessorCustomizerTest.java#L109

I've included this PR in https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2031 where the error is happening now

MikeGoldsmith (approved):

laurit:
@zeitlinger thanks for sorting this out, I have high hopes of this resolving the flaky test issue

PR #2163, opened 2025-08-22
Fix release workflow

trask:
space after the trailing `\` apparently causes problems


breedx-splk (approved):
Weird...and surprising!

PR #2165, opened 2025-08-22
Add already-published option to release workflow

trask:
This happens from time to time (including this time), so this seems useful to have available.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,6 +1,11 @@
 name: Release
 on:
   workflow_dispatch:
+    inputs:
+      already-published:
+        description: 'Skip publishing, download artifacts from Maven Central instead'
+        default: false
+        type: boolean
 
 permissions:
   contents: read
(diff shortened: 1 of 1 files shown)


laurit (approved):

.github/workflows/release.yml
```diff
@@ -1,6 +1,11 @@
 name: Release
 on:
   workflow_dispatch:
+    inputs:
+      already-published:
+        description: 'Skip publishing, download artifacts from Maven Central instead'
+        default: false
+        type: boolean
```
breedx-splk, line 8: Just so I understand this -- this allows one to manually pass in "true" for this in order to skip the actual build+publish step when doing a release. This allows the rest of the release workflow to run to completion without failure....including the generation of release notes and the release in GH.

Is that right? Does sound useful!

.github/workflows/release.yml
```diff
@@ -125,14 +130,31 @@ jobs:
...
+          mkdir -p jmx-scraper/build/libs
+          
+          curl -L -o jmx-metrics/build/libs/opentelemetry-jmx-metrics-$VERSION-alpha.jar \
+            "https://repo1.maven.org/maven2/io/opentelemetry/contrib/opentelemetry-jmx-metrics/$VERSION-alpha/opentelemetry-jmx-metrics-$VERSION-alpha.jar"
+          curl -L -o jmx-metrics/build/libs/opentelemetry-jmx-metrics-$VERSION-alpha.jar.asc \
+            "https://repo1.maven.org/maven2/io/opentelemetry/contrib/opentelemetry-jmx-metrics/$VERSION-alpha/opentelemetry-jmx-metrics-$VERSION-alpha.jar.asc"
+          curl -L -o jmx-scraper/build/libs/opentelemetry-jmx-scraper-$VERSION-alpha.jar \
+            "https://repo1.maven.org/maven2/io/opentelemetry/contrib/opentelemetry-jmx-scraper/$VERSION-alpha/opentelemetry-jmx-scraper-$VERSION-alpha.jar"
+          curl -L -o jmx-scraper/build/libs/opentelemetry-jmx-scraper-$VERSION-alpha.jar.asc \
+            "https://repo1.maven.org/maven2/io/opentelemetry/contrib/opentelemetry-jmx-scraper/$VERSION-alpha/opentelemetry-jmx-scraper-$VERSION-alpha.jar.asc"
```
breedx-splk, line 156: Is the scraper a new build output now? I didn't see it in 1.48 nor 1.47...

breedx-splk (approved):
Seems helpful! ❤️

.github/workflows/release.yml
trask: #2085

.github/workflows/release.yml
trask: yup! it's useful for when it succeeds publishing but fails during creating the github release (where we introduce bugs from time-to-time)

PR #2168, opened 2025-08-22
Sync with instrumentation workflows

trask:


.github/workflows/build-common.yml
```diff
@@ -18,6 +18,25 @@ permissions:
   contents: read
 
 jobs:
+  spotless:
```
trask, line 21: this splits spotless out into separate job so if it fails you can still see if the other jobs succeed or fail

probably not that important anymore with the autospotless automation

UPDATE I think it's helpful after all, for people running jobs in their forks (autospotless only runs in this repo)

.github/workflows/build-common.yml
```diff
@@ -44,8 +63,6 @@ jobs:
       fail-fast: false
       matrix:
         os:
-          - macos-latest
-          - macos-13
```
trask, line 48: we don't test macos in instrumentation repo, I think its probably ok not to test in this repo too(?)

laurit (approved):

PR #2173, opened 2025-08-24
Fix link check failures

trask:


.github/scripts/lychee-config.toml
```diff
@@ -6,11 +6,11 @@ max_concurrency = 4
...
-# excluding links to pull requests and issues is done for performance
-# sonatype snapshots are currrently unbrowseable
 exclude = [
+  # excluding links to pull requests and issues is done for performance
   "^https://github.com/open-telemetry/opentelemetry-java-contrib/(issues|pull)/\\d+$",
-  '^https://central.sonatype.com/service/rest/repository/browse/maven-snapshots/io/opentelemetry/contrib/$',
+  # groovy-lang.org links are timing out when accessed via lychee
+  "^http://(docs.|)groovy-lang.org",
+  # mvnrepository.com links are returning 403 when accessed via lychee
+  "^https://mvnrepository.com",
```
laurit: currently link check seems to pass, are these needed?

trask: there were failing 😅

let's revert them for now and see...

```suggestion
```

PR #2183, opened 2025-08-25
Disk buffering api implementation

LikeTheSalad:
Creation of a default implementation for the new API that was previously defined [here](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2084).

> [!NOTE]
> It looks like there are a lot of changes, but **27** of those _"files changes"_ are deleted files from the old implementation types and their test types. Also, a couple of file changes are just moving around files and/or adding minor changes.

## New usage

The same old API behavior is kept for the new API's default implementation.

### Writing

Setting up the signal storage objects as well as their `toDisk` exporters to be used in an OpenTelemetry instance.

```java
// Root dir
File rootDir = new File("/some/root");

// Setting up span storage
SignalStorage.Span spanStorage = FileSpanStorage.create(new File(rootDir, "spans"));

// Setting up metric storage
SignalStorage.Metric metricStorage = FileMetricStorage.create(new File(rootDir, "metrics"));

// Setting up log storage
SignalStorage.LogRecord logStorage = FileLogRecordStorage.create(new File(rootDir, "logs"));

// Setting up span to disk exporter
SpanToDiskExporter spanToDiskExporter =
    SpanToDiskExporter.builder(spanStorage).setExporterCallback(spanCallback).build();
// Setting up metric to disk
MetricToDiskExporter metricToDiskExporter =
    MetricToDiskExporter.builder(metricStorage).setExporterCallback(metricCallback).build();
// Setting up log to disk exporter
LogRecordToDiskExporter logToDiskExporter =
    LogRecordToDiskExporter.builder(logStorage).setExporterCallback(logCallback).build();
```

### Reading

```java
// Example of reading an exporting spans from disk
OtlpHttpSpanExporter networkExporter;
Iterator<Collection<SpanData>> spanCollections = spanStorage.iterator();
while (spanCollections.hasNext()) {
    networkExporter.export(spanCollections.next());
}
```

More info in the README.md and DESIGN.md files.

### DIFF
--- a/disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/FileSignalStorage.java
+++ b/disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/FileSignalStorage.java
@@ -49,7 +49,8 @@ public final class FileSignalStorage<T> implements SignalStorage<T> {
         return CompletableFuture.completedFuture(WriteResult.successful());
       }
       logger.fine("Could not store batch in disk.");
-      return CompletableFuture.completedFuture(WriteResult.error(null));
+      return CompletableFuture.completedFuture(
+          WriteResult.error(new Exception("Could not store batch in disk for an unknown reason.")));
     } catch (IOException e) {
       logger.log(
           Level.WARNING,
--- a/disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/FolderManager.java
+++ b/disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/FolderManager.java
@@ -31,7 +31,8 @@ public final class FolderManager implements Closeable {
       File destinationDir, FileStorageConfiguration configuration, Clock clock) {
     if (destinationDir.isFile()) {
       throw new IllegalArgumentException("destinationDir must be a directory");
-    } else if (!destinationDir.exists()) {
+    }
+    if (!destinationDir.exists()) {
       if (!destinationDir.mkdirs()) {
         throw new IllegalStateException("Could not create dir: " + destinationDir);
       }


LikeTheSalad:
It's ready for review. Please take a look, @zeitlinger.

LikeTheSalad:
> Can you create a PR that just moves things around? That would make it easier to review.

It depends. I was actually trying to do it in 2 parts, but I'm reusing a lot of logic from the old API, while at the same time, I had to change internal tools to adapt them to the new one. So if I don't delete the old API in this PR, then the project won't compile, because of the changes to the old internal tools needed to work with the new implementation. What parts of the current PR are you referring to moving around?

marandaneto:
> > Can you create a PR that just moves things around? That would make it easier to review.
> 
> It depends. I was actually trying to do it in 2 parts, but I'm reusing a lot of logic from the old API, while at the same time, I had to change internal tools to adapt them to the new one. So if I don't delete the old API in this PR, then the project won't compile, because of the changes to the old internal tools needed to work with the new implementation. What parts of the current PR are you referring to moving around?

an option would be to stack PRs on top of each other

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/Storage.java
```diff
@@ -53,91 +43,151 @@ public boolean isDebugEnabled() {
...
    * Attempts to read an item from a ready-to-read file.
    *
-   * @param processing Is passed over to {@link ReadableFile#readAndProcess(Function)}.
    * @throws IOException If an unexpected error happens.
    */
-  public ReadableResult readAndProcess(Function<byte[], ProcessResult> processing)
-      throws IOException {
-    return readAndProcess(processing, 1);
+  @Nullable
+  public ReadableResult<T> readNext(SignalDeserializer<T> deserializer) throws IOException {
```
tylerbenson, line 89: I think it's a mistake to move the deserialization deeper into the storage as it prevents what I've done in #2190 to avoid unnecessary re-serialization. Perhaps if this class had an additional access method that returned an InputStream factory against the file (so it could be read twice, first to count the items and second to stream it to a network output stream).

LikeTheSalad: > I don't think this new Storage API is as flexible as you suggest. It locks the re-serialization into place and forces extra inefficiencies. It would be better if Storage dealt with raw byte arrays or IO streams.

I think there might be a misunderstanding of the new API, so I'll clarify it here. In a nutshell, `Storage` is not part of the API; it's an internal tool. This is the core of the new API: [SignalStorage](https://github.com/open-telemetry/opentelemetry-java-contrib/blob/1ac0bcfed4687239214d6fe48af2d6d2bb751702/disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/storage/SignalStorage.java#L21).

Screenshot 2025-08-28 at 08 18 30

It provides a way to write and read a collection of signals, as it's also the way that upstream exporters receive data.

> It locks the re-serialization into place and forces extra inefficiencies

The API isn't aware of its implementations, so it's not aware of the default implementation that uses serialization and deserialization. The reason why the API provides an iterator of a collection of signals as a way to read stored data is not to enforce anything around serialization, but simply because that's also what it receives in its `write()` method.

> I think it's a mistake to move the deserialization deeper into the storage as it prevents what I've done in https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2190 to avoid unnecessary re-serialization.

Serialization is just an internal part of what the default implementation needs to do to receive and return the same kind of data that it receives, and it was never the intention to make it publicly supported in this module; the same way it is also not publicly supported in the core Java repo. So it shouldn't mean a mistake to move it around, as the users of the default impl shouldn't be aware of it.

As I mentioned to you a couple of days ago when you suggested this approach over DM, I wouldn't recommend that you work on top of the old API, as a lot of it is going to be changed or removed.

> Perhaps if this class had an additional access method that returned an InputStream factory against the file (so it could be read twice, first to count the items and second to stream it to a network output stream).

Maybe in the future, not this class, because it's not part of the API, but we might be able to somehow extend `SignalStorage` to allow consuming the data already serialized, so that it can be forwarded directly to a network exporter. This is definitely something that can help with performance for the specific use case where the storage serializes signals in the same way that they are going to be needed by the network exporter impl that will consume them. The downside of this approach, and the reason why deserializing it is more generic, is that it ties the storage impl to a specific data format and protocol for exporting. So if we wanted to do so in the future, it must be done in a way that custom implementations can provide their own serialized versions of the data, in case the format we use is not what they need, to enable them to work on their own impls.

---

## Alternative with the new API

Since the new API is just an interface, it shouldn't prevent you from creating your own implementation that adds the functionality that you need. To make it work with your use case, you could wrap the default implementation around your custom implementation for the `write()` part, and then add a method that provides you with an imputstream where you manually read the files to provide the serialized data. Alternatively, if you'd like to reuse disk buffering's internal `Storage` class, which is the one that manages the stored data, you can add a method to it that does something around the lines of what you mentioned here: 

> Perhaps if this class had an additional access method that returned an InputStream factory against the file (so it could be read twice, first to count the items and second to stream it to a network output stream).

And use it for your custom implementation. Just bear in mind that, since Storage is internal, there's no guarantee to maintain it as is forever (though I don't see why it should change tbh).

bidetofevil:
Looking at the changes proposed by @tylerbenson in #2190 along side the changes proposed by @LikeTheSalad in this PR, it seems like the conflicting issue is that in the former, the deserialization functionality was hoisted out from inside of `FromDiskExporter` into something that is settable via the builder.

Tyler's change allowed not only `SignalDeserializer` to be set, but also how it's actually used under the hood. Whereas the serializer was originally used uniformly (so the custom serializer just changes the serialized form), the updated implementation allowed you to change how it the implementation interacts with it by allow you to specify the implementation of the `BiFunction` in which it was used. Does that sound right, @tylerbenson?

Meanwhile, in order to streamline the usage workflow, Cesar's change took way the `FromDiskExporter` altogether, and made public a tweaked version of `SignalStorageExporter` that allowed the user to specify the `SignalStorage` and `ExporterCallback` for a given type. While this makes usage easier, it takes away the ability for alternate implementations to override how the transformation is done because the interfaces enforce that the input of `write()` is a concrete `Collection` of `T`, and what it exposes is an iterator on an iterable of the same type, `Collection<T>`. In other words, the interface doesn't only guarantee outcome, but it forces the input and output type in terms of how it'll be used. Does that correctly characterize your change, @LikeTheSalad?

With the elimination of that previously overridable layer, you're focused to hand off deserialized Collections of the signal, which eliminates the ability to override the behaviour of that hand off, as is done in `setExportFunction()` in #2190. To do that in the design proposed in this PR,  you'd have to rewrite pretty much everything. The parameters passed between the interface doesn't afford you any flexibility; you basically need to redo `SignalStorageExporter`, `SignalStorage`, and how they interact. At that point, you might as well rewrite the whole package.

Assuming that's not what we want (re: that's not what we want), I think we can go in two different directions:

1. Try to open up a layer in the new interface that can be overwritten, i.e. so the implementation of how the interfaces interact with each other can be overridden.

2. Revert to the previous interface and accomplish the goals of the refactor in a different way. Perhaps there is a way to lock down and streamline that doesn't blow out the permutation of variables that need supporting?

Thoughts, @breedx-splk?

tylerbenson:
@LikeTheSalad
In my experience, I've been involved in a lot of lengthy discussions within OpenTelemetry. I didn't feel the discussion this week was aggressive or overly negative. To be clear, I never intended to cause you grief and I'm sorry for the headache this has caused.  It's not my intention to alienate a positive contributor from the community.

I see your point about Http/GrpcExporter being internal and agree it is indeed a problematic roadblock for my PR, so I'm going to close it until the SDK has a better option for how to proceed. Even the base class `Marshaler` is internal, so we would likely need to work with the SDK to expose some new interfaces to achieve my goal.

Meanwhile I'm going to unblock myself and copy over what I need of these classes to my internal implementation until the new API has been stabilized enough and I can see a way to reimplement it.

breedx-splk:
(Disclaimer: I _still_ need to give the code in this PR a proper review!)

Since I was mentioned a few times above and I was part of the discussion at the SIG meeting(s), I thought I would open an issue to help track this: https://github.com/open-telemetry/opentelemetry-java-contrib/issues/2203. The goal of that issue is to try and capture the common ground approach that I think we all agree is possible with some additional future work.

The timing was indeed unfortunate -- we basically had two competing and conflicting and disruptive changes get proposed at the same time. 

@tylerbenson thanks for being understanding about this timing challenge and flexible enough to close your PR and circle back when the smoke clears.

@LikeTheSalad thank you for taking the time to write a thorough response. Same to you @bidetofevil.

LikeTheSalad:
Thank you for understanding, @tylerbenson. It would indeed make things easier if the SDK made some of the APIs you mentioned public. In the meantime, while those are still internal, I think we should still be able to enable your use case by tweaking the internal disk buffering API in a further update so that you can reuse and repurpose a lot of it, which I can help with if needed.

Thanks for chiming in, @breedx-splk, I like the idea of keeping track of a potential new feature in that issue and taking our time to discuss all the details there.

bidetofevil:
I think for me, it was unclear what was APIs were meant to be internal and what was meant to be external, especially given the use case Tyler was doing is explicitly about redoing some things that were considered internal, and that there was a desire to support it without needing to fork meant that we had to expose that part. If that was explicitly not wanted, then we probably started off on the wrong foot.

But Jason was right in pointing out there are conflicting change. We probably shouldn't have just merged the API changes, then found a way to build an acceptable solution on top that would work for Tyler. I basically encouraged scope creep, and I apologize for that. Lets finish this one and then figure out the other.

tylerbenson:
@bidetofevil the main issue with my PR is that it uses (and requires users to provide) internal classes from the SDK, not anything specific to this codebase's structure.  Until the API/SDK changes, I don't think what I wanted to do is reasonable for a contrib library.  (I'm continuing to implement it privately, but that's just because I'm willing to accept breaking changes in the API/SDK.)

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/exporters/callback/ExporterCallback.java
```diff
@@ -0,0 +1,34 @@
...
+   * @param items The items that couldn't get stored in disk.
+   * @param error Optional - provides more information of why the operation failed.
+   */
+  void onExportError(Collection<T> items, @Nullable Throwable error);
+
+  /** Called when the exporter is closed. */
+  void onShutdown();
+
+  static <T> ExporterCallback<T> noop() {
+    return new NoopExporterCallback<>();
```
breedx-splk: In other parts of otel we try and minimize object creation. It might be nice to do that here too (like if the NoopExporterCallback had an INSTANCE), but the generics might make that challenging. 🤷🏻

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/exporters/LogRecordToDiskExporter.java
```diff
@@ -0,0 +1,77 @@
...
+  @Override
+  public CompletableResultCode shutdown() {
+    callback.onShutdown();
+    return CompletableResultCode.ofSuccess();
+  }
+
+  public static final class Builder {
+    private final SignalStorage.LogRecord storage;
+    private ExporterCallback<LogRecordData> callback = ExporterCallback.noop();
+    private Duration writeTimeout = Duration.ofSeconds(10);
```
breedx-splk: I like to surface these as constants at the top of the file to make it very visible what the default is.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/exporters/MetricToDiskExporter.java
```diff
@@ -0,0 +1,96 @@
...
+  public AggregationTemporality getAggregationTemporality(InstrumentType instrumentType) {
+    return aggregationTemporalitySelector.getAggregationTemporality(instrumentType);
+  }
+
+  public static final class Builder {
+    private final SignalStorage.Metric storage;
+    private AggregationTemporalitySelector aggregationTemporalitySelector =
+        AggregationTemporalitySelector.alwaysCumulative();
+    private ExporterCallback<MetricData> callback = ExporterCallback.noop();
+    private Duration writeTimeout = Duration.ofSeconds(10);
```
breedx-splk: factor out a constant for visibility.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/files/reader/DelimitedProtoStreamReader.java
```diff
@@ -28,7 +28,7 @@ public ReadResult readNext() throws IOException {
     if (inputStream.read(bytes) < 0) {
```
breedx-splk: This wasn't changed in this PR, but I noticed this when reviewing (and is related) -- if 0 was returned, we'd have a byte array of non-zero length returned but without any data read into it. 

```suggestion
    if (inputStream.read(bytes) <= 0) {
```

Also, now that we're not returning the `ReadResult` the caller has no indication how many bytes were actually read into the byte array! I think this could be troubling. Take the case where the `getNextItemSize()` returned `n` bytes, but then the actual `inputStream.read(bytes)` returns `n-3` (or some other smaller number, doesn't matter). In that case, the caller gets back a byte array that isn't "full" (ie. the bytes at the end weren't read from the stream).

Any ideas what to do about that?

LikeTheSalad: > This wasn't changed in this PR, but I noticed this when reviewing (and is related) -- if 0 was returned, we'd have a byte array of non-zero length returned but without any data read into it.

Thanks, it's a great point. I've applied your suggestion.

> Also, now that we're not returning the ReadResult the caller has no indication how many bytes were actually read into the byte array! I think this could be troubling. Take the case where the getNextItemSize() returned n bytes, but then the actual inputStream.read(bytes) returns n-3 (or some other smaller number, doesn't matter). In that case, the caller gets back a byte array that isn't "full" (ie. the bytes at the end weren't read from the stream).
> 
> Any ideas what to do about that?

If you're referring to `ReadResult`'s `totalReadLength` field, it was removed in [this other PR](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1917/files#diff-fb0920b798b73be9fdbda266cabf3752079478b8d8ac91983c3ee11b754da5fcL104}) a while ago. It used to help tell how many bytes were processed so that they could get deleted on success, but after [this PR](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1917) it wasn't needed anymore, since the `FileStream` class added there keeps track of the position from the input file to know where to truncate it on success.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/exporters/MetricToDiskExporter.java
LikeTheSalad: Thanks, I've applied this change.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/exporters/LogRecordToDiskExporter.java
LikeTheSalad: Thanks, I've applied this change.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/exporters/callback/ExporterCallback.java
LikeTheSalad: I see what you mean. It's tricky to have a singleton due to the generics, so I've just created them in their respective exporters as singletons and removed this `noop()` function to avoid encouraging its use, or at least to make object creation in this case less straightforward.

breedx-splk: Sounds like a good approach for now, thanks!

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/files/reader/DelimitedProtoStreamReader.java
breedx-splk: I appreciate that, but I think there's still the potential for a bug here. The size of the next item (`getNextItemSize()`) comes from a byte in the protobuffer, and it is an indication of how large (how many bytes) the next "item" (object, whatever) is. There's no guarantee, however, that a single read on the input stream will have that many bytes available...which is exactly the reason it returns an int -- so that the caller can know how may bytes were read.

Unfortunately, to do this the right way, I think you have to loop.

```java
int offset = 0;
int readCt;
do {
  readCt = inputStream.read(bytes, offset, itemSize - offset);
  offset += readCt;
} while (readCt > 0);
if(offset != itemSize){
  return null;  // unable to read the whole item correctly
}
return bytes;
```

LikeTheSalad: Ah, seems like I misunderstood the first comment. You're right, I've just applied these changes. Thanks.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/FileSignalStorage.java
```diff
@@ -0,0 +1,91 @@
...
+  @Override
+  public CompletableFuture<WriteResult> write(Collection<T> items) {
+    logger.finer("Intercepting batch.");
+    try {
+      serializer.initialize(items);
+      if (storage.write(serializer)) {
+        return CompletableFuture.completedFuture(WriteResult.successful());
+      }
+      logger.fine("Could not store batch in disk.");
+      return CompletableFuture.completedFuture(WriteResult.error(null));
```
breedx-splk: Not blocking for this PR, but it would be nice to think about how we might give better information about the _why_ when a write fails. As written here, it's a bit vague and unactionable (hard to troubleshoot).

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/FolderManager.java
```diff
@@ -7,38 +7,47 @@
...
+  private final FileStorageConfiguration configuration;
   @Nullable private ReadableFile currentReadableFile;
   @Nullable private WritableFile currentWritableFile;
 
-  public FolderManager(File folder, StorageConfiguration configuration, Clock clock) {
+  public static FolderManager create(
+      File destinationDir, FileStorageConfiguration configuration, Clock clock) {
+    if (destinationDir.isFile()) {
+      throw new IllegalArgumentException("destinationDir must be a directory");
+    } else if (!destinationDir.exists()) {
```
breedx-splk: nit: `else` is redundant because the above `if` throws.

LikeTheSalad: Good point, thanks. I've removed it.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/FileSignalStorage.java
LikeTheSalad: I see what you mean. It's unlikely that this line is reached if there's an error, because if there is, then the catch would be called, where we provide the error details on line 59. Still, I've updated it to pass a generic error message at least, so that there's some information in case this line is executed.

breedx-splk (approved):
Thanks for staying with it.

breedx-splk:
> > Thanks for staying with it.
> 
> Cheers! Just checking in — is there anything else needed before we can move forward with these changes, @breedx-splk?

I don't think so, nah. Let's move this forward. Thanks again.

PR #2190, opened 2025-08-27
Make the disk buffered exporting more efficient

tylerbenson:
Following on from #2138...

The previous implementation would load the protobuf encoded bytes from disk, deserialize it, then send it through the delegate exporter pipeline which would likely re-serialize it into the exact same format.

This PR reduces the CPU burden by providing an optional alternate path which bypasses the exporter pipeline. Instead it takes the serialized bytes from disk and sends it straight to an HttpExporter or GrpcExporter, which only takes a `Marshaler` class allowing the serialized bytes to be passed straight through.

Since these exporters like to know the number of serialized elements that are being passed in, I added a function (with the help of Claude) that counts the serialized elements within the protobuf encoded byte array.

A future optimization would be to allow streaming the bytes directly from disk without reading them back into a byte array.


LikeTheSalad (changes requested):
Blocking as it's not compatible with the new API. More details [here](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2183#discussion_r2306413275).

PR #2211, opened 2025-09-06
Pin npm dependency version

trask:
Resolves https://github.com/open-telemetry/opentelemetry-java-contrib/security/code-scanning/102


.github/workflows/assign-issue-owners.yml
```diff
@@ -20,7 +20,7 @@ jobs:
         uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
 
       - name: Install js-yaml
-        run: npm install js-yaml
+        run: npm install js-yaml@4.0.0
```
trask, line 23: intentionally using version prior to latest so we can see right away if renovate custom manager is working

laurit (approved):

PR #2214, opened 2025-09-06
Apply final to public API classes where possible - ibm-mq-metrics

trask:


jaydeluca (approved):
lgtm. cc @atoulme

PR #2217, opened 2025-09-06
Apply final to public API classes where possible - kafka-exporter, micrometer-meter-provider, noop-api, processors, resource-providers, samplers

trask:


processors/src/main/java/io/opentelemetry/contrib/filter/FilteringLogRecordProcessor.java
```diff
@@ -11,10 +11,10 @@
 import io.opentelemetry.sdk.logs.data.LogRecordData;
 import java.util.function.Predicate;
 
-public class FilteringLogRecordProcessor implements LogRecordProcessor {
+public final class FilteringLogRecordProcessor implements LogRecordProcessor {
 
-  public final LogRecordProcessor delegate;
-  public final Predicate<LogRecordData> predicate;
+  private final LogRecordProcessor delegate;
+  private final Predicate<LogRecordData> predicate;
```
jaydeluca, line 17: it seems reasonable to me to expect that people would not be relying on these being available

PR #2219, opened 2025-09-06
Minor updates to mise/lychee workflow

trask:
Link check is failing due to #2221


mise.toml
```diff
@@ -5,7 +5,8 @@ lychee = "0.18.1"
 run = 'lychee --verbose --scheme file --include-fragments {{arg(name="files", default=".")}}'
 
 [tasks.lint-links]
-run = 'lychee --verbose --config .github/scripts/lychee-config.toml {{arg(name="files", default=".")}}'
+run = 'lychee --verbose --config .github/config/lychee.toml {{arg(name="files", var=true, default=".")}}'
```
trask, line 8: `var=true` is needed to handle when there are multiple .md files updated in a single PR

laurit (approved):

PR #2220, opened 2025-09-06
Fix some links

trask:
Depends on fix in #2219

### DIFF
--- a/jmx-scraper/README.md
+++ b/jmx-scraper/README.md
@@ -12,7 +12,7 @@ This project is released as part of the [OpenTelemetry Java Contrib](https://git
 The latest release is available from:
 
 - [GitHub Release assets](https://github.com/open-telemetry/opentelemetry-java-contrib/releases/latest/download/opentelemetry-jmx-scraper.jar)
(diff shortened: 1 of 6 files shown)


jmx-scraper/README.md
```diff
@@ -12,7 +12,7 @@ This project is released as part of the [OpenTelemetry Java Contrib](https://git
 The latest release is available from:
 
 - [GitHub Release assets](https://github.com/open-telemetry/opentelemetry-java-contrib/releases/latest/download/opentelemetry-jmx-scraper.jar)
-- [Maven Central](https://central.sonatype.com/artifact/io.opentelemetry.contrib/opentelemetry-jmx-scraper) or can be browsed via [MVN Repository](https://mvnrepository.com/artifact/io.opentelemetry.contrib/opentelemetry-jmx-scraper).
+- [Maven Central](https://central.sonatype.com/artifact/io.opentelemetry.contrib/opentelemetry-jmx-scraper)
```
trask, line 15: the mvnrepository link isn't broken, but they don't allow bots. we could exclude the link from bot checker, but in this case I think the maven central link allows browsing and is what we use elsewhere

jaydeluca (approved):

PR #2255, opened 2025-09-12
Pin npm hash

trask:
Resolves https://github.com/open-telemetry/opentelemetry-java-contrib/security/code-scanning/107


.github/scripts/package.json
trask, line 1: renovate will update this file and regenerate the `package-lock.json`

jaydeluca (approved):

PR #2261, opened 2025-09-15
Update otel

zeitlinger:
Replaces #2260

### DIFF
--- a/dependencyManagement/build.gradle.kts
+++ b/dependencyManagement/build.gradle.kts
@@ -2,7 +2,7 @@ plugins {
   `java-platform`
 }
 
-val otelInstrumentationVersion = "2.19.0-alpha"
+val otelInstrumentationVersion = "2.20.0-alpha"
 val semconvVersion = "1.37.0"
 
 javaPlatform {


azure-resources/src/main/java/io/opentelemetry/contrib/azure/resource/IncubatingAttributes.java
```diff
@@ -24,10 +24,10 @@ final class IncubatingAttributes {
...
 
   public static final class CloudPlatformIncubatingValues {
-    public static final String AZURE_VM = "azure_vm";
-    public static final String AZURE_AKS = "azure_aks";
-    public static final String AZURE_FUNCTIONS = "azure_functions";
-    public static final String AZURE_APP_SERVICE = "azure_app_service";
+    public static final String AZURE_VM = "azure.vm";
+    public static final String AZURE_AKS = "azure.aks";
+    public static final String AZURE_FUNCTIONS = "azure.functions";
+    public static final String AZURE_APP_SERVICE = "azure.app_service";
```
SylvainJuge: Those are semconv-related, why wasn't this needed when the semconv update to 1.37.0 was done ? I haven't looked in detail, but do the tests rely directly on those attributes instead of relying on the semconv incubating artifact equivalent ?

zeitlinger: good question - it is indeed related to semconv update - it looks like we got the config wrong and agent overwrites the version

zeitlinger: it's not working for incubating - I'll open a separate PR
```
+--- io.opentelemetry.semconv:opentelemetry-semconv-incubating -> 1.34.0-alpha
|    \--- io.opentelemetry.semconv:opentelemetry-semconv:1.34.0 -> 1.37.0
```

zeitlinger: here it is: https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2263

laurit: > why wasn't this needed when the semconv update to 1.37.0 was done

apparently the way we enforce the instrumentation bom overrides the semconv version declared in this project

laurit (approved):

Issue #2028, opened 2025-07-09
declarative config: span stacktrace


zeitlinger:
Port StackTraceAutoConfig to declarative config

Blocked by https://github.com/open-telemetry/opentelemetry-java/pull/7480 : uses `otel.java.experimental.span-stacktrace` - see https://github.com/open-telemetry/opentelemetry-java-contrib/issues/2013

PR #2262, opened 2025-09-15
declarative config: support Span stacktrace

zeitlinger:
Fixes https://github.com/open-telemetry/opentelemetry-java-contrib/issues/2028


span-stacktrace/src/main/java/io/opentelemetry/contrib/stacktrace/StackTraceComponentProvider.java
```diff
@@ -0,0 +1,34 @@
...
+import io.opentelemetry.instrumentation.config.bridge.DeclarativeConfigPropertiesBridgeBuilder;
+import io.opentelemetry.sdk.autoconfigure.spi.internal.ComponentProvider;
+import io.opentelemetry.sdk.trace.SpanProcessor;
+
+@SuppressWarnings("rawtypes")
+@AutoService(ComponentProvider.class)
+public class StackTraceComponentProvider implements ComponentProvider<SpanProcessor> {
+  @Override
+  public String getName() {
+    return "experimental-stacktrace";
```
jaydeluca: not sure if this has been discussed, but when using multi-word names, when should we use hyphens vs underscores? In [cloudfoundry we use underscore](https://github.com/open-telemetry/opentelemetry-java-contrib/blob/main/cloudfoundry-resources/src/main/java/io/opentelemetry/contrib/cloudfoundry/resources/CloudFoundryResourceDetector.java#L21), [hyphen for xray lambda](https://github.com/open-telemetry/opentelemetry-java-contrib/blob/main/aws-xray-propagator/src/main/java/io/opentelemetry/contrib/awsxray/propagator/internal/AwsXrayLambdaComponentProvider.java#L21), [underscore for Rule based routing sampler](https://github.com/open-telemetry/opentelemetry-java-contrib/blob/main/samplers/src/main/java/io/opentelemetry/contrib/sampler/internal/RuleBasedRoutingSamplerComponentProvider.java#L37)

Should we have some consistent convention?

zeitlinger: yes, should be underscores

span-stacktrace/README.md
```diff
@@ -25,6 +24,22 @@ SDK when included in the application runtime dependencies.
...
+
+You can enable the stacktrace span processor using declarative YAML configuration with the OpenTelemetry SDK. For example:
+
+```yaml
+file_format: 1.0-rc.1
+tracer_provider:
+  processors:
+    - experimental-stacktrace:
+        min:
+          duration: 10 # minimal duration in ms, default is 5, MUST be an integer
```
SylvainJuge: I know this might be related to the existing configuration option, but having `min` as an intermediate node feels weird, could we instead have `min_duration` ?

span-stacktrace/README.md
```diff
@@ -25,6 +24,22 @@ SDK when included in the application runtime dependencies.
...
 
+### Usage with declarative configuration
+
+You can enable the stacktrace span processor using declarative YAML configuration with the OpenTelemetry SDK. For example:
+
+```yaml
+file_format: 1.0-rc.1
+tracer_provider:
+  processors:
+    - experimental-stacktrace:
```
SylvainJuge:
```suggestion
    - experimental_stacktrace:
```

span-stacktrace/README.md
zeitlinger: yes, we can do that

zeitlinger: done

zeitlinger:
@SylvainJuge please check again

SylvainJuge (approved):

span-stacktrace/src/main/java/io/opentelemetry/contrib/stacktrace/StackTraceComponentProvider.java
```diff
@@ -0,0 +1,35 @@
...
+  @Override
+  public String getName() {
+    return "experimental_stacktrace";
+  }
+
+  @Override
+  public SpanProcessor create(DeclarativeConfigProperties config) {
+    return StackTraceAutoConfig.create(
+        new DeclarativeConfigPropertiesBridgeBuilder()
+            .addMapping(StackTraceAutoConfig.CONFIG_MIN_DURATION, "min_duration")
```
robsunday, line 26: PR looks good to me and the comment below is not directly related to it.
When looking at this real life sample use of mappings I have an impression that it is easy to make a mistake in some cases. This is because order of adding mappings matters. If more strict mapping is added after more general, then it will never be executed, like
```
addMapping("aa.bb", "v1");
addMapping("aa.bb.cc", "v2");
```
What do you think about adding validation to `DeclarativeConfigProperitesBridgeBuilder` class to make sure that more specific mapping is always added earlier?
Instead of validation, we could sort the entries automatically in the builder by using TreeMap instead of LinkedHashMap.

trask: @robsunday can you open an issue for this so we don't lose it?

PR #2263, opened 2025-09-15
fix version resolution for semconv

zeitlinger:
Found in https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2261

### DIFF
--- a/buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
@@ -19,6 +19,7 @@ base.archivesName.set("opentelemetry-${project.name}")
 
 // Version to use to compile code and run tests.
 val DEFAULT_JAVA_VERSION = JavaVersion.VERSION_17
+val semconvVersion = "1.37.0"
 
 java {
   toolchain {
@@ -77,7 +78,7 @@ tasks {
       exceptionFormat = TestExceptionFormat.FULL
       showStandardStreams = true
     }
-    
+
     configure<JacocoTaskExtension> {
       // only care about code coverage for code in this repository
       // (in particular avoiding netty classes which sometimes end up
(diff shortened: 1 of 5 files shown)


buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
```diff
@@ -19,6 +19,7 @@ base.archivesName.set("opentelemetry-${project.name}")
 
 // Version to use to compile code and run tests.
 val DEFAULT_JAVA_VERSION = JavaVersion.VERSION_17
+val semconvVersion = "1.37.0"
```
SylvainJuge: Do you think we could keep the version in `dependencyManagement/build.gradle.kts` to keep only having one place where versions are set and updated ? I'm not familiar with gradle resolution details but is there any particular reason we have to do this in `otel.java-conventions` ?

zeitlinger: it seems to be a bug in gradle dependency resolution - at least I didn't find a way to keep the version where it is now.

@laurit do you have an idea?

laurit: I'd replace `enforcedPlatform` with `platform`. I fail to see a reason why we should enforce this. cc @trask

zeitlinger: this also doesn't help 

```kotlin
configurations.configureEach {
      val b = isCanBeResolved && !isCanBeConsumed
      println("Configuring $name: add=$b")
      if (b || name.startsWith("test")) {
        extendsFrom(dependencyManagement)
      }
    }
```

zeitlinger: thanks @laurit - that was the issue!

zeitlinger:
@laurit it's ready

laurit (approved):

PR #2270, opened 2025-09-17
jmx-scraper implement stable service.instance.id

### LINKED ISSUE #2206: [jmx-scraper] implement stable service.instance.id
### Component(s)

jmx-scraper

JMX Scraper role is to collect metrics from a remote JVM, so the resource attributes of the captured metrics should reflect the remote JVM identity. For example, the `service.name` should be set on the JMX Scraper JVM to reflect the name of the observed remote JVM service.

Currently, a random `service.instance.id` ([semconv](https://opentelemetry.io/docs/specs/semconv/registry/attributes/service/#service-instance-id)) is randomly generated by the SDK, which means that a distinct value of `service.instance.id` is used for every JMX Scraper process, even if the target JVM process remains the same.

As a result, whenever the JMX Scraper is restarted (for example to update its configuration), it will report metrics that _appear_ to come from distinct service instances when the target JVM hasn't been restarted and there is a single instance, hence creating potentially confusing metrics.

In order to solve this, I would suggest to use a stable UUID value that depends only on the target JVM state, for example by querying the following MBean attributes values before the metrics are collected:
- `java.lang:type=Runtime` > `StartTime` attribute ([API](https://docs.oracle.com/en/java/javase/15/docs/api/java.management/java/lang/management/RuntimeMXBean.html#getStartTime()))
- `java.lang:type=Runtime` > `Pid` attribute ([API](https://docs.oracle.com/en/java/javase/15/docs/api/java.management/java/lang/management/RuntimeMXBean.html#getPid()))
- `java.lang:type=Runtime` > `Name` attribute ([API](https://docs.oracle.com/en/java/javase/15/docs/api/java.management/java/lang/management/RuntimeMXBean.html#getName())) which combines PID + hostname (at least on hotspot).

_No response_

### Additional context

_No response_

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

SylvainJuge:
Fixes #2206 

JMX scraper initialization have to be modified as
- we need to have the effective configuration from the properties customizer
- we need the effective configuration to establish connection to the remote JVM, and thus generate a stable UUID.

This is using an UUID v3 which relies on MD5, however this should be fine because it's only used to fingerprint JVM identity, apart from potentially confusing duplicated metrics there is no potential security risk involved.

Testing of this feature is done through what the tool reports on the standard output, like the "connection test" feature. Having proper end-to-end testing would require to validate the data sent to the otlp collector, and there is currently no such test, which means it would likely require a larger effort to be implemented. I think this shortcut is however fine given the feature only removes the natural randomness of `service.instance.id` resource attribute in the context of metrics captured through JMX.

### DIFF
--- a/jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/JmxScraper.java
+++ b/jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/JmxScraper.java
@@ -20,6 +20,7 @@ import io.opentelemetry.instrumentation.jmx.engine.JmxMetricInsight;
 import io.opentelemetry.instrumentation.jmx.engine.MetricConfiguration;
 import io.opentelemetry.instrumentation.jmx.yaml.RuleParser;
 import io.opentelemetry.sdk.autoconfigure.AutoConfiguredOpenTelemetrySdk;
+import io.opentelemetry.sdk.autoconfigure.spi.ConfigProperties;
 import io.opentelemetry.sdk.autoconfigure.spi.ConfigurationException;
 import io.opentelemetry.sdk.resources.Resource;
 import java.io.DataInputStream;
@@ -36,7 +37,7 @@ import java.util.Map;
 import java.util.Properties;
 import java.util.UUID;
 import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.logging.Level;
+import java.util.function.BiFunction;
 import java.util.logging.Logger;
 import javax.annotation.Nullable;
 import javax.management.MBeanServerConnection;
@@ -75,23 +76,24 @@ public final class JmxScraper {
 
       PropertiesCustomizer configCustomizer = new PropertiesCustomizer();
 
+      // we rely on the config customizer to be executed first to get effective config.
+      BiFunction<Resource, ConfigProperties, Resource> resourceCustomizer =
+          (resource, configProperties) -> {
+            UUID instanceId = getRemoteServiceInstanceId(configCustomizer.getConnectorBuilder());
+            if (resource.getAttribute(SERVICE_INSTANCE_ID) != null || instanceId == null) {
+              return resource;
+            }
+            logger.log(INFO, "remote service instance ID: " + instanceId);
+            return resource.merge(
+                Resource.create(Attributes.of(SERVICE_INSTANCE_ID, instanceId.toString())));
+          };
+
       // auto-configure SDK
       OpenTelemetry openTelemetry =
           AutoConfiguredOpenTelemetrySdk.builder()
               .addPropertiesSupplier(new PropertiesSupplier(argsConfig))
               .addPropertiesCustomizer(configCustomizer)
-              // we rely on the config customizer to be executed first to get effective config
-              .addResourceCustomizer(
-                  (resource, configProperties) -> {
-                    UUID instanceId =
-                        getRemoteServiceInstanceId(configCustomizer.getConnectorBuilder());
-                    if (resource.getAttribute(SERVICE_INSTANCE_ID) != null || instanceId == null) {
-                      return resource;
-                    }
-                    logger.log(Level.INFO, "remote service instance ID: " + instanceId);
-                    return resource.merge(
-                        Resource.create(Attributes.of(SERVICE_INSTANCE_ID, instanceId.toString())));
-                  })
+              .addResourceCustomizer(resourceCustomizer)
               .build()
               .getOpenTelemetrySdk();
 
@@ -146,7 +148,7 @@ public final class JmxScraper {
   }
 
   @Nullable
-  static UUID getRemoteServiceInstanceId(JmxConnectorBuilder connectorBuilder) {
+  private static UUID getRemoteServiceInstanceId(JmxConnectorBuilder connectorBuilder) {
     try (JMXConnector jmxConnector = connectorBuilder.build()) {
       MBeanServerConnection connection = jmxConnector.getMBeanServerConnection();
 


jmx-scraper/src/test/java/io/opentelemetry/contrib/jmxscraper/config/PropertiesCustomizerTest.java
```diff
@@ -17,16 +17,20 @@
 
 class PropertiesCustomizerTest {
 
+  private static final String DUMMY_URL = "service:jmx:rmi:///jndi/rmi://host:999/jmxrmi";
```
SylvainJuge, line 20: [for reviewer] we now have to test with an URL that can be parsed because the parsing is triggered, while it used to not be parsed previously.

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/JmxConnectionTest.java
```diff
@@ -175,6 +176,39 @@ void serverSslClientSsl(JmxScraperContainer.ConfigSource configSource) {
...
+            if (logLine.contains("remote service instance ID")) {
+              serviceId = UUID.fromString(logLine.substring(logLine.lastIndexOf(":") + 1).trim());
+            }
+          }
+          assertThat(serviceId).isNotNull();
+          if (expectedServiceId == null) {
+            expectedServiceId = serviceId;
+          } else {
+            assertThat(serviceId).isEqualTo(expectedServiceId);
+          }
```
breedx-splk: nit: this is a little clumsy to me -- mostly because the `expectedServiceId` is declared so far away. I might just factor out the body of the loop into a method that returns the id, and then remove the loop and just call it twice and compare the results. Not a big deal tho.

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/config/PropertiesCustomizer.java
```diff
@@ -45,6 +50,19 @@ public Map<String, String> apply(ConfigProperties config) {
...
+    logger.log(Level.INFO, "metrics export interval (seconds) =  " + exportSeconds);
+
+    connectorBuilder = JmxConnectorBuilder.createNew(scraperConfig.getServiceUrl());
+
+    Optional.ofNullable(scraperConfig.getUsername()).ifPresent(connectorBuilder::withUser);
+    Optional.ofNullable(scraperConfig.getPassword()).ifPresent(connectorBuilder::withPassword);
+
+    if (scraperConfig.isRegistrySsl()) {
+      connectorBuilder.withSslRegistry();
+    }
```
breedx-splk, line 70: This change seems unrelated to the intention of this PR?

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/JmxScraper.java
```diff
@@ -133,6 +145,30 @@ private static boolean testConnection(JmxConnectorBuilder connectorBuilder) {
...
+      StringBuilder id = new StringBuilder();
+      try {
+        ObjectName objectName = new ObjectName("java.lang:type=Runtime");
+        for (String attribute : Arrays.asList("StartTime", "Name")) {
+          Object value = connection.getAttribute(objectName, attribute);
+          if (id.length() > 0) {
+            id.append(" ");
+          }
+          id.append(value);
+        }
```
breedx-splk, line 164: The description in #2206 includes the `Pid` but it's missing here. Is that intentional? Maybe the start time is enough....but I like the idea of including the pid, because time might introduce an edge case.

jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/JmxConnectionTest.java
SylvainJuge: I agree, I've re-written it and now the test is easier to understand.

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/config/PropertiesCustomizer.java
SylvainJuge: No, this is definitely related and a side effect of how things are being initialized: we need to have the `ScraperConfig` instance to configure the `JmxConnectorBuilder`, and we need to have the `JmxConnectorBuilder` instance to set the resource attribute, the simplest was to move the `JmxConnectorBuilder` configuration to the same place where the config is being built. I've slightly reworked the implementation to make this intent more clear.

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/JmxScraper.java
SylvainJuge: Including the `Pid` is the first thing that failed when testing with Java8, so it must have been added in a later JVM version. From what I've seen so far it's implicitly included in the `Name` attribute which is usually in the `1234@hostname` format, where `1234` is the PID.

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/JmxScraper.java
```diff
@@ -64,36 +73,40 @@ public static void main(String[] args) {
...
-          .build();
 
+      // auto-configure SDK
+      OpenTelemetry openTelemetry =
+          AutoConfiguredOpenTelemetrySdk.builder()
+              .addPropertiesSupplier(new PropertiesSupplier(argsConfig))
+              .addPropertiesCustomizer(configCustomizer)
+              // we rely on the config customizer to be executed first to get effective config
+              .addResourceCustomizer(
+                  (resource, configProperties) -> {
```
robsunday: The code would be more readable if this customizer got extracted to some private method

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/JmxScraper.java
```diff
@@ -133,6 +145,30 @@ private static boolean testConnection(JmxConnectorBuilder connectorBuilder) {
     }
   }
 
+  @Nullable
+  static UUID getRemoteServiceInstanceId(JmxConnectorBuilder connectorBuilder) {
```
robsunday: Could be private

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/JmxScraper.java
SylvainJuge: I've moved it to a local variable, which is already helping a bit readability. Let me know if that's not enough.

jmx-scraper/src/main/java/io/opentelemetry/contrib/jmxscraper/config/PropertiesCustomizer.java
breedx-splk: Ah yes, ok, thank you for clarifying. That makes a lot more sense now.

PR #2280, opened 2025-09-19
[opamp-client] Refactor opamp client callbacks api

### LINKED ISSUE #2250: [opamp] Example in README is confusing
...and it doesn't compile.

The callback is passed to the builder, but one of the callback implementation methods references that same client that is being built. 🐔 vs 🥚 circular dependency.

Not sure what to really do about it, but I think it suggests a gap in the design.

breedx-splk:
Resolves #2250.

This is one approach that I thought was a reasonable compromise. The client implementation has a mutable/volatile `Callbacks` instance which is set in the `create()` factory method. Rather than passing a `Callbacks` instantly, the user passes a `Function` that accepts the client instance a returns a `Callbacks` instance.

In practice, I suspect that most users won't be creating anonymous `Callback` instances like in the `README.md` example, and that instead they will pass the client to the constructor of their concrete implementations.

### DIFF
--- a/opamp-client/README.md
+++ b/opamp-client/README.md
@@ -11,7 +11,6 @@ client [spec](https://github.com/open-telemetry/opamp-spec/blob/main/specificati
 
 ```java
 // Initializing it
-
 RequestService requestService = HttpRequestService.create(OkHttpSender.create("[OPAMP_SERVICE_URL]"));
 // RequestService requestService = WebSocketRequestService.create(OkHttpWebSocket.create("[OPAMP_SERVICE_URL]")); // Use this instead to connect to the server via WebSocket.
 OpampClient client =
@@ -19,8 +18,7 @@ OpampClient client =
         .putIdentifyingAttribute("service.name", "My service name")
         .enableRemoteConfig()
         .setRequestService(requestService)
-        .build(
-            new OpampClient.Callbacks() {
+        .build(cx -> new OpampClient.Callbacks() {
               @Override
               public void onConnect() {}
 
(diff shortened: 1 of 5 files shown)


LikeTheSalad:
Thank you for the detailed response and for understanding, @breedx-splk.

> If we want to go with the other alternatives suggested in https://github.com/open-telemetry/opentelemetry-java-contrib/issues/2250, I think I would choose to add the client as a parameter in the callback methods...but it sounded like you didn't really like that idea. Maybe I'm misunderstanding...but that solves the problem pretty cleanly as well without making the API unwieldy.

It's not my favorite one, although I can't find an easy way to mess things up with it if I were a user. That's usually how I pick between API approaches, by trying to think of ways to easily misuse them or to get into roadblocks that would require "creative solutions" on behalf of the users. Though I can always miss things, as it happened with the existing API when I was asked to remove its start/stop methods. It's nicer to have just a `close()` method from our perspective tbh, as we don't need to validate things such as avoiding `start()` to be called more than once, and also it isn't needed to have a nullable field for the callback, but I didn't think of what it could mean from a practical point of view when needed to interact with the client from within the callback, which should be quite common. That's why I added an open question to the "client in the callback methods" option on my comment, because I can't find something particularly bad with it, but was wondering if someone else was aware of something. 

So far, I've only gotten feedback on it from @trask, which may or may not have been due to some potential issues that he might be aware of, so I'll loop him in just in case. Though if there's no issue, I agree that the "client in the callback methods" is an option that doesn't break too much of the existing API and fully fixes the problem, so I'd be more inclined to go with that approach instead of the one from this PR.

PR #2309, opened 2025-09-26
Escape back-ticks

trask:


breedx-splk (approved):
👿 Evil backticks are evil.

PR #2318, opened 2025-09-30
Use mise to check for modified files

zeitlinger:
- make it easy to run locally
- makes it also easy to sync the task file, if we decide to do it (see https://github.com/open-telemetry/community/issues/3024)


trask:
looks like link check is failing

trask (approved):

PR #2324, opened 2025-10-01
Modernize JUnit Gradle configuration

trask:


ibm-mq-metrics/build.gradle.kts
```diff
@@ -58,23 +37,29 @@ dependencies {
...
-  classpath = sourceSets["integrationTest"].runtimeClasspath
-  shouldRunAfter("test")
-
-  useJUnitPlatform()
-
-  testLogging {
-    events("passed")
+tasks.shadowJar {
+  dependencies {
+    exclude(dependency("com.ibm.mq:com.ibm.mq.allclient:9.4.3.1"))
```
laurit: is the version number needed here?

PR #2325, opened 2025-10-01
Modernize/simplify JUnit Gradle configuration

trask:


buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
```diff
@@ -147,10 +147,11 @@ dependencies {
 
 testing {
   suites.withType(JvmTestSuite::class).configureEach {
+    useJUnitJupiter("5.13.4")
```
trask, line 150: moving back to draft, doesn't look like Renovate will update this

buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
```diff
@@ -147,10 +147,15 @@ dependencies {
 
 testing {
   suites.withType(JvmTestSuite::class).configureEach {
+    useJUnitJupiter()
```
laurit: what if you replaced
```
withType<Test>().configureEach {
  useJUnitPlatform()
```
with `useJUnitPlatform`? Doing it only for suites and leaving the rest of the test tasks as they are feels a bit weird.

trask: sorry, I'm not following what you are suggesting to do

trask: oh, do you mean replacing that with `useJUnitJupiter`? I just tried that and it seems like `useJUnitJupiter` only works for test suites...

buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
trask, line 150: intentionally one version behind to verify that Renovate picks it up after merging

buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
trask: bump

PR #2333, opened 2025-10-06
fix link check

zeitlinger:


trask (approved):
checked, works on Windows

PR #2342, opened 2025-10-10
Convert gcp-auth-extension to use test suites

trask:
Related to https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2325

### DIFF
--- a/gcp-auth-extension/build.gradle.kts
+++ b/gcp-auth-extension/build.gradle.kts
@@ -75,7 +75,6 @@ testing {
       targets.all {
         testTask.configure {
           dependsOn(tasks.shadowJar)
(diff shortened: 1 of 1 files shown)


gcp-auth-extension/build.gradle.kts
```diff
@@ -101,35 +144,3 @@ tasks.register<Copy>("copyAgent") {
     rename("opentelemetry-javaagent(.*).jar", "otel-agent.jar")
   })
 }
-
-tasks.register<Test>("IntegrationTestUserCreds") {
```
trask, line 105: noticed that this test wasn't running in the build, but fixing in different PR: #2344

PR #2354, opened 2025-10-13
Return Profiler old interval from setInterval

jackshirazi:
**Description:**

setInterval() was void, but it's useful when resetting the interval to know what the previous interval was. We could add a new getInterval, but it seems more efficient to just return it from the setInterval()

**Existing Issue(s):**

-

**Testing:**

-

**Documentation:**

-

**Outstanding items:**

Note this technically breaks the existing InferredSpans class API, but shouldn't impact anything because existing calls to setInterval() are not assigning a return value, so will continue to work fine

**Alternative Considered**

We could instead add a getInterval() call


inferred-spans/src/main/java/io/opentelemetry/contrib/inferredspans/InferredSpans.java
```diff
@@ -42,10 +42,12 @@ public static boolean isEnabled() {
    *
    * @param interval the new profiler interval
    */
-  public static void setProfilerInterval(Duration interval) {
+  @Nullable
+  public static Duration setProfilerInterval(Duration interval) {
     InferredSpansProcessor p = instance;
     if (p != null) {
```
breedx-splk, line 48: nit: I prefer this flipped, but not a huge deal.

breedx-splk (approved):
LGTM

Issue #2072, opened 2025-08-02
Create weaver model for jmx-scraper metrics


atoulme:
Please help create a Weaver model for the different target systems of the metrics defined by jmx-scraper, so we can work to standardize them. This issue should be decomposed into an issue by target system. A simple litmus test on a simple target model should be enough to demonstrate first how hard this is to do, and we can go from there.

The idea is that eventually we can move some of the weaver models to semconv, as metric references.

atoulme:
Here is an example of use that introduces a Weaver model to back the existing yaml files rather than directly replacing them: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/40698

We build this in a backwards-compatible, non-breaking way by making sure we can express everything we could in weaver before we do anything else.

SylvainJuge:
I've opened https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/19139 PR to start adding a weaver registry for JMX metrics.

SylvainJuge:
This will be covered in instrumentation with steps described in https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/19227

With the new "delegated registry" model, maybe those metrics will be defined elsewhere in the future, but for now they are being described (in weaver yaml) and defined (in jmx yaml) in instrumentation.

So I'll close this issue for now, unless someone really needs to have "semantic convention definitions" for the "legacy" metrics in JMX Scraper (the ones that are close to the now deprecated jmx-gatherer).

PR #2362, opened 2025-10-17
[jmx-scraper] Create weaver model for jmx-scraper metrics #2072

mcmho:
**Description:**
Feature / Improvement: use weaver model to describe metrics so that they can be standardized.
 - created metrics.yaml weaver model for jvm metrics
 - created attriutes.yaml
 - tested with weaver registry generate
 
**Testing:**
 - tested with weaver registry generate;  metadata.yaml is successfully generated and content is valided

**Documentation:**
 - customer facing docs not yet needed 
  
**Outstanding items:**
 - once this looks good, there will be other models to work on for other target systems.

### DIFF
--- a/jmx-scraper/templates/registry/rules/jvm.yaml.j2
+++ b/jmx-scraper/templates/registry/rules/jvm.yaml.j2
@@ -98,19 +98,19 @@ rules:
       {%- if m.id == jvm_memory_prefix + "pool.committed" %}
       Usage.committed:
         metric: {{ m.id | safe }}
-        desc: current memory pool usage
+        desc: {{ m.brief | safe }}
       {%- elif m.id == jvm_memory_prefix + "pool.init" %}
       Usage.init:
         metric: {{ m.id | safe }}
-        desc: current memory pool usage
+        desc: {{ m.brief | safe }}
       {%- elif m.id == jvm_memory_prefix + "pool.max" %}
       Usage.max:
         metric: {{ m.id | safe }}
-        desc: current memory pool usage
+        desc: {{ m.brief | safe }}
       {%- elif m.id == jvm_memory_prefix + "pool.used" %}
       Usage.used:
         metric: {{ m.id | safe }}
-        desc: current memory pool usage
+        desc: {{ m.brief | safe }}
       {%- endif %}
       {%- endfor %}
 
@@ -121,4 +121,4 @@ rules:
         {%- set m_threads_count = (ctx.groups | selectattr("id", "equalto", "jvm.threads.count") | first) %}
         metric: {{ m_threads_count.id | safe}}
         unit: "{{ m_threads_count.unit | safe }}"
-        desc: number of threads
+        desc: {{ m_threads_count.brief | safe }}
--- a/jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/JvmIntegrationTest.java
+++ b/jmx-scraper/src/integrationTest/java/io/opentelemetry/contrib/jmxscraper/target_systems/JvmIntegrationTest.java
@@ -186,7 +186,7 @@ class JvmIntegrationTest extends TargetSystemIntegrationTest {
             "jvm.threads.count",
             metric ->
                 metric
-                    .hasDescription("number of threads")
+                    .hasDescription("current number of thread count")
                     .hasUnit("{thread}")
                     .isGauge()
                     .hasDataPointsWithoutAttributes());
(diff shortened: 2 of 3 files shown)


mcmho:
Thanks all.  sorry for the delay!   The tests are fixed now. 

Re: @SylvainJuge  
yes, the rule configuration will be generated from jinja2 template and registry.

Perhaps @atoulme  can chime in since he created the original request.   As far as I understand, one of the main benefits coming out of this is standardization of metrics.   Ultimately, we can have a centralized registry of metrics.   That would become the single source of truth / semconv.     I think that's what @atoulme  has in mind "The idea is that eventually we can move some of the weaver models to semconv, as metric references."

atoulme:
> With that, modifying the JMX metrics yaml would require expertise in jinja and in running weaver with a makefile, which is not how things are usually done in this java-only project.

Please upvote and consider participating in https://github.com/open-telemetry/weaver/issues/792

I'm sure we can work out a gradle task to run a docker container, might take a bit to get the kinks out.

The choice of jinja2 as a templating language seems like the least of all choices at this point, but please feel free to offer alternatives in the weaver repository?

> What I think would be simpler here is to keep the JMX metrics yaml as source of truth and then generate the weaver yaml files from it as part of the build process.

It can work, but the weaver model is richer than the JMX metrics yaml though.

jmx-scraper/templates/registry/rules/jvm.yaml.j2
```diff
@@ -0,0 +1,124 @@
...
+      {%- endfor %}
+
+  # Rule for Heap and Non-Heap Memory metrics
+  - bean: java.lang:type=Memory
+    unit: By
+    prefix: {{ jvm_memory_prefix | safe }}
+    mapping:
+      {%- for m in ctx.groups if m.id.startswith(jvm_memory_prefix + "heap") %}
+      {%- if m.id == jvm_memory_prefix + "heap.committed" %}
+      HeapMemoryUsage.committed:
```
jsuereth, line 45: You could put an annotation on the metric and use it here if you wanted.

So your metrics.yaml would be something like:

```
groups:
  - id: jvm.memory.head.committed
    type: metric
    ...
    annotations:
    - jmx_name: HeapMemoryUsage.committed
```

Then here you could have the mapping generically pull the annotation:
```
{{ m.annotatiions.jmx_name }}:
  metric: {{ m.id | ... }
  ...
```

jsuereth: You could even go further and record the Bean / property and group-by bean, etc.

SylvainJuge:
I think the weaver yaml should be used as a "contract" to describe and document what signals are captured by a specific compolent, this allows to automate documentation and maybe generate other artifacts derived from it.

In a sense, this is quite similar to the work around metadata in instrumentation, what is needed is the _ability_ to provide an up-to-date weaver model to allow consumers to diff/migrate using weaver tool.

My intent here would be to provide the following:
- create a reusable tool included in `instrumentation/jmx-metrics` that allows to generate weaver yaml from any JMX metrics yaml definition https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/15196
- use this tool to generate metrics definitions for current and past versions of JMX metrics in this repository, those would be checked-in git repo (to be later consumable by weaver)
- integrate this tool within build to automate the process to update generated weaver yaml
- do the same in contrib/jmx-scraper for the "legacy" metrics definitions, this will also include the metrics definitions inherited from instrumentation.

In addition, the users of JMX metrics are also likely to provide their own metrics definitions, so having the ability to generate the weaver yaml for them would also provide the ability for end-users to also generate documentation, versioning and migration of their custom data.

I don't think using weaver to generate the JMX metrics yaml provides anything beyond generating one yaml from another, it just makes the maintenance of those JMX metrics yaml extra complicated due to the templating, also it does not allow to provide to end-users the ability to generate their own weaver yaml. In a sense, it would be trying to re-implement the JMX metrics yaml into the weaver yaml (we could even push it further by making JMX metrics reuse weaver yaml format, but that would be a breaking change and likely an extra challenge to document and maintain and help users migrating.

However, I'm not against using the weaver itself, for example I agree that it would make sense to use it to generate documentation from the weaver yaml, this is currently done manually.

I plan to experiment a bit this week trying to implement https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/15196 as I think this is a simpler approach.

trask:
> We need to standardize to a model, make our documentation easier, and publish them out there.

Can't this be done without changing the primary configuration format (which aligns with declarative configuration)?

atoulme:
> > We need to standardize to a model, make our documentation easier, and publish them out there.
> 
> Can't this be done without changing the primary configuration format (which aligns with declarative configuration)?

In the case of the collector, we just put the weaver model upstream from metadata.yaml, and used it to generate it. That's the approach offered here too. You don't need to change anything about the model you use right now, and keeping those aligned helps make sure weaver is aligned and we can entirely generate the breadth of the existing model using it. So far, so good.

PR #2383, opened 2025-10-22
Update sampler docs

jaydeluca:
A question came up in [slack](https://cloud-native.slack.com/archives/C014L2KCTE3/p1761104868320289?thread_ts=1761004110.264759&cid=C014L2KCTE3) around whether this should be added as a dependency to the application classpath or if it's an agent extension, so I wanted to make this more clear

### DIFF
--- a/cel-sampler/README.md
+++ b/cel-sampler/README.md
@@ -7,6 +7,7 @@ The `CelBasedSampler` supports [declarative configuration](https://opentelemetry
 To use:
 
 * Add a dependency on `io.opentelemetry.contrib:opentelemetry-cel-sampler:<version>`
+  * See the [extension documentation](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/examples/extension/README.md#build-and-add-extensions) for how to add extensions when using the java agent.
 * Follow the [instructions](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk-extensions/incubator/README.md#declarative-configuration) to configure OpenTelemetry with declarative configuration.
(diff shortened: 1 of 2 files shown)


cel-sampler/README.md
```diff
@@ -7,6 +7,7 @@ The `CelBasedSampler` supports [declarative configuration](https://opentelemetry
 To use:
 
 * Add a dependency on `io.opentelemetry.contrib:opentelemetry-cel-sampler:<version>`
+  * See the [extension documentation](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/examples/extension/README.md#build-and-add-extensions) for how to add extensions when using the java agent.
```
laurit, line 10: Since this module has an external dependency it can't just be added as an extension. User would first need to merge everything into single jar or also pass the `cel` jar as an extension (haven't tried that but since all the extensions are in the same class loader passing the dependencies as extensions should work). We could make this simpler by publishing shaded jars like in https://github.com/open-telemetry/opentelemetry-java-contrib/blob/5809da024bf450aa388d37a7ca2c398c2ca2cac0/gcp-auth-extension/build.gradle.kts#L114 See also https://github.com/open-telemetry/opentelemetry-java-contrib/blob/5809da024bf450aa388d37a7ca2c398c2ca2cac0/gcp-auth-extension/README.md#with-opentelemetry-java-agent

dol: @laurit Good point. I will have a look into this. Definitly an improvement.

jaydeluca: I created a pr to package the jar: https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2516

PR #2386, opened 2025-10-25
chore: Added more project links to the pom.xml

dol:
**Description:**

According to https://maven.apache.org/pom.html#SCM the `scm.url` should be a publicly browsable repository.
Replaced git URI with URL to project page on Github.

Added `issueManagement` and `ciManagement` information to the scm section.

Took inspiration from https://central.sonatype.com/artifact/org.apache.maven/maven/4.0.0-rc-4 and their [pom.xml](https://repo1.maven.org/maven2/org/apache/maven/maven/4.0.0-rc-4/maven-4.0.0-rc-4.pom).

image

The link on https://central.sonatype.com/artifact/io.opentelemetry.contrib/opentelemetry-cel-sampler/1.51.0-alpha ends up in a 404.

image

### Before the change

image

### After the change

The new pom.xml would look like this. Generated with `./gradlew clean publishToMavenLocal`.
The `scm.url` points to the release version of the source code.

image

### DIFF
--- a/buildSrc/src/main/kotlin/otel.publish-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/otel.publish-conventions.gradle.kts
@@ -3,6 +3,8 @@ plugins {
   signing
 }
 
+val tagVersion: String by rootProject.extra
+
 publishing {
   publications {
     register<MavenPublication>("maven") {
@@ -55,7 +57,18 @@ publishing {
         scm {
           connection.set("scm:git:git@github.com:open-telemetry/opentelemetry-java-contrib.git")
           developerConnection.set("scm:git:git@github.com:open-telemetry/opentelemetry-java-contrib.git")
-          url.set("git@github.com:open-telemetry/opentelemetry-java-contrib.git")
+          tag.set(tagVersion)
+          url.set("https://github.com/open-telemetry/opentelemetry-java-contrib/tree/${tagVersion}")
+        }
+
+        issueManagement {
+          system.set("GitHub Issues")
+          url.set("https://github.com/open-telemetry/opentelemetry-java-contrib/issues")
+        }
+
+        ciManagement {
+          system.set("GitHub Actions")
+          url.set("https://github.com/open-telemetry/opentelemetry-java-contrib/actions")
         }
 
         withXml {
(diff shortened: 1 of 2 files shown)


trask:
makes sense to me 👍

> The new pom.xml would look like this. Generated with `./gradlew clean publishToMavenLocal`. The `scm.url` points to the release version of the source code.
> 
> image

this repos's tags don't have `-alpha` and start with `v`, e.g. https://github.com/open-telemetry/opentelemetry-java-contrib/tree/v1.51.0

trask (approved):
@open-telemetry/java-approvers @open-telemetry/java-instrumentation-approvers if this looks good, I'll port to the other Java repos after merging

buildSrc/src/main/kotlin/otel.publish-conventions.gradle.kts
```diff
@@ -55,7 +57,18 @@ publishing {
         scm {
           connection.set("scm:git:git@github.com:open-telemetry/opentelemetry-java-contrib.git")
           developerConnection.set("scm:git:git@github.com:open-telemetry/opentelemetry-java-contrib.git")
-          url.set("git@github.com:open-telemetry/opentelemetry-java-contrib.git")
+          tag.set(tagVersion)
+          url.set("https://github.com/open-telemetry/opentelemetry-java-contrib/tree/${tagVersion}")
```
laurit, line 61: these probably aren't correct for snapshot builds, does that matter?

breedx-splk (approved):
Thanks. Having a working link on central would be nice indeed.

To @laurit 's question about snapshot builds -- I think it's fine if it's not accurate for snapshots in the short term. The snapshots aren't even browsable, and the chances of someone using a url from a downloaded `pom.xml` I think are minimal. If browsing ever comes back for snapshots, we can circle back then.

version.gradle.kts
```diff
@@ -1,5 +1,6 @@
 val stableVersion = "1.52.0-SNAPSHOT"
 val alphaVersion = "1.52.0-alpha-SNAPSHOT"
+val tagVersion by extra { "v$stableVersion" }
```
jack-berg, line 3: Why not make this a local variable if its only used in `otel.publish-conventions.gradle.kts`?

dol: I did not want to change the behavior of the release process to update the version number.
https://github.com/open-telemetry/opentelemetry-java-contrib/blob/077cd6ee15a3c5f5938484c90745f80eac66ce95/.github/scripts/update-version.sh#L11 .

Since the `tagVersion` is a reference to the `stableVersion` no other change is require. But if the other reviewers agree, that the tag reference should only be in the `otel.publish-conventions.gradle.kts` I'm open to also update the `update-version.sh` file.

PR #2408, opened 2025-10-31
Simplify OpAmp client api

laurit:
Simplify OpAmp client api by allowing user to specify an endpoint. We'll choose whether to use http or websocket based on the endpoint scheme. This way users don't need to create the `RequestService` and `OkHttpSender` or `OkHttpWebSocket` themselves. This PR also moves `RequestService` back to `internal` package as most users won't be needing it. For those who need to provide a custom `RequestService` I have added an experimental api that lets users set a factory `Function<URI, RequestService>` for creating `RequestService` instances.


opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/Experimental.java
```diff
@@ -0,0 +1,42 @@
...
+public final class Experimental {
+
+  @Nullable
+  private static volatile BiConsumer<OpampClientBuilder, Function<URI, RequestService>>
+      setServiceFactory;
+
+  /**
+   * Sets factory function for creating {@link RequestService} instances form a given server URI.
+   */
+  public static void setServiceFactory(
```
LikeTheSalad, line 29: > This PR also moves RequestService back to internal package as most users won't be needing it.

I'm not too worried about the internal package, as this lib is not stable anyway, though I currently need to pass a [custom service](https://github.com/elastic/apm-agent-android/blob/83e511e9762fb307e1f07ab5bc16a6ef22e78c37/agent-sdk/src/main/java/co/elastic/otel/android/internal/features/centralconfig/CentralConfigurationSource.kt#L95) with a customized okhttp client instance to intercept all the requests, so having to rely on statics for this factory and within an experimental scope seems a bit unnecessarily inconvenient to me, given the current status of this lib.

What if this setter is moved into the builder and it's called from within the `setEndpoint(URI)` function instead? I agree there's value in having convenient functions such as `setEndpoint(String/URI)` so I was thinking if it could be done in a way that presents different options in the same place (the builder) and people would choose the one that better fits their needs by looking at the options from an ide autocomplete list.

laurit: This is the pattern we are using in the instrumentation repository to introduce experimental functionality. Vast majority of users shouldn't really need the `RequestService` abstraction. For your use case it doesn't really change much. You'd just change
```java
requestService = HttpRequestService.create(OkHttpSender.create(connectivity.getUrl(), okhttpClient))
builder.setRequestService(requestService)
```
to
```java
Experimental.setServiceFactory(builder, endpoint -> HttpRequestService.create(OkHttpSender.create(endpoint, okhttpClient)));
```
I'd say it isn't too bad usability wise. Explicitly having to go through a class named `Experimental` shouldn't be a problem for those who are fine with using classes from a package named `internal`.

LikeTheSalad: Got it, thanks for the example. I agree it's not too bad, I just have the following questions:

- Why bothering with an experimental/hidden API when the whole lib is essentially experimental? Hiding RequestService away might just discourage people from using it, and I guess ideally people should use the apis to see if there're no issues with them. Though I guess there's also the counter argument that, if it's hidden, then we'll know if it's actually needed whenever people ask for it to be included in the "main api", is that the reason for hiding it?
- I'm not sure about the vast majority not needing it. Without it, the only configuration for the OpAMP server connection they can provide is the URL, but what if they need other often needed features, such as payload compression, tls config, custom periodic request delay for when http is used?

laurit: > Why bothering with an experimental/hidden API when the whole lib is essentially experimental?

because eventually we'd like it to be not experimental, got to start from somewhere

> I'm not sure about the vast majority not needing it. Without it, the only configuration for the OpAMP server connection they can provide is the URL, but what if they need other often needed features, such as payload compression, tls config, custom periodic request delay for when http is used?

we could expose these in the `OpampClientBuilder` and change the factory to `BiFunction<URI, Options, RequestService>`

LikeTheSalad: > because eventually we'd like it to be not experimental, got to start from somewhere

This is the confusing part for me, I think we both want the same, but from my point of view, hiding the API away will actually slow this process down.

> we could expose these in the OpampClientBuilder and change the factory to BiFunction<URI, Options, RequestService

Sounds good. Still, I'd suggest adding the factory method right away. If anything, having to provide a preconfigured object to do the network connectivity shouldn't be weird for OTel users, as that's essentially what we do with the `[Signal]Exporter`s already. Also, the `Options` object would probably have to be an interface too, as there's not a full overlap across the options we can pass when using HTTP vs WebSocket.

opamp-client/src/main/java/io/opentelemetry/opamp/client/OpampClientBuilder.java
```diff
@@ -30,27 +34,57 @@
...
 public final class OpampClientBuilder {
+  private static final String DEFAULT_ENDPOINT_URL = "http://localhost:4320/v1/opamp";
+  private static final URI DEFAULT_ENDPOINT = URI.create(DEFAULT_ENDPOINT_URL);
+
   private final Map<String, AnyValue> identifyingAttributes = new HashMap<>();
   private final Map<String, AnyValue> nonIdentifyingAttributes = new HashMap<>();
   private long capabilities = 0;
-  private RequestService service =
-      HttpRequestService.create(OkHttpSender.create("http://localhost:4320/v1/opamp"));
+  private Function<URI, RequestService> serviceFactory =
```
breedx-splk, line 43: It seems a little unconventional in Java to use a function for this kind of thing. I'd prefer that we just stick with using a method...which I think can even be static.

laurit: Currently we allow users replace this factory function, so they could customize things. Ideally eventually this will go away when we can provide users an option to customize the okhttp client without having to mess with the RequestService or HttpSender implementations.

LikeTheSalad:
> Now you'd probably say that you'd prefer it to be non-experimental, but I think in this case it is easier to move in small steps.

Can you elaborate on what about this case makes it easier to move in small steps? I'd argue the opposite, due to the current state of the project (incubating), how long it has been available for use (not so long that a big change would affect many users), and how quickly we'd like to receive feedback from users (who tend to avoid APIs marked as "experimental" or hidden away).

> not really, could just throw UnsupportedOperationException

Making users go with this approach, to me, is often times an indication of a poorly designed API. So I'd rather avoid going with it unless there's no other option.

> I don't really get why setEndpoint would be problematic. Setting the endpoint seems to me like the most common customization users would like to make.

It is not problematic when no other config option is needed. But I'm certain that an endpoint is not the only option needed for real world use cases, so introducing a `setEndpoint` setter separate from another that's more complete (e.g. `setHttpSender/setConnectionFactory/setRequestService`), will cause it to become redundant and even confusing, such as in the case of the `setServiceFactory` which receives the URL provided by `setEndpoint` to be used in the RequestService instance, but that can be easily ignored, defeating the purpose of a setEndpoint/factory combination.

I'd also argue that I don't see why making a call like `setHttpSender(OkHttpSender.create(endpoint))` would cause any issues or be problematic in any way, while keeping the config simple (as the same setter can be used to add more complex configs) and keeps the API surface leaner.

> It depends on whether you wish to have an abstraction for passing http client options or whether you let users provide a preconfigured http client.

I prefer the latter. And I think `HttpSender` and `Websocket` are meant for that job.

> I didn't completely remove the setServiceFactory api precisely because it is currently needed for real world use cases.

I agree. That's why I mentioned later that a good way of avoiding to expose the `RequestService` interface, as it has been the main point of debate for this PR, is to go with the 2 builders option that you mentioned.

> given that there is currently 1 option that is not shared between these I feel like separate builders would be an overkill
> assuming that we'd let user provide a preconfigured http client the TransportFactory api could look like

There are 2 options, the periodic polling and the network layer interface (HttpSender and Websocket). This is a pretty new lib and HTTP and Websocket are pretty different, so I'd argue that keeping them properly separate will make them flexible for any future changes. Plus, I don't see why it would be problematic in terms of api semantics to call one builder vs another. It sounds like a better option than having to place another interface layer on top, such as `ConnectionFactory`, only to having to create an instance of it first where users would have to provide either an `HttpSender` or a `WebSocket` implementation via a specific implementation such as `OkHttpConnection/TransportFactory`, which to be frank just sounds like doing the same it's done right now with `RequestService` implementations.

> It could extend from a base class would make supporting alternative http clients easy.

A specific client implementation would have to duplicate a lot of logic that's opamp-specific, so having to create one just to change the underlying http client doesn't make sense to me when there's an abstraction that only focuses on the http connectivity (HttpSender).

> Currently the OpampClientBuilder is tied to okhttp because it creates the OkHttpSender in constructor. After this PR this would be resolved.

I don't understand this statement. The OpampClientBuilder constructor doesn't create an OkHttpSender, plus, `OkHttpSender` is an implementation of `HttpSender` which is what's required to be provided by users, so they can provide any other implementation.

> I think the experimental apis would let us move faster since there should be less friction in getting them merged. They also serve as a good indication of areas that need further work.

I think this is related to the first comment that I addressed at the top.

PR #2413, opened 2025-11-03
Gradle plugin for generating weaver files

jaydeluca:
I took a stab at creating a gradle plugin to replace the makefiles for the weaver generation tasks, inspired by the [semantic-conventions-java](https://github.com/open-telemetry/semantic-conventions-java/blob/8df5e9f2050ea4cbe7295539314653b1761b6c89/build.gradle.kts#L115) approach

I imagine this will need to evolve as we figure out the needs of running it for multiple modules (like #2362), aside from just the ibm-mq-metrics, which this was mostly focused on solving for.

I'm still not sure if this is the best approach, there are some others that are still being discussed in https://github.com/open-telemetry/weaver/issues/792 but wanted to see what this might look like

cc @atoulme

### DIFF
--- a/buildSrc/src/main/kotlin/otel.weaver-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/otel.weaver-conventions.gradle.kts
@@ -17,6 +17,16 @@ val dockerPlatform = System.getProperty("os.arch").let { arch ->
   }
 }
 
+// Docker executable path: use env var DOCKER_EXECUTABLE or usual platform defaults
+val dockerExecutablePath = System.getenv("DOCKER_EXECUTABLE") ?: run {
+  val os = System.getProperty("os.name").lowercase()
+  when {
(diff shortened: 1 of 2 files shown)


atoulme (approved):
@mcmho please review as well

trask:
just double checking y'all ready to have this merged and I'll hit the button

jaydeluca:
> just double checking y'all ready to have this merged and I'll hit the button

It's been a while, but i just went back and re-tested everything and its good to go from my perspective. 

I just added one more small change to allow invoking the weaverGenerate jobs from anywhere (so you can do `./gradlew :ibm-mq-metrics:generateWeaver` from root or `../gradlew :ibm-mq-metrics:generateWeaver` from within the ibm-mq-metrics directory

@atoulme @mcmho is this good to go from your perspective?

mcmho (approved):
I think this is a job well done!   Thanks for the gradle plugin.

PR #2447, opened 2025-11-20
Remove auto spotless workflow

trask:
Even though I think it's been designed to be secure, security scanners really don't like it, and I don't want to fight with them...


breedx-splk (approved):
🪦

PR #2469, opened 2025-11-24
Fix flaky test

### LINKED ISSUE #2458: Flaky tests in inferred-spans
During the last contrib release, there were a few flaky tests that cause some problems.

https://develocity.opentelemetry.io/s/kzarodjoam5t6/tests/task/:inferred-spans:test/details/io.opentelemetry.contrib.inferredspans.InferredSpansTest/testSetProfilerInterval()?top-execution=1

Image

Same basic thing with the [SamplingProfilerTest](https://develocity.opentelemetry.io/s/kzarodjoam5t6/tests/task/:inferred-spans:test/details/io.opentelemetry.contrib.inferredspans.internal.SamplingProfilerTest).

Seems like maybe the profiler isn't being reset correctly between tests or there is some other unexpected test interaction.

jackshirazi:
closes #2458

### DIFF
--- a/inferred-spans/src/main/java/io/opentelemetry/contrib/inferredspans/InferredSpansProcessorBuilder.java
+++ b/inferred-spans/src/main/java/io/opentelemetry/contrib/inferredspans/InferredSpansProcessorBuilder.java
@@ -195,13 +195,13 @@ public class InferredSpansProcessorBuilder {
   }
 
   /** For testing only. */
-  public InferredSpansProcessorBuilder activationEventsFile(@Nullable File activationEventsFile) {
+  InferredSpansProcessorBuilder activationEventsFile(@Nullable File activationEventsFile) {
     this.activationEventsFile = activationEventsFile;
     return this;
   }
 
   /** For testing only. */
-  public InferredSpansProcessorBuilder jfrFile(@Nullable File jfrFile) {
+  InferredSpansProcessorBuilder jfrFile(@Nullable File jfrFile) {
     this.jfrFile = jfrFile;
     return this;
   }
--- a/inferred-spans/src/main/java/io/opentelemetry/contrib/inferredspans/InferredSpansProcessorBuilder.java
+++ b/inferred-spans/src/main/java/io/opentelemetry/contrib/inferredspans/InferredSpansProcessorBuilder.java
@@ -52,6 +52,7 @@ public class InferredSpansProcessorBuilder {
   private boolean startScheduledProfiling = true;
   @Nullable private File activationEventsFile = null;
   @Nullable private File jfrFile = null;
+  @Nullable private File tempDir = null;
   private BiConsumer<SpanBuilder, SpanContext> parentOverrideHandler =
       CallTree.DEFAULT_PARENT_OVERRIDE;
 
@@ -75,7 +76,7 @@ public class InferredSpansProcessorBuilder {
             parentOverrideHandler);
     InferredSpansProcessor processor =
         new InferredSpansProcessor(
-            config, clock, startScheduledProfiling, activationEventsFile, jfrFile);
+            config, clock, startScheduledProfiling, activationEventsFile, jfrFile, tempDir);
     InferredSpans.setInstance(processor);
     return processor;
   }
@@ -194,14 +195,20 @@ public class InferredSpansProcessorBuilder {
   }
 
   /** For testing only. */
-  InferredSpansProcessorBuilder activationEventsFile(@Nullable File activationEventsFile) {
+  public InferredSpansProcessorBuilder activationEventsFile(@Nullable File activationEventsFile) {
     this.activationEventsFile = activationEventsFile;
     return this;
   }
 
   /** For testing only. */
-  InferredSpansProcessorBuilder jfrFile(@Nullable File jfrFile) {
+  public InferredSpansProcessorBuilder jfrFile(@Nullable File jfrFile) {
     this.jfrFile = jfrFile;
+    return this;
+  }
+
+  /** For testing only. */
+  public InferredSpansProcessorBuilder tempDir(@Nullable File tempDir) {
+    this.tempDir = tempDir;
     return this;
   }
 
--- a/inferred-spans/src/main/java/io/opentelemetry/contrib/inferredspans/internal/SamplingProfiler.java
+++ b/inferred-spans/src/main/java/io/opentelemetry/contrib/inferredspans/internal/SamplingProfiler.java
@@ -532,6 +532,9 @@ public class SamplingProfiler implements Runnable {
   }
 
   public void processTraces() throws IOException {
+    if (!config.isPostProcessingEnabled()) {
+      return;
+    }
     if (jfrParser == null) {
       jfrParser = new JfrParser();
     }
--- a/inferred-spans/src/main/java/io/opentelemetry/contrib/inferredspans/internal/SamplingProfiler.java
+++ b/inferred-spans/src/main/java/io/opentelemetry/contrib/inferredspans/internal/SamplingProfiler.java
@@ -764,10 +764,10 @@ public class SamplingProfiler implements Runnable {
     }
   }
 
+  @SuppressWarnings({"FutureReturnValueIgnored", "Interruption"})
   public void stop() throws InterruptedException, IOException {
     // cancels/interrupts the profiling thread
     if (profilingTask != null) {
-      @SuppressWarnings("Interruption")
       profilingTask.cancel(true);
     }
     // implicitly clears profiled threads
--- a/inferred-spans/src/main/java/io/opentelemetry/contrib/inferredspans/internal/SamplingProfiler.java
+++ b/inferred-spans/src/main/java/io/opentelemetry/contrib/inferredspans/internal/SamplingProfiler.java
@@ -398,7 +398,7 @@ public class SamplingProfiler implements Runnable {
 
     profilerLock.lock();
     try {
-      if (!isInterrupted && !scheduler.isShutdown()) {
+      if (!Thread.currentThread().isInterrupted() && !scheduler.isShutdown()) {
         long delay = config.getProfilingInterval().toMillis() - profilingDuration.toMillis();
         profilingTask = scheduler.schedule(this, delay, TimeUnit.MILLISECONDS);
       }
--- a/inferred-spans/src/main/java/io/opentelemetry/contrib/inferredspans/internal/SamplingProfiler.java
+++ b/inferred-spans/src/main/java/io/opentelemetry/contrib/inferredspans/internal/SamplingProfiler.java
@@ -766,6 +766,9 @@ public class SamplingProfiler implements Runnable {
 
   public void stop() throws InterruptedException, IOException {
     // cancels/interrupts the profiling thread
+    if (profilingTask != null) {
+      profilingTask.cancel(true);
+    }
     // implicitly clears profiled threads
     scheduler.shutdown();
     scheduler.awaitTermination(10, TimeUnit.SECONDS);
--- a/inferred-spans/src/main/java/io/opentelemetry/contrib/inferredspans/internal/SamplingProfiler.java
+++ b/inferred-spans/src/main/java/io/opentelemetry/contrib/inferredspans/internal/SamplingProfiler.java
@@ -101,7 +101,7 @@ import org.agrona.collections.Long2ObjectHashMap;
  * stack trace}.
  */
 public class SamplingProfiler implements Runnable {
-  // fake comment to bump for more test runs
+
   private static final String LIB_DIR_PROPERTY_NAME = "one.profiler.extractPath";
 
   private static final Logger logger = Logger.getLogger(SamplingProfiler.class.getName());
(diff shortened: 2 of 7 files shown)


inferred-spans/src/main/java/io/opentelemetry/contrib/inferredspans/internal/SamplingProfiler.java
```diff
@@ -394,9 +396,14 @@ public void run() {
...
-      long delay = config.getProfilingInterval().toMillis() - profilingDuration.toMillis();
-      profilingTask = scheduler.schedule(this, delay, TimeUnit.MILLISECONDS);
+    profilerLock.lock();
+    try {
+      if (!isInterrupted && !scheduler.isShutdown()) {
+        long delay = config.getProfilingInterval().toMillis() - profilingDuration.toMillis();
+        profilingTask = scheduler.schedule(this, delay, TimeUnit.MILLISECONDS);
+      }
+    } finally {
+      profilerLock.unlock();
```
jackshirazi, line 417: for reviewer, same code block, just pushed into try-finally block to use a lock. This ensures that setProfilerInterval is correct. The race is that run() finishes a profiling session and prepares to schedule the next one. It reads the old interval (e.g., 10s). Concurrently, the test calls setProfilerInterval (e.g., 100ms) and then reschedule(). reschedule() tries to cancel the current task. If they race, run() might successfully schedule the next task with the old 10s delay, effectively ignoring the update. The test then waits for the next session (expecting it in 100ms), but it doesn't happen for 10s, causing a timeout.

it's actually possible for an interruption to occur just before the lock was acquired. This should is handled by re-reading Thread.currentThread().isInterrupted()  to ensure no task is scheduled if an interruption occurred just before acquiring the lock

inferred-spans/src/main/java/io/opentelemetry/contrib/inferredspans/internal/SamplingProfiler.java
```diff
@@ -394,9 +396,14 @@ public void run() {
         config.isNonStopProfiling() && !interrupted && postProcessingEnabled;
     setProfilingSessionOngoing(continueProfilingSession);
 
-    if (!interrupted && !scheduler.isShutdown()) {
-      long delay = config.getProfilingInterval().toMillis() - profilingDuration.toMillis();
-      profilingTask = scheduler.schedule(this, delay, TimeUnit.MILLISECONDS);
+    profilerLock.lock();
+    try {
+      if (!isInterrupted && !scheduler.isShutdown()) {
```
jackshirazi: See below comment on whether to make this change
```suggestion
      if (!Thread.currentThread().isInterrupted() && !scheduler.isShutdown()) {
```

inferred-spans/src/main/java/io/opentelemetry/contrib/inferredspans/internal/SamplingProfiler.java
```diff
@@ -739,13 +746,18 @@ public void start() {
...
+      Future<?> future = this.profilingTask;
+      if (future != null) {
+        if (future.cancel(true)) {
+          Duration profilingDuration = config.getProfilingDuration();
+          long delay = config.getProfilingInterval().toMillis() - profilingDuration.toMillis();
+          profilingTask = scheduler.schedule(this, delay, TimeUnit.MILLISECONDS);
+        }
       }
+    } finally {
+      profilerLock.unlock();
```
jackshirazi, line 789: for reviewer, same code block just wrapped in the lock try-finally

inferred-spans/src/main/java/io/opentelemetry/contrib/inferredspans/internal/SamplingProfiler.java
```diff
@@ -406,25 +413,25 @@ private void profile(Duration profilingDuration) throws Exception {
...
+        profilingSessions++;
+
+        // When post-processing is disabled activation events are ignored, but we still need to
+        // invoke this method as it is the one enforcing the sampling session duration. As a side
+        // effect it will also consume residual activation events if post-processing is disabled
+        // dynamically
+        consumeActivationEventsFromRingBufferAndWriteToFile(profilingDuration);
+      } finally {
+        String stopMessage = profiler.execute("stop");
+        logger.fine(stopMessage);
```
jackshirazi, line 458: for reviewer. same code block, but restructured to ensure that `profiler.execute("stop");` is definitely called as an earlier interrupt will go straight to the catch block (below) bypassing the stop - this is most likely the test flakiness ("Profiler already started") - the profiler not correctly stopped

inferred-spans/src/main/java/io/opentelemetry/contrib/inferredspans/internal/SamplingProfiler.java
```diff
@@ -394,9 +396,14 @@ public void run() {
         config.isNonStopProfiling() && !interrupted && postProcessingEnabled;
```
SylvainJuge, line 404: Maybe we could inline `interrupted` variable and add a comment why we need the double call to `Thread.currentThread().isInterrupted()` to get an up-to-date status without using the value from the previous call.

SylvainJuge (approved):

inferred-spans/src/main/java/io/opentelemetry/contrib/inferredspans/internal/SamplingProfiler.java
laurit: you could add a comment that explains that this is in a finally block because the preceding code be interrupted

inferred-spans/src/main/java/io/opentelemetry/contrib/inferredspans/internal/SamplingProfiler.java
```diff
@@ -739,13 +746,18 @@ public void start() {
...
-    if (future != null) {
-      if (future.cancel(true)) {
-        Duration profilingDuration = config.getProfilingDuration();
-        long delay = config.getProfilingInterval().toMillis() - profilingDuration.toMillis();
-        profilingTask = scheduler.schedule(this, delay, TimeUnit.MILLISECONDS);
+    profilerLock.lock();
+    try {
+      Future<?> future = this.profilingTask;
+      if (future != null) {
+        if (future.cancel(true)) {
```
laurit: could be merged

jackshirazi: done

inferred-spans/src/main/java/io/opentelemetry/contrib/inferredspans/internal/SamplingProfiler.java
jackshirazi: done

inferred-spans/src/main/java/io/opentelemetry/contrib/inferredspans/internal/SamplingProfiler.java
jackshirazi: done

trask:
@jackshirazi does the CI failure on this PR look like same or different issue than you were trying to fix here? https://scans.gradle.com/s/gfqf5jlzidn6q

inferred-spans/src/main/java/io/opentelemetry/contrib/inferredspans/InferredSpansProcessorBuilder.java
```diff
@@ -194,17 +195,23 @@ public InferredSpansProcessorBuilder startScheduledProfiling(boolean startSchedu
...
 
   /** For testing only. */
-  InferredSpansProcessorBuilder jfrFile(@Nullable File jfrFile) {
+  public InferredSpansProcessorBuilder jfrFile(@Nullable File jfrFile) {
     this.jfrFile = jfrFile;
     return this;
   }
 
+  /** For testing only. */
+  public InferredSpansProcessorBuilder tempDir(@Nullable File tempDir) {
```
SylvainJuge, line 210: [minor] not related to this PR, but we could probably replace usage of `File` with `Path` to prevent having to call `toFile`.

inferred-spans/src/main/java/io/opentelemetry/contrib/inferredspans/internal/SamplingProfiler.java
```diff
@@ -739,18 +777,25 @@ public void start() {
...
+      profilerLock.unlock();
     }
   }
 
+  @SuppressWarnings({"FutureReturnValueIgnored", "Interruption"})
   public void stop() throws InterruptedException, IOException {
     // cancels/interrupts the profiling thread
+    if (profilingTask != null) {
+      profilingTask.cancel(true);
+    }
```
SylvainJuge, line 798: [minor] I would suggest to also wrap the access to `profilingTask` and potentially modifying its state with the `profilerLock`, at least for consistency with the other call to `.cancel(true)` above. There is however no side effect on potentially calling this twice as the return value of `cancel(...)` is ignored.

inferred-spans/src/main/java/io/opentelemetry/contrib/inferredspans/internal/asyncprofiler/JfrParser.java
```diff
@@ -93,6 +93,9 @@ public void parse(
     this.includedClasses = includedClasses;
     bufferedFile.setFile(file);
     long fileSize = bufferedFile.size();
+    if (fileSize == 0) {
+      return;
+    }
```
SylvainJuge, line 98: Is the case of empty file covered by any existing test ? I haven't seen any change related to this in the tests.

inferred-spans/src/test/java/io/opentelemetry/contrib/inferredspans/internal/SamplingProfilerTest.java
```diff
@@ -44,6 +43,10 @@
 @DisabledOnOpenJ9
 class SamplingProfilerTest {
 
+  static {
+    ProfilingActivationListener.ensureInitialized();
```
SylvainJuge, line 48: what is the benefit of explicitly calling this in the tests and not in production code ? is there any benefit of doing this and if yes maybe a comment could be welcome.

inferred-spans/src/main/java/io/opentelemetry/contrib/inferredspans/internal/SamplingProfiler.java
```diff
@@ -394,37 +404,62 @@ public void run() {
...
+          logger.fine("Profiler already started. Stopping and restarting.");
+          try {
+            profiler.stop();
+          } catch (RuntimeException ignore) {
+            logger.log(Level.FINE, "Ignored error on stopping profiler", ignore);
+          }
+          startMessage = profiler.execute(startCommand);
+        } else {
+          throw e;
+        }
```
SylvainJuge, line 439: Is is a known failure mode of async profiler here ? If so, then maybe having a dedicated method to wrap this "when exception is thrown stop, start and try again" logic could make it slightly more readeable.

inferred-spans/src/test/java/io/opentelemetry/contrib/inferredspans/internal/SamplingProfilerTest.java
jackshirazi: the ProfilingActivationListener has a static initializer which has to be executed before any Context access. This happens in production, but is not guaranteed in test because we start things out of order to setup tests

inferred-spans/src/main/java/io/opentelemetry/contrib/inferredspans/internal/SamplingProfiler.java
jackshirazi: this is testing fragility under concurrent test runs, I'd be surprised to see it get hit in production. Especially as dynamic updates to the test durations will be few and far in between in reality

inferred-spans/src/main/java/io/opentelemetry/contrib/inferredspans/internal/asyncprofiler/JfrParser.java
jackshirazi: good point, test added now

inferred-spans/src/main/java/io/opentelemetry/contrib/inferredspans/InferredSpansProcessorBuilder.java
jackshirazi: leaving this for another time

inferred-spans/src/main/java/io/opentelemetry/contrib/inferredspans/internal/SamplingProfiler.java
jackshirazi: leaving this for another time

SylvainJuge (approved):

inferred-spans/src/main/java/io/opentelemetry/contrib/inferredspans/InferredSpansProcessorBuilder.java
```diff
@@ -194,17 +195,23 @@ public InferredSpansProcessorBuilder startScheduledProfiling(boolean startSchedu
   }
 
   /** For testing only. */
-  InferredSpansProcessorBuilder activationEventsFile(@Nullable File activationEventsFile) {
+  public InferredSpansProcessorBuilder activationEventsFile(@Nullable File activationEventsFile) {
```
laurit: are you sure these need to be added to the public api?

jackshirazi: good catch, thanks. They don't

laurit (approved):

PR #2472, opened 2025-11-24
Simplify mise task

trask:
Follow-up to simplification introduced in https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/15447 (still need to go back after this is merged and remove the `--event pull_request` from those callers


.mise/tasks/lint/links-in-modified-files.sh
trask, line 8: need to leave this line to avoid breaking usages that pass in `--event`, will remove the line later after updating all the callers

zeitlinger:
we could also just fix the script to only check for modifications if it's a PR, but I'm fine either way

PR #2475, opened 2025-11-25
Update to markdownlint-cli2

trask:


jmx-metrics/docs/target-systems/kafka-consumer.md
```diff
@@ -1,56 +1,70 @@
 # Kafka Consumer Metrics
 
 The JMX Metric Gatherer provides built in Kafka consumer metric gathering capabilities for versions v0.8.2.x and above.
-These metrics are sourced from Kafka's exposed Yammer metrics for each instance: https://kafka.apache.org/documentation/#monitoring
+These metrics are sourced from Kafka's exposed Yammer metrics for each instance: <https://kafka.apache.org/documentation/#monitoring>
```
laurit: Just wondering whether it would make sense to remove `Yammer`. The linked document does mentions `Yammer Metrics` but considering that yammer -> codahale -> dropwizard rename happened more than a decade ago I wonder whether they really use yammer or forgot to update the documentation.

laurit (approved):

PR #2478, opened 2025-11-25
Use mise for running markdownlint

trask:
Also moves the config file to the root dir where it will be picked up by editors


zeitlinger:
great :heart:

PR #2516, opened 2025-12-10
Publish jar for cel extension

jaydeluca:
**Description:**

@laurit [pointed out](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2383#discussion_r2451930086) that we are not publishing a jar for the new cel sampler.

I've added the extension jar creation here, as well as an integration test.

**Testing:**

Integration tests that use a basic app and package and test the CEL sampler functionality

cc @dol

### DIFF
--- a/.github/workflows/oats-tests.yml
+++ b/.github/workflows/oats-tests.yml
@@ -5,6 +5,8 @@ on:
     branches:
       - main
     paths:
+      - .mise.toml
+      - .github/workflows/oats-tests.yml
       - cel-sampler/**
       - .mise/tasks/oats-tests.sh
   workflow_dispatch:


.github/workflows/oats-tests.yml
```diff
@@ -0,0 +1,36 @@
+name: OATS Tests
+
+on:
+  pull_request:
+    branches:
+      - main
+    paths:
+      - cel-sampler/**
+      - .mise/tasks/oats-tests.sh
```
zeitlinger, line 11: also `.mise.toml`, `.github/workflows/oats-tests.yml‎`

cel-sampler/build.gradle.kts
```diff
@@ -18,3 +19,32 @@ dependencies {
   testImplementation("io.opentelemetry:opentelemetry-sdk-extension-autoconfigure")
   testImplementation("io.opentelemetry:opentelemetry-sdk-extension-incubator")
 }
+
+tasks {
+  shadowJar {
+    /**
```
jaydeluca, line 25: I copied this from the GCP auth extension as-is
https://github.com/open-telemetry/opentelemetry-java-contrib/blob/5809da024bf450aa388d37a7ca2c398c2ca2cac0/gcp-auth-extension/build.gradle.kts

cel-sampler/oats/Dockerfile
```diff
@@ -0,0 +1,24 @@
+FROM eclipse-temurin:21-jre
+
+WORKDIR /usr/src/app/
+
+# renovate: datasource=github-releases depName=open-telemetry/opentelemetry-java-instrumentation
+ENV OPENTELEMETRY_JAVA_INSTRUMENTATION_VERSION=v2.22.0
```
zeitlinger, line 6: using `FROM` is more cache friendly and renovate works out of the box

zeitlinger: see https://github.com/grafana/grafana-opentelemetry-java/blob/bad369155b754fdb36887fc2082bb283ecdb76be/Dockerfile#L7

mise.toml
```diff
@@ -10,4 +11,5 @@ idiomatic_version_file_enable_tools = []
 # Based on: https://github.com/jdx/mise/discussions/4461
 windows_executable_extensions = ["sh"]
 windows_default_file_shell_args = "bash"
+unix_default_file_shell_args = "bash"
```
zeitlinger, line 14: what's that needed for?

jaydeluca: was getting [this](https://github.com/open-telemetry/opentelemetry-java-contrib/actions/runs/20107383942/job/57695095789?pr=2516):

```
[oats-tests] $ ~/work/opentelemetry-java-contrib/opentelemetry-java-contrib/.mise/tasks/oats-tests.sh
/home/runner/work/opentelemetry-java-contrib/opentelemetry-java-contrib/.mise/tasks/oats-tests.sh: 4: set: Illegal option -o pipefail
```

cel-sampler/oats/Dockerfile
jaydeluca: i might be misunderstanding something, but what would i be using as the base image using "FROM"? this is pulling the agent

zeitlinger: yes, that also what the example does.
it's then copied in https://github.com/grafana/grafana-opentelemetry-java/blob/bad369155b754fdb36887fc2082bb283ecdb76be/Dockerfile#L62

jaydeluca: the grafana one is pulling [a docker image](https://console.cloud.google.com/artifacts/docker/grafanalabs-global/us/docker-grafana-opentelemetry-java-prod/grafana-opentelemetry-java), it is not pulling the agent directly.

Doing something like this doesn't work/make sense since it isn't a container:

```
ARG OPENTELEMETRY_JAVA_INSTRUMENTATION_VERSION=v2.22.0
FROM github.com/open-telemetry/opentelemetry-java-instrumentation/releases/download/$OPENTELEMETRY_JAVA_INSTRUMENTATION_VERSION/opentelemetry-javaagent.jar as agent
```

I'm not aware of a container I could use to pull with the agent, and I'm not sure if it would be beneficial compared to just pulling it directly like this?

dol:
@jaydeluca Sorry, for being late to the party. I was experimenting the last few days on a proper setup on how to distribute the shadow jar of the cel-sampler, which took longer than expected.
As per legal requirement, the license agreement of the distributed dependencies should be checked before each release or dependency update. The PR #2518 tries to detect the licenses of all direct and transient dependencies and compares it to an allow list.
It also adds the SBOM and all the third party license agreement to the uber/shadow jar. This way all the legal obligations should be met.

Your PR adds the shadowJar bundling capability and an integration test to the cel-sampler. I agree, that an end2end integration test is missing. As https://github.com/open-telemetry/opentelemetry-java-contrib/blob/b7a327218c8cf0928c8cb2198e5fa764d530afda/gcp-auth-extension/build.gradle.kts#L70 already provides a good base for an integration test, I would take a similar approach and would do an implementation in pure Gradle and Java. From a developer perspective it's now required to have the docker, docker compose and mise tools installed to run the integration tests.

For the shadowJar bundling I would wait how the discussion in #2518 unfolds. Depending on the outcome it would be ideal if the integration tests could based on the proposed shadowJar convention.

cel-sampler/oats/Dockerfile
zeitlinger: - image: https://github.com/open-telemetry/opentelemetry-operator/pkgs/container/opentelemetry-operator%2Fautoinstrumentation-java
- why: the docker layer can be cached and renovate knows how to update

PR #2518, opened 2025-12-10
feat: add otel.shadow-conventions plugin to centralize dist packaging and license handling

dol:
Pull requests and discussion's involved:

- #2383
- #2516 . This PR is an overlap. The end2end test makes sense, but I would set it up similar to https://github.com/open-telemetry/opentelemetry-java-contrib/blob/b7a327218c8cf0928c8cb2198e5fa764d530afda/gcp-auth-extension/build.gradle.kts#L70 rather then relying on docker to be present for the test.

# Summary

A new in-repo convention plugin `otel.shadow-conventions` was added to centralize creation of
runtime distribution jars and legal/SBOM handling. The plugin applies and configures Shadow
(fat-jar), CycloneDX (SBOM), and the dependency-license-report plugin, and wires generated outputs
into the distribution packaging in a lazy, incremental-friendly way.

Inspired by https://cwiki.apache.org/confluence/display/FLINK/Licensing# .

# What changed

A new convention file was added under
`buildSrc/src/main/kotlin/otel.shadow-conventions.gradle.kts` which:

- applies the `com.gradleup.shadow`, `org.cyclonedx.bom`, and
  `com.github.jk1.dependency-license-report` plugins;
- centralizes `licenseReport` configuration and points it at the repo
  `config/dependency-license/*` policy files;
- configures CycloneDX to emit JSON-only SBOM and registers the BOM as a lazy input
  for packaging;
- adds a provider-based `copyLegalDocs` task that copies generated
  `THIRD-PARTY-NOTICES.txt` and the CycloneDX BOM into `build/generated/legal-docs` and
  renames the BOM to `SBOM.json`;
- configures `shadowJar` to set classifier `dist`, exclude dependency-provided
  `LICENSE`/`NOTICE` files, depend on the license and SBOM generator tasks, and include only the
  curated legal artifacts and project `LICENSE` into `META-INF` in the distribution jar.

`buildSrc/build.gradle.kts` was updated to include implementation dependencies for the Shadow,
dependency-license-report, and CycloneDX Gradle plugins so the new convention plugin compiles.

The `cel-sampler` module now applies `otel.shadow-conventions` and its README was updated to
recommend depending on the runtime distribution artifact (artifact with `-dist` classifier) for
extension usage.

Two configuration files were added under `config/dependency-license/`:

- `allowed-licenses.json` — an allowlist of acceptable licenses used by the license report check
- `license-normalizer-bundle.json` — rules to normalize license names across inputs

# Rationale

Centralizing dist packaging and license handling in a convention plugin provides a consistent
workflow across modules: the normal library JAR remains clean for library consumers, while the
self-contained distribution jar (classifier `dist`) bundles dependencies, SBOM, and curated legal
materials for runtime use. The plugin uses Gradle Providers and `dependsOn` wiring so generated
outputs are included without eager file resolution, supporting up-to-date checks and
configuration-cache friendliness.

Dependency-provided `LICENSE`/`NOTICE` files are excluded from the fat jar to avoid duplicates and
conflicts; the plugin includes only the authorized/generated artifacts.

### DIFF
--- a/gcp-auth-extension/build.gradle.kts
+++ b/gcp-auth-extension/build.gradle.kts
@@ -1,7 +1,7 @@
 plugins {
   id("otel.java-conventions")
+  id("otel.shadow-conventions")
   id("otel.publish-conventions")
-  id("com.gradleup.shadow")
 }
 
 description = "OpenTelemetry extension that provides GCP authentication support for OTLP exporters"
--- a/ibm-mq-metrics/build.gradle.kts
+++ b/ibm-mq-metrics/build.gradle.kts
@@ -1,7 +1,7 @@
 plugins {
   application
-  id("com.gradleup.shadow")
   id("otel.java-conventions")
+  id("otel.shadow-conventions")
   id("otel.publish-conventions")
 }
 
@@ -60,6 +60,10 @@ testing {
       }
     }
   }
+}
+
+licenseReport {
+  excludes = arrayOf("com.ibm.mq:com.ibm.mq.allclient")
 }
 
 tasks.shadowJar {
--- a/jmx-metrics/build.gradle.kts
+++ b/jmx-metrics/build.gradle.kts
@@ -1,7 +1,7 @@
 plugins {
   id("otel.java-conventions")
   application
-  id("com.gradleup.shadow")
+  id("otel.shadow-conventions")
 
   id("otel.groovy-conventions")
   id("otel.publish-conventions")
--- a/maven-extension/build.gradle.kts
+++ b/maven-extension/build.gradle.kts
@@ -1,6 +1,6 @@
 plugins {
   id("java")
-  id("com.gradleup.shadow")
+  id("otel.shadow-conventions")
   id("otel.java-conventions")
   id("otel.publish-conventions")
 }


buildSrc/src/main/kotlin/otel.shadow-conventions.gradle.kts
```diff
@@ -0,0 +1,97 @@
...
+import com.github.jk1.license.filter.LicenseBundleNormalizer
+import com.github.jk1.license.render.ReportRenderer
+import com.github.jk1.license.render.TextReportRenderer
+import com.github.jk1.license.task.CheckLicensePreparationTask
+import com.github.jk1.license.task.CheckLicenseTask
+import com.github.jk1.license.task.ReportTask
+
+plugins {
+  id("com.gradleup.shadow")
+  id("org.cyclonedx.bom")
```
breedx-splk, line 11: Hmmm, I'm fairly sure we don't use this `org.cyclonedx.bom` plugin in other java projects. It would be nice if we could remain consistent....have you tried looking at how the other repos are doing bom conventions?

Issue #2540, opened 2026-01-04
Improve disk buffering reading paradigm


bencehornak:
### Component(s)

disk-buffering

### Is your feature request related to a problem? Please describe.

I find the design of `StorageIterator` counterintuitive, because as opposed to most `Iterator`s, its [`next` method](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/Iterator.html#next() has a side-effect, namely that it deletes a file from the disk. I would even say that it violates the "contract" of the `Iterator.next()` method, because according to its javadoc it is only supposed to return the next element, so changes to the underlying data structure are unexpectable.

In the end I think the counterintuitive design contributed to the provided example in the README and the implementation in the opentelemetry-android project being logically incorrect (#2539 & https://github.com/open-telemetry/opentelemetry-android/pull/1493).

### Describe the solution you'd like

I would find an API for the `SignalStorage` more fool-proof, that requires the explicit notation of successful or unsuccessful consumption of the batches.

### Describe alternatives you've considered

 e.g.

**Alternative 1:**

```java
spanStorage.consume((batch) -> {
    CompletableResultCode resultCode = networkExporter.export(spanData);
    resultCode.join(timeout, TimeUnit.MILLISECONDS);

    return resultCode.isSuccess(); // if the return value is false, the file won't be deleted and optionally the iteration is aborted
});
```

**Alternative 2:**

```java
for (Iterator<Collection<SpanData>> iterator = spanStorage.iterator(); iterator.hasNext(); ) {
    Collection<SpanData> spanData = iterator.next();
    CompletableResultCode resultCode = networkExporter.export(spanData);
    resultCode.join(timeout, TimeUnit.MILLISECONDS);

    if (resultCode.isSuccess()) {
        iterator.remove(); // .remove() is called explicitly
    } else {
        break;
    }
}
```

Alternative 2 leaves more flexibility to the user, because the user can decide to continue or to abort the iteration. However, it would certainly surprise devs just bumping the version of the disk buffering lib without inserting the `remove()` logic into the code.

LMK what you think, @LikeTheSalad, @breedx-splk 

### Additional context

- #2084 
- #2183

### Tip

[...]

PR #2560, opened 2026-01-16
Disk buffering iterator explicit removal config option

LikeTheSalad:
Addresses https://github.com/open-telemetry/opentelemetry-java-contrib/issues/2540 by making item deletion during iteration configurable.

A new `deleteItemsOnIteration` option has been added to `FileStorageConfiguration`, defaulting to `true` to preserve backward compatibility. When set to `false`, items are no longer implicitly removed from disk during iteration, so users must call `Iterator.remove()` explicitly to delete items in that scenario.

Example usage with explicit deletion:

```java
FileStorageConfiguration config = FileStorageConfiguration.builder()
    .setDeleteItemsOnIteration(false)
    .build();
SignalStorage.Span spanStorage = FileSpanStorage.create(spansDir, config);

Iterator<Collection<SpanData>> spanIterator = spanStorage.iterator();
while (spanIterator.hasNext()) {
  Collection<SpanData> spans = spanIterator.next();
  // ... export spans ...
  spanIterator.remove(); // Items stay on disk unless this is called when "deleteItemsOnIteration" is set to "false".
}
```

### DIFF
--- a/disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/Storage.java
+++ b/disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/Storage.java
@@ -113,7 +113,6 @@ public final class Storage<T> implements Closeable {
     }
     ReadableFile readableFile = readableFileRef.get();
     if (readableFile != null && readableFile.isClosed()) {
-      // The file was deleted from the iterator
       readableFileRef.set(null);
       readableFile = null;
     }


disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/FolderManager.java
```diff
@@ -46,19 +52,29 @@ public FolderManager(File folder, FileStorageConfiguration configuration, Clock
     this.clock = clock;
   }
 
+  public static class CacheFile {
```
jaydeluca: could this be package private or do you think we need to expose it?

```suggestion
  static class CacheFile {
```

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/FolderManager.java
```diff
@@ -46,19 +52,29 @@ public FolderManager(File folder, FileStorageConfiguration configuration, Clock
...
+    public final File file;
+    public final long createdTimeMillis;
```
jaydeluca: could these be private, and then add a getter for `createdTimeMillis` ?

```suggestion
    private final File file;
    private final long createdTimeMillis;
```

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/files/ReadableFile.java
```diff
@@ -87,6 +89,10 @@ public synchronized boolean hasExpired() {
     return nowMillis(clock) >= expireTimeMillis;
   }
 
+  public long getCreatedTimeMillis() {
+    return createdTimeMillis;
+  }
```
jaydeluca: it looks like it might be possible that we try and [access this](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2560/files#diff-4ab92b18edc8e550a570fd5e6f577e0d5d08df1f1aed47fd700ffe3c79e0694bR141) if the file isn't valid. Should we synchronize this and add a check to make sure the file itself isn't closed?

```suggestion
  public synchronized long getCreatedTimeMillis() {
    if (isClosed.get()) {
      throw new IllegalStateException("File is closed");
    }
    return createdTimeMillis;
  }
```

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/FolderManager.java
```diff
@@ -93,29 +109,57 @@ public synchronized void clear() throws IOException {
     }
   }
 
+  @SuppressWarnings("MixedMutabilityReturnType")
```
jaydeluca: do we have a preference for returning mutable vs immutable? i think we could probably get rid of this annotation if we landed on one

```java
  private List<CacheFile> listCacheFiles(Predicate<CacheFile> exclude) {
    File[] existingFiles = folder.listFiles();
    if (existingFiles == null) {
      return new ArrayList<>();
    }
    ArrayList<CacheFile> files = new ArrayList<>();
    for (File file : existingFiles) {
      CacheFile cacheFile = fileToCacheFile(file);
      if (cacheFile != null && !exclude.test(cacheFile)) {
        files.add(cacheFile);
      }
    }
    return files;
  }
```

or 

```java
  private List<CacheFile> listCacheFiles(Predicate<CacheFile> exclude) {
    File[] existingFiles = folder.listFiles();
    if (existingFiles == null) {
      return Collections.emptyList();
    }
    ArrayList<CacheFile> files = new ArrayList<>();
    for (File file : existingFiles) {
      CacheFile cacheFile = fileToCacheFile(file);
      if (cacheFile != null && !exclude.test(cacheFile)) {
        files.add(cacheFile);
      }
    }
    return Collections.unmodifiableList(files);
  }
```

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/Storage.java
```diff
@@ -21,13 +21,16 @@
...
+import java.util.function.Predicate;
 import java.util.logging.Logger;
 import javax.annotation.Nullable;
 
 public final class Storage<T> implements Closeable {
   private static final int MAX_ATTEMPTS = 3;
   private final Logger logger = Logger.getLogger(Storage.class.getName());
   private final FolderManager folderManager;
+  private final AtomicReference<Predicate<FolderManager.CacheFile>> fileExclusion =
+      new AtomicReference<>(file -> false);
```
jaydeluca: it looks like we are just doing wholesale assignments with this instead of compare and set etc, could we just make it volatile?

```suggestion
  private volatile Predicate<FolderManager.CacheFile> fileExclusion = file -> false;
```

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/FolderManager.java
LikeTheSalad: Sure

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/FolderManager.java
LikeTheSalad: I think it's ok to make it package private. I didn't think much about exposing it because it's in an internal package, but it's probably better to change it and make sure it's not usable from the outside.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/Storage.java
LikeTheSalad: Sure

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/files/ReadableFile.java
LikeTheSalad: Sure 👍

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/FolderManager.java
LikeTheSalad: Thanks. I went with the "unmodifiableList" option.

disk-buffering/README.md
```diff
@@ -129,11 +132,52 @@ public boolean exportSpansFromDisk(SpanExporter networkExporter, long timeout) {
...
+    } else {
+      return false;
+    }
+  }
+  return true;
+}
+```
+
+Note that even with explicit deletion, disk usage is still bounded by the configured max folder size and max file
+age, so stale files are automatically purged when there's not enough space available before new data is written.
```
jaydeluca, line 176: 👍

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/Storage.java
```diff
@@ -117,19 +128,27 @@ private ReadableResult<T> doReadNext(SignalDeserializer<T> deserializer, int att
...
+      // Proto data corrupted, clear file.
+      readableFile.clear();
     }
 
-    // Retry with new file
+    // Search for newer files than the current one.
+    fileExclusion = file -> file.getCreatedTimeMillis() <= currentFileCreatedTime;
+    readableFile.close();
     readableFileRef.set(null);
     return doReadNext(deserializer, ++attemptNumber);
```
bencehornak, line 152: Not related, but loops are better than recursion as they don't pollute the stack.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/Storage.java
```diff
@@ -106,9 +112,14 @@ private ReadableResult<T> doReadNext(SignalDeserializer<T> deserializer, int att
       return null;
     }
     ReadableFile readableFile = readableFileRef.get();
+    if (readableFile != null && readableFile.isClosed()) {
+      // The file was deleted from the iterator
+      readableFileRef.set(null);
+      readableFile = null;
+    }
```
bencehornak, line 118: I don't get why `readableFile.isClosed()` means that the file was deleted from the iterator. Is this comment up-to-date?

LikeTheSalad: The comment was added in this PR, and it was meant to explain the use case introduced here (people explicitly deleting the content). But on a second thought, it can be closed due to other reasons as well, so it's ultimately confusing. I've just removed it.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/Storage.java
LikeTheSalad: It's been ages since I wrote this part. I honestly don't remember why I chose recursion, but I think it's because on a happy path, it should only be executed once, so the recursion would only be present on exceptions, hence an iteration was not the expected, common behavior. I think I understand what you're saying, but so far it doesn't seem to cause an issue in practice, and it also doesn't seem to be a trivial change, so it wouldn't fall under my current set of priorities. Still, if you're interested in finding an alternative solution, feel free to open a PR, and I'll take a look 👍

PR #2619, opened 2026-02-09
feat(lambda): read cloud.account.id from symlink in Lambda detector

RaphaelManke:
## Summary

- Reads `cloud.account.id` from the symlink at `/tmp/.otel-aws-account-id` created by the OTel Lambda Extension
- Uses `Files.readSymbolicLink()` to read the symlink target
- Logs at FINE level if the symlink doesn't exist (`catch IOException | UnsupportedOperationException`)

## Depends on

- open-telemetry/opentelemetry-lambda#2127 (extension creates the symlink)

## Changes

- `LambdaResource.java` — Added readlink logic with `ACCOUNT_ID_SYMLINK_PATH` constant, injectable `Path` for testing
- `LambdaResourceTest.java` — Happy path, missing symlink, leading zeros tests

## Test plan

- [x] Unit test: symlink exists → `cloud.account.id` set correctly
- [x] Unit test: leading zeros preserved as string
- [x] Unit test: missing symlink → attribute absent, no error


aws-resources/src/test/java/io/opentelemetry/contrib/aws/resource/LambdaResourceTest.java
```diff
@@ -65,6 +70,53 @@ void shouldAddAllAttributes() {
             entry(FAAS_VERSION, "1.2.3"));
   }
 
+  @Test
+  void shouldReadCloudAccountIdFromSymlink(@TempDir Path tempDir) throws Exception {
```
laurit, line 77: it would be nice to skip the test when run on windows

aws-resources/src/main/java/io/opentelemetry/contrib/aws/resource/LambdaResource.java
```diff
@@ -60,6 +76,15 @@ static Resource buildResource(Map<String, String> environmentVariables) {
       builder.put(FAAS_VERSION, functionVersion);
     }
 
+    try {
+      String accountId = Files.readSymbolicLink(accountIdSymlink).toString();
```
laurit: you could check with `Files.isSymbolicLink` to avoid triggering exceptions when the file does not exist

aws-resources/src/test/java/io/opentelemetry/contrib/aws/resource/LambdaResourceTest.java
RaphaelManke: Done — used `@EnabledOnOs(OS.LINUX)` instead of `@DisabledOnOs(OS.WINDOWS)` since Lambda only runs on Linux, so it's more precise to enable only there.

aws-resources/src/main/java/io/opentelemetry/contrib/aws/resource/LambdaResource.java
```diff
@@ -60,6 +76,15 @@ static Resource buildResource(Map<String, String> environmentVariables) {
...
+      if (!accountId.isEmpty()) {
+        builder.put(CLOUD_ACCOUNT_ID, accountId);
+      }
+    } catch (IOException | UnsupportedOperationException e) {
+      logger.log(Level.FINE, "cloud.account.id not available via symlink", e);
```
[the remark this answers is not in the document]
RaphaelManke: This was already using a static import (`import static java.util.logging.Level.FINE`) — no change needed.

aws-resources/src/main/java/io/opentelemetry/contrib/aws/resource/LambdaResource.java
RaphaelManke: Done — added `Files.isSymbolicLink()` guard before the read. Also added an `isLinux()` OS check so the symlink detection is skipped entirely on non-Linux platforms, since Lambda is Linux-only.

aws-resources/src/main/java/io/opentelemetry/contrib/aws/resource/LambdaResource.java
```diff
@@ -5,24 +5,36 @@
...
+import java.util.logging.Logger;
 import java.util.stream.Stream;
 
 /** A factory for a {@link Resource} which provides information about the AWS Lambda function. */
 public final class LambdaResource {
 
+  private static final Logger logger = Logger.getLogger(LambdaResource.class.getName());
+
+  private static final String ACCOUNT_ID_SYMLINK_PATH = "/tmp/.otel-aws-account-id";
+
```
laurit:
```suggestion
```

laurit (approved):

jaydeluca:
@srprash @wangzlei this does this look good to go to you?

Issue #2546, opened 2026-01-09
[Dynamic] Add basic Telemetry Policy support to dynamic-control


jackshirazi:
### Component(s)

dynamic-control

### Is your feature request related to a problem? Please describe.

The draft [Telemetry Policy](https://github.com/open-telemetry/opentelemetry-specification/pull/4738) seems a reasonable approach to provide a framework for dynamically updating features, so is appropriate to implement here experimentally

The current proposal provides for a process from some source of a change definition through to applying that change in the agent. At a high level, it is
- message -> provider -> policy -> policy aggregator -> implementer

An example could be
- message to make a change to the "sampling rate" -> OpampPolicyProvider (reads the message from OpAMP) -> TraceSamplingRatePolicy (the policy that changes the sampling rate, with the target traceid rate) -> PolicyStore (handles priority and potential merges of policy types) -> TraceSamplingRatePolicyImplementer (takes a new traceid rate and applies it to the sampler)

The policy body data structure sent to the provider is non-specific

Overall, the current proposal suggests these elements
```
FileProvider -----\
HTTPProvider ----- \  
OpAMPProvider ----- > PolicyHandler --> PolicyImplementers
CustomProvider ----/
```

### Describe the solution you'd like

Add in a basic example using the proposed framework, to periodically read from a file and applying it to dynamically changing the sampling rate. Expected additions:
- [x] TelemetryPolicy 
- [x] PolicyProvider interface for providers common behaviours
- [x] LinePerPolicyFileProvider to read a file and create TelemetryPolicys
- [x] PolicyImplementer interface to handle changes to policies, so that the ...
- [x] PolicyStore handles changes and propagates them to appliers
- [x] A DelegatingSampler so that changes can be applied generically in the sampler chain
- [x] A specific TraceSamplingRatePolicyImplementer to apply changes to the sampling rate
- [ ] env var/property/config option to specify an init file
- [ ] A TelemetryPolicyManager which manages the full process
- [ ] extension handling so that this can be used as an extension

To keep things easy for reviews, I'll add things piecewise. The initial implementation is nearing completion, there remain elements to be added to fully setup the policy pipeline described above, and to make the implementation a working extension 

### Additional context

Part of https://github.com/open-telemetry/opentelemetry-java-contrib/issues/2416

### Tip

[...]

PR #2634, opened 2026-02-16
[dynamic-control] Add trace sampling-rate implementer

jackshirazi:
Per https://github.com/open-telemetry/opentelemetry-java-contrib/issues/2546

A basic trace sampling-rate implementer which simplistically applies changes to the sample rate

A full working implementation of the https://github.com/open-telemetry/opentelemetry-specification/pull/4738 is [here](https://github.com/jackshirazi/opentelemetry-java-contrib/commit/01d10270cae6c6d55cbb1364f6b5767d27d5a78e) in this [branch](https://github.com/jackshirazi/opentelemetry-java-contrib/tree/sampler-example)

Note the final dynamic-control sampling management needs to be more sophisticated, to dynamically handle route exclusion/inclusions as well as sampling rate and needs to use the recently added composable samplers (by default), but that's all for later addition

### DIFF
--- a/dynamic-control/src/test/java/io/opentelemetry/contrib/dynamic/policy/TraceSamplingRatePolicyImplementerTest.java
+++ b/dynamic-control/src/test/java/io/opentelemetry/contrib/dynamic/policy/TraceSamplingRatePolicyImplementerTest.java
@@ -22,7 +22,7 @@ import java.util.Collections;
 import java.util.List;
 import org.junit.jupiter.api.Test;
 
-public class TraceSamplingRatePolicyImplementerTest {
+class TraceSamplingRatePolicyImplementerTest {
 
   private static final ObjectMapper MAPPER = new ObjectMapper();
 


dynamic-control/src/test/java/io/opentelemetry/contrib/dynamic/policy/TraceSamplingRatePolicyImplementerTest.java
```diff
@@ -0,0 +1,107 @@
...
+  private static final ObjectMapper MAPPER = new ObjectMapper();
+
+  @Test
+  void nullSpecFallsBackToAlwaysOn() {
+    DelegatingSampler delegatingSampler = new DelegatingSampler(Sampler.alwaysOff());
+    TraceSamplingRatePolicyImplementer implementer =
+        new TraceSamplingRatePolicyImplementer(delegatingSampler);
+
+    implementer.onPoliciesChanged(
+        Collections.singletonList(new TelemetryPolicy("trace-sampling", null)));
```
jaydeluca: could do a static import for all of these usages. we recently added this to spotless in instrumentation so i imagine we'll do something similar here soon too
```suggestion
       singletonList(new TelemetryPolicy("trace-sampling", null)));
```

jaydeluca (approved):

dynamic-control/src/test/java/io/opentelemetry/contrib/dynamic/policy/TraceSamplingRatePolicyImplementerTest.java
jackshirazi: good suggestion, thanks. done now

PR #2639, opened 2026-02-19
Fix release workflow

trask:
And fix up apidiffs from past post-release updates.

Supersedes #2638


.github/workflows/release.yml
```diff
@@ -261,7 +264,7 @@ jobs:
 
       - name: Create pull request against main
         env:
-          VERSION: ${{ needs.release.outputs.version }}
+          VERSION: ${{ needs.release.outputs.version }}gi
```
jaydeluca:
```suggestion
          VERSION: ${{ needs.release.outputs.version }}
```

PR #2652, opened 2026-02-24
[dynamic control] Move away from JSON requirement

jackshirazi:
**Description:**

The [Telemetry Policy OTEP](https://github.com/open-telemetry/opentelemetry-specification/pull/4738) has moved away from a specific JSON requirement to a non-specific data structure. First step in moving to this updated spec is to migrate away from JSON in the telemetry policy. As a by product, it's cleaner in many ways

**Existing Issue(s):**

**Testing:**

Tests updated

**Documentation:**

Documentation has been changed accordingly but the module documentation is the same as it's the same architecture

**Outstanding items:**

Although this is cleaner, it means there will be an additional mapping infrastructure needed to map custom structures into providers

### DIFF
--- a/dynamic-control/src/main/java/io/opentelemetry/contrib/dynamic/policy/TelemetryPolicy.java
+++ b/dynamic-control/src/main/java/io/opentelemetry/contrib/dynamic/policy/TelemetryPolicy.java
@@ -18,6 +18,9 @@ import java.util.Objects;
  * <p>As an example, policy type {@code trace-sampling} indicates that trace sampling behavior
  * should be configured.
  *
+ * <p>Direct instantiation of this base class is intentionally supported for type-only policy
+ * signals (for example, to indicate policy removal/reset without policy-specific values).
+ *
  * @see io.opentelemetry.contrib.dynamic.policy
  */
 public class TelemetryPolicy {
@@ -25,6 +28,9 @@ public class TelemetryPolicy {
 
   /**
    * Constructs a new TelemetryPolicy.
+   *
+   * <p>This constructor is used by type-specific subclasses and also directly for type-only policy
+   * signals such as policy removal/reset.
    *
    * @param type the type of the policy (e.g., "trace-sampling"), must not be null.
    */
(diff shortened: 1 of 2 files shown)


dynamic-control/src/main/java/io/opentelemetry/contrib/dynamic/policy/TelemetryPolicy.java
```diff
@@ -5,43 +5,38 @@
...
- * sampling-probability be set to 50%
+ * <p>As an example, policy type {@code trace-sampling} indicates that trace sampling behavior
+ * should be configured.
+ *
+ * <p>Direct instantiation of this base class is intentionally supported for type-only policy
+ * signals (for example, to indicate policy removal/reset without policy-specific values).
  *
  * @see io.opentelemetry.contrib.dynamic.policy
  */
 public class TelemetryPolicy {
```
LikeTheSalad, line 26: Not a blocker, but I'm curious if it makes sense to turn this into an interface to make it more flexible/easier to extend? I don't have a specific use case in mind where more flexibility would be needed; it's just that the rules around dynamic support are changing so often that it might be useful to have such flexibility.

jackshirazi: for now its useful as a concrete class for resetting to default. I've considered having it abstract or as an interface but it seems most useful at the moment as a concrete class. My aim is to have a working extension that supports sampling rate changes, then add further functionality and see how it changes as needed. Since this is all experimental, letting it evolve as needed rather than trying to anticipate the end state, is the preferred way to go (I think)

LikeTheSalad (approved):

dynamic-control/src/main/java/io/opentelemetry/contrib/dynamic/policy/TelemetryPolicy.java
breedx-splk: If it's a public class, let's make it `final` so that users don't try and subclass it (they will!). 🤣

breedx-splk: oh wait, that's apparently the intended use....oof. I don't love that.

jackshirazi: Each policy will be a subclass of TelemetryPolicy. TelemetryPolicy itself could potentially become an abstract superclass or an interface, and we could switch to a "DeleteTelemetryPolicy" subclass/implementation to handle deletions. Please can we do that in a different PR if that's the way we want to go

jackshirazi: Yes?

PR #2738, opened 2026-04-02
[dynamic control] add config parsing for both json and yaml

jackshirazi:
**Description:**

Adding parsing for json and yaml of initialization of the Policy pipeline for examples like the following json or yaml

```
{
  "sources": [
    {
      "kind": "opamp",
      "format": "jsonkeyvalue",
      "location": "vendor-specific",
      "mappings": [
        { "sourceKey": "sampling_rate", "policyType": "trace_sampling_rate_policy" },
        { "sourceKey": "send_logs", "policyType": "log_export_enabled_policy" }
      ]
    },
    {
      "kind": "file",
      "format": "keyvalue",
      "location": "/some/path/conf.file",
      "mappings": [
        { "sourceKey": "send_metrics", "policyType": "metric_export_enabled_policy" }
      ]
    }
  ]
}
```

```
sources:
  - kind: opamp
    format: jsonkeyvalue
    location: vendor-specific
    mappings:
      - sourceKey: sampling_rate
        policyType: trace_sampling_rate_policy
      - sourceKey: send_logs
        policyType: log_export_enabled_policy

  - kind: file
    format: keyvalue
    location: /some/path/conf.file
    mappings:
      - sourceKey: send_metrics
        policyType: metric_export_enabled_policy
```

**Existing Issue(s):**

**Testing:**

Added

**Documentation:**

to be added

**Outstanding items:**

Still to hook up config to actual policy and build the pipeline

### DIFF
--- a/dynamic-control/src/main/java/io/opentelemetry/contrib/dynamic/policy/registry/yaml/YamlPolicyInitConfigReader.java
+++ b/dynamic-control/src/main/java/io/opentelemetry/contrib/dynamic/policy/registry/yaml/YamlPolicyInitConfigReader.java
@@ -17,7 +17,7 @@ import java.nio.charset.StandardCharsets;
 import java.util.Objects;
 
 /** Reads {@link PolicyInitConfig} from YAML (registry initialization file or payload). */
-public final class PolicyInitConfigReader {
+public final class YamlPolicyInitConfigReader {
 
   private static final ObjectMapper MAPPER = new ObjectMapper(new YAMLFactory());
 
--- a/dynamic-control/src/main/java/io/opentelemetry/contrib/dynamic/policy/registry/yaml/YamlPolicyInitConfigReader.java
+++ b/dynamic-control/src/main/java/io/opentelemetry/contrib/dynamic/policy/registry/yaml/YamlPolicyInitConfigReader.java
@@ -21,13 +21,15 @@ public final class YamlPolicyInitConfigReader {
 
   private static final ObjectMapper MAPPER = new ObjectMapper(new YAMLFactory());
 
-  public PolicyInitConfig read(InputStream in) throws IOException {
+  private YamlPolicyInitConfigReader() {}
+
+  public static PolicyInitConfig read(InputStream in) throws IOException {
     Objects.requireNonNull(in, "in cannot be null");
     JsonNode root = MAPPER.readTree(in);
     return JsonNodePolicyInitConfigParser.parse(root);
   }
 
-  public PolicyInitConfig read(String yaml) throws IOException {
+  public static PolicyInitConfig read(String yaml) throws IOException {
     Objects.requireNonNull(yaml, "yaml cannot be null");
     return read(new ByteArrayInputStream(yaml.getBytes(StandardCharsets.UTF_8)));
   }
--- a/dynamic-control/src/test/resources/io/opentelemetry/contrib/dynamic/policy/registry/json/policy-init-example.json
+++ b/dynamic-control/src/test/resources/io/opentelemetry/contrib/dynamic/policy/registry/json/policy-init-example.json
@@ -3,7 +3,7 @@
     {
       "kind": "opamp",
       "format": "jsonkeyvalue",
-      "location": "elastic",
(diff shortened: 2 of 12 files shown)


dynamic-control/src/test/resources/io/opentelemetry/contrib/dynamic/policy/registry/json/policy-init-example.json
```diff
@@ -0,0 +1,15 @@
+{
+  "sources": [
+    {
+      "kind": "opamp",
+      "format": "jsonkeyvalue",
+      "location": "elastic",
```
jackshirazi:
```suggestion
      "location": "vendor-specific",
```

dynamic-control/src/test/resources/io/opentelemetry/contrib/dynamic/policy/registry/yaml/policy-init-example.yaml
```diff
@@ -0,0 +1,13 @@
+sources:
+  - kind: opamp
+    format: jsonkeyvalue
+    location: elastic
```
jackshirazi:
```suggestion
    location: vendor-specific
```

jackshirazi:
```suggestion
    location: vendor-specific
```

dynamic-control/src/main/java/io/opentelemetry/contrib/dynamic/policy/registry/yaml/YamlPolicyInitConfigReader.java
```diff
@@ -0,0 +1,34 @@
...
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
+import java.util.Objects;
+
+/** Reads {@link PolicyInitConfig} from YAML (registry initialization file or payload). */
+public final class PolicyInitConfigReader {
+
+  private static final ObjectMapper MAPPER = new ObjectMapper(new YAMLFactory());
+
+  public PolicyInitConfig read(InputStream in) throws IOException {
```
LikeTheSalad: I find the 2 `PolicyInitConfigReader` types easy to mix up by users, in case they have to reference them directly, because not only are they named the same but they also have the same method names, so I'm worried about potential human-error kind of situations where they might think they are parsing a JSON while they are in fact using the YAML parser or vice versa. I noticed that the one hint they are able to see is the type's package name, to tell whether it's for json or yaml, which I'm not sure is enough.

jackshirazi: Good point! I've renamed them to YamlPolicy... and JsonPolicy...

LikeTheSalad (approved):

dynamic-control/src/main/java/io/opentelemetry/contrib/dynamic/policy/registry/yaml/YamlPolicyInitConfigReader.java
```diff
@@ -0,0 +1,34 @@
...
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
+import java.util.Objects;
+
+/** Reads {@link PolicyInitConfig} from YAML (registry initialization file or payload). */
+public final class YamlPolicyInitConfigReader {
+
+  private static final ObjectMapper MAPPER = new ObjectMapper(new YAMLFactory());
+
+  public PolicyInitConfig read(InputStream in) throws IOException {
```
jaydeluca: since these reader classes don't have any state, could they be static utility classes like `JsonNodePolicyInitConfigParser`?

jackshirazi: Yes, good call. If they need to be polymorphic we can move them back to instance methods but per your point I've made them static now

jaydeluca (approved):

PR #2745, opened 2026-04-07
Add FilteringSpanExporter with composable SpanFilter and TraceFilter interfaces

udaysagar2177:
## Summary

Adds a trace-level filtering `SpanExporter` to the `processors` module. `FilteringSpanExporter` wraps a delegate exporter and only forwards spans belonging to traces that match at least one configured filter. Filtering is trace-aware: if any filter matches, all spans sharing that trace ID are exported together.

Two filter interfaces enable composable, user-extensible filtering:

- **`SpanFilter`** — per-span evaluation (e.g., error status, slow duration)
- **`TraceFilter`** — batch-level evaluation over all spans sharing a trace ID (e.g., overall trace wall-clock duration)

Built-in implementations:

| Class | Type | Behavior |
|-------|------|----------|
| `ErrorSpanFilter` | `SpanFilter` | Keeps traces containing any span with `StatusCode.ERROR` |
| `DurationSpanFilter` | `SpanFilter` | Keeps traces containing any span exceeding a duration threshold |
| `TraceDurationFilter` | `TraceFilter` | Keeps traces whose wall-clock duration (max end - min start) exceeds a threshold |

An optional `Meter` parameter emits a `filtering.span.exporter.dropped` counter with a `reason` attribute for observability into dropped spans.

### Motivation

Services with heavy sampling often want to prioritize exporting traces that contain errors or exhibit unusual latency, without increasing overall sampling rate. This exporter wrapper provides that capability as a composable, upstream component that any OTel Java user can plug into their SDK configuration.

The existing `InterceptableSpanExporter` in this module operates per-span via the `Interceptor` interface, but has no trace-level grouping. `FilteringSpanExporter` fills this gap by grouping spans by trace ID and making keep/drop decisions at the trace level.

### Usage

```java
SpanExporter filtered = new FilteringSpanExporter(
    delegate,
    Arrays.asList(new ErrorSpanFilter(), new DurationSpanFilter(2000)),
    Collections.singletonList(new TraceDurationFilter(10000)));

// Custom filters via lambda
SpanFilter nameFilter = span -> span.getName().contains("important");

// With metrics
SpanExporter withMetrics = new FilteringSpanExporter(
    delegate, spanFilters, traceFilters, meter);
```

## Changes

| File | Description |
|------|-------------|
| `SpanFilter.java` | Interface for per-span filtering |
| `TraceFilter.java` | Interface for batch-level trace filtering |
| `ErrorSpanFilter.java` | Built-in: keeps traces with error spans |
| `DurationSpanFilter.java` | Built-in: keeps traces with slow spans |
| `TraceDurationFilter.java` | Built-in: keeps traces with long wall-clock duration |
| `FilteringSpanExporter.java` | Core exporter wrapper with single-pass filtering |
| `ErrorSpanFilterTest.java` | 3 tests |
| `DurationSpanFilterTest.java` | 3 tests |
| `TraceDurationFilterTest.java` | 5 tests |
| `FilteringSpanExporterTest.java` | 13 tests (composition, grouping, metrics, delegation) |
| `README.md` | Documentation |

## Test plan

- [x] `ErrorSpanFilterTest` — error, OK, UNSET status codes
- [x] `DurationSpanFilterTest` — over/at/under threshold boundary cases
- [x] `TraceDurationFilterTest` — over/at/under threshold, single-span trace
- [x] `FilteringSpanExporterTest` — trace grouping, mixed traces, span-only/trace-only/custom filter composition, dropped-span metrics, flush/shutdown delegation, empty batch
- [x] Spotless formatting passes
- [x] Compiles with NullAway and Error Prone checks

### DIFF
--- a/processors/README.md
+++ b/processors/README.md
@@ -32,6 +32,47 @@ logger_provider:
 
 `FilteringLogRecordProcessor` is a `LogRecordProcessor` that only keep logs  based on a predicate
 
+## Filtering Span Exporter
+
+`FilteringSpanExporter` is a `SpanExporter` wrapper that filters spans at the trace level before delegating to the underlying exporter. Filtering is composable via two interfaces:
+
+- `SpanFilter` - evaluates individual spans (e.g., error status, slow duration)
+- `TraceFilter` - evaluates all spans belonging to a trace within the batch (e.g., overall trace wall-clock duration)
+
+A trace is kept if any `SpanFilter` matches any of its spans, OR any `TraceFilter` matches the trace's span group. All spans sharing a trace ID are exported together.
+
+Built-in filters:
+- `ErrorSpanFilter` - keeps traces containing any span with error status
+- `DurationSpanFilter` - keeps traces containing any span exceeding a duration threshold
+- `TraceDurationFilter` - keeps traces whose wall-clock duration (max end - min start) exceeds a threshold
+
+Usage:
+
+```java
+SpanExporter delegate = OtlpGrpcSpanExporter.getDefault();
+
+// Export traces with errors, individual spans > 2s, or trace duration > 10s
+SpanExporter filtering = new FilteringSpanExporter(
+    delegate,
+    Arrays.asList(new ErrorSpanFilter(), new DurationSpanFilter(2000)),
+    Collections.singletonList(new TraceDurationFilter(10000)));
+
+// Custom filters
+SpanFilter nameFilter = span -> span.getName().contains("important");
+SpanExporter custom = new FilteringSpanExporter(
+    delegate,
+    Collections.singletonList(nameFilter),
+    Collections.emptyList());
+
+// Optionally pass a Meter to emit dropped-span metrics
+Meter meter = openTelemetry.getMeter("my-service");
+SpanExporter withMetrics = new FilteringSpanExporter(
+    delegate,
+    Arrays.asList(new ErrorSpanFilter(), new DurationSpanFilter(2000)),
+    Collections.singletonList(new TraceDurationFilter(10000)),
+    meter);
+```
+
 ## Component owners
 
 - [Cesar Munoz](https://github.com/LikeTheSalad), Elastic
--- a/processors/README.md
+++ b/processors/README.md
@@ -34,29 +34,31 @@ logger_provider:
 
 ## Filtering Span Exporter
 
-`FilteringSpanExporter` is a `SpanExporter` wrapper that filters spans at the trace level before delegating to the underlying exporter. Filtering is composable via two interfaces:
+`FilteringSpanExporter` is a `SpanExporter` wrapper that filters spans within each export batch before delegating to the underlying exporter. Filtering is composable via two interfaces:
 
 - `SpanFilter` - evaluates individual spans (e.g., error status, slow duration)
 - `TraceFilter` - evaluates all spans belonging to a trace within the batch (e.g., overall trace wall-clock duration)
 
-A trace is kept if any `SpanFilter` matches any of its spans, OR any `TraceFilter` matches the trace's span group. All spans sharing a trace ID are exported together.
+Within a batch, if any `SpanFilter` matches any span or any `TraceFilter` matches a trace's span group, all spans sharing that trace ID in the batch are exported together.
+
+**Note:** Filtering decisions are scoped to a single `export()` call. Spans from the same trace arriving in different batches are evaluated independently, so a trace split across batches may be partially exported.
 
 Built-in filters:
 
-- `ErrorSpanFilter` - keeps traces containing any span with error status
-- `DurationSpanFilter` - keeps traces containing any span exceeding a duration threshold
-- `TraceDurationFilter` - keeps traces whose wall-clock duration (max end - min start) exceeds a threshold
+- `ErrorSpanFilter` - matches spans with error status
+- `DurationSpanFilter` - matches spans exceeding a duration threshold
+- `TraceDurationFilter` - matches when a trace's wall-clock duration (max end - min start) in the batch exceeds a threshold
 
 Usage:
 
 ```java
 SpanExporter delegate = OtlpGrpcSpanExporter.getDefault();
 
-// Export traces with errors, individual spans > 2s, or trace duration > 10s
+// Export spans whose batch-colocated trace has errors, individual spans > 2s, or trace duration > 10s
 SpanExporter filtering = new FilteringSpanExporter(
     delegate,
-    Arrays.asList(new ErrorSpanFilter(), new DurationSpanFilter(2000)),
-    Collections.singletonList(new TraceDurationFilter(10000)));
+    Arrays.asList(new ErrorSpanFilter(), new DurationSpanFilter(Duration.ofSeconds(2))),
+    Collections.singletonList(new TraceDurationFilter(Duration.ofSeconds(10))));
 
 // Custom filters
 SpanFilter nameFilter = span -> span.getName().contains("important");
@@ -69,8 +71,8 @@ SpanExporter custom = new FilteringSpanExporter(
 Meter meter = openTelemetry.getMeter("my-service");
 SpanExporter withMetrics = new FilteringSpanExporter(
     delegate,
-    Arrays.asList(new ErrorSpanFilter(), new DurationSpanFilter(2000)),
-    Collections.singletonList(new TraceDurationFilter(10000)),
+    Arrays.asList(new ErrorSpanFilter(), new DurationSpanFilter(Duration.ofSeconds(2))),
+    Collections.singletonList(new TraceDurationFilter(Duration.ofSeconds(10))),
     meter);
 ```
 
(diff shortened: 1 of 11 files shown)


processors/README.md
```diff
@@ -32,6 +32,47 @@ logger_provider:
...
+
+- `SpanFilter` - evaluates individual spans (e.g., error status, slow duration)
+- `TraceFilter` - evaluates all spans belonging to a trace within the batch (e.g., overall trace wall-clock duration)
+
+A trace is kept if any `SpanFilter` matches any of its spans, OR any `TraceFilter` matches the trace's span group. All spans sharing a trace ID are exported together.
+
+Built-in filters:
+- `ErrorSpanFilter` - keeps traces containing any span with error status
+- `DurationSpanFilter` - keeps traces containing any span exceeding a duration threshold
+- `TraceDurationFilter` - keeps traces whose wall-clock duration (max end - min start) exceeds a threshold
```
breedx-splk: (this should fix the linter)

```suggestion
Built-in filters:

- `ErrorSpanFilter` - keeps traces containing any span with error status
- `DurationSpanFilter` - keeps traces containing any span exceeding a duration threshold
- `TraceDurationFilter` - keeps traces whose wall-clock duration (max end - min start) exceeds a threshold
```

udaysagar2177: Done!

LikeTheSalad (approved):
LGTM! Thank you 🙏

processors/src/main/java/io/opentelemetry/contrib/filter/TraceDurationFilter.java
```diff
@@ -0,0 +1,44 @@
...
+   *
+   * @param thresholdMs the trace duration threshold in milliseconds; traces with wall-clock
+   *     duration strictly greater than this are considered interesting
+   */
+  public TraceDurationFilter(long thresholdMs) {
+    this.thresholdNanos = TimeUnit.MILLISECONDS.toNanos(thresholdMs);
+  }
+
+  @Override
+  public boolean shouldKeep(String traceId, Collection<SpanData> spans) {
```
jaydeluca, line 34:
```suggestion
  public boolean shouldKeep(String traceId, Collection<SpanData> spans) {
    if (spans.isEmpty()) {
      return false;
    }
```

we can add an early return for an empty list of spans otherwise you end up with `maxEnd - minStart = Long.MIN_VALUE - Long.MAX_VALUE = 1` down on line 42 and it'll return true.

Related test that currently would fail:

```java
  @Test
  void worksWithEmptyList() {
    TraceDurationFilter filter = new TraceDurationFilter(0);
    assertThat(filter.shouldKeep("trace-1", emptyList())).isFalse();
  }
```

processors/src/main/java/io/opentelemetry/contrib/filter/FilteringSpanExporter.java
```diff
@@ -0,0 +1,164 @@
...
+   *     disable metrics
+   */
+  public FilteringSpanExporter(
+      SpanExporter delegate,
+      List<SpanFilter> spanFilters,
+      List<TraceFilter> traceFilters,
+      @Nullable Meter meter) {
+    this.delegate = delegate;
+    this.spanFilters = Collections.unmodifiableList(new ArrayList<>(spanFilters));
+    this.traceFilters = Collections.unmodifiableList(new ArrayList<>(traceFilters));
```
jaydeluca: should we do any kind of validation here?

```suggestion
    this.delegate = Objects.requireNonNull(delegate, "delegate");
    Objects.requireNonNull(spanFilters, "spanFilters");
    Objects.requireNonNull(traceFilters, "traceFilters");
    this.spanFilters = Collections.unmodifiableList(new ArrayList<>(spanFilters));
    this.traceFilters = Collections.unmodifiableList(new ArrayList<>(traceFilters));
```

processors/src/main/java/io/opentelemetry/contrib/filter/TraceDurationFilter.java
```diff
@@ -0,0 +1,44 @@
...
+  private final long thresholdNanos;
+
+  /**
+   * Creates a new {@code TraceDurationFilter}.
+   *
+   * @param thresholdMs the trace duration threshold in milliseconds; traces with wall-clock
+   *     duration strictly greater than this are considered interesting
+   */
+  public TraceDurationFilter(long thresholdMs) {
+    this.thresholdNanos = TimeUnit.MILLISECONDS.toNanos(thresholdMs);
```
jaydeluca:
```suggestion
    if (thresholdMs < 0) {
      throw new IllegalArgumentException("thresholdMs must be non-negative, got: " + thresholdMs);
    }
    this.thresholdNanos = TimeUnit.MILLISECONDS.toNanos(thresholdMs);
```

Also, same comment as `DurationSpanFilter` about using `Duration` instead of long

processors/src/main/java/io/opentelemetry/contrib/filter/DurationSpanFilter.java
```diff
@@ -0,0 +1,34 @@
...
+  private final long thresholdNanos;
+
+  /**
+   * Creates a new {@code DurationSpanFilter}.
+   *
+   * @param thresholdMs the duration threshold in milliseconds; spans with duration strictly greater
+   *     than this are considered interesting
+   */
+  public DurationSpanFilter(long thresholdMs) {
+    this.thresholdNanos = TimeUnit.MILLISECONDS.toNanos(thresholdMs);
```
jaydeluca: we should validate the input here

```suggestion
    if (thresholdMs < 0) {
      throw new IllegalArgumentException("thresholdMs must be non-negative, got: " + thresholdMs);
    }
    this.thresholdNanos = TimeUnit.MILLISECONDS.toNanos(thresholdMs);
```

Also, perhaps it would make the API a bit more intuitive if we accepted a Duration instead of a long? `new DurationSpanFilter(Duration.ofSeconds(2))` feels more intuitive than `new DurationSpanFilter(2000)`

processors/src/main/java/io/opentelemetry/contrib/filter/FilteringSpanExporter.java
```diff
@@ -0,0 +1,164 @@
...
+    if (meter != null) {
+      this.droppedSpansCounter =
+          meter
+              .counterBuilder("filtering.span.exporter.dropped")
```
jaydeluca: in the [sdk](https://opentelemetry.io/docs/specs/semconv/otel/sdk-metrics/#metric-otelsdkexporterspanexported) they have metrics that look like `otel.sdk.exporter.span.exported` and `otel.sdk.processor.span.processed`, so perhaps we should use something closer to that pattern like `otel.contrib.exporter.span.filtered` or `otel.contrib.processor.span.dropped`?  Or maybe having `contrib` doesn't make sense in this context

@trask  wdyt?

```suggestion
              .counterBuilder("otel.contrib.processor.span.filtered")
```

processors/src/main/java/io/opentelemetry/contrib/filter/FilteringSpanExporter.java
```diff
@@ -0,0 +1,164 @@
...
+              .setDescription("Number of spans dropped by the filtering span exporter")
```
jaydeluca, line 79:
```suggestion
              .setDescription("Number of spans dropped by the filtering span exporter")
              .setUnit("{span}")
```

processors/src/main/java/io/opentelemetry/contrib/filter/TraceDurationFilter.java
udaysagar2177: Good catch on the overflow — fixed with an early return. Added a test for this case too.

processors/src/main/java/io/opentelemetry/contrib/filter/FilteringSpanExporter.java
udaysagar2177: Done, added `Objects.requireNonNull` for all three parameters.

processors/src/main/java/io/opentelemetry/contrib/filter/TraceDurationFilter.java
udaysagar2177: Added both — validation and switched to `Duration`. Agreed it reads much better: `new TraceDurationFilter(Duration.ofSeconds(10))`.

processors/src/main/java/io/opentelemetry/contrib/filter/DurationSpanFilter.java
udaysagar2177: Done on both — validation added and switched to `Duration`. `new DurationSpanFilter(Duration.ofSeconds(2))` is much more intuitive.

processors/src/main/java/io/opentelemetry/contrib/filter/FilteringSpanExporter.java
udaysagar2177: Used `otel.contrib.processor.span.filtered` — aligns well with the SDK naming conventions. Happy to adjust if @trask has a different preference.

processors/src/main/java/io/opentelemetry/contrib/filter/FilteringSpanExporter.java
udaysagar2177: Done, added `.setUnit("{span}")`.

jaydeluca:
@udaysagar2177 looks like there's some formatting issues, could you run spotless to fix that up?

breedx-splk (approved):

PR #2753, opened 2026-04-09
[dynamic control] Add pipeline config initialization from file

jackshirazi:
**Description:**

Adding a method to read a pipeline initialization file in yaml (preferred) or json format.

**Existing Issue(s):**

**Testing:**

Added

**Documentation:**

Added

**Outstanding items:**

Wiring up the policy pipeline: 
- Generic: message -> provider -> policy -> policy handler -> implementer
- eg change sampling rate message -> OpampPolicyProvider -> TraceSamplingRatePolicy -> PolicyStore -> TraceSamplingRatePolicyImplementer

Steps needed for the wiring:
- configuring the pipeline
   - DONE in this PR
- providers reading policies, eg OpampPolicyProvider, FilePolicyProvider, etc
   - DONE: OpampPolicyProvider
- implementers applying policies, eg TraceSamplingRatePolicyImplementer
   - DONE: TraceSamplingRatePolicyImplementer
- policy structures (eg TraceSamplingRatePolicy) that the provider converts messages into
   - DONE: TraceSamplingRatePolicy
- registering config to policy structures (eg "trace_sampling_rate_policy" registered to TraceSamplingRatePolicy)
- initializing policy classes (eg TraceSamplingRatePolicy needs to install a custom sampler)
   - DONE TraceSamplingRatePolicy initialization
- activate configured providers (eg start OpampPolicyprovider reading from it's source)
- register implementers for policies (eg a new TraceSamplingRatePolicy is applied by a TraceSamplingRatePolicyImplementer)
- link the provider to processing policies and applying implementers

### DIFF
--- a/dynamic-control/src/main/java/io/opentelemetry/contrib/dynamic/policy/registry/PolicyInitConfig.java
+++ b/dynamic-control/src/main/java/io/opentelemetry/contrib/dynamic/policy/registry/PolicyInitConfig.java
@@ -46,10 +46,13 @@ public final class PolicyInitConfig {
   /**
    * Reads policy-init configuration based on config properties.
    *
-   * <p>YAML takes precedence over JSON when both are present.
+   * <p>YAML takes precedence over JSON when both are present. If both are present, and the YAML
+   * file is invalid, the JSON file is still ignored. If the file parsed is invalid, a warning is
+   * logged and null is returned.
    *
    * @param config OpenTelemetry config properties
-   * @return parsed init config, or null when no init-config path is configured
+   * @return parsed init config, or null when no init-config path is configured or the file is
+   *     invalid
    * @throws NullPointerException if config is null
    */
   @Nullable
--- a/dynamic-control/src/main/java/io/opentelemetry/contrib/dynamic/policy/registry/PolicyInitConfig.java
+++ b/dynamic-control/src/main/java/io/opentelemetry/contrib/dynamic/policy/registry/PolicyInitConfig.java
@@ -5,13 +5,28 @@
 
 package io.opentelemetry.contrib.dynamic.policy.registry;
 
+import io.opentelemetry.contrib.dynamic.policy.registry.json.JsonPolicyInitConfigReader;
+import io.opentelemetry.contrib.dynamic.policy.registry.yaml.YamlPolicyInitConfigReader;
+import io.opentelemetry.sdk.autoconfigure.spi.ConfigProperties;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.Files;
+import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import javax.annotation.Nullable;
 
 /** Top-level registry initialization model containing per-source mapping config. */
 public final class PolicyInitConfig {
+  static final String POLICY_INIT_CONFIG_PROPERTY_JSON =
+      "otel.java.experimental.telemetry.policy.init.json";
+  static final String POLICY_INIT_CONFIG_PROPERTY_YAML =
+      "otel.java.experimental.telemetry.policy.init.yaml";
+  private static final Logger logger = Logger.getLogger(PolicyInitConfig.class.getName());
 
   private final List<PolicySourceConfig> sources;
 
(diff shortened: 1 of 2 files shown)


dynamic-control/src/main/java/io/opentelemetry/contrib/dynamic/policy/registry/PolicyInitConfig.java
```diff
@@ -28,6 +43,49 @@ public List<PolicySourceConfig> getSources() {
     return sources;
   }
 
+  /**
+   * Reads policy-init configuration based on config properties.
+   *
+   * <p>YAML takes precedence over JSON when both are present.
+   *
+   * @param config OpenTelemetry config properties
+   * @return parsed init config, or null when no init-config path is configured
```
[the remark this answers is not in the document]
jackshirazi: valid, changed to similar

dynamic-control/src/main/java/io/opentelemetry/contrib/dynamic/policy/registry/PolicyInitConfig.java
```diff
@@ -28,6 +43,49 @@ public List<PolicySourceConfig> getSources() {
...
+        logger.log(
+            Level.WARNING,
+            "Failed to load telemetry policy init config from {0}",
+            mappingPathYaml.trim());
+        logger.log(Level.WARNING, "Policy init config read failed", e);
+        return null;
+      }
+    }
+  }
+
```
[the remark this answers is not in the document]
jackshirazi: valid, changed documentation

LikeTheSalad (approved):

dynamic-control/src/main/java/io/opentelemetry/contrib/dynamic/policy/registry/PolicyInitConfig.java
```diff
@@ -28,6 +43,52 @@ public List<PolicySourceConfig> getSources() {
...
+   *     invalid
+   * @throws NullPointerException if config is null
+   */
+  @Nullable
+  public static PolicyInitConfig readFromConfigProperties(ConfigProperties config) {
+    Objects.requireNonNull(config, "config cannot be null");
+    String mappingPathYaml = config.getString(POLICY_INIT_CONFIG_PROPERTY_YAML);
+    if (mappingPathYaml == null || mappingPathYaml.trim().isEmpty()) {
+      String mappingPathJson = config.getString(POLICY_INIT_CONFIG_PROPERTY_JSON);
+      if (mappingPathJson == null || mappingPathJson.trim().isEmpty()) {
```
trask, line 64: can we use declarative config api instead? (and rely on the java agent declarative config bridge for the system property support)

jackshirazi: I'm not sure that gives any benefit. The system property doesn't provide the pipeline configuration, it points to a file that provides pipeline configuration. We'd need to change the property to be an ugly string encoded yaml to use that. json would work because that can be a single string for the full config, but yaml would be ugly.

````
jsonconfig='{"sources": [{"kind": "opamp", "format": "jsonkeyvalue", "location": "elastic", "mappings": [ { "sourceKey": "sampling_rate", "policyType": "trace_sampling_rate_policy" }, ] } ]}'
````

```
yamlconfig=sources:\n  - kind: opamp\n    format: jsonkeyvalue\n    location: elastic\n    mappings:\n      - sourceKey: sampling_rate\n        policyType: trace_sampling_rate_policy
```

Even to define a small pipeline this is painful

The yaml is deliberately setup so that when the telemetry policy otep is accepted and we then add `policy` to declarative config, it slots right in, eg
 ```
policy:
  sources:
    - kind: opamp
      format: jsonkeyvalue
      location: elastic
      mappings:
        - sourceKey: sampling_rate
          policyType: trace_sampling_rate_policy
```

The node under policy is exactly what this is supporting. To use the declarative config api just to get the file path seems pointless?

trask: > The yaml is deliberately setup so that when the telemetry policy otep is accepted and we then add `policy` to declarative config, it slots right in, eg
> 
> ```
> policy:
>  sources:
>    - kind: opamp
>      format: jsonkeyvalue
>      location: elastic
>      mappings:
>        - sourceKey: sampling_rate
>          policyType: trace_sampling_rate_policy
> ```

does it make sense to slot it in now, even if the location under declarative config changes later when it gets an official location?

jackshirazi: We can do that for the declarative config, since this is experimental it would be fine. But how would you define the config when declarative config is not used.

```
List<DeclarativeConfigProperties> sources = config.getStructuredList("sources");
if (sources == null) {//true when not using declarative config
  //how to get the structure? read a file? use the ugly stringified file?
```

trask: could this be a declarative config only feature? (similar to rule based sampling, some of the new method instrumentation, ..)

jackshirazi: Allowing the structure to be read by failing over to how this PR does it would work. So it would look like this. I need to add the additional methods if this plan looks okay

```
PolicyInitConfig policyPipeline;
List<DeclarativeConfigProperties> sources = config.getStructuredList("sources");
if (sources == null) {//true when not using declarative config
  policyPipeline = PolicyInitConfig.readFromConfigProperties();
} else {
  policyPipeline = convertToPipeline(sources);
}
```

jackshirazi: I've done a different implementation in https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2766 (I'll close this PR when this discussion is complete or shifts to that).

I think having only declarative config supported imposes a very high burden here. You already need to include this as an extension, insisting that you must also wire everything up using declarative config in order to use this makes it a huge shift instead of the smaller one I'd like to support

trask: > I think having only declarative config supported imposes a very high burden here.

yeah, I understand. at the same time, the whole reason we (in Java at least) have been looking forward to declarative configuration is that it allows us to have a unified story for these kinds of complex configurations, instead of having lots of different config files that are pointed to by various configuration properties (e.g. having separate metric view yaml configuration file, rule-based sampling configuration file, jmx configuration file, opamp policy configuration file, methods instrumentation configuration file, etc etc..)

jackshirazi: I've made the declarative config the default in the [new PR](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2766), with support for experimental config files only if declarative config is not used.

Without the config file failover, it just forces deployments that don't use declarative config to use another coded extension to create the pipeline which seems like imposing a lot of additional pain to avoid including an option and a method

jackshirazi:
superseded by https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2766

PR #2767, opened 2026-04-17
feat(gcp-auth-extension): Support custom credentials via configuration properties

keshavdandeva:
# feat(gcp-auth-extension): Support custom credentials via configuration properties

## Description
Currently, `gcp-auth-extension` is hardcoded to use Application Default Credentials (ADC). This PR adds support for specifying custom Service Account credentials via configuration properties (file path or JSON string), while maintaining backward compatibility by falling back to ADC.

### `gcp-auth-extension`

*   **New Properties:** Added `google.cloud.credentials.path` and `google.cloud.credentials.json` options in `ConfigurableOption.java`.
*   **Lazy & Cached Resolution:** Moved credentials resolution to be lazy and cached per `ConfigProperties` instance in `GcpAuthAutoConfigurationCustomizerProvider.java` to avoid redundant I/O.
*   **Type-Safe Loading:** Uses `ServiceAccountCredentials.fromStream` instead of the obsolete generic `GoogleCredentials.fromStream` to address security concerns.
*   **Improved Error Handling:** Distinguishes between a missing file (`ConfigurationException`) and a failure to create credentials (`GoogleAuthException`).
*   **Tests:** Added unit tests for file path, JSON string, precedence rules, missing files, and malformed JSON.
*   **Docs:** Updated Javadoc to clarify that both relative and absolute paths are supported, and to reflect the new capabilities.

## Verification Results
*   Executed unit tests: `./gradlew :gcp-auth-extension:test` -> **BUILD SUCCESSFUL**

### DIFF
--- a/gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/ConfigurableOption.java
+++ b/gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/ConfigurableOption.java
@@ -56,16 +56,16 @@ enum ConfigurableOption {
 
   /**
    * Specifies the path to a Google Cloud service account JSON key file. The path can be either
-   * absolute or relative to the current working directory. Can be configured using the
-   * environment variable `GOOGLE_CLOUD_CREDENTIALS_PATH` or the system property
+   * absolute or relative to the current working directory. Can be configured using the environment
+   * variable `GOOGLE_CLOUD_CREDENTIALS_PATH` or the system property
    * `google.cloud.credentials.path`.
    */
   GOOGLE_CLOUD_CREDENTIALS_PATH("Google Cloud Credentials Path"),
 
   /**
    * Specifies the raw JSON content of a Google Cloud service account key. This is useful when
-   * credentials are not stored in a file but are available as a string. Can be configured using
-   * the environment variable `GOOGLE_CLOUD_CREDENTIALS_JSON` or the system property
+   * credentials are not stored in a file but are available as a string. Can be configured using the
+   * environment variable `GOOGLE_CLOUD_CREDENTIALS_JSON` or the system property
    * `google.cloud.credentials.json`.
    */
   GOOGLE_CLOUD_CREDENTIALS_JSON("Google Cloud Credentials JSON String");
--- a/gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/ConfigurableOption.java
+++ b/gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/ConfigurableOption.java
@@ -52,7 +52,19 @@ enum ConfigurableOption {
    * configured using the environment variable `GOOGLE_OTEL_AUTH_TARGET_SIGNALS` or the system
    * property `google.otel.auth.target.signals`.
    */
-  GOOGLE_OTEL_AUTH_TARGET_SIGNALS("Target Signals for Google Authentication Extension");
+  GOOGLE_OTEL_AUTH_TARGET_SIGNALS("Target Signals for Google Authentication Extension"),
+
+  /**
+   * Represents the Google Cloud Credentials Path option. Can be configured using the environment
+   * variable `GOOGLE_CLOUD_CREDENTIALS_PATH` or the system property `google.cloud.credentials.path`.
+   */
+  GOOGLE_CLOUD_CREDENTIALS_PATH("Google Cloud Credentials Path"),
+
+  /**
+   * Represents the Google Cloud Credentials JSON option. Can be configured using the environment
+   * variable `GOOGLE_CLOUD_CREDENTIALS_JSON` or the system property `google.cloud.credentials.json`.
+   */
+  GOOGLE_CLOUD_CREDENTIALS_JSON("Google Cloud Credentials JSON String");
 
   private final String userReadableName;
   private final String environmentVariableName;
--- a/gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/ConfigurableOption.java
+++ b/gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/ConfigurableOption.java
@@ -56,13 +56,15 @@ enum ConfigurableOption {
 
   /**
    * Represents the Google Cloud Credentials Path option. Can be configured using the environment
-   * variable `GOOGLE_CLOUD_CREDENTIALS_PATH` or the system property `google.cloud.credentials.path`.
+   * variable `GOOGLE_CLOUD_CREDENTIALS_PATH` or the system property
+   * `google.cloud.credentials.path`.
    */
   GOOGLE_CLOUD_CREDENTIALS_PATH("Google Cloud Credentials Path"),
 
   /**
    * Represents the Google Cloud Credentials JSON option. Can be configured using the environment
-   * variable `GOOGLE_CLOUD_CREDENTIALS_JSON` or the system property `google.cloud.credentials.json`.
+   * variable `GOOGLE_CLOUD_CREDENTIALS_JSON` or the system property
+   * `google.cloud.credentials.json`.
    */
   GOOGLE_CLOUD_CREDENTIALS_JSON("Google Cloud Credentials JSON String");
 
--- a/gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/ConfigurableOption.java
+++ b/gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/ConfigurableOption.java
@@ -55,15 +55,17 @@ enum ConfigurableOption {
   GOOGLE_OTEL_AUTH_TARGET_SIGNALS("Target Signals for Google Authentication Extension"),
 
   /**
-   * Represents the Google Cloud Credentials Path option. Can be configured using the environment
-   * variable `GOOGLE_CLOUD_CREDENTIALS_PATH` or the system property
+   * Specifies the path to a Google Cloud service account JSON key file. The path can be either
+   * absolute or relative to the current working directory. Can be configured using the
+   * environment variable `GOOGLE_CLOUD_CREDENTIALS_PATH` or the system property
    * `google.cloud.credentials.path`.
    */
   GOOGLE_CLOUD_CREDENTIALS_PATH("Google Cloud Credentials Path"),
 
   /**
-   * Represents the Google Cloud Credentials JSON option. Can be configured using the environment
-   * variable `GOOGLE_CLOUD_CREDENTIALS_JSON` or the system property
+   * Specifies the raw JSON content of a Google Cloud service account key. This is useful when
+   * credentials are not stored in a file but are available as a string. Can be configured using
+   * the environment variable `GOOGLE_CLOUD_CREDENTIALS_JSON` or the system property
    * `google.cloud.credentials.json`.
    */
   GOOGLE_CLOUD_CREDENTIALS_JSON("Google Cloud Credentials JSON String");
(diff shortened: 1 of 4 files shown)


jaydeluca:
@keshavdandeva  are you able to sign the CLA?

keshavdandeva:
> @keshavdandeva are you able to sign the CLA?

Yes, done

gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/ConfigurableOption.java
```diff
@@ -52,7 +52,21 @@ enum ConfigurableOption {
...
+   * `google.cloud.credentials.path`.
+   */
+  GOOGLE_CLOUD_CREDENTIALS_PATH("Google Cloud Credentials Path"),
+
+  /**
+   * Represents the Google Cloud Credentials JSON option. Can be configured using the environment
+   * variable `GOOGLE_CLOUD_CREDENTIALS_JSON` or the system property
+   * `google.cloud.credentials.json`.
+   */
+  GOOGLE_CLOUD_CREDENTIALS_JSON("Google Cloud Credentials JSON String");
```
psx95, line 72: To improve clarity, the documentation should indicate if the path should be relative or absolute. 
Current documentation comments simply state - "Represents the Google Cloud Credentials Path option", but does not explain what these options are.

gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
```diff
@@ -57,6 +63,8 @@ public class GcpAuthAutoConfigurationCustomizerProvider
 
   private static final Logger logger =
       Logger.getLogger(GcpAuthAutoConfigurationCustomizerProvider.class.getName());
+  private static final Map<ConfigProperties, GoogleCredentials> credentialsCache =
```
psx95, line 68: Does this need to be a map? The map implicates that there could be multiple `GoogleCredentials` object for authenticating to the same backend.

IIUC, the same `configProperties` are used, so the Map will only ever have a single entry.

gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
```diff
@@ -99,22 +107,62 @@ public class GcpAuthAutoConfigurationCustomizerProvider
...
+        return cachedCredentials;
+      }
+
+      GoogleCredentials resolvedCredentials = loadCredentials(configProperties);
+      credentialsCache.put(configProperties, resolvedCredentials);
+      return resolvedCredentials;
+    }
+  }
+
+  private static GoogleCredentials loadCredentials(ConfigProperties props) {
```
psx95: Nit: Rename `props` to `configProperties` to maintain consistency.

gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
```diff
@@ -99,22 +107,62 @@ public class GcpAuthAutoConfigurationCustomizerProvider
...
+    Optional<String> credsPath =
+        ConfigurableOption.GOOGLE_CLOUD_CREDENTIALS_PATH.getConfiguredValueAsOptional(props);
+    if (credsPath.isPresent()) {
+      try (FileInputStream fis = new FileInputStream(credsPath.get())) {
+        return GoogleCredentials.fromStream(fis);
```
psx95: This method is marked as obsolete citing potential security risk.

See [google-auth-library-java](https://github.com/googleapis/google-auth-library-java/blob/9ac2d4340ebc6a8582b898e97f65aeed3c1776d6/oauth2_http/java/com/google/auth/oauth2/GoogleCredentials.java#L233).

gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
```diff
@@ -99,22 +107,62 @@ public class GcpAuthAutoConfigurationCustomizerProvider
...
+    }
+
+    Optional<String> credsJson =
+        ConfigurableOption.GOOGLE_CLOUD_CREDENTIALS_JSON.getConfiguredValueAsOptional(props);
+    if (credsJson.isPresent()) {
+      try (ByteArrayInputStream bais =
+          new ByteArrayInputStream(credsJson.get().getBytes(StandardCharsets.UTF_8))) {
+        return GoogleCredentials.fromStream(bais);
+      } catch (IOException e) {
+        throw new ConfigurationException("Failed to load credentials from JSON string", e);
```
psx95: Since this error would be generated from the line `GoogleCredentials.fromStream` - I would think to wrap this and throw a `GoogleAuthException` instead.

You can add a new `Reason`:  `FAILED_CREDENTIAL_CREATION` and then re-use it in above case as well.

gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
```diff
@@ -99,22 +107,62 @@ public class GcpAuthAutoConfigurationCustomizerProvider
...
+  private static GoogleCredentials loadCredentials(ConfigProperties props) {
+    Optional<String> credsPath =
+        ConfigurableOption.GOOGLE_CLOUD_CREDENTIALS_PATH.getConfiguredValueAsOptional(props);
+    if (credsPath.isPresent()) {
+      try (FileInputStream fis = new FileInputStream(credsPath.get())) {
+        return GoogleCredentials.fromStream(fis);
+      } catch (IOException e) {
+        throw new ConfigurationException(
+            "Failed to load credentials from file: " + new File(credsPath.get()).getName(), e);
+      }
```
psx95: This handling should distinguish between 2 different scenarios: 
1. `credsPath` is invalid/wrong: `ConfigurationException`. 
2. Unable to create `GoogleCredentials` from a valid `credsPath`: `GoogleAuthException`.

gcp-auth-extension/src/test/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProviderTest.java
```diff
@@ -243,6 +248,166 @@ void testTraceCustomizerOtlpGrpc() {
     }
   }
 
+  @Test
+  void testCredentialsLoadedFromFilePath() throws IOException {
+    File tempFile = File.createTempFile("gcp-creds", ".json");
```
psx95, line 254: WDYT about adding a dummy creds file in test/resources. 
The file could contain valid looking fields to see if GoogleCredentials can be actually parsed from it.

gcp-auth-extension/src/test/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProviderTest.java
psx95, line 1: There could also be tests that exercise the case where either the file (JSON/Creds) is not found or the content in the file is malformed/invalid.

psx95 (changes requested):

gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/ConfigurableOption.java
keshavdandeva: Updated the comments

gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
keshavdandeva: You are correct that within a single auto-configuration pass, the same `ConfigProperties` instance is shared, so the map will indeed only have a single entry.

I used a map (specifically a `WeakHashMap`) to defensively support scenarios where multiple SDK instances might be created in the same JVM or sequential unit tests with different configurations, ensuring they don't conflict.

However, if you disagree and prefer to simplify it to a single cached instance, I am happy to change it back

gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
keshavdandeva: Done

gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
keshavdandeva: Thanks for the pointer. I have removed the use of `GoogleCredentials.fromStream` and switched to `ServiceAccountCredentials.fromStream` for now, as service accounts might be fine for our use case.

However, I'm wondering how we should ideally support other credential types (like external accounts or user credentials) in the future without re-implementing the auto-detection logic of the obsolete method. Do you have a recommendation on whether we should stick to Service Accounts for these properties, or if we should implement a custom detector to route to other specific load methods?

gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
keshavdandeva: Done

gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
keshavdandeva: Done

gcp-auth-extension/src/test/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProviderTest.java
keshavdandeva: Done

gcp-auth-extension/src/test/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProviderTest.java
keshavdandeva: I looked into adding a dummy file, but generating a valid private key structure without real keys proved to be complicated and prone to JSON parsing issues. I think the existing unit tests with mocks are sufficient to verify the logic, but I can try to add it if you insist.

gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
psx95: I don't think the case of multiple SDKs apply here. The extension is designed to be used with autoconfigure module which uses the `AutoConfiguredOpenTelemetrySdk`, which [sets the GlobalOpenTelemetry instance upon initialization](https://github.com/open-telemetry/opentelemetry-java/blob/f2ed2ed4abc21e3fe975f43eec20f169a34f641e/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdk.java#L37), which is a singleton.

If the user does end up using multiple `OpenTelemetrySdk` instances besides what is autoconfigured, this extension won't apply to those use-cases and such instances would need to be manually configured.

Also, could you clarify the use case around unit tests here? I don't quite get it.

psx95:
Discussed offline, will be closing this PR. 

cc: @keshavdandeva, @jsuereth

Issue #2777, opened 2026-04-22
[opamp] HTTP redirects cause havoc


breedx-splk:
If the initial URL for the opamp client points to a resource that returns a redirect (HTTP 301 or 302), the underlying okhttp client will turn around and issue a `GET` request to the redirected Location. This is not normal expected behavior, but can happen if a resource moves or if the client is misconfigured.

See https://github.com/square/okhttp/issues/6627 for additional context about this behavior.
You can reproduce this by setting the opamp endpoint url to http://cisco.com/

The end result is that the connection is considered successful and the callback `onConnect()` is called...which appears as success...even though we are not connected to a valid opamp server.

What can we do about it? I'm not sure yet, but there are a couple options:

* maybe we should always fail in the event that a redirect is returned to the POST?
* ...or maybe add an okhttp interceptor that somehow detects this 301/302 response to a POST and ensures that the following  requests are also POST (maybe all requests should always be POST?)
* ...or maybe (also?) wait until after the response is successfully decoded as protobuf before determining connection success.

PR #2778, opened 2026-04-22
[opamp] Improve error handling

breedx-splk:
Related to #2777, but probably does not resolve it.

There are two main ideas here, the main one is that the `onConnectionSuccess()` is moved into the `isSuccessful()` check and is also moved under the decoding of the server proto. With this, if the response is non-200 we don't consider this a successful connection, and if the server's response is invalid, we also do not consider this a valid connection.

There's another situation where the response body parsing can generate an ISE in the protobuf libs, and we turn that into a request failure.

There should be sufficient test coverage to handle these oddities. The semantics here are admittedly tricky/confusing, so I appreciate another set of eyes.

### DIFF
--- a/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
+++ b/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
@@ -154,8 +154,8 @@ public final class HttpRequestService implements RequestService {
     try (HttpSender.Response response = future.get(30, TimeUnit.SECONDS)) {
       getCallback().onConnectionSuccess();
       if (isSuccessful(response)) {
-        handleHttpSuccess(
-            Response.create(ServerToAgent.ADAPTER.decode(response.bodyInputStream())));
+        ServerToAgent serverToAgent = ServerToAgent.ADAPTER.decode(response.bodyInputStream());
+        handleHttpSuccess(Response.create(serverToAgent));
       } else {
         handleHttpError(response);
       }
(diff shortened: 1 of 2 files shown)


opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
```diff
@@ -152,13 +152,16 @@ private void doSendRequest() {
...
+        handleHttpSuccess(Response.create(serverToAgent));
       } else {
         handleHttpError(response);
       }
+    } catch (IllegalStateException e) {
+      getCallback().onRequestFailed(e);
+      connectionStatus.retryAfter(null);
     } catch (IOException | InterruptedException | TimeoutException e) {
       getCallback().onConnectionFailed(e);
       connectionStatus.retryAfter(null);
```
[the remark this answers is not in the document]
breedx-splk: While not inaccurate, this comment is not addressing code that was modified in this PR.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
```diff
@@ -152,13 +152,16 @@ private void doSendRequest() {
...
-            Response.create(ServerToAgent.ADAPTER.decode(response.bodyInputStream())));
+        ServerToAgent serverToAgent = ServerToAgent.ADAPTER.decode(response.bodyInputStream());
+        getCallback().onConnectionSuccess();
+        handleHttpSuccess(Response.create(serverToAgent));
       } else {
         handleHttpError(response);
       }
+    } catch (IllegalStateException e) {
+      getCallback().onRequestFailed(e);
+      connectionStatus.retryAfter(null);
```
jackshirazi: I think there are other parsing errors that could cause other exceptions that are not caught. Instead of the up front `IllegalStateException`, maybe it's better to have a catch all `Exception` as the last catch clause which does the onRequestFailed

jackshirazi (approved):
Approved as is, there's one feedback suggestion but it's fine to go with this as is, too

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/request/service/HttpRequestService.java
breedx-splk: Yeah that sounds like a good idea.

The sheer number of permutations of which exception should map into a connectionFailed and/or requestFailed under what circumstances is a bit daunting...so I like that simplifying that.

jaydeluca (approved):

PR #2802, opened 2026-04-29
baggage delegate to IncludeExcludePredicate and allow wildcards in auto-configuration

SylvainJuge:
Baggage processor already delegates to `IncludeExcludePredicate`.

When using declarative configuration:
- `include` and `exclude` configuration attributes can be used to include/exclude and directly delegate to `IncludeExcludePredicate`

When using auto-configuration
- baggage attributes are only opt-in, equivalent to only having `include` set in declarative configuration
- exact matching is being used
- the `*` value can be used as "include-all"
- no other form of wildcard matching is supported

This PR:
- makes the delegation to `IncludeExcludePredicate` applied in both configuration types
- adds the ability to use wildcards when not using declarative configuration (not limited anymore to `*`)
- refactor implementation and tests following this change
- deprecates the following constructors to push using the `include/exclude` variants.
  - `BaggageSpanProcessor(Predicate<String> baggageKeyPredicate)`
  - `BaggageLogRecordProcessor(Predicate<String> baggageKeyPredicate)`


baggage-processor/src/main/java/io/opentelemetry/contrib/baggage/processor/BaggageProcessorCustomizer.java
```diff
@@ -60,13 +53,6 @@ private static void addLogRecordProcessor(
     }
 
     // need to add before the batch log processor
-    sdkLoggerProviderBuilder.addLogRecordProcessorFirst(createBaggageLogRecordProcessor(keys));
-  }
-
-  static BaggageLogRecordProcessor createBaggageLogRecordProcessor(List<String> keys) {
-    if (keys.size() == 1 && keys.get(0).equals("*")) {
```
SylvainJuge, line 67: [for reviewer] this optimization is not really necessary as an equivalent is implemented in `GlobUtils`, we could however do this in `IncludeExcludePredicate` as a follow-up improvement.

baggage-processor/src/main/java/io/opentelemetry/contrib/baggage/processor/BaggageProcessorCustomizer.java
```diff
@@ -40,14 +40,7 @@ private static void addSpanProcessor(
     }
 
     // need to add before the batch span processor
-    sdkTracerProviderBuilder.addSpanProcessorFirst(createBaggageSpanProcessor(keys));
-  }
-
-  static BaggageSpanProcessor createBaggageSpanProcessor(List<String> keys) {
-    if (keys.size() == 1 && keys.get(0).equals("*")) {
-      return BaggageSpanProcessor.allowAllBaggageKeys();
-    }
```
SylvainJuge, line 49: [for reviewer] this optimization is not really necessary as an equivalent is implemented in `GlobUtils`, we could however do this in `IncludeExcludePredicate` as a follow-up improvement.

baggage-processor/README.md
```diff
@@ -63,32 +63,25 @@ To configure the span and log processors to copy all baggage entries during conf
...
-For example, to only copy baggage entries that match the regex `^key.+`:
+For example, to only copy baggage entries that start with `my-key` and ignore keys that end with `*-ignored`
 
 ```java
-Pattern pattern = Pattern.compile("^key.+");
-new BaggageSpanProcessor(baggageKey -> pattern.matcher(baggageKey).matches());
-new BaggageLogRecordProcessor(baggageKey -> pattern.matcher(baggageKey).matches());
+new BaggageSpanProcessor(singletonList("my-key*"), singletonList("*-ignored"));
+new BaggageLogRecordProcessor(singletonList("my-key*"), singletonList("*-ignored"));
 ```
```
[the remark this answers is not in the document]
SylvainJuge: @copilot apply changes based on [this feedback](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2802#discussion_r3160686637)

baggage-processor/src/test/java/io/opentelemetry/contrib/baggage/processor/BaggageProcessorCustomizerTest.java
```diff
@@ -218,4 +215,23 @@ void test_baggageLogRecordProcessor_adds_attributes_to_spans_when_key_filter_mat
       }
     }
   }
+
+  @Test
+  void test_baggageLogRecordProcessor_adds_attributes_to_spans_include_exclude(
+      @Mock ReadWriteLogRecord logRecord) {
```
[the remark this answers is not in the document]
SylvainJuge: @copilot apply changes based on [this feedback](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2802#discussion_r3160686672)

baggage-processor/src/main/java/io/opentelemetry/contrib/baggage/processor/BaggageLogRecordProcessor.java
```diff
@@ -21,19 +24,37 @@ public final class BaggageLogRecordProcessor implements LogRecordProcessor {
...
     this.baggageKeyPredicate = baggageKeyPredicate;
   }
 
+  /**
+   * Creates a new {@link BaggageLogRecordProcessor} that copies baggage entries with keys that pass
+   * the provided include/exclude filtering into the newly created log record, when both arguments
+   * are null or empty all baggage are included.
+   *
+   * @param included list of included attribute patterns to include
+   * @param excluded list of excluded attribute patterns to exclude
```
MikeGoldsmith:
```suggestion
   * @param included list of included baggage key patterns to include
   * @param excluded list of excluded baggage key patterns to exclude
```

baggage-processor/src/main/java/io/opentelemetry/contrib/baggage/processor/BaggageSpanProcessor.java
```diff
@@ -20,17 +23,35 @@ public final class BaggageSpanProcessor implements SpanProcessor {
...
   }
 
+  /**
+   * Creates a new {@link BaggageSpanProcessor} that copies baggage entries with keys that pass the
+   * provided include/exclude filtering into the newly created {@link
+   * io.opentelemetry.api.trace.Span}, when both arguments are null or empty all baggage are
+   * included.
+   *
+   * @param included list of included attribute patterns to include
+   * @param excluded list of excluded attribute patterns to exclude
```
MikeGoldsmith:
```suggestion
   * @param included list of included baggage key patterns to include
   * @param excluded list of excluded baggage key patterns to exclude
```

baggage-processor/src/main/java/io/opentelemetry/contrib/baggage/processor/BaggageLogRecordProcessor.java
```diff
@@ -21,19 +24,37 @@ public final class BaggageLogRecordProcessor implements LogRecordProcessor {
...
    * created log record.
+   *
+   * @return baggage log processor including all attributes
+   * @deprecated use {@code new BaggageLogRecordProcessor(Collections.singletonList("*"), null)}
+   *     instead
    */
+  @Deprecated
   public static BaggageLogRecordProcessor allowAllBaggageKeys() {
-    return new BaggageLogRecordProcessor(baggageKey -> true);
+    return new BaggageLogRecordProcessor(excluded -> true);
```
MikeGoldsmith:
```suggestion
    return new BaggageLogRecordProcessor(baggageKey -> true);
```

MikeGoldsmith (approved):
Over all looks good, thanks @SylvainJuge 👍🏻 

Some docs notes:

The switch from exact matching (`keys::contains`) to a glob pattern match via `IncludeExcludePredicate` is a behavioural change for auto-configuration users. Config values containing glob-special characters (`?`, `[`, `]`) would now be interpreted as patterns instead of literal key names. I think it's unlikely to affect anyone but worth noting it.

Deprecating the `Predicate<String>` constructor removes the ability to express some filtering logic options (eg `startsWith`). The glob pattern constructor covers common cases but isn't a full replacement. I think we should document the trade-off in the deprecation notice.

breedx-splk (approved):
Looks good to me. I'm not sure if this was previously covered, but if both the include and exclude are provided, and an entry matches both, which one wins out? Is there a test that shows/covers that?

SylvainJuge:
>I'm not sure if this was previously covered, but if both the include and exclude are provided, and an entry matches both, which one wins out? Is there a test that shows/covers that?

I've opened https://github.com/open-telemetry/opentelemetry-java/pull/8371 to add a dedicated test for the case where both include and exclude match the provided value, [the spec](https://github.com/open-telemetry/opentelemetry-configuration/blob/804c36185ee9e6e12c61e889e49eb9487568a568/CONTRIBUTING.md#properties-requiring-pattern-matching) is quite clear in this case: when both match then the exclusion has priority and thus the value is excluded.

jaydeluca:
@MikeGoldsmith just wanted to check in to see if you had any other concerns or if this is now good to go? thanks!

MikeGoldsmith (approved):
I think only thing I'd like to see is a deprecation notice for `Predicate<String>` to describe how it's different from the new one. They don't work the same, so we should let people know.

SylvainJuge:
> I think only thing I'd like to see is a deprecation notice for `Predicate<String>` to describe how it's different from the new one. They don't work the same, so we should let people know.

@MikeGoldsmith, can you clarify what and where would you like to see this added ?
- in the code examples in the readme
- in this PR description
- in the release notes
- somewhere in the javadoc comments that isn't yet covered
- all of the above

baggage-processor/src/main/java/io/opentelemetry/contrib/baggage/processor/BaggageLogRecordProcessor.java
```diff
@@ -21,17 +24,35 @@ public final class BaggageLogRecordProcessor implements LogRecordProcessor {
...
    * Creates a new {@link BaggageLogRecordProcessor} that copies all baggage entries into the newly
    * created log record.
+   *
+   * @return baggage log processor including all attributes
+   * @deprecated use {@code new BaggageLogRecordProcessor(Collections.singletonList("*"), null)}
+   *     instead
    */
+  @Deprecated
   public static BaggageLogRecordProcessor allowAllBaggageKeys() {
     return new BaggageLogRecordProcessor(baggageKey -> true);
```
MikeGoldsmith: Instead of marking this deprecated, should we just update to the suggested value? The `allowAllBaggageKeys` is a nice user helper.

```suggestion
  */
public static BaggageLogRecordProcessor allowAllBaggageKeys() {
    return new BaggageLogRecordProcessor(Collections.singletonList("*"), null);
}
```

baggage-processor/src/main/java/io/opentelemetry/contrib/baggage/processor/BaggageLogRecordProcessor.java
```diff
@@ -21,17 +24,35 @@ public final class BaggageLogRecordProcessor implements LogRecordProcessor {
   private final Predicate<String> baggageKeyPredicate;
 
   /**
-   * Creates a new {@link BaggageLogRecordProcessor} that copies only baggage entries with keys that
-   * pass the provided filter into the newly created log record.
+   * @deprecated Use {@link #BaggageLogRecordProcessor(Collection, Collection)} instead.
```
MikeGoldsmith: As we're redirecting users to use the new constructor, I think we should include a note to say not all existing predicates can be migrated and must be based on string glob patterns.

MikeGoldsmith (approved):
This looks good to me, thanks @SylvainJuge.

Sorry the ask for docs wasn't clear. I've left a note for where I think it would be useful. In the changelog / PR desc would also be good for easier discoverability.

baggage-processor/src/main/java/io/opentelemetry/contrib/baggage/processor/BaggageLogRecordProcessor.java
SylvainJuge: covered in [61ef1cb](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2802/commits/61ef1cbc69ed068b5404b186a38c70c6044d31f8)

baggage-processor/src/main/java/io/opentelemetry/contrib/baggage/processor/BaggageLogRecordProcessor.java
SylvainJuge: good point, this covered in [97844ff](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2802/commits/97844ff27ce06917598810c542752c9d6e2ee8e2)

MikeGoldsmith:
Thanks @SylvainJuge 👍🏻

MikeGoldsmith (approved):

jaydeluca (approved):

PR #2803, opened 2026-04-29
make span stacktrace ignore inferred spans by default

SylvainJuge:
The span stacktrace span processor captures the current thread stack trace when the span ends.

However, the inferred spans span processor produces spans that are not executed when the span is active as they are artificially generated from sampling data.

When the span stacktrace span processor is executed before the inferred spans, the inferred spans are naturally excluded, however when the stacktrace span processor is after then we have to explicitly exclude such spans. We can argue that this may be a mis-configuration in the span processor pipeline, but this creates an extra burden on user.

This is why the span stacktrace processor was created with the ability to provide a custom filter. The only known usage and filter implementation in that I'm aware of is in the Elastic (EDOT) distribution which is using the `is_inferred` span attribute to filter inferred spans.

With this change:
- inferred-spans are ignored by default for span stacktraces, hence making it easier for users
- we remove the need to use the `is_inferred` span attribute to detect span attributes, we can rely on the scope name (in a sense we trade one implementation detail to another)
- this exclusion is being tested in unit tests

Possible follow-up tasks
- remove the `is_inferred` span attribute in inferred spans: it is not part of semconv and should not be used by anyone after this change is effective
- deprecate the `filter` option in span stacktrace span processor.

### DIFF
--- a/span-stacktrace/build.gradle.kts
+++ b/span-stacktrace/build.gradle.kts
@@ -15,7 +15,6 @@ dependencies {
 
   compileOnly("io.opentelemetry:opentelemetry-sdk-extension-autoconfigure")
   compileOnly("io.opentelemetry:opentelemetry-sdk-extension-autoconfigure-spi")
-  compileOnly("io.opentelemetry:opentelemetry-sdk-extension-autoconfigure")
   compileOnly("io.opentelemetry:opentelemetry-sdk-extension-incubator")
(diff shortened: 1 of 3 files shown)


JonasKunz (approved):
Seems reasonable to me.

jaydeluca (approved):

PR #2810, opened 2026-05-01
[opamp-client] Remove CompletableFuture usage

breedx-splk:
[See this comment ](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2790#issuecomment-4344961356)about the gummy bears upgrade change for context. 

The tools were quick to point out that the async ritual and use of CompletableFuture wasn't really that helpful because the caller immediately turned around and called `.get(30)` or whatever on the future anyway....effectively making it synchronous.

(slopvomited with codex)

### DIFF
--- a/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/HttpSender.java
+++ b/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/HttpSender.java
@@ -9,11 +9,11 @@ import java.io.Closeable;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
-import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeoutException;
 
 public interface HttpSender {
 
-  CompletableFuture<Response> send(BodyWriter writer, int contentLength);
+  Response send(BodyWriter writer, int contentLength) throws IOException, TimeoutException;
 
   interface BodyWriter {
     void writeTo(OutputStream outputStream) throws IOException;
--- a/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/HttpSender.java
+++ b/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/HttpSender.java
@@ -9,11 +9,10 @@ import java.io.Closeable;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
-import java.util.concurrent.TimeoutException;
 
 public interface HttpSender {
 
-  Response send(BodyWriter writer, int contentLength) throws IOException, TimeoutException;
+  Response send(BodyWriter writer, int contentLength) throws IOException;
 
   interface BodyWriter {
     void writeTo(OutputStream outputStream) throws IOException;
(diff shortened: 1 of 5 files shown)


opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/OkHttpSender.java
```diff
@@ -39,36 +40,40 @@ private OkHttpSender(String url, OkHttpClient client) {
...
-              }
-            });
+  private Response doSendRequest(Request request) throws IOException {
+    okhttp3.Response response =
+        client
+            .newBuilder()
+            .callTimeout(REQUEST_TIMEOUT)
+            .build()
+            .newCall(request)
+            .execute();
```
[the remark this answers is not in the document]
breedx-splk: yeah agreed this sucks. will fix.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/HttpSender.java
```diff
@@ -9,11 +9,11 @@
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
-import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeoutException;
 
 public interface HttpSender {
 
-  CompletableFuture<Response> send(BodyWriter writer, int contentLength);
+  Response send(BodyWriter writer, int contentLength) throws IOException, TimeoutException;
```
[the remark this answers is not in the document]
breedx-splk: yup, on it.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/connectivity/http/HttpSender.java
```diff
@@ -9,11 +9,10 @@
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
-import java.util.concurrent.CompletableFuture;
 
 public interface HttpSender {
 
-  CompletableFuture<Response> send(BodyWriter writer, int contentLength);
+  Response send(BodyWriter writer, int contentLength) throws IOException;
```
LikeTheSalad, line 15: Do you think it's possible to use `CompletableResultCode` from OTel Java? I understand the argument that the current okhttp implementation can live without a "future" result; it's just that, since this is an interface, I think the "future result" allows for more options for both the implementer and the consumer of this API, and it's a shame that our min android version doesn't support it tbh.

This is not a blocker, though. A future result is a nice-to-have for this use case.

breedx-splk: I understand what you're saying....so lemme see what that looks like.

breedx-splk: Ok, so I can't find a straightforward way to use the `CompletableResultCode` from java core, because that class doesn't carry data. It's simply a success/failure, and can only hold a throwable in the failure case...it can't carry the actual `Response`.

LikeTheSalad: Got it. Still, thank you for checking 👍

LikeTheSalad (approved):
Cheers!

jackshirazi (approved):

jaydeluca (approved):

PR #2818, opened 2026-05-05
Fix jfr-events tests

jaydeluca:
While reviewing #2739 i was trying to run the tests locally and they were not working due to a bug in the gradle config beacuse Kotlin DSL requires the is-prefixed form for boolean properties.

Once the tests were fixed to run, they were failing, so i updated the assertions accordingly

### DIFF
--- a/jfr-events/build.gradle.kts
+++ b/jfr-events/build.gradle.kts
@@ -23,7 +23,7 @@ tasks {
 
     // Disabled due to https://bugs.openjdk.java.net/browse/JDK-8245283
     configure<JacocoTaskExtension> {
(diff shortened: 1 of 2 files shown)


laurit (approved):

jfr-events/build.gradle.kts
```diff
@@ -23,7 +23,7 @@ tasks {
 
     // Disabled due to https://bugs.openjdk.java.net/browse/JDK-8245283
```
marschall, line 24: This is fixed in JDK 15 and backported to 11.0.11. Is this still needed?

PR #2834, opened 2026-05-13
Remove mateusz from resource providers.

breedx-splk:
Removing @mateuszrzeszutek from the component owners for the `resource_providers` module.

We will welcome you back from any org, just say the word.


trask (approved):
👋

PR #2835, opened 2026-05-14
feat(ibm-mq-metrics): report queue manager uptime (#2823)

### LINKED ISSUE #2823: [ibm-mq-metrics] Report queue manager uptime
### Component(s)

ibm-mq-metrics

The IBM MQ metrics integration reports a heartbeat metric. This metric returns 1 when the queue manager is up.
This metric could be more useful if it reported the cumulative number of seconds the queue manager is up, so we can catch any restarts.

Adopt a new metric to report uptime of the queue manager.

Optionally, can be a followup issue: deprecate and remove the heartbeat metric.

_No response_

### Additional context

_No response_

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

harshitt13:
**Description:**

Feature addition - This PR introduces a new metric to report the cumulative uptime of the queue manager in seconds, providing better visibility into queue manager restarts compared to the static heartbeat metric. 

* Added `ibm_mq.queue_manager.uptime` metric (Gauge, seconds) calculating uptime from PCF attributes `CMQCFC.MQCACF_Q_MGR_START_DATE` and `CMQCFC.MQCACF_Q_MGR_START_TIME`.
* Eliminated timezone drift by securely anchoring the parsed IBM MQ local time to `ZoneOffset.UTC`.
* Enabled the new metric in the default configuration.
* Deprecated the existing `heartbeat` metric per the issue discussion, keeping it active with `@SuppressWarnings` during the transition period so users aren't immediately broken.

**Existing Issue(s):**

Fixes #2823

**Testing:**

* Updated `QueueManagerMetricsCollectorTest` to inject mock start dates/times into the PCF response and verify the new gauge emission.
* Created `MessageBuddyTest` to explicitly verify that the uptime calculation remains perfectly stable regardless of the collector host's underlying JVM default timezone. 
* Executed `./gradlew :ibm-mq-metrics:generateWeaver` and verified existing module tests pass cleanly.

**Documentation:**

* Updated `model/metrics.yaml` and `config.yml` to define and enable the new `ibm_mq.queue_manager.uptime` metric.
* Added standard Java `@Deprecated` annotations and Javadoc for the `heartbeat` metric.

**Outstanding items:**

* The `heartbeat` metric is currently deprecated but still emitting. It will need to be completely removed in a future major release.


ibm-mq-metrics/model/metrics.yaml
```diff
@@ -528,11 +528,24 @@ groups:
     metric_name: ibm.mq.heartbeat
     stability: development
     brief: "Queue manager heartbeat"
+    description: |
+      Deprecated. Use ibm.mq.queue_manager.uptime instead.
```
atoulme: I would deprecate in a separate PR, no rush.

ibm-mq-metrics/model/metrics.yaml
```diff
@@ -528,11 +528,24 @@ groups:
...
     unit: "1"
     attributes:
       - ref: ibm.mq.queue.manager
         requirement_level: required
+  - id: ibm.mq.queue_manager.uptime
+    type: metric
+    metric_name: ibm.mq.queue_manager.uptime
+    stability: development
+    brief: "Queue manager uptime"
+    description: "Cumulative time the Queue Manager has been up"
```
atoulme: If this is a cumulative value, can it be a counter?

ibm-mq-metrics/model/metrics.yaml
harshitt13: I've reverted the deprecation changes in this PR.

ibm-mq-metrics/model/metrics.yaml
harshitt13: I've updated the model to use a counter instrument, regenerated the Weaver classes, and updated the collector to use add().

atoulme:
Please rebase, we have changed the way we record metrics to use a MetricProducer.

harshitt13:
> Please rebase, we have changed the way we record metrics to use a MetricProducer.

Rebased and refactored the implementation to utilize the new MetricProducer pattern as requested. 
Ready for re-review.

ibm-mq-metrics/src/main/java/io/opentelemetry/ibm/mq/metrics/MetricProducer.java
```diff
@@ -668,6 +670,36 @@ public void recordIbmMqHeartbeat(long value, Attributes attributes) {
...
+  public void addIbmMqQueueManagerUptime(long value, Attributes attributes) {
+    long cumulativeValue =
+        this.counterIbmMqQueueManagerUptime.compute(
+            attributes,
+            (k, v) -> {
+              if (v == null) {
+                return value;
+              } else {
+                return v + value;
+              }
```
breedx-splk: nitpick:
```suggestion
            (k, v) -> v == null ? value : v + value
```

breedx-splk: Oh, is the value coming in a cumulative value already? If so, doesn't this math make it incorrect?

I'm thinking we should just take the value that we're given and report it as a gauge. I'm not sure that he guidance to use a counter is correct, merely because it's monotonic. Since we're reporting a "current value" of the measured uptime, wouldn't a gauge be better? 

Also, what if the manager restarted between collects, causing its uptime to drop?

harshitt13: You are right. Because the value is already an absolute cumulative value. Adding it to the previous value (`v + value`) across collection intervals was a logic error.

I agree that an asynchronous Gauge is correct here. By simply reporting the latest calculated value as a Gauge, it also perfectly handles the restart scenario you mentioned: if the queue manager restarts between collections, the new `start_time` will simply result in a smaller reported value, which a Gauge handles natively without issue.

I will push a fix to remove the addition logic so it just reports the latest measured value.

breedx-splk (approved):
Had a pending review open for a while, so sorry about that. Looking good now. Cheers and thanks again for the contribution.

PR #2837, opened 2026-05-15
[chore] Automate changelog generation via Python LLM pipeline (#2787)

### LINKED ISSUE #2787: [Tooling] Automate changelog generation
Right now as part of the [release process](https://github.com/open-telemetry/opentelemetry-java-contrib/blob/main/RELEASING.md), we run a few steps locally and have to do some manual work. It would be great to automate all of this so that we can kick off a github action that will generate the changelog, have it reviewed by copilot or equivalent to touch it up, and then present it for review by the release captain.

harshitt13:
**Description:**

Feature addition - Following maintainer direction and consensus from the issue thread, this PR ports the robust, Python-based LLM drafting pipeline directly from the main `opentelemetry-java-instrumentation` repository. This automates the Release Captain's workload at release time without requiring contributors to manually write YAML fragments or alter their standard workflow.

* **Architecture Update & Porting:**
    * Ported the 5 core pipeline scripts (`fetch.py`, `classify.py`, `merge.py`, `draft-release-notes.py`, and `rules.md`) into `.github/scripts/draft-release-notes/`.
    * Updated the internal target variables within the scripts to correctly target `open-telemetry/opentelemetry-java-contrib`.
* **Release Automation (`.github/workflows/draft-release-notes.yml`):**
    * Introduced a new manual `workflow_dispatch` trigger intended for the Release Captain.
    * Configured the job to run in the `protected` environment to securely leverage the `COPILOT_GITHUB_TOKEN` secret provisioned by the administrators.
    * Executes the end-to-end Python drafting script which fetches PRs, classifies them using the Copilot CLI, and merges them directly into the `CHANGELOG.md` `## Unreleased` block.
* **Security & OpenSSF Compliance:**
    * All third-party GitHub Actions (`actions/checkout` and `actions/setup-python`) are strictly pinned to immutable commit SHAs.
    * Explicitly set `persist-credentials: false` during checkout to resolve `zizmor` / GitHub Advanced Security alerts regarding token persistence.

**Existing Issue(s):**

Resolves #2787 

**Testing:**

* Performed local syntax and execution validation on the ported Python scripts.

**Documentation:**

* Added `.github/scripts/draft-release-notes/rules.md` which acts as the single source of truth and inline documentation for the LLM classification schema and categorization logic.

**Outstanding items:**

* **Initial Test Run:** The workflow and repository secrets are fully configured. A maintainer with access to the `protected` environment will need to manually trigger the `Draft Release Notes` workflow via `workflow_dispatch` to perform an initial live end-to-end test of the classification outputs.


harshitt13:
@jaydeluca, could you please review the PR whenever you have a moment?

breedx-splk (changes requested):
I hate `chloggen`. It's unnecessarily complicated and makes the process of submitting PRs worse.

Who is asking for this? I don't think the maintainers want this.

jack-berg:
For things like this, please open an issue to gather consensus from the codeowners (approvers + maintainers) before going and opening a PR.

harshitt13:
@jack-berg 
Understood! I've converted this to a draft. I will move the conversation over to the issue so we can figure out how the maintainers actually want to handle this.

.github/workflows/draft-release-notes.yml
```diff
@@ -0,0 +1,40 @@
+name: Draft Release Notes
+
+on:
+  workflow_dispatch:
```
jaydeluca, line 4: Worth adding a concurrency group here, matching upstream, to prevent overlapping runs from racing on `CHANGELOG.md`/the git push.

```suggestion
on:
  workflow_dispatch:

concurrency:
  group: draft-release-notes
```

.github/workflows/draft-release-notes.yml
```diff
@@ -0,0 +1,40 @@
...
+
+permissions:
+  contents: write
+  pull-requests: write
```
jaydeluca: Granting `contents: write` + `pull-requests: write` directly on the default token isn't this repo's convention, and upstream explicitly avoids it: "not using secrets.GITHUB_TOKEN since pull requests from that token do not run workflows" — PRs opened with the default token won't trigger CI on themselves.

`prepare-release-branch.yml`/`prepare-patch-release.yml` narrow this to `contents: read` at the top and grant `contents: write` only at the job level (for the git push), then mint a short-lived app token for the actual `gh pr create` call — see the suggestions on the job block and the last step below.

```suggestion
permissions:
  contents: read
```

.github/workflows/draft-release-notes.yml
```diff
@@ -0,0 +1,40 @@
...
+
+jobs:
+  draft-notes:
+    runs-on: ubuntu-latest
+    environment: protected
```
jaydeluca, line 20: Add the fork guard and job-level write permission, matching `prepare-release-branch.yml`/`prepare-patch-release.yml`:

```suggestion
  draft-notes:
    if: github.repository == 'open-telemetry/opentelemetry-java-contrib'
    permissions:
      contents: write  # for git push to PR branch
    runs-on: ubuntu-latest
    environment: protected
```

.github/workflows/draft-release-notes.yml
```diff
@@ -0,0 +1,40 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```
jaydeluca: This should be updated to `v6.0.2` to be consistent with the others in this repo.

```suggestion
        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
```

.github/workflows/draft-release-notes.yml
```diff
@@ -0,0 +1,40 @@
...
+        with:
+          fetch-depth: 0
+          persist-credentials: false
```
jaydeluca, line 26: This needs to come out. With `persist-credentials: false`, the checkout won't leave a token in the git config, so the `git push` in the final step won't have anything to authenticate with once `contents: write` is granted at the job level. None of the other workflows in this repo that push from the checked-out ref set this.

```suggestion
```

.github/workflows/draft-release-notes.yml
```diff
@@ -0,0 +1,40 @@
...
+
+      - name: Set up Python 3.11
+        uses: actions/setup-python@0a5c61591373683505ea898e09a3ea4f39ef2b9c # v5.0.0
+        with:
+          python-version: '3.11'
+
+      - name: Install GitHub CLI
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y gh
```
jaydeluca: `gh` ships preinstalled on `ubuntu-latest` runners — no other workflow in this repo (or upstream's) installs it explicitly. This step can be dropped.

```suggestion
```

.github/workflows/draft-release-notes.yml
```diff
@@ -0,0 +1,40 @@
...
+      - name: Install Copilot CLI
+        run: npm install -g @githubnext/github-copilot-cli
+
+      - name: Authenticate Copilot CLI
+        run: github-copilot-cli auth "${{ secrets.COPILOT_GITHUB_TOKEN }}"
+
+      - name: Draft release notes
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: python3 .github/scripts/draft-release-notes/draft-release-notes.py
```
jaydeluca, line 42: `classify.py` invokes the `copilot` binary directly (`subprocess.run(["copilot", "-p", ...])`). `@githubnext/github-copilot-cli` is a different, long-deprecated package whose binary is named `github-copilot-cli` (a completely different CLI surface — no `-p`/`--allow-all-tools`/`--model`), so as written `copilot` is never on `PATH` and the classify step fails with `FileNotFoundError`.

Also, nothing after this commits, pushes, or opens a PR — `draft-release-notes.py` only rewrites `CHANGELOG.md` on disk (per its own docstring: "edit CHANGELOG.md directly... git diff CHANGELOG.md"), so today the change is made in the ephemeral runner workspace and discarded when the job ends.

```suggestion
      - name: Install Copilot CLI
        run: |
          npm ci --prefix .github/scripts/copilot-cli
          echo "$GITHUB_WORKSPACE/.github/scripts/copilot-cli/node_modules/.bin" >> "$GITHUB_PATH"

      - name: Draft release notes
        env:
          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
          COPILOT_GITHUB_TOKEN: ${{ secrets.COPILOT_GITHUB_TOKEN }}
        run: python3 .github/scripts/draft-release-notes/draft-release-notes.py

      - name: Use CLA approved bot
        run: .github/scripts/use-cla-approved-bot.sh

      - uses: actions/create-github-app-token@1b10c78c7865c340bc4f6099eb2f838309f1e8c3 # v3.1.1
        id: otelbot-token
        with:
          app-id: ${{ vars.OTELBOT_APP_ID }}
          private-key: ${{ secrets.OTELBOT_PRIVATE_KEY }}

      - name: Create pull request against main
        env:
          # not using secrets.GITHUB_TOKEN since pull requests from that token do not run workflows
          GH_TOKEN: ${{ steps.otelbot-token.outputs.token }}
        run: |
          message="Draft release notes"
          body="Automated draft of the \`## Unreleased\` section of \`CHANGELOG.md\`. Review the diff to recover any hand-written content that should be retained."
          branch="otelbot/draft-release-notes-${GITHUB_RUN_ID}"

          git checkout -b $branch
          git commit -a -m "$message"
          git push --set-upstream origin $branch
          gh pr create --title "$message" \
                       --body "$body" \
                       --base main
```

One piece this needs outside of this diff: `.github/scripts/copilot-cli/package.json` pinning `@github/copilot`, mirroring `opentelemetry-java-instrumentation`'s — that's what actually provides the `copilot` binary `classify.py` shells out to.

jaydeluca (commented):
Compared this against the equivalent `draft-release-notes` workflow in `opentelemetry-java-instrumentation` and this repo's own `prepare-release-branch.yml`/`prepare-patch-release.yml` conventions. The Python pipeline (`fetch.py`/`classify.py`/`merge.py`/`rules.md`) is fine, but the issues below are all in the workflow YAML, which was written independently rather than ported.

.github/scripts/draft-release-notes/merge.py
```diff
@@ -0,0 +1,184 @@
...
+CHANGELOG = Path("CHANGELOG.md")
+
+SECTION_ORDER = [
+    ("breaking", "### ⚠️ Breaking changes to non-stable APIs"),
+    ("deprecations", "### 🚫 Deprecations"),
+    ("new-javaagent", "### 🌟 New javaagent instrumentation"),
+    ("new-library", "### 🌟 New library instrumentation"),
+    ("enhancements", "### 📈 Enhancements"),
+    ("bug-fixes", "### 🛠️ Bug fixes"),
+]
```
jaydeluca: This `SECTION_ORDER` — and the matching `section` enum in `classify.py` (see comment there) — writes `opentelemetry-java-instrumentation`'s changelog taxonomy (`breaking`/`deprecations`/`new-javaagent`/`new-library`/`enhancements`/`bug-fixes`) straight into `CHANGELOG.md`. But this repo's changelog has never used change-type headers — every `###` header in `CHANGELOG.md`'s history groups by module/component name instead (`### Disk buffering`, `### AWS X-Ray SDK support`, `### Azure resources - New 🌟`), with a `- New 🌟` suffix marking brand-new modules rather than a separate section.

Recommend matching that convention directly: group entries by module name (derived from the changed file paths) and replace `new-javaagent`/`new-library` with a single `new-module` label, applied together with the `- New 🌟` suffix the same way this repo already marks brand-new modules. `breaking`/`deprecations`/`enhancements`/`bug-fixes` still make sense as labels within a module's entries, but they need to stop being top-level section headers.

This is bigger than this constant, though — it means `classify.py` needs to emit a `module` per PR alongside `section` (inferred from the diff's file paths), and this grouping/rendering logic needs to key off `module` first. Flagging here since this is the file that actually writes to `CHANGELOG.md`.

.github/scripts/draft-release-notes/classify.py
```diff
@@ -0,0 +1,451 @@
...
+
+
+VALID_SECTIONS = {
+    "breaking",
+    "deprecations",
+    "new-javaagent",
+    "new-library",
+    "enhancements",
+    "bug-fixes",
+    None,
```
jaydeluca, line 68: Following from the `merge.py` recommendation: swap `new-javaagent`/`new-library` for a single `new-module` value here (and in the schema docstring above, lines 19-21).

```suggestion
VALID_SECTIONS = {
    "breaking",
    "deprecations",
    "new-module",
    "enhancements",
    "bug-fixes",
    None,
}
```

That's only the enum, though — grouping by module name also needs a `module` field added to `decision.json`'s schema (inferred from the diff's file paths) so `merge.py` has something to group by. This enum change alone isn't sufficient.

.github/scripts/draft-release-notes/rules.md
```diff
@@ -0,0 +1,170 @@
...
+
+If an unlinked summary bullet at the top of Deprecations already covers
+the rename, do not add a duplicate PR-linked bullet.
+
+## New javaagent / library instrumentation
+
+Only for a brand-new module under `instrumentation/<name>/javaagent/**`
+or `instrumentation/<name>/library/**` — new `build.gradle.kts`, new
+sources, and a new `settings.gradle.kts` entry. Renames or extractions
+do not qualify.
```
jaydeluca: Nearly every section of this file assumes a javaagent/bytecode-instrumentation repo, which contrib isn't — there's no `javaagent`/`muzzle` directory anywhere in this repo, no `@Advice` usage, and no `v3-preview`/`SemconvStability` references. This section checks for `instrumentation/<name>/javaagent/**` / `instrumentation/<name>/library/**` paths that can never exist here, so it can never fire — dead weight in the prompt at best, and it pairs with the `new-javaagent`/`new-library` section values in `classify.py`/`merge.py` (see those comments) that have no meaning in this repo's changelog. Should be rewritten around how this repo actually introduces new modules: a new top-level module directory with its own `build.gradle.kts` and `settings.gradle.kts` entry, classified as `new-module` and tagged `- New 🌟` in the changelog per existing convention.

Other agent-only criteria scattered through the rest of the file worth sweeping in the same pass:
- Line 34: `javaagent-extension-api` as a non-stable-module example (breaking changes)
- Lines 48-49, 83, 86-87, 152-154: `otel.instrumentation.common.v3-preview` / `otel.semconv-stability.opt-in=...` gating — this config property doesn't exist in contrib
- Line 93: muzzle `versionRange` as a bug-fix criterion — no muzzle in this repo
- Line 136: `SemconvStability.emitStable…` — same, doesn't exist here
- Lines 139-140: "fixing an agent deadlock" as an example — there's no agent
- Lines 141-142: `@Advice` inline vs indy, helper-class exposure to the application class loader — bytecode-weaving concepts specific to the javaagent

Probably worth a holistic pass rather than spot edits — this doc likely needs rewriting example-by-example against contrib's actual module set (samplers, resource providers, disk buffering, JMX scraping, etc.).

.github/scripts/draft-release-notes/fetch.py
```diff
@@ -0,0 +1,893 @@
...
+# A file is "user-facing runtime" iff it sits under `/src/main/` and is not
+# inside a testing, smoke-test, or docs module. Everything else (tests, docs,
+# build scripts, CI config, gradle wrappers, top-level properties, etc.) has
+# no `/src/main/` in its path by construction and is therefore non-runtime
+# without needing an explicit allow-list.
+_NON_USER_FACING_PREFIXES = (
+    "smoke-tests/",
+    "smoke-tests-otel-starter/",
+    "instrumentation-docs/",
+)
```
jaydeluca, line 210: None of these three paths exist in this repo (no `smoke-tests/`, `smoke-tests-otel-starter/`, or `instrumentation-docs/` dirs here), so this filter is currently a no-op carried over from `opentelemetry-java-instrumentation`. Meanwhile contrib has its own non-user-facing test-harness/sample dirs that aren't covered — `jmx-scraper/test-app/`, `jmx-scraper/test-webapp/`, `cel-sampler/testapp/`, `static-instrumenter/test-app/`, and `example/` — so a PR touching only one of those is currently (mis)treated as touching user-facing `src/main` by this fast-path check.

```suggestion
_NON_USER_FACING_PREFIXES = (
    "jmx-scraper/test-app/",
    "jmx-scraper/test-webapp/",
    "cel-sampler/testapp/",
    "static-instrumenter/test-app/",
    "example/",
)
```

.github/scripts/draft-release-notes/fetch.py
```diff
@@ -0,0 +1,893 @@
...
+            actual_range,
+            candidates,
+            breaking_prs,
+            deprecation_prs,
+            refetch=refetch,
+        )
+        relative_bundle = bundle_dir.relative_to(REPO_ROOT).as_posix()
+        warn(f"Prepared changelog bundle at {relative_bundle}")
+
+    sys.stdout.write(render_draft_output(breaking_prs, deprecation_prs, candidates))
```
jaydeluca: `render_draft_output()` (defined at line 789) is called unconditionally here and its output is written to stdout / the CI log every run. It hardcodes the same instrumentation-flavored headers as `merge.py`'s `SECTION_ORDER` (`### 🌟 New javaagent instrumentation`, `### 🌟 New library instrumentation`) plus an extra `### 🧰 Tooling` header that isn't even in `classify.py`'s `VALID_SECTIONS`, so it always renders empty. It doesn't feed `CHANGELOG.md` (that's `merge.py --splice`'s job), so it's log noise rather than a correctness bug, but it looks like leftover scaffolding from before the fetch/classify/merge split. `fetch_labeled_prs`/`breaking_prs`/`deprecation_prs` are still needed elsewhere (`prepare_bundle` writes them into the manifest), but `render_labeled_prs` would become unused if this call is dropped. Worth removing once the section-taxonomy question above is settled.

jaydeluca (commented):
This pass looks at the classification pipeline itself: several pieces (the CHANGELOG section taxonomy, some of the classification rules, and a couple of filters) were ported from `opentelemetry-java-instrumentation` without being adapted to how this repo actually organizes its changelog and modules.

.github/workflows/draft-release-notes.yml
```diff
@@ -0,0 +1,71 @@
...
+  group: draft-release-notes
+
+permissions:
+  contents: read
+
+jobs:
+  draft-notes:
+    if: github.repository == 'open-telemetry/opentelemetry-java-contrib'
+    permissions:
+      contents: write  # for git push to PR branch
```
jaydeluca, line 16: Once a job sets its own `permissions:` block, it replaces the workflow-level block for that job — unlisted scopes become `none`, they don't inherit the workflow-level `contents: read`. This job only has `contents: write`, but the `Draft release notes` step below runs `draft-release-notes.py` → `fetch.py`, which shells out to:
- `gh pr list --label "breaking change" / "deprecation" ...`
- `gh pr view <n> --json ...,comments,reviews,...` (once per PR)
- `gh issue view <n> ...` (once per linked issue)

All of those need `pull-requests: read` / `issues: read` on the token driving `GH_TOKEN` (`secrets.GITHUB_TOKEN`, set a few lines down).

```suggestion
    permissions:
      contents: write  # for git push to PR branch
      pull-requests: read  # for `gh pr list`/`gh pr view` in fetch.py
      issues: read  # for `gh issue view` in fetch.py
```

.github/scripts/draft-release-notes/fetch.py
```diff
@@ -0,0 +1,894 @@
...
+        f"- deprecated_added: `{candidate.deprecated_added}`",
+        f"- deprecated_removed: `{candidate.deprecated_removed}`",
+        f"- patch: [{bundle_ref}/patch.diff]({bundle_ref}/patch.diff)",
+        f"- meta: [{bundle_ref}/meta.json]({bundle_ref}/meta.json)",
+        f"- body: [{bundle_ref}/body.md]({bundle_ref}/body.md)",
+        "",
+    ])
+
+
+def render_draft_output(
```
jaydeluca: `render_draft_output()` (and `render_labeled_prs()` at line 326, and `format_subject_as_entry()` at line 235) look like leftovers from an earlier commit-subject-based design. None of them are called from `do_draft()` or anywhere else — the live output path today is `classify.py` → `decision.json` → `merge.py --splice`, which never touches these functions. Confirmed with a repo-wide grep: no caller outside this file.

This also makes the two `fetch_labeled_prs(actual_range, "breaking change" / "deprecation")` calls in `do_draft()` pure overhead — extra `gh pr list` calls whose results only get written into `manifest.json` (`breaking_change_prs`, `deprecation_prs`) and are never read back by `classify.py` or `merge.py`.

Suggest deleting `render_draft_output`, `render_labeled_prs`, `format_subject_as_entry`, and the `fetch_labeled_prs` calls + `breaking_prs`/`deprecation_prs` plumbing, since breaking/deprecation detection is already handled per-PR via `count_deprecated_deltas` and the LLM's `section` field.

.github/scripts/draft-release-notes/merge.py
```diff
@@ -0,0 +1,202 @@
...
+    errors = 0
+    for d in decisions:
+        pr = d.get("pr")
+        decision = d.get("decision")
+        section = d.get("section")
+        
+        if decision == "omit":
+            continue
+            
+        reason: str | None = None
```
jaydeluca, line 120: This comment reads backwards relative to the current schema. `classify.py`'s own `decision.json` schema docstring (lines 16-28 there) only ever produces a `bullet` key — there's no `summary` key anywhere in `classify.py`. So `d.get("summary")` is always `None` today and every entry falls through to `bullet`, which works, but the comment describes `bullet` as the *legacy* fallback when it's actually the only field the pipeline currently writes.

Simplify to `d.get("bullet")` (and doing the same at line 167)

jaydeluca (changes requested):

jaydeluca:
Hey @harshitt13, I want to thank you for taking this on and for working through the various rounds of code review. I really appreciate the time you put into it and your willingness to iterate on the feedback.

As I’ve been working on the next release, I started using the implementation and came across a number of additional things that I’d like to change. There have also been several iterations on the upstream example in the instrumentation repo that I want to incorporate.

At this point, I think it will be more efficient for me to work through those changes directly as I prepare the release rather than continuing to iterate through additional review rounds on this PR. So I’m going to close this PR in favor of a fresh branch that I’ve started.

Again, thanks for taking the time to work on this and for sticking with the review process. I appreciate your effort here, and apologies that I’m going to take a different approach for the release.

PR #2856, opened 2026-05-21
Add zizmor workflow

jaydeluca:


.github/workflows/zizmor.yml
```diff
@@ -0,0 +1,27 @@
...
+  zizmor:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      security-events: write
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          filter: 'tree:0'
```
trask, line 23: What does this do?

trask (approved):

.github/workflows/zizmor.yml
jaydeluca: Copied it from the .NET repo, but it makes the fetch more efficient and only downloads objects/trees as needed.

[reference comment](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/4419/files#r3276148773)

[Docs say:](https://git-scm.com/docs/partial-clone?utm_source=chatgpt.com)

> Partial clone allows us to avoid downloading such unneeded objects in advance during clone and fetch operations and thereby reduce download times and disk usage. Missing objects can later be "demand fetched" if/when needed.

PR #2867, opened 2026-05-26
[OpAMP client] Added health reporting support

robsunday:
Add support for reporting component health to the OpAMP server.
The following actions have to be taken to report agent's health:
1) Enable health reporting by calling `OpampClientBuilder#enableHealthReporting(initialHealth)` - this will ensure that client will send health report in the first message.
2) New health report should be sent every time agent failure/recovery happen by calling `OpampClient#setHealth(ComponentHealth health)`. Original agent start time should be preserved.


opamp-client/src/main/java/io/opentelemetry/opamp/client/OpampClientBuilder.java
```diff
@@ -362,6 +363,19 @@ public OpampClientBuilder enableEffectiveConfigReporting() {
...
+   * status report, as explained <a
+   * href="https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#agenttoserverhealth">here</a>.
+   *
+   * @return this
+   */
+  @CanIgnoreReturnValue
+  public OpampClientBuilder enableHealthReporting() {
+    capabilities = capabilities | AgentCapabilities.AgentCapabilities_ReportsHealth.getValue();
+    return this;
+  }
```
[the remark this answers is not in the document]
robsunday: Initial value is now set

opamp-client/src/main/java/io/opentelemetry/opamp/client/OpampClientBuilder.java
```diff
@@ -389,6 +417,9 @@ public OpampClient build(OpampClient.Callbacks callbacks) {
     if (effectiveConfigState == null) {
       effectiveConfigState = createEffectiveConfigNoop();
     }
```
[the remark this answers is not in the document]
robsunday: Handled differently now

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/impl/OpampClientImpl.java
```diff
@@ -144,6 +149,14 @@ public void setRemoteConfigStatus(@Nonnull RemoteConfigStatus remoteConfigStatus
     }
   }
 
+  @Override
+  public void setHealth(@Nonnull ComponentHealth health) {
+    if (!Objects.equals(state.health.get(), health)) {
+      state.health.set(health);
+      addFieldAndSend(Field.HEALTH);
+    }
+  }
```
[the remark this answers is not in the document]
robsunday: Agreed, but this is consistent with current approach and is out of scope for this PR. Separate refactoring should be done to make states thread safe.

opamp-client/src/main/java/io/opentelemetry/opamp/client/OpampClient.java
```diff
@@ -35,6 +36,13 @@ static OpampClientBuilder builder() {
    */
   void setRemoteConfigStatus(RemoteConfigStatus remoteConfigStatus);
 
+  /**
+   * Sets the current Agent health which will be sent in the next agent to server request.
+   *
+   * @param health The new component health.
+   */
+  void setHealth(ComponentHealth health);
```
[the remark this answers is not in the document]
robsunday: It is consistent with other methods

opamp-client/src/main/java/io/opentelemetry/opamp/client/OpampClientBuilder.java
```diff
@@ -415,6 +449,13 @@ public opamp.proto.EffectiveConfig get() {
     };
   }
 
+  private static ComponentHealth createInitialHealth() {
```
LikeTheSalad: If I understand correctly, when `setHealth` is not called, this will be the default value. Which by default leaves the status as "unhealthy". Also, since the value is never null, even when `enableHealthReporting()` is not called, a health field will be sent.

If the above is correct, I think we should provide `null` as the default to avoid sending misleading health status reports.

opamp-client/src/main/java/io/opentelemetry/opamp/client/OpampClientBuilder.java
```diff
@@ -362,6 +366,19 @@ public OpampClientBuilder enableEffectiveConfigReporting() {
...
 
+  /**
+   * Adds the ReportsHealth capability to the Client so that the Server expects the Client's health
+   * status report, as explained <a
+   * href="https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#agenttoserverhealth">here</a>.
+   *
+   * @return this
+   */
+  @CanIgnoreReturnValue
+  public OpampClientBuilder enableHealthReporting() {
```
LikeTheSalad: Given that an initial health value must be present when this flag is enabled, I was thinking about enabling it as part of the builder's `setHealth` call, and maybe making this function private. That way, the API helps users properly comply with the spec as the easiest path to take.

The way you did it is more consistent with other flags, but it seems to me that so far this is the first one that requires sending an initial data in the first agent to server message, so it might make the usage less error-prone to combine the health setter with the flag being enabled. In other cases, such as for `enableRemoteConfig`, the flag helps prompt the server to offer configurations to the agent, but just by enabling that flag, the agent doesn't have to send anything immediately, so keeping the flag enabling function public makes more sense in that case imo.

LikeTheSalad (commented):
Thank you, @robsunday 🙏 

It looks good overall, and it follows the same patterns as other fields. Thank you for keeping that consistency. I only have some comments regarding the default value for this new field.

opamp-client/src/main/java/io/opentelemetry/opamp/client/OpampClientBuilder.java
robsunday: Passing `null` was actually my first attempt but I had to make `State.Health` class extending `ObservableState` to allow `null` initial value.

There is one issue with `ObservableState`. Its javadoc says:

> Implementations of state that cannot be stored in memory

In my opinion this is not strictly the case for `State.Health`, so that's why I decided to switch to extending `InMemoryState` which disallows `null` values.

I agree with what you said here - "unhealthy" status is not an optimal solution, but on the other hand it forces developer to set the healthy status consciously on startup time when health reporting is enabled.

To be honest I'd also prefer 'null' solution, but would you be OK with relaxing 'null' validations in InMemeoryState ? Or maybe it would be to introduce one more base class for states?

opamp-client/src/main/java/io/opentelemetry/opamp/client/OpampClientBuilder.java
robsunday: Yes, I did it that way to be consistent with other states and I agree that your recommendation makes sense, however it requires using 'null' for initial value of ComponentHealth, as recommended [above](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2867#discussion_r3310847869)

opamp-client/src/main/java/io/opentelemetry/opamp/client/OpampClientBuilder.java
LikeTheSalad: Thanks for the detailed explanation, I understand it better now.

> would you be OK with relaxing 'null' validations in InMemeoryState ?

I'm not closed to this option. Right now, the APIs are in a very early stage and the lib is not stable, so it should be the right time to add these kinds of changes. I've thought about a couple of alternatives, such as creating another implementation that allows nullables, or passing "Optional" to InMemoryState, but they don't fully convince me.

For now, I don't want to block you with these API design issues, and it seems like the simplest option is to make `InMemoryState` nullable, so I'm up for going down that path. So please go ahead, and if you happen to run into compiler warnings when doing so for places that might currently rely on the `@Nonnull` annotation added to `InMemoryState.get`, please replace those calls with `State.mustGet` to avoid the compiler warnings.

opamp-client/src/main/java/io/opentelemetry/opamp/client/OpampClientBuilder.java
LikeTheSalad: Got it. Please take a look at my recent response in the other thread.

robsunday: I actually added a parameter to the `enableHealthReporting` and got rid of setHealth from builder class. It is more meaningful to me, but I'm also ok to use `setHealth`.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/impl/OpampClientImpl.java
```diff
@@ -130,20 +135,28 @@ public void close() {
...
-    if (!state.remoteConfigStatus.get().equals(remoteConfigStatus)) {
+    if (!state.remoteConfigStatus.mustGet().equals(remoteConfigStatus)) {
       state.remoteConfigStatus.set(remoteConfigStatus);
       addFieldAndSend(Field.REMOTE_CONFIG_STATUS);
     }
   }
 
+  @Override
+  public void setHealth(@Nonnull ComponentHealth health) {
+    if (!Objects.equals(state.health.get(), health)) {
```
robsunday: [for reviewer] I think this needs to be refactored because it is not really thread safe. Other thread may manipulate the state between `state.health.get()` `and state.health.set()` calls.
I'm thinking about adding `boolean update(@Nonnull T value)` method to State interface. It would do atomic get and set, and return `true` if value was updated, so `OpampClientImpl` can call `addFieldAndSend` if value was changed.

LikeTheSalad: Sounds good. I just think it'd be a bit strange to have 2 methods that set the state's value. Maybe the atomic update could be done in the existing `set`, and change its signature to return a boolean.

LikeTheSalad (approved):
Thanks again, @robsunday 🙏

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/impl/OpampClientImpl.java
robsunday: Done!

LikeTheSalad: Cheers! I already approved your changes, so we just have to wait for one of the repo maintainers to have a look.

robsunday:
Thanks for quick reviews @LikeTheSalad !

robsunday:
@jackshirazi - could you please review?

jackshirazi:
setHealth() can send health without ReportsHealth capability

robsunday:
I added capability verification, also for remote config status reporting

laurit (approved):

PR #2875, opened 2026-05-27
Adjust jmxmetrics, jmxscraper to accept char[] password, and clear after use

jack-berg:


jack-berg:
Not worth it since we have to read password from config as string, and convert the password to string to obtain the jmx connection

PR #2877, opened 2026-05-27
[disk-buffering] Add guard for non-numeric files

jaydeluca:
**Description:**

`purgeExpiredFilesIfAny` was calling `Long.parseLong(...` with no guard, which could throw `NumberFormatException` for non-numeric files. The `fileToCacheFile` method already had a pattern check so made this method consistnt with that approach.

Also a small update for the `clear()` method, just adding a more descriptive error message instead of a generic NPE if a directory is unreachable.


jack-berg (approved):

LikeTheSalad (approved):
Thank you 🙏

Issue #2868, opened 2026-05-27
[dynamic control][Meta] Phase 2 steps


jackshirazi:
### Component(s)

dynamic-control

### Is your feature request related to a problem? Please describe.

The initial POC implementation is essentially complete (#2546). The next steps needed to make it more fully featured and aligned with the spec are listed here.

- [x] Documentation on how to use the extension  #2885
- [x] Enabling the extension to be used with declarative config as well as autoconfiguration (needs a ComponentProvider and a DeclarativeCustomizerProvider) #2967 
- [x] Per feedback change TelemetryPolicy to be interface (or abstract?), with a special instance for deleted elements #2943 #2944 #2948
- [x] Align to spec - policy identity is incomplete. Spec requires every policy to have an id and name #2926 #2942 #2949
- [x] TelemetryPolicy should retain the source as, when comparing, the sources are prioritized, so eg a change from OpAMP source cannot be overridden except from another OpAMP source change, but a change from a file can be overridden, etc #2970 #2972 #2973 #2974 #3004 
- [x] refactor TelemetryPolicy implementations to enforce some abstract methods that get called from equals and hashCode or whatever is required for comparison #2960 #2968
- [ ] Align to spec, support [acceptable message structure](https://github.com/open-telemetry/opentelemetry-specification/pull/4738#issuecomment-4766896657) #3037 
- [ ] Align to spec [clarification](https://github.com/open-telemetry/opentelemetry-specification/issues/5218)
- [ ] Add further available policies
  - [ ] trace sampling percentage & ratio #3050 #3070 #3071 
  - [ ] disabling instrumentations
  - [ ] disabling all instrumentations
  - [ ] stop exporting per signal
  - [ ] change log level
  - [ ] enable inferred spans
  - [ ] change opamp polling interval
- [ ] Redo how DelegatingSampler works to handle more than one sampler, eg both probability and rule-based.
  - [ ] Then add a rule-based sampler policy to dynamic control
- [ ] Add further provider implementations #2975 #2987 #3010 #3038 #3039 
  - [ ] Add provider implementation for File
  - [ ] Add provider implementation for HTTP #3040 
- [ ] Align to spec (current implementation has drifted in some areas) #2957
- [x] ~reconsider parsing per [suggestion](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2881#discussion_r3463086780)~ - reviewed and it's a different approach but has downsides, as well as some simplification. Won't do
[...]

PR #2881, opened 2026-05-27
[dynamic control] add initializing from declarative config

jackshirazi:
**Description:**

Sorry for the change set being a little larger than normal, but I thought it was better to combine all the declarative config support changes together, rather than adding them piecemeal. I think it may be easier to understand the whole than the parts

This contains
- a `DeclarativeConfigurationCustomizerProvider` that reads top-level `telemetry_policy/development` from declarative config and routes it into `PolicyInit`
- a declarative `ComponentProvider` for `telemetry_policy/development` to add the delagating sampler - initially for the rate sampling policy, but will later make more generic (includes a TODO for that)
- Changes to PolicyInit to support the declarative config and also make it a one-shot execution so that a later/repeated/accidental auto-config pass does not accidentally re-use stale declarative config

**Existing Issue(s):**

part of #2868

**Testing:**

Unit tests added, plus a manual end-to-end test using declarative config was performed 

**Documentation:**

TBD see #2868

**Outstanding items:**

see #2868


jackshirazi:
@jack-berg I mentioned a while back that the documentation for this type of declarative config addition was minimal, so would love for you to review the implementation here.

LikeTheSalad (approved):

jack-berg:
I'm struggling to visualize the UX for this. 

From what I can tell:

- You include this dynamic control jar on your class path.
- If declarative config is used `TelemetryPolicyDeclarativeCustomizerProvider` is invoked, which looks for the presence of top level `.telemetry_policy/development`.
- If not present, do nothing.
- If present, try to initialize:
  - Parse PolicyInitConfig from the node
  - Modify the OpenTelemetryConfiguration model to set the tracer provider sampler to be `telemetry_policy/development`, and some other properties set (sources, resource_attributes, otel.resource.attributes - not sure what these do what its a code smell).
- TelemetryPolicyComponentProvider sees the sampler set to `telemetry_policy/development`, and returns a sampler with a dynamic implementation wired up to the policy

If that's about right, then it seems convoluted. Maybe necessarily so based on the hooks available today, but if so, maybe the answer is to go and fix the hooks so this flows more cleanly.

In my head, a user should opt into dynamic control with a config file something like:
```
file_format: "1.0"
tracer_provider:
  sampler: ..
meter_provider: ...
logger_provider: ...
telemetry_policy/development: ...
```

When the `.telemetry_policy/development` is present, the telemetry policy machinery initializes, and ultimately gets a reference to the OpenTelemetrySdk instance returned by declarative config. From there, it can listen for changes to the policy (however those happen) and call update methods on the OpenTelemetrySdk instance. 

I.e. 
- if a policy changes that results in a new sampler config, call `OpenTelemetrySdk.getSdkTracerProvider().setSampler(newSampler)`
- if a policy changes that results loggers being turned off or on, call `OpenTelemetrySdk.getSdkLoggerProvider().setLoggerConfigurator(newLoggerConfigurator)`
- etc

Caveat: I'm not currently doing any dynamic config tasks. Just trying to help and make sure that we're not trying to solve this problem strictly through the tools we already have. We can change / add tools if needed.

jackshirazi:
> I'm struggling to visualize the UX for this.

The [README](https://github.com/open-telemetry/opentelemetry-java-contrib/tree/main/dynamic-control#readme) has the details

>   some other properties set (sources, resource_attributes, otel.resource.attributes - not sure what these do what its a code smell).

Some components need specific resource items at initialization. In the current particular case an OpAMP client needs to know both the service name and the environment

> If that's about right, then it seems convoluted. Maybe necessarily so based on the hooks available today, but if so, maybe the answer is to go and fix the hooks so this flows more cleanly.

There are two different considerations

1. Getting this extension working now with the current hooks
2. Making the hooks better to make the implementation nicer

Please keep these separate. I want this extension functioning then we can evolve it. (Note it is already fully functioning for autoconfiguration, this particular PR enables it to function in the same way for declarative config).
 
> In my head, a user should opt into dynamic control with a config file something like:
> 
> ```
> file_format: "1.0"
> tracer_provider:
>   sampler: ..
> meter_provider: ...
> logger_provider: ...
> telemetry_policy/development: ...
> ```

This is exactly how it works with this PR.

> When the `.telemetry_policy/development` is present, the telemetry policy machinery initializes, and ultimately gets a reference to the OpenTelemetrySdk instance returned by declarative config. From there, it can listen for changes to the policy (however those happen) and call update methods on the OpenTelemetrySdk instance.
> 
> I.e.
> 
> * if a policy changes that results in a new sampler config, call `OpenTelemetrySdk.getSdkTracerProvider().setSampler(newSampler)`
> * if a policy changes that results loggers being turned off or on, call `OpenTelemetrySdk.getSdkLoggerProvider().setLoggerConfigurator(newLoggerConfigurator)`
> * etc

These don't currently work. Sampler can't be set there, and even if you added it, it's currently final at initialization. Logging is much worse, you need to install your preferred logging framework at exactly the right point, not too early (or you hit cyclic dependencies) nor too late (it just won't install correctly), and definitely not dynamically currently

I agree this is the way it *should* work. We can open a separate issue on changing the SDK to make it work, but that's really extensive changes even for these two features, and there are many more to consider

What I'm hoping for here in review terms is whether the declarative config implementation is the correct way for what needs to be done here, ie 

- parse the `telemetry_policy/development` to initialize the policy pipeline
- install the required sampler needed to support dynamically changing sampling rate

dynamic-control/src/main/java/io/opentelemetry/contrib/dynamic/TelemetryPolicyDeclarativeCustomizerProvider.java
```diff
@@ -0,0 +1,138 @@
...
+      return;
+    }
+    Map<String, Object> additionalProperties = model.getAdditionalProperties();
+    Object telemetryPolicy = additionalProperties.remove(TELEMETRY_POLICY_TOP_LEVEL_KEY);
+    if (telemetryPolicy == null) {
+      return;
+    }
+
+    JsonNode telemetryPolicyNode = MAPPER.valueToTree(telemetryPolicy);
+    PolicyInitConfig initConfig = JsonNodePolicyInitConfigParser.parse(telemetryPolicyNode);
```
jack-berg, line 78: If you annotate PolicySourceConfig with the jackson /@JsonProperty annotations, you can just call MAPPER.convertValue for a one line conversion between Map<String, Object> and PolicySourceConfig

dynamic-control/src/main/java/io/opentelemetry/contrib/dynamic/TelemetryPolicyDeclarativeCustomizerProvider.java
```diff
@@ -0,0 +1,138 @@
...
+          return model;
+        });
+  }
+
+  private static void registerTopLevelTelemetryPolicy(OpenTelemetryConfigurationModel model) {
+    if (model == null || model.getAdditionalProperties() == null) {
+      return;
+    }
+    Map<String, Object> additionalProperties = model.getAdditionalProperties();
+    Object telemetryPolicy = additionalProperties.remove(TELEMETRY_POLICY_TOP_LEVEL_KEY);
```
jack-berg, line 65: Code might be a little simpler if you verify this is a map here and error / exit early if not

dynamic-control/src/main/java/io/opentelemetry/contrib/dynamic/TelemetryPolicyComponentProvider.java
```diff
@@ -0,0 +1,181 @@
...
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import javax.annotation.Nullable;
+
+/** Declarative component that bootstraps top-level telemetry policy wiring. */
+@AutoService(ComponentProvider.class)
+public final class TelemetryPolicyComponentProvider implements ComponentProvider {
```
jack-berg: Consider naming this to reflect that its the telemetry policy _sampler_ component provider. If this is the pattern you go with, I suspect there will end up being other `telemetry_policy/development` components (maybe exporters, processors, etc). Will want to differentiate them.

dynamic-control/src/main/java/io/opentelemetry/contrib/dynamic/TelemetryPolicySamplerComponentProvider.java
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+  @Override
+  public Sampler create(DeclarativeConfigProperties config) {
+    logger.log(
+        Level.INFO,
+        "Dynamic control extension has been loaded by the agent via declarative config");
+    ConfigProperties bridgedConfig =
+        new SystemPropertyFallbackConfigProperties(
+            new DeclarativeConfigPropertiesBridgeBuilder().build(config));
```
jack-berg, line 45: I'm not familiar with the DeclarativeConfigPropertiesBridgeBuilder. Double check this matches the semantics you need and doesn't have things that are instrumentation specific.

Since create is called with the [data you define](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2881/changes#diff-1b72ad81de2f1d7f4925e121595ceb70be30550ffe0ed045045f703925cdbb37R99-R102), it seems a strange to have to use a bridge rather than just reconstructing the data previously set.

jack-berg:
> Please keep these separate.

👍 

> Sampler can't be set there,

Just because we haven't made it mutable yet. I don't think there is any fundamental limitation from making it mutable like the logger / tracer / meter configurators. 

> Logging is much worse,

I'm not following. Updating the otel log SDK with the set of loggers that should be enabled / disabled is an independent problem from installing / initializing your preferred log framework.

> What I'm hoping for here in review terms is whether the declarative config implementation is the correct way for what needs to be done here, ie
> 
>     parse the telemetry_policy/development to initialize the policy pipeline
>     install the required sampler needed to support dynamically changing sampling rate

I won't claim there is a correct way at this point, but it does look like you've managed to make it work despite subpar hooks. 

See recent comments for thoughts. None are blocking. Just things to consider since I'm not an owner of this component.

dynamic-control/src/main/java/io/opentelemetry/contrib/dynamic/TelemetryPolicyDeclarativeCustomizerProvider.java
jackshirazi: It's more extensive needing similar changes of Jackson creators/properties to PolicyInitConfig, PolicySourceConfig, PolicySourceMappingConfig and some other changes. But useful to consider for a future cleanup, I've added the suggestion to the todo list in the meta issue. Thanks

dynamic-control/src/main/java/io/opentelemetry/contrib/dynamic/TelemetryPolicySamplerComponentProvider.java
jackshirazi: DeclarativeConfigPropertiesBridgeBuilder can bridge either declarative config nodes into ConfigProperties, or
in its build(AutoConfiguredOpenTelemetrySdk) path, fall back to SDK autoconfigure ConfigProperties when declarative config is not available.

I'm wary of changing this to a local construct since DeclarativeConfigPropertiesBridgeBuilder seems to be the supported way to provide config to instrumentation

dynamic-control/src/main/java/io/opentelemetry/contrib/dynamic/TelemetryPolicyComponentProvider.java
jackshirazi: good suggestion, applied

dynamic-control/src/main/java/io/opentelemetry/contrib/dynamic/TelemetryPolicyDeclarativeCustomizerProvider.java
jackshirazi: good suggestion, applied

jackshirazi:
> Just because we haven't made it mutable yet. I don't think there is any fundamental limitation from making it mutable like the logger / tracer / meter configurators.

I'll open an issue in the SDK with changes suggested

> > Logging is much worse,
> 
> I'm not following. Updating the otel log SDK with the set of loggers that should be enabled / disabled is an independent problem from installing / initializing your preferred log framework.

In the Elastic distribution we install log4j as the preferred logging framework. When you do that you also have to initialize it, otherwise weird things happen. If you do that too early, you get a class circularity error (and in fact when too early is changes with different SDK versions because the class circularity is with MethodHandles and is painful). If you leave it too late, log4j doesn't correctly install into the SDK and you get a sort of partial logging.

dynamic-control/src/main/java/io/opentelemetry/contrib/dynamic/TelemetryPolicyDeclarativeCustomizerProvider.java
```diff
@@ -0,0 +1,138 @@
...
+  private static final Logger logger =
+      Logger.getLogger(TelemetryPolicyDeclarativeCustomizerProvider.class.getName());
+  private static final ObjectMapper MAPPER = new ObjectMapper();
+
+  @Override
+  public void customize(DeclarativeConfigurationCustomizer customizer) {
+    customizer.addModelCustomizer(
+        model -> {
+          registerTopLevelTelemetryPolicy(model);
+          return model;
```
jack-berg: Should comment or have registerTopLevelTelemetryPolicy return a model - this works because the `with*` methods mutate the model instance rather than returning new immutable instances. We have discussed changing them to be proper with-ers (i.e. immutable) instances, which would break this.

dynamic-control/src/main/java/io/opentelemetry/contrib/dynamic/TelemetryPolicySamplerComponentProvider.java
jack-berg: But this isn't instrumentation. Is falling back to SDK autoconfigure ConfigProperties meaningful here given that this code path is only activated when you're using declalrative config?

dynamic-control/build.gradle.kts
```diff
@@ -23,8 +23,9 @@ dependencies {
 
   compileOnly("io.opentelemetry:opentelemetry-sdk-extension-autoconfigure")
   compileOnly("io.opentelemetry:opentelemetry-sdk-extension-autoconfigure-spi")
-  compileOnly("io.opentelemetry:opentelemetry-sdk-extension-declarative-config")
+  compileOnly("io.opentelemetry:opentelemetry-sdk-extension-declarative-config:1.63.0-alpha")
```
jack-berg, line 27: Shouldn't need this anymore. (Also applies below)

```suggestion
  compileOnly("io.opentelemetry:opentelemetry-sdk-extension-declarative-config")
```

jack-berg (approved):
Seems reasonable enough. I'm not a code owner and have just scanned this, so take the approval with a grain of salt

dynamic-control/src/main/java/io/opentelemetry/contrib/dynamic/TelemetryPolicyDeclarativeCustomizerProvider.java
jackshirazi: Thanks, good point! applied

dynamic-control/src/main/java/io/opentelemetry/contrib/dynamic/TelemetryPolicySamplerComponentProvider.java
jackshirazi: Currently useful for the underlying properties access. I'll leave as is for now, it is easily changed in the future

dynamic-control/build.gradle.kts
jackshirazi: still needed on this branch, I've made a note to remove from main

jackshirazi:
moved change set to https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2967 (unless someone knows how to authorize cursor as co-author)

PR #2883, opened 2026-05-28
gcp-resources: add response body limit

jack-berg:


gcp-resources/src/main/java/io/opentelemetry/contrib/gcp/resource/GcpMetadataConfig.java
```diff
@@ -146,13 +152,17 @@ private String fetchAttribute(String attributeName) {
...
         InputStream input = connection.getInputStream();
         try (BufferedReader reader =
             new BufferedReader(new InputStreamReader(input, StandardCharsets.UTF_8))) {
-          return reader.readLine();
+          char[] buffer = new char[MAX_METADATA_VALUE_CHARS];
+          int n = reader.read(buffer, 0, buffer.length);
+          return n > 0 ? new String(buffer, 0, n) : null;
         }
       }
     } catch (IOException ignore) {
```
psx95, line 167: When adding the read connect timeouts, I think the catch block should explicitly check for `SocketTimeoutException` and log a warning. 

Might be better to place the `connection.getResponseCode()` and `connection.getInputStream()` in individual try catch blocks to explicitly log if it was a read timeout or connect timeout. 

I'm hoping if this becomes a frequent issue, the users can inform us about the exact cause.

gcp-resources/src/main/java/io/opentelemetry/contrib/gcp/resource/GcpMetadataConfig.java
```diff
@@ -146,13 +152,17 @@ private String fetchAttribute(String attributeName) {
...
+      connection.setConnectTimeout(CONNECT_TIMEOUT_MS);
+      connection.setReadTimeout(READ_TIMEOUT_MS);
       connection.setRequestProperty("Metadata-Flavor", "Google");
       if (connection.getResponseCode() == 200
           && "Google".equals(connection.getHeaderField("Metadata-Flavor"))) {
         InputStream input = connection.getInputStream();
         try (BufferedReader reader =
             new BufferedReader(new InputStreamReader(input, StandardCharsets.UTF_8))) {
-          return reader.readLine();
+          char[] buffer = new char[MAX_METADATA_VALUE_CHARS];
```
psx95, line 156: Not sure if this required.  Would `readLine` really be that inefficient?

gcp-resources/src/main/java/io/opentelemetry/contrib/gcp/resource/GcpMetadataConfig.java
```diff
@@ -27,6 +27,12 @@ final class GcpMetadataConfig {
   static final GcpMetadataConfig DEFAULT_INSTANCE = new GcpMetadataConfig();
 
   private static final String DEFAULT_URL = "http://metadata.google.internal/computeMetadata/v1/";
+  // google-auth-library-java uses 500 ms connect timeout (with up to 3 retries) for its metadata
+  // ping and sets no read timeout; 1 s is generous for an attribute fetch on the link-local
+  // metadata server.
+  private static final int CONNECT_TIMEOUT_MS = 1_000;
+  private static final int READ_TIMEOUT_MS = 1_000;
```
psx95: While I think these timeouts make sense (since we are only doing simple gets), I am wondering if it would be better to expose them as optional configuration.

gcp-resources/src/main/java/io/opentelemetry/contrib/gcp/resource/GcpMetadataConfig.java
```diff
@@ -27,6 +27,12 @@ final class GcpMetadataConfig {
   static final GcpMetadataConfig DEFAULT_INSTANCE = new GcpMetadataConfig();
 
   private static final String DEFAULT_URL = "http://metadata.google.internal/computeMetadata/v1/";
+  // google-auth-library-java uses 500 ms connect timeout (with up to 3 retries) for its metadata
```
psx95: Question: Why does auth-library matter here? It does not seem to be a dependency of this component.

psx95 (commented):
Hi @jack-berg ,
Thanks for the PR, adding a connect and read timeout makes sense. I'm just wondering if something specific prompted this change.

gcp-resources/src/main/java/io/opentelemetry/contrib/gcp/resource/GcpMetadataConfig.java
jack-berg: Its a reference for another library that interacts with these same APIs to inform the decisions here

gcp-resources/src/main/java/io/opentelemetry/contrib/gcp/resource/GcpMetadataConfig.java
jack-berg: This is all about limiting the size of the payload read into memory in the event the server is malicious. I don't think readLine would impose any limits if there's a huge payload all on one line. 

Its unlikely, since we're connecting to google servers, but its just a bit more cautious. (I don't think there's technically anything that guarantees that the server at `http://metadata.google.internal/computeMetadata/v1/` is googles in the case that the network / DNS has been compromised.

breedx-splk:
@jack-berg do you wanna come back and wrap this one up? The line termination comment seems legit, the rest looks good to me if @psx95 agrees. Thanks!

gcp-resources/src/main/java/io/opentelemetry/contrib/gcp/resource/GcpMetadataConfig.java
```diff
@@ -146,13 +152,17 @@ private String fetchAttribute(String attributeName) {
...
       connection.setRequestProperty("Metadata-Flavor", "Google");
       if (connection.getResponseCode() == 200
           && "Google".equals(connection.getHeaderField("Metadata-Flavor"))) {
         InputStream input = connection.getInputStream();
         try (BufferedReader reader =
             new BufferedReader(new InputStreamReader(input, StandardCharsets.UTF_8))) {
-          return reader.readLine();
+          char[] buffer = new char[MAX_METADATA_VALUE_CHARS];
+          int n = reader.read(buffer, 0, buffer.length);
+          return n > 0 ? new String(buffer, 0, n) : null;
```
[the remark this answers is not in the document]
psx95: I don't think the values returned by the metadata server are returned with trailing newline or carriage return characters, but adding a defensive check to strip away trailing newline characters would technically make sure the updated code's behavior is unchanged.

psx95:
> @jack-berg do you wanna come back and wrap this one up? The line termination comment seems legit, the rest looks good to me if @psx95 agrees. Thanks!

Yeah, I think this is a good idea, but I do think having some way to externally control/configure the timeout might be a better. Just like the `gcp-auth-extension`, we could support an Optional config:  `GOOGLE_OTEL_RESOURCE_DETECTOR_TIMEOUT` - which would accept timeout in ms.

By default it will be enabled and be 1 second as proposed in this PR, but will be user configurable - 0ms would mean the current behavior (no timeout).

Adding support for this could be tackled in a separate PR and I'm happy to do it, but I would not want to ship the next version without it.

gcp-resources/src/main/java/io/opentelemetry/contrib/gcp/resource/GcpMetadataConfig.java
jack-berg: Decided to pull these out in https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2883/commits/e2766d6433672a536d75a3c5ee7c164682b430fc. Configurable timeouts are problematic in a few ways:

- Hard to discover / use
- Add complexity which may very well never get used
- Inconsistent with other network based resource detectors

Rather than haggle over the correct timeout or increase scope by adding configuration, I'm withdrawing the change.

gcp-resources/src/main/java/io/opentelemetry/contrib/gcp/resource/GcpMetadataConfig.java
jack-berg: Moot after https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2883/commits/e2766d6433672a536d75a3c5ee7c164682b430fc

gcp-resources/src/main/java/io/opentelemetry/contrib/gcp/resource/GcpMetadataConfig.java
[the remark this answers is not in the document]
jack-berg: Resolved in https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2883/commits/e2766d6433672a536d75a3c5ee7c164682b430fc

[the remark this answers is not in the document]
jack-berg: Resolved in https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2883/commits/e2766d6433672a536d75a3c5ee7c164682b430fc

jack-berg:
Sorry I let this linger so long. I've pushed a commit and changed the scope as described in this [comment](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2883#discussion_r3634106884).

psx95 (approved):
LGTM! Thanks @jack-berg ! 

#### Updated Scope: 
- The PR is about limiting the size of the payload read into the memory at once.
- See [this comment](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2883#discussion_r3363319936) that justifies this change.

jaydeluca (approved):

PR #2884, opened 2026-05-28
Add opamp command processor

breedx-splk:
The OpAmp [spec](https://opentelemetry.io/docs/specs/opamp/) provides a way for the `ServerToAgent` message to contain a `ServerToAgentCommand`, but this client doesn't yet provide a way to handle these commands.

This PR introduces a new inner interface called `CommandProcessor` which can be configured to handle commands with the client.

### DIFF
--- a/opamp-client/src/main/java/io/opentelemetry/opamp/client/OpampClient.java
+++ b/opamp-client/src/main/java/io/opentelemetry/opamp/client/OpampClient.java
@@ -11,6 +11,7 @@ import javax.annotation.Nullable;
 import opamp.proto.AgentDescription;
 import opamp.proto.RemoteConfigStatus;
 import opamp.proto.ServerErrorResponse;
+import opamp.proto.ServerToAgentCommand;
 
 public interface OpampClient extends Closeable {
 
@@ -78,5 +79,15 @@ public interface OpampClient extends Closeable {
      * @param messageData The server response data that needs processing.
      */
     void onMessage(OpampClient client, MessageData messageData);
+  }
+
+  @FunctionalInterface
+  interface CommandProcessor {
+    /**
+     * Called when the Agent receives a command from the Server.
+     *
+     * @param command The command requested by the Server.
+     */
+    void onCommand(ServerToAgentCommand command);
   }
 }
--- a/opamp-client/src/main/java/io/opentelemetry/opamp/client/OpampClientBuilder.java
+++ b/opamp-client/src/main/java/io/opentelemetry/opamp/client/OpampClientBuilder.java
@@ -37,6 +37,7 @@ public final class OpampClientBuilder {
       HttpRequestService.create(OkHttpSender.create("http://localhost:4320/v1/opamp"));
   @Nullable private byte[] instanceUid;
   @Nullable private State.EffectiveConfig effectiveConfigState;
+  @Nullable private OpampClient.CommandProcessor commandProcessor;
 
   OpampClientBuilder() {}
 
@@ -376,6 +377,20 @@ public final class OpampClientBuilder {
     return this;
   }
 
+  /**
+   * Sets the command processor. This command processor will be called for every ServerToAgent
+   * message that contains a non-null command.
+   *
+   * @param commandProcessor The command processor.
+   * @return this
+   */
+  @CanIgnoreReturnValue
+  public OpampClientBuilder setCommandProcessor(
+      @Nullable OpampClient.CommandProcessor commandProcessor) {
+    this.commandProcessor = commandProcessor;
+    return this;
+  }
+
   public OpampClient build(OpampClient.Callbacks callbacks) {
     List<KeyValue> protoIdentifyingAttributes = new ArrayList<>();
     List<KeyValue> protoNonIdentifyingAttributes = new ArrayList<>();
@@ -402,7 +417,7 @@ public final class OpampClientBuilder {
             new State.InstanceUid(instanceUid),
             new State.Flags(0L),
             effectiveConfigState);
-    return OpampClientImpl.create(service, state, callbacks);
+    return OpampClientImpl.create(service, state, callbacks, commandProcessor);
   }
 
   private static State.EffectiveConfig createEffectiveConfigNoop() {
--- a/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/impl/OpampClientImpl.java
+++ b/opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/impl/OpampClientImpl.java
@@ -31,6 +31,7 @@ import java.util.List;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Supplier;
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import okio.ByteString;
 import opamp.proto.AgentDescription;
 import opamp.proto.AgentToServer;
@@ -45,12 +46,16 @@ import opamp.proto.ServerToAgentFlags;
  */
 public final class OpampClientImpl
     implements OpampClient, ObservableState.Listener, RequestService.Callback, Supplier<Request> {
+
+  private static final CommandProcessor NOOP_COMMAND_PROCESSOR = command -> {};
+
   private final RequestService requestService;
   private final AgentToServerAppenders appenders;
   private final OpampClientState state;
   private final RecipeManager recipeManager;
   private final AtomicBoolean hasStopped = new AtomicBoolean(false);
   private final Callbacks callbacks;
+  private final CommandProcessor commandProcessor;
 
   /** Fields that must always be sent. */
   private static final List<Field> REQUIRED_FIELDS;
@@ -83,6 +88,16 @@ public final class OpampClientImpl
 
   public static OpampClientImpl create(
       RequestService requestService, OpampClientState state, Callbacks callbacks) {
+    return create(requestService, state, callbacks, NOOP_COMMAND_PROCESSOR);
+  }
+
+  public static OpampClientImpl create(
+      RequestService requestService,
+      OpampClientState state,
+      Callbacks callbacks,
+      @Nullable CommandProcessor commandProcessor) {
+    CommandProcessor nonNullCommandProcessor =
+        commandProcessor == null ? NOOP_COMMAND_PROCESSOR : commandProcessor;
     AgentToServerAppenders appenders =
         new AgentToServerAppenders(
             AgentDescriptionAppender.create(state.agentDescription),
@@ -95,7 +110,12 @@ public final class OpampClientImpl
             AgentDisconnectAppender.create());
     OpampClientImpl client =
         new OpampClientImpl(
-            requestService, appenders, state, RecipeManager.create(REQUIRED_FIELDS), callbacks);
+            requestService,
+            appenders,
+            state,
+            RecipeManager.create(REQUIRED_FIELDS),
+            callbacks,
+            nonNullCommandProcessor);
 
     // Start
     requestService.start(client, client);
@@ -111,12 +131,14 @@ public final class OpampClientImpl
       AgentToServerAppenders appenders,
       OpampClientState state,
       RecipeManager recipeManager,
-      Callbacks callbacks) {
+      Callbacks callbacks,
+      CommandProcessor commandProcessor) {
     this.requestService = requestService;
     this.appenders = appenders;
     this.state = state;
     this.recipeManager = recipeManager;
     this.callbacks = callbacks;
+    this.commandProcessor = commandProcessor;
   }
 
   @Override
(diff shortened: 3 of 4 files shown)


opamp-client/src/main/java/io/opentelemetry/opamp/client/OpampClientBuilder.java
```diff
@@ -376,6 +377,20 @@ public OpampClientBuilder setEffectiveConfigState(State.EffectiveConfig effectiv
...
+   * Sets the command processor. This command processor will be called for every ServerToAgent
+   * message that contains a non-null command.
+   *
+   * @param commandProcessor The command processor.
+   * @return this
+   */
+  @CanIgnoreReturnValue
+  public OpampClientBuilder setCommandProcessor(
+      @Nullable OpampClient.CommandProcessor commandProcessor) {
+    this.commandProcessor = commandProcessor;
```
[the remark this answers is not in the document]
breedx-splk: This is a reasonable assessment, however what I don't like is the assumption that ALL command processors will handle restartability. The spec is intentionally extensible, and even tho the protobufs only have a single value in the enum today (restart), we would expect it to expand in the future.

Coupling the command processor to this functionality (like in the suggestion) is more brittle than the separate enable method.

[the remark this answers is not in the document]
breedx-splk: Also, since the protos are v3, it _should_ be possible (tolerable) to use other integer values for the `CommandType`. Unfortunately, our currently generated proto classes seem like they forbid/exclude this ability to override...and since the java instrumentation agent isn't exactly restartable, maybe this entire effort is for nought. 😢

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/impl/OpampClientImpl.java
```diff
@@ -186,16 +208,15 @@ private void handleResponsePayload(ServerToAgent response) {
     }
     handleAgentIdentification(response);
 
-    boolean notifyOnMessage = false;
     MessageData.Builder messageBuilder = MessageData.builder();
 
     if (response.remote_config != null) {
-      notifyOnMessage = true;
       messageBuilder.setRemoteConfig(response.remote_config);
+      callbacks.onMessage(this, messageBuilder.build());
```
LikeTheSalad, line 215: I understand that, given the current state of this project where only remote config is supported, it makes sense to get rid of the boolean flag. However, the reason why it's there is that, as we add support for more serverToAgent fields, we're going to need to aggregate them into a single call to `callbacks.onMessage` and avoid calling it multiple times. There might be more elegant ways to achieve the same goal than using a flag, though, if you can think of an alternative that allows the aggregation, I'm up for it. If not, I'd like to at least add some tests that ensure that this method is called only once per serverToAgent message, even though a remote config isn't present in the payload (but another field for `onMessage` is).

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/impl/OpampClientImpl.java
```diff
@@ -186,16 +208,15 @@ private void handleResponsePayload(ServerToAgent response) {
...
     }
 
-    if (notifyOnMessage) {
-      callbacks.onMessage(this, messageBuilder.build());
+    if (response.command != null) {
+      commandProcessor.onCommand(response.command);
```
LikeTheSalad, line 219: According to [the spec](https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#servertoagentcommand-message), this field is special in the sense that it represents a terminal operation, where any other field from the serverToAgent payload, aside from `instance_uid` and `capabilities`, must be ignored. That leaves out the possibility of calling `callbacks.onMessage`, so it's a bit strange to have it handled after checking for calls to `callbacks.onMessage`. I believe the ideal way to handle it is before starting to build a `MessageData` object, and return early in case a command is found.

opamp-client/src/main/java/io/opentelemetry/opamp/client/OpampClient.java
```diff
@@ -79,4 +80,14 @@ interface Callbacks {
      */
     void onMessage(OpampClient client, MessageData messageData);
   }
+
+  @FunctionalInterface
+  interface CommandProcessor {
```
LikeTheSalad, line 85: I like this idea in general, but I'd like to try to keep as much parity as possible with the Go client implementation, for which commands are handled within the same `Callbacks` interface in a function named [onCommand](https://github.com/open-telemetry/opamp-go/blob/ddc92c06085e62df061feb5f566eb9558e9c7f9d/client/types/callbacks.go#L125).

I'm not opposed to doing things differently, though, but I think we should be consistent. If we feel like Go's style is causing us issues, then I think we should consider changing our approach, but in an internally consistent way, by refactoring existing callbacks too. However, right now I don't see why we should do things differently than Go, though.

opamp-client/src/main/java/io/opentelemetry/opamp/client/OpampClientBuilder.java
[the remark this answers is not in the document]
LikeTheSalad: > it should be possible (tolerable) to use other integer values for the CommandType

I'm not sure I'm following, but it sounds like the goal of these changes is to provide a generic way of receiving commands from the server?

If the above is true, then I'm a bit confused, because my understanding from the spec is that commands are fixed and defined in the spec, and only `restart` is available now.

Also, according to [the spec](https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#servertoagentcommand-message), if an agent receives a `restart` command and can't handle it, the agent can ignore it. So, by adding support for commands and then not enabling the one currently available, it seems like we're adding effectively dead code (unless I missed something from the spec that allows for custom commands or the like).

Based on the above, and following up from [this other comment](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2884#discussion_r3324175543), I think a way forward could be to move the `onCommand` function to the existing `Callbacks` interface, which will allow us to keep consistency with the Go client while at the same time avoiding the need to pass a new nullable object around and/or to decide whether we should couple this setter with some capabilities flag setting, and instead replace the setter by a specific command flag enabling option, like `enableRestartCommand()`, where only the flag is added to capabilities, which will allow us to keep on adding these enabler methods as new commands are added into the spec later on.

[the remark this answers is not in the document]
breedx-splk: > > it should be possible (tolerable) to use other integer values for the CommandType
> 
> I'm not sure I'm following, but it sounds like the goal of these changes is to provide a generic way of receiving commands from the server?

That's right. This is the direction that this is trying take. The protobuf enum only has one value, but proto3 allows any values in enums...even if our generated classes don't. This allows custom/bespoke commands that are not (yet) spec'd to be implemented as a demo/prototype.
 
> If the above is true, then I'm a bit confused, because my understanding from the spec is that commands are fixed and defined in the spec, and only `restart` is available now.

Going strictly by the spec yes. Looking at the protos, then no, proto3 allows arbitrary int values in enums.

> if an agent receives a `restart` command and can't handle it, the agent can ignore it. 

Right, but I don't actually care about the one thing that is spec'd currently. I want to use it for other stuff.

> I think a way forward could be to move the `onCommand` function to the existing `Callbacks` interface

I don't care for the `Callbacks` interface because I think it does too much already. I intentionally chose to create a new interface so that it wouldn't make that problem worse...but also to keep this new/unused functionality more isolated.

>  which will allow us to keep consistency with the Go client 

Never been a priority for me. 🤷🏻

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/impl/OpampClientImpl.java
breedx-splk: I only simplified because it was clunky and reads funny. I follow your future-proofing explanation though. I can put it back if you'd like.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/impl/OpampClientImpl.java
breedx-splk: sure, we could move the command handling up and then return early I guess. I think it's silly that the spec both allows and disallows sending messages alongside commands, but whatever.

opamp-client/src/main/java/io/opentelemetry/opamp/client/OpampClient.java
breedx-splk: It's fine. I've never cared about trying to follow the Go client, but if you want to I can. That's fine.

opamp-client/src/main/java/io/opentelemetry/opamp/client/OpampClientBuilder.java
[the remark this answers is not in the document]
LikeTheSalad: > Going strictly by the spec yes. Looking at the protos, then no, proto3 allows arbitrary int values in enums.

Right. I'm still a bit confused by your comment, because my understanding is that we should follow the spec 🤷 The proto3 implementation detail is just how Wire decided to work with proto enums, by backing them with an integer field, but what's defined in the proto file [is an enum](https://github.com/open-telemetry/opamp-spec/blob/a36b47ecf160a8a741519bce036d238b814a1fc3/proto/opamp.proto#L681), hence it's fixed.

The fact we expose Wire's autogenerated types is not ideal, it was done due to time constraints, but ideally we should create Java abstractions that get mapped from the proto value, like OTel Java does with OTLP data, instead of relying on auto-generated types.

For comparison, I see this being similar to OTLP's [SpanKind](https://github.com/open-telemetry/opentelemetry-proto/blob/cb15b866bcfc4021a4bb7925d580fc472d466d7d/opentelemetry/proto/trace/v1/trace.proto#L153). It's defined as a proto enum, and if we generate the Java source using Wire, it will have an integer field in it, but that's Wire's implementation-specific approach, the intent of setting it as enum should be because it's not supposed to get any other values aside from the ones defined in the spec.

Whether it's a good or bad idea to constrain the OpAMP command to an enum is another topic, though, and I think it's a question for the OpAMP spec SIG. Maybe they didn't have use cases for custom command values when they defined it, so if you mention yours to them, maybe they can make it customizable, and I'd be fine with that.

> > which will allow us to keep consistency with the Go client
> 
> Never been a priority for me. 🤷🏻

This client was built based on the Go client impl's architecture in an attempt to keep some API consistency that could become an SDK standard in the future. It is important to me because I believe it's better to strive for some consistency across platforms. In any case, as I mentioned previously [here](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2884#discussion_r3324175543), I'm open to changing the current approach, but not by mixing up two different approaches; it's strange to me to have commands arrive as a separate interface, whereas messageData is part of callbacks. If you'd like to propose a different approach, I think that's a work that warrants its own PR with an overall refactor that, ideally, we should mention to the Go implementers to see if they can benefit from it too, or at least to take that other architecture into account should an SDK API spec is created later on how SDKs should implement the spec, similarly to the OTel SDK API one.

opamp-client/src/main/java/io/opentelemetry/opamp/client/OpampClient.java
LikeTheSalad: I understand that. It's also probably on me because I didn't explain anywhere in the docs here that this client is based on the Go one. I expanded further on refactoring the architecture if needed in [this other thread](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2884#discussion_r3335407423). But for the purposes of these changes, I'd prefer to keep the existing architecture of adding these callbacks into the Callbacks interface.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/impl/OpampClientImpl.java
LikeTheSalad: > sure, we could move the command handling up and then return early I guess

That'd be ideal, thanks.

> I think it's silly that the spec both allows and disallows sending messages alongside commands

I see what you mean. I think they should be open to any feedback on the spec right now, if you have some suggestions, as it's still in development.

opamp-client/src/main/java/io/opentelemetry/opamp/client/internal/impl/OpampClientImpl.java
LikeTheSalad: If you have an idea for an alternative mechanism to avoid the boolean while still ensuring that this method is called once per response, even if it's by adding some unit tests that ensure that, I'm up for it. Otherwise, I prefer to restore the boolean flag.

LikeTheSalad (commented):
Thanks for the follow-up comments, @breedx-splk 

I added some comments around the callbacks approach. I think we can move forward when those are addressed, and it probably enables your use case to use custom commands. **However**, I want to clarify that (if this use case works today) there's no guarantee that it won't get broken in future releases of this client if the spec doesn't change to allow for it, because the reason why it might work right now, if at all, would be by chance, given that the current proto implementation (Wire) is backed by an integer field.

If the spec doesn't change to allow for custom commands, when we do a proper Java data type mapping from proto to Java interfaces (like OTel Java does) then we wouldn't expose the proto-generated types that allow this use case. So I'd suggest you discuss your use case with the OpAMP SIG to see if they relax this field for custom values.

breedx-splk:
I suppose the command approach for things that are not restart is currently unrealistic. I'll close this for now and, if necessary, look into creating something for [custom messages](https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#custom-messages) instead. 🤷🏻

PR #2885, opened 2026-05-28
[dynamic control] Update README on instructions for using the project as an extension

jackshirazi:
**Description:**

Update README on instructions for using the project as an extension. I've included the declarative config instructions although they aren't valid until #2881 is merged.

**Existing Issue(s):**

part of #2868

**Testing:**

Manually tested

**Documentation:**

That's what this is

**Outstanding items:**

see #2868

### DIFF
--- a/dynamic-control/README.md
+++ b/dynamic-control/README.md
@@ -148,6 +148,12 @@ sources:
 
 ```
 
+### Policies supported
+
+- `trace-sampling`
+   - IMPORTANT: if this policy is included in the config, then the sampler installed is overridden and a consistent sampling sampler is installed (technically the ComposableSampler.parentThreshold(ComposableSampler.probability()) sampler)
+   - Expects a value between 0.0 an 1.0 (including both end values), and will apply that sampling rate to the agent's sampler where 0.0 is 0% head sampling and 1.0 is 100% sampling
+
 ### Config example
 
 Working through the following example may be helpful (it assumes the `file` source has been implemented, this is not yet the case).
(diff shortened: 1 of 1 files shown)


dynamic-control/README.md
```diff
@@ -10,6 +10,180 @@ Adding dynamic control of some specific features of the Java agent.
...
+### Gradle execution
+
+You can execute gradlew either from the repo root, or the project root, eg
+
+```
+git clone https://github.com/open-telemetry/opentelemetry-java-contrib.git
+cd opentelemetry-java-contrib
+#/REM Unix or windows, execute the gradlew from this directory
+./gradlew dynamic-control:<TARGET>
+.\gradlew dynamic-control:<TARGET>
```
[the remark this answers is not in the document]
jackshirazi: Actually valid in unix, but doesn't matter I'm leaving it as is

LikeTheSalad (approved):

dynamic-control/README.md
```diff
@@ -10,6 +10,222 @@ Adding dynamic control of some specific features of the Java agent.
...
+cd opentelemetry-java-contrib
+cd dynamic-control
+#/REM Unix or windows, execute the gradlew from the directory above
+../gradlew <TARGET>
+..\gradlew <TARGET>
+```
+
+Useful values for `<TARGET>` are `jar` for the jar containing the classes from this project,
+and `shadowJar` to create a jar containing the classes from this project plus all dependencies.
+The latter target will produce a `*-all.jar` jar.
```
jaydeluca: could we make this more clear, in a way that helps guide the user which one of these two options is right for them? When should they use one vs the other?

jackshirazi: good idea, done

jaydeluca (approved):

Issue #2889, opened 2026-06-01
[disk-buffering] Replace the read-delay synchronisation with atomic rename


mamazzol:
### Component(s)

disk-buffering

### Is your feature request related to a problem? Please describe.

The current disk-buffering writer/reader synchronisation is time-based: a file is written for up to maxFileAgeForWriteMillis (default 30 s), and the reader is forbidden from touching it until minFileAgeForReadMillis (default 33 s) has elapsed since creation. 
This protects the reader from observing a file before it's ready, but it also means that there is a time window where the file is there but invisible to the reader and there is also a strong coupling and order between those two settings.

### Describe the solution you'd like

Move from a time-based to a rename-based sync mechanism, writing to a tmp file and atomic renaming.

Benefits:

- The reader sees buffered data as soon as the writer rolls a file — no extra 3 s+ wait, and the perceived end-to-end latency drops by up to ~33 s in the common case.
- The writer and reader work on disjoint files by construction (different name shapes), which is much simpler to reason about than two interlocking time windows.
- The "in-process only" caveat in the README ("this tool assumes that both the reading and the writing actions are executed within the same application process") is no longer necessary for correctness — atomic rename is a cross-process primitive.
- Configuration becomes orthogonal: write-window and read-window settings no longer constrain each other.

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

breedx-splk:
I like this suggestion. One question that this brings up to me tho is what happens when the process exits or crashes? Are the temp files able to be recovered and resumed/exported?

Right now, with the current approach, I think the files are nearly always able to be picked up when the app restarts. Would this rename approach change that for the worse?

breedx-splk:
I think this was completed with #2897. Feel free to reopen if there's more to do on this.

PR #2897, opened 2026-06-02
Replace time based logic with atomic rename for write

mamazzol:
**Description:**

Feature addition: This PR replaces the time-based separation with an atomic rename

- The writer always writes to <timestamp>.tmp. On close the temporary file is atomically renamed to its final name via File.renameTo, which delegates to POSIX rename(2) for same-directory renames.
- FolderManager now sweeps for orphan temporary files at construction and promotes them.
- A read attempt that finds no ready file will force-close an expired writer instead of waiting for the next append.

As a consequence, `minFileAgeForReadMillis` is no longer required for correctness (default is now 0) and  the build-time validation minFileAgeForRead > maxFileAgeForWrite has been dropped. 

**Existing Issue(s):**

https://github.com/open-telemetry/opentelemetry-java-contrib/issues/2889

**Testing:**

`./gradlew :disk-buffering:check` is green.

- New WritableFileTest case for the staging-then-rename behavior.
- New FolderManagerTest cases for orphan recovery and force-close-on-read with an expired writer.

**Documentation:**

- README.md: updated the writer/reader synchronization explanation and the minFileAgeForReadMillis description.
- DESIGN.md: added a short section on the rename-on-close model.
- FileStorageConfiguration#getMinFileAgeForReadMillis Javadoc updated to describe the new default and use case.


LikeTheSalad (commented):
Thank you, @mamazzol 🙏 

Please take a look at Copilot's suggestions because I think those are valid points. But apart from those, I think it's good to go.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/files/WritableFile.java
```diff
@@ -80,19 +86,22 @@ public synchronized boolean isClosed() {
...
 
   @Override
   public synchronized void close() throws IOException {
     if (isClosed.compareAndSet(false, true)) {
       out.close();
+      if (!staging.renameTo(destination)) {
+        throw new IOException("Could not rename " + staging + " to " + destination);
+      }
     }
   }
```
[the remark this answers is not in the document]
LikeTheSalad: @mamazzol I think we should ignore this recommendation, as it caused the tests to fail due to the method `File#toPath()` not being supported by Android. We have an Android support check to make sure we don't break this tool for Android apps, which usually means we cannot use some of the Java 8 APIs, such as this one.

LikeTheSalad (approved):
Thank you! 🙏

breedx-splk (commented):
I haven't fully reviewed this yet, but I wanted to ask the same question I asked in the issue: What happens to the staging files when the process exits and restarts? Will the staged files be moved over or ?

breedx-splk:
> I haven't fully reviewed this yet, but I wanted to ask the same question I asked in the issue: What happens to the staging files when the process exits and restarts? Will the staged files be moved over or ?

Ok, I see it now, it does look like there is explicit code to cover this case and recover the "orphaned" files. Killer! Thanks.

mamazzol:
> > I haven't fully reviewed this yet, but I wanted to ask the same question I asked in the issue: What happens to the staging files when the process exits and restarts? Will the staged files be moved over or ?
> 
> Ok, I see it now, it does look like there is explicit code to cover this case and recover the "orphaned" files. Killer! Thanks.

Apologies, I missed the original question! Happy to see you liked the approach :)

mamazzol:
Hello @breedx-splk  is anything blocking this PR? Apologies for the ping, just checking!

LikeTheSalad:
I think @breedx-splk is OOO for a while. Can anyone else take a look? Or should we wait until @breedx-splk is back? @open-telemetry/java-contrib-approvers

breedx-splk:
I'm back and taking a look now, sorry for the lag. Were all of the copilot comments responded to with code? If so, can you please close those?

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/FolderManager.java
```diff
@@ -51,6 +52,36 @@ public FolderManager(File folder, FileStorageConfiguration configuration, Clock
     this.folder = folder;
     this.configuration = configuration;
     this.clock = clock;
+    recoverOrphanTempFiles();
```
breedx-splk: We should not escape the constructor to do this work! I'd like to make this constructor package-private (so that tests can see it), and then in the `create()` method just immediately call `recoverOrphanTempFiles()` on the new instance before returning it. That way it's taken care of for all new instances, but without escaping or doing work in the constructor.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/FolderManager.java
```diff
@@ -51,6 +52,36 @@ public FolderManager(File folder, FileStorageConfiguration configuration, Clock
...
+        continue;
+      }
+      if (file.renameTo(target)) {
+        logger.fine(
+            "Recovered orphan temp file: '" + file.getName() + "' -> '" + target.getName() + "'");
+      } else {
+        logger.warning("Could not promote orphan temp file: '" + file.getName() + "'");
+      }
+    }
   }
```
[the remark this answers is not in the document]
breedx-splk: this has been implemented/resolved. See line 65.

disk-buffering/src/main/java/io/opentelemetry/contrib/disk/buffering/internal/storage/FolderManager.java
```diff
@@ -51,6 +52,36 @@ public FolderManager(File folder, FileStorageConfiguration configuration, Clock
...
+        }
+        continue;
+      }
+      if (file.renameTo(target)) {
+        logger.fine(
+            "Recovered orphan temp file: '" + file.getName() + "' -> '" + target.getName() + "'");
+      } else {
+        logger.warning("Could not promote orphan temp file: '" + file.getName() + "'");
+      }
+    }
```
breedx-splk: nitpick: Rather than having a bunch of stateful `continue` checks, I'd much prefer to filter the list in advance, before looping.

disk-buffering/README.md
```diff
@@ -33,8 +33,10 @@ The available configuration parameters are the following:
 * Max folder size, defaults to 10MB.
 * Max age for file writing. It sets the time window where a file can get signals appended to it.
   Defaults to 30 seconds.
-* Min age for file reading. It sets the time to wait before starting to read from a file after
-  its creation. Defaults to 33 seconds. It must be greater that the max age for file writing.
+* Min age for file reading. Optional delay applied before a finalized file becomes eligible for
+  reading. Defaults to `0`. With the rename-on-close design the writer and reader are already
+  synchronized via atomic file rename, so this knob is no longer required for correctness; set it
+  if you want to give writers more time to accumulate larger batches before they're consumed.
```
breedx-splk, line 39: Thanks for adding this. I was very much wondering why there was still stuff dealing with "expiry", but this helps to explain that.

breedx-splk:
Looks good to me. I definitely favor this approach over the pervious timeout-based one. Not yet giving approval, because I want to ensure that the other comments are all addressed (including some one that @LikeTheSalad pointed out). 

Thanks for this contribution! 🍻

mamazzol:
Hi @breedx-splk  thanks for the feedback! 
I worked on the latest things highlighted and resolved all the comments. Everything is accounted for. 
Let me know what you think, happy to do more improvements

breedx-splk (approved):

breedx-splk:
@LikeTheSalad ready for merge? I'm into it. 

Thanks again @mamazzol for the contribution!

LikeTheSalad:
> @LikeTheSalad ready for merge? I'm into it.
> 
> Thanks again @mamazzol for the contribution!

I'm ready, thanks!

PR #2899, opened 2026-06-02
feat(gcp-auth): support 'none' option in GOOGLE_OTEL_AUTH_TARGET_SIGNALS

stankiewicz:
## Description
This PR introduces a new `none` option for the `GOOGLE_OTEL_AUTH_TARGET_SIGNALS` environment variable (and `google.otel.auth.target.signals` system property) in the GCP Authentication Extension.

Previously, there was no explicit way to completely disable the extension if it was loaded. This change allows users to explicitly set the target signals to `none`, which:
 
1.  Disables authentication (no authorization headers are added) for all signals.
2.  Suppresses the warning logs (e.g., "GCP Authentication Extension is not configured for signal type...") that would otherwise appear when signals are not targeted, ensuring a silent and intentional disablement.

Documentation in `README.md` and `ConfigurableOption` Javadoc has been updated to reflect this new option.

## Testing
*   Added a new parameterized test case in `GcpAuthAutoConfigurationCustomizerProviderTest` to verify that `none` correctly prevents modification of both traces and metrics exporters, and that warnings are suppressed.

### DIFF
--- a/gcp-auth-extension/README.md
+++ b/gcp-auth-extension/README.md
@@ -64,7 +64,7 @@ Here is a list of required and optional configuration available for the extensio
 * `GOOGLE_OTEL_AUTH_TARGET_SIGNALS`: Environment variable that specifies a
   comma-separated list of OpenTelemetry signals for which this authentication
   extension should be active.
-  Valid values are `metrics`, `traces`, and `all`.
+  Valid values are `metrics`, `traces`, `all`, and `none`.
   If left unspecified, `all` is assumed, meaning the extension will attempt to
   apply authentication to exports for all signals.
 
--- a/gcp-auth-extension/README.md
+++ b/gcp-auth-extension/README.md
@@ -66,7 +66,9 @@ Here is a list of required and optional configuration available for the extensio
   extension should be active.
   Valid values are `metrics`, `traces`, `all`, and `none`.
   If left unspecified, `all` is assumed, meaning the extension will attempt to
-  apply authentication to exports for all signals.
+  apply authentication to exports for all signals. If `none` is set, disables
+  authentication for all exports. If set alongside other signal types,
+  it takes precedence and all other signal types will be ignored.
 
   * Can also be configured using `google.otel.auth.target.signals` system property.
 
--- a/gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/ConfigurableOption.java
+++ b/gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/ConfigurableOption.java
@@ -46,7 +46,7 @@ enum ConfigurableOption {
    *   <li>{@code metrics} - Enables authentication for metric exports.
    *   <li>{@code traces} - Enables authentication for trace exports.
    *   <li>{@code all} - Enables authentication for all exports.
-   *   <li>{@code none} - Disables authentication for all exports.
+   *   <li>{@code none} - Disables authentication for all exports. If set alongside other signal types, it takes precedence and all other signal types will be ignored.
    * </ul>
    *
    * <p>The values are case-sensitive. Whitespace around commas and values is ignored. Can be
--- a/gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/ConfigurableOption.java
+++ b/gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/ConfigurableOption.java
@@ -46,7 +46,8 @@ enum ConfigurableOption {
    *   <li>{@code metrics} - Enables authentication for metric exports.
    *   <li>{@code traces} - Enables authentication for trace exports.
    *   <li>{@code all} - Enables authentication for all exports.
-   *   <li>{@code none} - Disables authentication for all exports. If set alongside other signal types, it takes precedence and all other signal types will be ignored.
+   *   <li>{@code none} - Disables authentication for all exports. If set alongside other signal
+   *       types, it takes precedence and all other signal types will be ignored.
    * </ul>
    *
    * <p>The values are case-sensitive. Whitespace around commas and values is ignored. Can be
--- a/gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/ConfigurableOption.java
+++ b/gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/ConfigurableOption.java
@@ -46,6 +46,7 @@ enum ConfigurableOption {
    *   <li>{@code metrics} - Enables authentication for metric exports.
    *   <li>{@code traces} - Enables authentication for trace exports.
    *   <li>{@code all} - Enables authentication for all exports.
+   *   <li>{@code none} - Disables authentication for all exports.
    * </ul>
    *
    * <p>The values are case-sensitive. Whitespace around commas and values is ignored. Can be
--- a/gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
+++ b/gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
@@ -70,6 +70,7 @@ public class GcpAuthAutoConfigurationCustomizerProvider
   static final String SIGNAL_TYPE_TRACES = "traces";
   static final String SIGNAL_TYPE_METRICS = "metrics";
   static final String SIGNAL_TYPE_ALL = "all";
+  static final String SIGNAL_TYPE_NONE = "none";
 
   /**
    * Customizes the provided {@link AutoConfigurationCustomizer} such that authenticated exports to
@@ -126,7 +127,7 @@ public class GcpAuthAutoConfigurationCustomizerProvider
       SpanExporter exporter, GoogleCredentials credentials, ConfigProperties configProperties) {
     if (isSignalTargeted(SIGNAL_TYPE_TRACES, configProperties)) {
       return addAuthorizationHeaders(exporter, credentials, configProperties);
-    } else {
+    } else if (!hasSignal(SIGNAL_TYPE_NONE, configProperties)) {
       String[] params = {SIGNAL_TYPE_TRACES, SIGNAL_TARGET_WARNING_FIX_SUGGESTION};
       logger.log(
           Level.WARNING,
@@ -140,7 +141,7 @@ public class GcpAuthAutoConfigurationCustomizerProvider
       MetricExporter exporter, GoogleCredentials credentials, ConfigProperties configProperties) {
     if (isSignalTargeted(SIGNAL_TYPE_METRICS, configProperties)) {
       return addAuthorizationHeaders(exporter, credentials, configProperties);
-    } else {
+    } else if (!hasSignal(SIGNAL_TYPE_NONE, configProperties)) {
       String[] params = {SIGNAL_TYPE_METRICS, SIGNAL_TARGET_WARNING_FIX_SUGGESTION};
       logger.log(
           Level.WARNING,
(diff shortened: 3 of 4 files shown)


gcp-auth-extension/src/test/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProviderTest.java
```diff
@@ -696,9 +678,9 @@ private static Stream<Arguments> provideTargetSignalBehaviorTestCases() {
                 .setUserSpecifiedOtelProperties(
                     ImmutableMap.of(
                         "otel.exporter.otlp.metrics.endpoint",
-                        "https://localhost:4813/v1/metrics",
```
psx95, line 699: Why was this test case altered ? IIRC, this one was intentionally different.

gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
```diff
@@ -152,6 +154,27 @@ private static MetricExporter customizeMetricExporter(
 
   // Checks if the auth extension is configured to target the passed signal for authentication.
   private static boolean isSignalTargeted(String checkSignal, ConfigProperties configProperties) {
```
psx95, line 176: To make this more readable, I suggest creating a dedicated function to check if endpoint is Google's Telemetry API. 

```java
private static boolean targetsGcpEndpoint(String checkSignal, ConfigProperties configProperties) {
 ... 
}
```

This can then be used in the current logic as: 

```java
// IMO this makes it clear the updated criteria for when 
// customization should be applied by this  extension.
// Note the updated warning message. You would also need to add an additional fix suggestion message to indicate possible misconfiguration.
private static SpanExporter customizeSpanExporter(
      SpanExporter exporter, GoogleCredentials credentials, ConfigProperties configProperties) {
    if (isSignalTargeted(SIGNAL_TYPE_TRACES, configProperties) && targetsGcpEndpoint(SIGNAL_TYPE_TRACES, configProperties)) {
      return addAuthorizationHeaders(exporter, credentials, configProperties);
    } else {
      String[] params = {SIGNAL_TYPE_TRACES, SIGNAL_TARGET_WARNING_FIX_SUGGESTION};
      logger.log(
          Level.WARNING,
          "GCP Authentication Extension is not configured for signal type: {0} or is not compatible with the specified endpoint. {1}",
          params);
    }
    return exporter;
  }
```

gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
```diff
@@ -152,6 +154,27 @@ private static MetricExporter customizeMetricExporter(
...
+    String endpoint = configProperties.getString("otel.exporter.otlp." + checkSignal + ".endpoint");
+    if (endpoint == null) {
+      endpoint = configProperties.getString("otel.exporter.otlp.endpoint");
+    }
+    if (endpoint == null) {
+      return false;
+    }
+    try {
+      URI uri = new URI(endpoint);
+      if (!"https".equalsIgnoreCase(uri.getScheme())) {
```
psx95: Do we need to check for http vs https on the client side? I think the backend would respond with the appropriate error in this case.

gcp-auth-extension/src/test/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProviderTest.java
```diff
@@ -122,6 +122,21 @@ class GcpAuthAutoConfigurationCustomizerProviderTest {
...
+          "otel.exporter.otlp.traces.endpoint",
+          "https://telemetry.googleapis.com/v1/traces",
+          "otel.traces.exporter",
+          "otlp",
+          "otel.metrics.exporter",
+          "otlp",
+          "otel.logs.exporter",
+          "none",
+          "otel.resource.attributes",
+          "foo=bar");
```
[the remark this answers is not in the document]
psx95: This seems wrong. [Javadoc for this](https://www.javadoc.io/doc/com.google.guava/guava/33.0.0-jre/com/google/common/collect/ImmutableMap.html#of(K,V,K,V,K,V,K,V,K,V,K,V,K,V,K,V,K,V,K,V)) shows a method with upto 10 key-value pairs.

gcp-auth-extension/src/test/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProviderTest.java
```diff
@@ -639,54 +654,21 @@ private static Stream<Arguments> provideTargetSignalBehaviorTestCases() {
         Arguments.of(
```
psx95, line 639: This is unrelated to your changes, but it would be great if we could use [Named arguments](https://docs.junit.org/5.9.0/api/org.junit.jupiter.api/org/junit/jupiter/api/Named.html) here to indicate what these test cases mean or semantically test.
This would make the intentions of what case is being tested clear to the reader.

Something like: 
```java
Arguments.of(
    Named.of(
        "Configured for all signals (traces and metrics)",
        TargetSignalBehavior.builder()
            .setConfiguredTargetSignals("all")
            .setUserSpecifiedOtelProperties(
                ImmutableMap.of(
                    "otel.exporter.otlp.metrics.endpoint",
                    "https://localhost:4813/v1/metrics",
                    "otel.exporter.otlp.traces.endpoint",
                    "https://localhost:4813/v1/traces",
                    "otel.traces.exporter",
                    "otlp",
                    "otel.metrics.exporter",
                    "otlp",
                    "otel.logs.exporter",
                    "none"))
            .setExpectedIsMetricsSignalModified(true)
            .setExpectedIsTraceSignalModified(true)
            .build())),
Arguments.of(
    Named.of(
        "Configured for traces only",
        TargetSignalBehavior.builder()
            .setConfiguredTargetSignals("traces")
            .setUserSpecifiedOtelProperties(defaultOtelPropertiesSpanExporter)
            .setExpectedIsMetricsSignalModified(false)
            .setExpectedIsTraceSignalModified(true)
            .build())),
 ...
```

This would also help evaluate if there are any additional test cases that should be tested.

psx95 (commented):
Hi @stankiewicz, thanks for the PR! Based on our discussions, I think adding a adding a `none` option in the existing [GOOGLE_OTEL_AUTH_TARGET_SIGNALS](https://github.com/open-telemetry/opentelemetry-java-contrib/tree/main/gcp-auth-extension#configuring-the-extension) property makes more sense and is less of a disruptive change for your particular use case.

stankiewicz:
> Hi @stankiewicz, thanks for the PR! Based on our discussions, I think adding a adding a `none` option in the existing [GOOGLE_OTEL_AUTH_TARGET_SIGNALS](https://github.com/open-telemetry/opentelemetry-java-contrib/tree/main/gcp-auth-extension#configuring-the-extension) property makes more sense and is less of a disruptive change for your particular use case.

done!

gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
```diff
@@ -152,14 +169,19 @@ private static MetricExporter customizeMetricExporter(
 
   // Checks if the auth extension is configured to target the passed signal for authentication.
   private static boolean isSignalTargeted(String checkSignal, ConfigProperties configProperties) {
```
psx95, line 176: This entire thing could be simplified by updating isSignalTargeted to consider `none`:

```java
 private static boolean isSignalTargeted(String checkSignal, ConfigProperties configProperties) {
    String userSpecifiedTargetedSignals =
        ConfigurableOption.GOOGLE_OTEL_AUTH_TARGET_SIGNALS.getConfiguredValueWithFallback(
            configProperties, () -> SIGNAL_TYPE_ALL);
    List<String> targetedSignals =
        stream(userSpecifiedTargetedSignals.split(",")).map(String::trim).collect(toList());
    if (targetedSignals.contains(SIGNAL_TYPE_NONE)) {
      return false;
    }
    return targetedSignals.contains(checkSignal) || targetedSignals.contains(SIGNAL_TYPE_ALL);
  }
```
Now, `isSignalTargeted` will return false if it finds `none` specified by the user. This would prevent adding the additional if statement in customize methods and eliminate the need for `hasSignal` method. 

*(Out of scope for this PR):*
As a follow-up the signals that could actually be checked (METRICS, TRACES) or accepted by this method should be converted to specific enums and isSignalTargeted should be updated to something like:
```java
 private static boolean isSignalTargeted(TelemetryTarget checkTelemetryTarget, ConfigProperties configProperties) {
 ...
 }
```

gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
psx95: Update this to: 
```java
String[] params = {SIGNAL_TYPE_METRICS, SIGNAL_TYPE_NONE, SIGNAL_TARGET_WARNING_FIX_SUGGESTION};
...
 "GCP Authentication Extension is not configured for signal type: {0} or is configured with signal type: {1}. {2}",
```

gcp-auth-extension/src/test/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProviderTest.java
```diff
@@ -743,6 +743,24 @@ private static Stream<Arguments> provideTargetSignalBehaviorTestCases() {
                         "none"))
                 .setExpectedIsMetricsSignalModified(true)
                 .setExpectedIsTraceSignalModified(false)
+                .build()),
+        Arguments.of(
+            TargetSignalBehavior.builder()
+                .setConfiguredTargetSignals("none")
```
psx95, line 749: The extension allows setting up multiple signal types at once, add more test case with conflicting signal types like - 
 - "metrics", "traces", "none"
 - "all", "none"
 - "metrics, none"

The expected behavior is that if the user sets none, every other signal type gets ignored.

gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/ConfigurableOption.java
```diff
@@ -46,6 +46,7 @@ enum ConfigurableOption {
    *   <li>{@code metrics} - Enables authentication for metric exports.
    *   <li>{@code traces} - Enables authentication for trace exports.
    *   <li>{@code all} - Enables authentication for all exports.
+   *   <li>{@code none} - Disables authentication for all exports.
```
psx95: Document the behavior for `none`: 

```
Disables authentication for all exports. If set alongside other signal types, it takes precedence and all other signal types will be ignored.
```

Also update README to reflect this behavior.

psx95 (changes requested):

gcp-auth-extension/README.md
```diff
@@ -64,7 +64,7 @@ Here is a list of required and optional configuration available for the extensio
 * `GOOGLE_OTEL_AUTH_TARGET_SIGNALS`: Environment variable that specifies a
   comma-separated list of OpenTelemetry signals for which this authentication
   extension should be active.
-  Valid values are `metrics`, `traces`, and `all`.
+  Valid values are `metrics`, `traces`, `all`, and `none`.
```
psx95, line 67: nit: Add information about how `none` behaves here as well.

psx95 (commented):
LGTM, pending one nit. Thanks for following up on the PR! 💯

psx95 (approved):

psx95:
Hold off on merging till the pending comment is addressed. GitHub does not allow revoking approval for some reason.

psx95 (approved):

psx95:
> Hold off on merging till the pending comment is addressed. GitHub does not allow revoking approval for some reason.

Ready to merge now.

jaydeluca (approved):

PR #2979, opened 2026-07-10
Submit Gradle dependencies to dependency graph

trask:
Submit resolved Gradle dependencies to GitHub’s dependency graph on each push to main, enabling GitHub to detect vulnerabilities in direct and transitive dependencies.

After merging, the next step is to verify that the workflow successfully populates the dependency graph, then check whether Renovate immediately opens an update for Jackson 2.22.1 from the resulting vulnerability alert. If it does not, follow up by configuring security-only Dependabot updates similar to the [semantic-conventions-genai dependency policy](https://github.com/open-telemetry/semantic-conventions-genai/blob/main/.github/DEPENDENCIES.md) and its [Dependabot configuration](https://github.com/open-telemetry/semantic-conventions-genai/blob/main/.github/dependabot.yml).

Related to https://github.com/open-telemetry/opentelemetry-java-contrib/issues/2927

### DIFF
--- a/.github/workflows/dependency-submission.yml
+++ b/.github/workflows/dependency-submission.yml
@@ -1,0 +1,31 @@
+name: Dependency submission
+
+on:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  submit:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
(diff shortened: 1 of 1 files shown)


laurit (approved):

laurit:
2.22.1 was released only a couple of days ago, in instrumentation repo through the dependency dashboard issue https://github.com/open-telemetry/opentelemetry-java-contrib/issues/969

trask:
yeah, dependency updates that fix CVEs are supposed to ignore our weekly schedule and get updated right away. I [got this working in another repo](https://github.com/open-telemetry/semantic-conventions-genai/blob/main/.github/DEPENDENCIES.md#why-the-split), seeing if I can get it working for Java repos.

PR #2982, opened 2026-07-10
Remove the in development notice of the IBM MQ metrics component and move it to beta

atoulme:
This removes the warning displayed in the README saying the component is in development. It also updates the stability of IBM MQ from alpha to beta.


breedx-splk (approved):
Do you want to also update the main README to say `beta` now?

PR #2988, opened 2026-07-13
Replace declarative config bridge usages in contrib

zeitlinger:
## Summary
- replace the remaining contrib `DeclarativeConfigPropertiesBridgeBuilder` usages in `span-stacktrace` and `inferred-spans`
- switch those modules to the new `ConfigPropertiesBackedConfigProvider` access-path API
- keep the declarative config access in `inferred-spans` on flat underscore keys
- remove duplicated duration parsing by using `DeclarativeConfigPropertiesDurationUtil.parseDuration(...)`

## Why
`DeclarativeConfigPropertiesBridge` is being deprecated in `opentelemetry-java-instrumentation`, and these remaining contrib usages need to move off it first.

This change depends on https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/19220, which supersedes #15835 and adds the bridge helpers used here.

Related deprecation PR: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/19202

The unrelated `SamplingProfilerTest` flake fix was split into https://github.com/open-telemetry/opentelemetry-java-contrib/pull/3000.

## How this uses the new bridge APIs
- `inferred-spans` bridges the component root directly to `otel.inferred.spans.` with `ConfigPropertiesBackedConfigProvider.builder().setAccessPath("", "otel.inferred.spans.")`
- `span-stacktrace` bridges the component root directly to `otel.java.experimental.span-stacktrace.` the same way
- both modules use `DeclarativeConfigPropertiesDurationUtil.parseDuration(...)` instead of carrying their own parsing copy

## Breaking behavior change
Declarative config YAML now only accepts integer millisecond values for these duration properties.

Examples:
- supported: `min_duration: 100`
- not supported in declarative YAML anymore: `min_duration: 100ms`

That matches the behavior documented on `DeclarativeConfigPropertiesDurationUtil`: flat `ConfigProperties` inputs can still use string durations such as `42ms`, but non-`ConfigProperties` declarative config only accepts integer milliseconds.

## Testing
- `GRADLE_USER_HOME=/tmp/gradle-home-otel-java-contrib ./gradlew --no-build-cache --no-configuration-cache -PuseLocalMaven :span-stacktrace:check :inferred-spans:check`

## CI note
The newly added `InferredSpansConfigTest` creates a real `InferredSpansProcessor`, which loads async-profiler. Async-profiler is not supported on Windows, so the test is disabled there with the same `@DisabledOnOs(WINDOWS)` guard used by the other inferred-spans tests that instantiate it. Earlier CI failures across all platforms masked this Windows-only failure until the dependency and deprecation-warning issues were fixed.

Additional verification:
- `GRADLE_USER_HOME=/tmp/gradle-home-otel-java-contrib-pr2988 ./gradlew --no-build-cache --no-configuration-cache :inferred-spans:check`


span-stacktrace/src/main/java/io/opentelemetry/contrib/stacktrace/StackTraceAutoConfig.java
```diff
@@ -26,44 +29,52 @@ public class StackTraceAutoConfig implements AutoConfigurationCustomizerProvider
...
   }
 
-  static StackTraceSpanProcessor create(ConfigProperties properties) {
+  static StackTraceSpanProcessor create(DeclarativeConfigProperties properties) {
     return new StackTraceSpanProcessor(getMinDuration(properties), getFilterPredicate(properties));
   }
 
+  static DeclarativeConfigProperties createDeclarativeConfig(ConfigProperties properties) {
+    return DeclarativeConfigBridge.createComponentProperties(
+        properties, "otel.java.experimental.span-stacktrace.");
```
breedx-splk, line 52: Oh interesting, I don't want to block this PR since this was here already, but should we be using a hyphen here? I think the underscore is more conventional and doesn't have some of the weird problems that hyphens might for env vars.

I'll open an issue to track it.

breedx-splk: Opened #3014.

breedx-splk (approved):
Thanks for picking this up, @zeitlinger.

span-stacktrace/src/main/java/io/opentelemetry/contrib/stacktrace/StackTraceAutoConfig.java
zeitlinger: this is the param "configPropertyPrefix" - not the declarative config syntax

jaydeluca (approved):

Issue #2996, opened 2026-07-14
gcp-auth-extension: support ID token authentication for IAM-protected OTLP endpoints


tknhs:
Component: gcp-auth-extension

The extension currently adds an OAuth 2.0 access token from ADC to OTLP exports. That works for Google Cloud endpoints like telemetry.googleapis.com, but not for OTLP endpoints protected by Google IAM — for example a Collector running on Cloud Run, or one behind IAP. Those expect a Google-signed OIDC ID token with a matching aud and reject access tokens.

The Collector's googleclientauthextension already handles this with token_type: id_token. It would be useful to have the same option on the Java agent side, especially since running the Collector on Cloud Run is a documented setup.

Proposed config:
- GOOGLE_OTEL_AUTH_TOKEN_TYPE: access_token (default) or id_token
- GOOGLE_OTEL_AUTH_ID_TOKEN_AUDIENCE: audience for the ID token, required when id_token

For id_token the extension would wrap ADC in IdTokenCredentials for the given audience, and skip the quota-project header and gcp.project_id resource attribute (both specific to the GCP APIs). Default behavior is unchanged.

I've implemented this with unit tests and tested it against a Collector on Cloud Run, over both HTTP and gRPC, including token refresh across expiry. I can open a PR if this sounds reasonable.

psx95:
Hi @tknhs ,
Thanks for opening the issue and the subsequent PR to address it. Seems like a valid feature request, but I'll take a closer look at it again and review your PR! 

Appreciate your patience on this! 👍🏻

psx95:
This issue fixed by #2999 .

See the [sample update in this PR](https://github.com/GoogleCloudPlatform/opentelemetry-samples/pull/104) on how to configure the extension for ID Token based auth.

PR #2999, opened 2026-07-14
Add ID token support to gcp-auth-extension

tknhs:
**Description:**

Feature addition. `gcp-auth-extension` currently attaches an OAuth 2.0 access token from ADC, which works for Google Cloud endpoints like `telemetry.googleapis.com` but not for OTLP endpoints protected by Google IAM (e.g. a Collector running on Cloud Run, or behind IAP), which require a Google-signed OIDC ID token with a matching `aud`.

This adds two config options:

- `GOOGLE_OTEL_AUTH_TOKEN_TYPE`: `access_token` (default) or `id_token`
- `GOOGLE_OTEL_AUTH_ID_TOKEN_AUDIENCE`: audience for the ID token, required when `id_token`

For `id_token`, the extension wraps ADC in `IdTokenCredentials` for the given audience, and skips the quota-project header and the `gcp.project_id` resource attribute (both specific to the GCP APIs). Default behavior is unchanged, so this is backward compatible. This mirrors the `token_type: id_token` option of the Collector's `googleclientauthextension`.

**Existing Issue(s):**

#2996 

**Testing:**

- Added unit tests covering the `id_token` path (audience present, HTTP/gRPC exporters, missing-audience and non-IdTokenProvider credential errors, unsupported token type).
- Existing tests pass unchanged, confirming backward compatibility.
- Also tested end-to-end against a Collector on Cloud Run over both HTTP and gRPC, including token refresh across the ~1h ID token expiry.

**Documentation:**

- Updated the module README with the two new options and when to use `id_token` (Cloud Run / IAP), including the audience semantics.

### DIFF
--- a/gcp-auth-extension/README.md
+++ b/gcp-auth-extension/README.md
@@ -72,6 +72,41 @@ Here is a list of required and optional configuration available for the extensio
 
   * Can also be configured using `google.otel.auth.target.signals` system property.
 
+* `GOOGLE_OTEL_AUTH_TOKEN_TYPE`: Environment variable that specifies the type
+  of token that the extension attaches to the exported telemetry.
+  Valid values are `access_token` and `id_token`:
+
+  * `access_token` (default): OAuth 2.0 access token retrieved from the
+    Application Default Credentials. Use this to export telemetry to Google
+    Cloud APIs, for example `telemetry.googleapis.com`.
+  * `id_token`: Google-signed OpenID Connect ID token minted from the
+    Application Default Credentials for the configured audience. Use this to
+    export telemetry to OTLP endpoints protected by Google IAM-based
+    authentication - for example, an OpenTelemetry Collector running on
+    [Cloud Run](https://cloud.google.com/run/docs/authenticating/service-to-service)
+    or behind [Identity-Aware Proxy](https://cloud.google.com/iap/docs/authentication-howto).
+    The Application Default Credentials must be able to mint ID tokens. This
+    is the case for service account based credentials - for example, the
+    default service account of a GCP compute environment, a service account
+    key, or impersonated credentials. When `id_token` is used, the
+    `GOOGLE_CLOUD_PROJECT` option is not required and the `gcp.project_id`
+    resource attribute is not added.
+
+  The value names are consistent with the `token_type` option of the
+  [Google Client Auth Extension](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/googleclientauthextension)
+  for the OpenTelemetry Collector.
+
+  * Can also be configured using `google.otel.auth.token.type` system property.
+
+* `GOOGLE_OTEL_AUTH_ID_TOKEN_AUDIENCE`: Environment variable that specifies
+  the audience used when minting Google-signed ID tokens. Required when
+  `GOOGLE_OTEL_AUTH_TOKEN_TYPE` is set to `id_token`, ignored otherwise.
+  For Cloud Run, this is the URL of the receiving service or one of its
+  [custom audiences](https://cloud.google.com/run/docs/configuring/custom-audiences).
+  For Identity-Aware Proxy, this is the OAuth 2.0 client ID.
+
+  * Can also be configured using `google.otel.auth.id.token.audience` system property.
+
 ## Usage
 
 ### With OpenTelemetry Java agent
--- a/gcp-auth-extension/README.md
+++ b/gcp-auth-extension/README.md
@@ -72,7 +72,7 @@ Here is a list of required and optional configuration available for the extensio
 
   * Can also be configured using `google.otel.auth.target.signals` system property.
 
-* `GOOGLE_OTEL_AUTH_TOKEN_TYPE`: Environment variable that specifies the type
+* `GOOGLE_AUTH_TOKEN_TYPE`: Environment variable that specifies the type
   of token that the extension attaches to the exported telemetry.
   Valid values are `access_token` and `id_token`:
 
@@ -91,20 +91,20 @@ Here is a list of required and optional configuration available for the extensio
     key, or impersonated credentials. When `id_token` is used, the
     `GOOGLE_CLOUD_PROJECT` option is not required and the `gcp.project_id`
     resource attribute is not added.
-  * Can also be configured using `google.otel.auth.token.type` system property.
+  * Can also be configured using `google.auth.token.type` system property.
 
   The value names are consistent with the `token_type` option of the
   [Google Client Auth Extension](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/googleclientauthextension)
   for the OpenTelemetry Collector.
 
-* `GOOGLE_OTEL_AUTH_ID_TOKEN_AUDIENCE`: Environment variable that specifies
+* `GOOGLE_AUTH_ID_TOKEN_AUDIENCE`: Environment variable that specifies
   the audience used when minting Google-signed ID tokens. Required when
-  `GOOGLE_OTEL_AUTH_TOKEN_TYPE` is set to `id_token`, ignored otherwise.
+  `GOOGLE_AUTH_TOKEN_TYPE` is set to `id_token`, ignored otherwise.
   For Cloud Run, this is the URL of the receiving service or one of its
   [custom audiences](https://cloud.google.com/run/docs/configuring/custom-audiences).
   For Identity-Aware Proxy, this is the OAuth 2.0 client ID.
 
-  * Can also be configured using `google.otel.auth.id.token.audience` system property.
+  * Can also be configured using `google.auth.id.token.audience` system property.
 
 ## Usage
 
--- a/gcp-auth-extension/README.md
+++ b/gcp-auth-extension/README.md
@@ -91,12 +91,11 @@ Here is a list of required and optional configuration available for the extensio
     key, or impersonated credentials. When `id_token` is used, the
     `GOOGLE_CLOUD_PROJECT` option is not required and the `gcp.project_id`
     resource attribute is not added.
+  * Can also be configured using `google.otel.auth.token.type` system property.
 
   The value names are consistent with the `token_type` option of the
   [Google Client Auth Extension](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/googleclientauthextension)
   for the OpenTelemetry Collector.
-
-  * Can also be configured using `google.otel.auth.token.type` system property.
 
 * `GOOGLE_OTEL_AUTH_ID_TOKEN_AUDIENCE`: Environment variable that specifies
   the audience used when minting Google-signed ID tokens. Required when
--- a/gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/ConfigurableOption.java
+++ b/gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/ConfigurableOption.java
@@ -54,7 +54,42 @@ enum ConfigurableOption {
    * configured using the environment variable `GOOGLE_OTEL_AUTH_TARGET_SIGNALS` or the system
    * property `google.otel.auth.target.signals`.
    */
-  GOOGLE_OTEL_AUTH_TARGET_SIGNALS("Target Signals for Google Authentication Extension");
+  GOOGLE_OTEL_AUTH_TARGET_SIGNALS("Target Signals for Google Authentication Extension"),
+
+  /**
+   * Specifies the type of token that the extension attaches to the exported telemetry. Can be
+   * configured using the environment variable `GOOGLE_OTEL_AUTH_TOKEN_TYPE` or the system property
+   * `google.otel.auth.token.type`.
+   *
+   * <p>Valid values are:
+   *
+   * <ul>
+   *   <li>{@code access_token} (default) - OAuth 2.0 access token retrieved from the Application
+   *       Default Credentials. Required for exporting to Google Cloud APIs, for example {@code
+   *       telemetry.googleapis.com}.
+   *   <li>{@code id_token} - Google-signed OpenID Connect ID token minted from the Application
+   *       Default Credentials for the audience configured via {@link
+   *       #GOOGLE_OTEL_AUTH_ID_TOKEN_AUDIENCE}. Required for exporting to OTLP endpoints protected
+   *       by IAM-based authentication, for example an OpenTelemetry Collector running on Cloud Run
+   *       or behind Identity-Aware Proxy.
+   * </ul>
+   *
+   * <p>The value names are consistent with the {@code token_type} option of the <a
+   * href="https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/googleclientauthextension">Google
+   * Client Auth Extension for the OpenTelemetry Collector</a>.
+   */
+  GOOGLE_OTEL_AUTH_TOKEN_TYPE("Token Type for Google Authentication Extension"),
+
+  /**
+   * The audience used when minting Google-signed ID tokens. Required when {@link
+   * #GOOGLE_OTEL_AUTH_TOKEN_TYPE} is set to {@code id_token}, ignored otherwise. Can be configured
+   * using the environment variable `GOOGLE_OTEL_AUTH_ID_TOKEN_AUDIENCE` or the system property
+   * `google.otel.auth.id.token.audience`.
+   *
+   * <p>For Cloud Run, this is the URL of the receiving service or one of its configured custom
+   * audiences. For Identity-Aware Proxy, this is the OAuth 2.0 client ID.
+   */
+  GOOGLE_OTEL_AUTH_ID_TOKEN_AUDIENCE("Google ID Token Audience");
 
   private final String userReadableName;
   private final String environmentVariableName;
--- a/gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/ConfigurableOption.java
+++ b/gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/ConfigurableOption.java
@@ -58,8 +58,8 @@ enum ConfigurableOption {
 
   /**
    * Specifies the type of token that the extension attaches to the exported telemetry. Can be
-   * configured using the environment variable `GOOGLE_OTEL_AUTH_TOKEN_TYPE` or the system property
-   * `google.otel.auth.token.type`.
+   * configured using the environment variable `GOOGLE_AUTH_TOKEN_TYPE` or the system property
+   * `google.auth.token.type`.
    *
    * <p>Valid values are:
    *
@@ -69,27 +69,27 @@ enum ConfigurableOption {
    *       telemetry.googleapis.com}.
    *   <li>{@code id_token} - Google-signed OpenID Connect ID token minted from the Application
    *       Default Credentials for the audience configured via {@link
-   *       #GOOGLE_OTEL_AUTH_ID_TOKEN_AUDIENCE}. Required for exporting to OTLP endpoints protected
-   *       by IAM-based authentication, for example an OpenTelemetry Collector running on Cloud Run
-   *       or behind Identity-Aware Proxy.
+   *       #GOOGLE_AUTH_ID_TOKEN_AUDIENCE}. Required for exporting to OTLP endpoints protected by
+   *       IAM-based authentication, for example an OpenTelemetry Collector running on Cloud Run or
+   *       behind Identity-Aware Proxy.
    * </ul>
    *
    * <p>The value names are consistent with the {@code token_type} option of the <a
    * href="https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/googleclientauthextension">Google
    * Client Auth Extension for the OpenTelemetry Collector</a>.
    */
-  GOOGLE_OTEL_AUTH_TOKEN_TYPE("Token Type for Google Authentication Extension"),
+  GOOGLE_AUTH_TOKEN_TYPE("Token Type for Google Authentication Extension"),
 
   /**
    * The audience used when minting Google-signed ID tokens. Required when {@link
-   * #GOOGLE_OTEL_AUTH_TOKEN_TYPE} is set to {@code id_token}, ignored otherwise. Can be configured
-   * using the environment variable `GOOGLE_OTEL_AUTH_ID_TOKEN_AUDIENCE` or the system property
-   * `google.otel.auth.id.token.audience`.
+   * #GOOGLE_AUTH_TOKEN_TYPE} is set to {@code id_token}, ignored otherwise. Can be configured using
+   * the environment variable `GOOGLE_AUTH_ID_TOKEN_AUDIENCE` or the system property
+   * `google.auth.id.token.audience`.
    *
    * <p>For Cloud Run, this is the URL of the receiving service or one of its configured custom
    * audiences. For Identity-Aware Proxy, this is the OAuth 2.0 client ID.
    */
-  GOOGLE_OTEL_AUTH_ID_TOKEN_AUDIENCE("Google ID Token Audience");
+  GOOGLE_AUTH_ID_TOKEN_AUDIENCE("Google ID Token Audience");
 
   private final String userReadableName;
   private final String environmentVariableName;
--- a/gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
+++ b/gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
@@ -127,7 +127,7 @@ public class GcpAuthAutoConfigurationCustomizerProvider
   }
 
   /**
-   * Lazily initializes and caches the {@link GoogleCredentials} used for authentication. The type
+   * Lazily initializes and caches the {@link OAuth2Credentials} used for authentication. The type
    * of the credentials is determined by the configured {@link
    * ConfigurableOption#GOOGLE_OTEL_AUTH_TOKEN_TYPE}.
    */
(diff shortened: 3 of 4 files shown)


gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
```diff
@@ -136,12 +126,86 @@ public int order() {
...
+    }
+    if (!(applicationDefaultCredentials instanceof IdTokenProvider)) {
+      throw new ConfigurationException(
+          String.format(
+              "GCP Authentication Extension not configured properly: the retrieved Application Default Credentials (%s) cannot mint ID tokens. Use a credential type implementing IdTokenProvider - for example the default service account of a GCP compute environment, a service account key, or impersonated credentials (gcloud auth application-default login --impersonate-service-account=<SERVICE_ACCOUNT>).",
+              applicationDefaultCredentials.getClass().getSimpleName()));
+    }
+    return IdTokenCredentials.newBuilder()
+        .setIdTokenProvider((IdTokenProvider) applicationDefaultCredentials)
+        .setTargetAudience(audience)
```
psx95: I think for impersonated credentials, you would also need to set
```java
.setOptions(asList(IdTokenProvider.Option.INCLUDE_EMAIL))
```

From [Authenticate with a service account OIDC token](https://docs.cloud.google.com/iap/docs/authentication-howto#authenticate_with_a_service_account_oidc_token), the note mentions
> Note: The generated ID token must have an email claim for IAP to accept it.

The PR description mentions that this was tested with Cloud Run - just to confirm, was the OTel Collector running in Cloud Run behind an IAP service? 

Based on my [testing](https://github.com/GoogleCloudPlatform/opentelemetry-samples/pull/88) this seems required.

gcp-auth-extension/README.md
```diff
@@ -72,6 +72,41 @@ Here is a list of required and optional configuration available for the extensio
...
+    [Cloud Run](https://cloud.google.com/run/docs/authenticating/service-to-service)
+    or behind [Identity-Aware Proxy](https://cloud.google.com/iap/docs/authentication-howto).
+    The Application Default Credentials must be able to mint ID tokens. This
+    is the case for service account based credentials - for example, the
+    default service account of a GCP compute environment, a service account
+    key, or impersonated credentials. When `id_token` is used, the
+    `GOOGLE_CLOUD_PROJECT` option is not required and the `gcp.project_id`
+    resource attribute is not added.
+
+  The value names are consistent with the `token_type` option of the
```
psx95, line 96: nit: IMO, a better position for this text would be either after the third bullet point or at the very beginning after the text: "Valid values are ..."

gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/ConfigurableOption.java
```diff
@@ -54,7 +54,42 @@ enum ConfigurableOption {
...
+   *       #GOOGLE_OTEL_AUTH_ID_TOKEN_AUDIENCE}. Required for exporting to OTLP endpoints protected
+   *       by IAM-based authentication, for example an OpenTelemetry Collector running on Cloud Run
+   *       or behind Identity-Aware Proxy.
+   * </ul>
+   *
+   * <p>The value names are consistent with the {@code token_type} option of the <a
+   * href="https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/googleclientauthextension">Google
+   * Client Auth Extension for the OpenTelemetry Collector</a>.
+   */
+  GOOGLE_OTEL_AUTH_TOKEN_TYPE("Token Type for Google Authentication Extension"),
```
psx95: Nit: Rename this to `GOOGLE_AUTH_TOKEN_TYPE`. 

Rationale: Auth being configured here is not specific to OpenTelemetry.

gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/ConfigurableOption.java
```diff
@@ -54,7 +54,42 @@ enum ConfigurableOption {
...
+  /**
+   * The audience used when minting Google-signed ID tokens. Required when {@link
+   * #GOOGLE_OTEL_AUTH_TOKEN_TYPE} is set to {@code id_token}, ignored otherwise. Can be configured
+   * using the environment variable `GOOGLE_OTEL_AUTH_ID_TOKEN_AUDIENCE` or the system property
+   * `google.otel.auth.id.token.audience`.
+   *
+   * <p>For Cloud Run, this is the URL of the receiving service or one of its configured custom
+   * audiences. For Identity-Aware Proxy, this is the OAuth 2.0 client ID.
+   */
+  GOOGLE_OTEL_AUTH_ID_TOKEN_AUDIENCE("Google ID Token Audience");
```
psx95: Nit: Rename this to ` GOOGLE_AUTH_ID_TOKEN_AUDIENCE`

Rationale: Auth being configured here is not specific to OpenTelemetry.

psx95:
Hi @tknhs, thanks for the PR! 
Looks mostly correct to me, just one concern and a couple of nits!

gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
tknhs: Thanks. To answer your question: when I tested it, the Collector was running on Cloud Run behind Cloud Run's IAM (run.invoker), not IAP.
You're right that IAP requires the email claim, so I've added INCLUDE_EMAIL.
[`bfc8e2d` (this PR)](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2999/commits/bfc8e2d1393b6bd9698a3b5e80b0f6baffc8d33f)

gcp-auth-extension/README.md
tknhs: Done, moved as suggested.

[`c04d13a` (this PR)](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2999/commits/c04d13ac76f96a55ec1d544e1c14acef6511e1d4)

gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/ConfigurableOption.java
tknhs: Thanks, agreed.

[`be647df` (this PR)](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2999/commits/be647dff4e0a316b6731251ec19e7fb3a09b1f94)

gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/ConfigurableOption.java
tknhs: Same as above.

[`be647df` (this PR)](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2999/commits/be647dff4e0a316b6731251ec19e7fb3a09b1f94)

tknhs:
@psx95 Thanks for the review! 
I've addressed the concern and the nits, so please take another look when you have a chance.

gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
```diff
@@ -136,12 +127,86 @@ public int order() {
...
+      }
+      return credentials;
+    }
+  }
+
+  // Creates the credentials used for authentication based on the configured token type. For the
+  // default token type (access_token) the Application Default Credentials are used as-is. For
+  // id_token, the Application Default Credentials are wrapped into IdTokenCredentials which mint
+  // Google-signed ID tokens for the configured audience.
+  private static OAuth2Credentials createCredentials(ConfigProperties configProperties) {
```
psx95: QQ -  Why not move this inside the `LazyCredentialsSupplier` ?

gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
```diff
@@ -136,12 +127,86 @@ public int order() {
     return Integer.MAX_VALUE - 1;
   }
 
+  /**
+   * Lazily initializes and caches the {@link OAuth2Credentials} used for authentication. The type
+   * of the credentials is determined by the configured {@link
+   * ConfigurableOption#GOOGLE_AUTH_TOKEN_TYPE}.
+   */
+  private static class LazyCredentialsSupplier {
```
psx95: Sorry missed this in the previous review - any reason this does not implement the standard `Supplier` interface? 

I'm wondering if we can use: 
```java
private static class LazyCredentialsSupplier implements Supplier<OAuth2Credentials> {
    private final ConfigProperties configProperties;
    @Nullable private OAuth2Credentials credentials;

    // Pass the config once when creating the supplier
    LazyCredentialsSupplier(ConfigProperties configProperties) {
        this.configProperties = configProperties;
    }

    @Override
    public synchronized OAuth2Credentials get() {
        if (credentials == null) {
            credentials = createCredentials(configProperties);
        }
        return credentials;
    }
}
```

gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
```diff
@@ -136,12 +127,86 @@ public int order() {
...
+        .setIdTokenProvider((IdTokenProvider) applicationDefaultCredentials)
+        .setTargetAudience(audience)
+        .setOptions(singletonList(IdTokenProvider.Option.INCLUDE_EMAIL))
+        .build();
+  }
+
+  // Checks whether the extension is configured to attach ID tokens instead of access tokens.
+  private static boolean isIdTokenType(ConfigProperties configProperties) {
+    String tokenType =
+        ConfigurableOption.GOOGLE_AUTH_TOKEN_TYPE.getConfiguredValueWithFallback(
```
psx95, line 202: You've used `getConfiguredValueWithFallback` so `tokenType` is guaranteed to be valid. So in this case the following switch could be simplified using a ternary if statement.

That will simplify this method a lot more.

gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
psx95: Thanks for verifying - and including email did not break your setup with Collector running in Cloud Run behind IAM - correct ?

gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
```diff
@@ -136,12 +127,86 @@ public int order() {
...
+    try {
+      applicationDefaultCredentials = GoogleCredentials.getApplicationDefault();
+    } catch (IOException e) {
+      throw new GoogleAuthException(Reason.FAILED_ADC_RETRIEVAL, e);
+    }
+    if (!useIdToken) {
+      return applicationDefaultCredentials;
+    }
+    if (!(applicationDefaultCredentials instanceof IdTokenProvider)) {
+      throw new ConfigurationException(
```
psx95: Looking at the design pattern here, we probably don't want to throw Exception in case Optional configuration is misconfigured.

We should silently fail the exception unless the configuration is required (Like unable to get project ID). 

You should use logger to log a warning message about possible misconfiguration of options here. See how `SIGNAL_TARGET_WARNING_FIX_SUGGESTION` is used in warnings. 
If the code does reach this exception - you can return `applicationDefaultCredentials` instead. 

Auth will fail, but the application won't crash.

psx95 (changes requested):
Thanks for following up! I think this should work functionally, but I do have some additional follow-up comments which I missed last time. My apologies for that.

gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
tknhs: Yes, that's right.
I added INCLUDE_EMAIL and re-tested it, and it worked without any issues.

gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
tknhs: No strong reason.
I moved `createCredentials` into it as a private method.

[`7f73905` (this PR)](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2999/commits/7f7390588abba1955a2803e5db3116016d881526)

gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
tknhs: Implemented `Supplier<OAuth2Credentials>` and take `ConfigProperties` in the constructor, as you suggested.
I also changed the implementation to reuse a single credentials instance.

[`7f73905` (this PR)](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2999/commits/7f7390588abba1955a2803e5db3116016d881526)

gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
tknhs: Updated as you suggested.

[`f252e91` (this PR)](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2999/commits/f252e9109c160498bbe23be90575dc9dbe1ecec5)

gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
tknhs: Changed it to log a warning and return access tokens instead of throwing.

[`f252e91` (this PR)](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2999/commits/f252e9109c160498bbe23be90575dc9dbe1ecec5) [`4da80db` (this PR)](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2999/commits/4da80db18a840560401bbca2cd1f0bdf51af0620)

tknhs:
@psx95
No worries at all, and thanks for taking another look.
I've addressed the additional comments as well.

gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
```diff
@@ -103,42 +111,120 @@ public class GcpAuthAutoConfigurationCustomizerProvider
...
-              try {
-                credentials = GoogleCredentials.getApplicationDefault();
-              } catch (IOException e) {
-                throw new GoogleAuthException(Reason.FAILED_ADC_RETRIEVAL, e);
-              }
-            }
-            return credentials;
-          }
-        };
+    AtomicReference<LazyCredentialsSupplier> supplierRef = new AtomicReference<>();
```
psx95: I am not sure if `AtomicReference` is required here. The customize method is typically only called once and there won't be multiple threads trying to call this at once. 
Additionally, the Supplier object is not going to be updated even if this method is called multiple times.

Any specific use-case for which you think AtomicReference is required?

gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
```diff
@@ -103,42 +111,120 @@ public class GcpAuthAutoConfigurationCustomizerProvider
...
+            Level.WARNING,
+            "{0} is set to {1} but {2} is not configured; falling back to access tokens."
+                + " Configure it by exporting environment variable {3} or system property {4}.",
+            params);
+        return applicationDefaultCredentials;
+      }
+      if (!(applicationDefaultCredentials instanceof IdTokenProvider)) {
+        logger.log(
+            Level.WARNING,
+            "The retrieved Application Default Credentials ({0}) cannot mint ID tokens; falling back to access tokens. Use a credential type implementing IdTokenProvider, for example the default service account of a GCP compute environment, a service account key, or impersonated credentials.",
```
psx95: nit: Mention the configuration in the message as well like the previous warning:

```java
logger.log(
            Level.WARNING,
            "{0} is set to {1} but the retrieved Application Default Credentials ({2}) cannot mint ID tokens; 
falling back to access tokens. Use a credential type implementing IdTokenProvider, for example the default service account of a GCP compute environment, a service account key, or impersonated credentials.", 
params[0],
paramas[1], 
applicationDefaultCredentials.getClass().getSimpleName());
```

gcp-auth-extension/src/test/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProviderTest.java
```diff
@@ -1167,6 +1173,142 @@ abstract static class Builder {
 
   // Mockito.lenient is used here because this method is used with parameterized tests where based
```
psx95: nit: This comment is misplaced now. It should be moved above the private method `prepareMockBehaviorForGoogleCredentials()`

gcp-auth-extension/src/test/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProviderTest.java
```diff
@@ -1167,6 +1173,142 @@ abstract static class Builder {
...
+  @Test
+  @SetSystemProperty(key = "google.auth.token.type", value = "unsupported_token")
+  @SetSystemProperty(key = "google.otel.auth.target.signals", value = "traces")
+  @SetSystemProperty(key = "google.cloud.project", value = "test-gcp-project")
+  void testUnsupportedTokenTypeFallsBackToAccessToken() throws IOException {
+    assertConfiguredCredentialsFallBackToAccessToken();
+  }
+
+  @SuppressWarnings("CannotMockMethod")
+  private void assertConfiguredCredentialsFallBackToAccessToken() throws IOException {
```
psx95: nit: Could you rename this method to `assertConfiguredCredentialsForTraceFallbackToAccessToken` - seeing as we are only verifying spans here. 

And adding a small comment might help too.

gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
tknhs: @psx95 
My intent behind this was to reuse a single credentials instance (and therefore a single token) across the span/metric/resource customizers, rather than minting and refreshing a separate token per exporter.

Since Supplier<OAuth2Credentials> has a no-arg get(), the ConfigProperties has to be a constructor field, and it is only available inside the customizer callbacks. So to keep a single shared instance I create it lazily on the first callback and reuse it through the holder. You're right that the atomicity isn't needed here (single-threaded, called once) — a plain holder would do; I only used AtomicReference as a convenient shared box.

Given that, which would you prefer?
- (a) Keep the single shared instance via a simple (non-atomic) holder, so one token is reused across exporters.
- (b) Construct a new LazyCredentialsSupplier(config) per callback (no holder) — simpler, but each exporter gets its own credentials/token.

psx95: So IIUC, the use of a Supplier (in the way it was [originally written](https://github.com/open-telemetry/opentelemetry-java-contrib/blob/bfef2a59eee8ae5899479125827884c9a402f8b2/gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java#L106-L121)) is already ensuring that a **single credentials object** is being used as long as the `customize` method is called once per lifetime of the extension.
 - The same supplier is passed to span, metrics and resource customizers - so it doesn't mint or refreshes different tokens per exporter. 
 - The only thing that should change here is the type of `Supplier<GoogleCredentials>` -> `LazyCredentialsSupplier`. In fact, come to think of it - if you compare the [original supplier used in the extension code](https://github.com/open-telemetry/opentelemetry-java-contrib/blob/bfef2a59eee8ae5899479125827884c9a402f8b2/gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java#L106-L121) to what you have in the `LazyCredentialsSupplier` in this PR, you'll notice that even the original supplier was supplying the credentials lazily and was reusing a single credentials object. 
*Slight tangent: I'd recommend to update the name to `LazyOAuth2CredentialsSupplier`.*

Using `AtomicReference` to ensure that a **single supplier object** (*and by extension, a single credentials object*) is used has a lot of overhead with not much benefit given the constraints in which this extension operates. 

Additionally, if *for some reason* the `customize` is called again,  you'd end up creating a new `AtomicReference` since the `AtomicReference` is defined as a method local variable.

I think the setup we had in the original code was simpler and worked. I think this is what you recommend in (a)? If so, I'd prefer (a). 

Let me know if I'm missing something here.

gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
psx95: I think this was missed ?

psx95 (changes requested):
Hey @tknhs , not sure if you saw the last round of reviews, but only a couple of comments left around the use of `AtomicReference`. 

Thanks!

tknhs:
@psx95 Thanks for the ping, and sorry for the delay. I was on a short vacation.

gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
tknhs: This is addressed in [`7b2c25f` (this PR)](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2999/commits/7b2c25fadd3485d59e7fc6110ffdc9f035abd91f)

gcp-auth-extension/src/main/java/io/opentelemetry/contrib/gcp/auth/GcpAuthAutoConfigurationCustomizerProvider.java
tknhs: Thanks, and agreed the original already reused a single credentials object.

One clarification on (a): as I framed it, (a) still keeps a small non-atomic holder to carry the single instance across the callbacks. That's lighter than the AtomicReference, but it's still a holder, whereas the original you liked had none. The original could avoid a holder because its get() was no-arg and needed no config, so it was constructed once up front and captured by the lambdas.

The id_token path needs ConfigProperties, which is only available inside the callbacks, so the way to match the original (single instance, no holder) is to let get take the config:

```java
LazyOAuth2CredentialsSupplier credentialsSupplier = new LazyOAuth2CredentialsSupplier();
// ...passed to all three customizers...

private static class LazyOAuth2CredentialsSupplier {
  @Nullable private OAuth2Credentials credentials;

  synchronized OAuth2Credentials get(ConfigProperties config) {
    if (credentials == null) {
      credentials = createCredentials(config);
    }
    return credentials;
  }
}
```

This is the closest to the original (single instance, no holder). The one trade-off is that it doesn't implement `Supplier<OAuth2Credentials>`, since get takes the config. Which would you prefer: this simpler original-style version (get(config), no holder), or keeping the Supplier<OAuth2Credentials> interface with the small non-atomic holder?

psx95: I see, thanks for the clarification! It makes sense now. 

I'm honestly not sure if one is better than the others in terms of performance, but I think keeping it close to the original one (using `get` with `ConfigProperties` param and letting go of implementing `Supplier` interface).

My rationale - 
 - Keeps the code simple and we could always choose to not call this class a "Supplier" so as to avoid confusion as to why this is not implementing `Supplier` interface. (E.g.,`LazyOAuth2CredentialHolder`).

tknhs: Sounds good, thanks.
I'll go with `get(ConfigProperties)` and `no holder`, and rename it away from "Supplier" to `LazyOAuth2CredentialsHolder` since it no longer implements the interface.

tknhs: Done. [`7c4a01b` (this PR)](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/2999/commits/7c4a01b5f0c0f34a660f0f5eae7c2f7d7c6d8168)
I've also re-verified the IAM-authenticated flow on Cloud Run.

psx95 (approved):

psx95:
@tknhs Thanks for the contribution!

psx95 (approved):

jaydeluca (approved):

PR #3040, opened 2026-08-06
[dynamic control] Http provider implementation

jackshirazi:
**Description:**

Implementing the HttpPolicyProvider

**Existing Issue(s):**

#2868

**Testing:**

Included

**Documentation:**

To be added

**Outstanding items:**

Docs need changing slightly to say http provider is implemented rather than a no-op, will do in a later PR

#2868


dynamic-control/src/main/java/io/opentelemetry/contrib/dynamic/policy/HttpPolicyProvider.java
```diff
@@ -0,0 +1,125 @@
...
+      new AtomicReference<>();
+  private final AtomicReference<Closeable> pollRegistration = new AtomicReference<>();
+
+  public HttpPolicyProvider(
+      URI endpoint,
+      SourceFormat format,
+      List<PolicySourceMappingConfig> mappings,
+      List<PolicyValidator> validators) {
+    this.endpoint = Objects.requireNonNull(endpoint, "endpoint cannot be null");
+    validateEndpoint(endpoint);
```
LikeTheSalad, line 43: nit: I know `validateEndpoint` is static, but still there's something about validating things by calling methods in the constructor that bothers me a little, I usually prefer factory methods to do the preparations for constructing an object.

dynamic-control/src/main/java/io/opentelemetry/contrib/dynamic/policy/HttpPolicyProvider.java
```diff
@@ -0,0 +1,125 @@
...
+    }
+    Closeable registration = PolicyProviderPoller.registerUrl(endpoint, this::reloadAndNotify);
+    if (!pollRegistration.compareAndSet(null, registration)) {
+      closeRegistration(registration);
+    }
+    return this::stopWatching;
+  }
+
+  /** Resets shared polling state used by tests. */
+  public static void resetForTest() {
```
LikeTheSalad, line 73: I'm guessing this method could help by having a single place for this object to get reset for tests, where we can add more stuff if needed later, and to serve as a style pattern for other provider implementations. It's just that it's a bit strange to see a static reset-object method.

dynamic-control/src/test/java/io/opentelemetry/contrib/dynamic/policy/HttpPolicyProviderTest.java
```diff
@@ -0,0 +1,190 @@
...
+    List<TelemetryPolicy> policies = latestPolicies.get();
+    assertThat(policies).hasSize(1);
+    TraceSamplingRatePolicy policy = (TraceSamplingRatePolicy) policies.get(0);
+    assertThat(policy.getProbability()).isEqualTo(0.25);
+    assertThat(policy.getSourceKind()).isEqualTo(SourceKind.HTTP);
+    watch.close();
+  }
+
+  /*
+   * TODO: Uncomment after full-policy JSON source support is merged.
```
LikeTheSalad, line 96: Maybe we need to keep this tracked in an issue?

LikeTheSalad (approved):
Cheers! Some minor comments but lgtm overall

dynamic-control/src/test/java/io/opentelemetry/contrib/dynamic/policy/HttpPolicyProviderTest.java
jackshirazi: There's already a tracking item in the meta issue to come back and enact all the TODOs

dynamic-control/src/main/java/io/opentelemetry/contrib/dynamic/policy/HttpPolicyProvider.java
jackshirazi: I've generally not used the factory pattern for this as it's a prototype rather than something that will move in-place into the SDK. When Telemetry policy becomes a spec and the SDK starts implementing, it will need factory methods, and SPIs for the policy implementations and more. I think we don't need that here. At least not unless people start needing these

dynamic-control/src/main/java/io/opentelemetry/contrib/dynamic/policy/HttpPolicyProvider.java
jackshirazi: agreed. the tests could call `PolicyProviderPoller.reset()` directly rather than this,. I've left it like this for maintenance so you think about it when amending the class, but I'm fine with removing it

LikeTheSalad: I think it makes sense to have it as a standard.

jaydeluca (approved):

PR #3073, opened 2026-08-25
Ensure RuleBasedRoutingSampler can use number valued attributes

rmannibucau:
**Description:**

Trying to use declarative configuration to remove spans on spring boot management server using port filtering it fails cause value is not a string.

**Existing Issue(s):**

-

**Testing:**

mainly unit tests


jaydeluca (approved):

jack-berg:
Saw this and opened https://github.com/open-telemetry/opentelemetry-java-contrib/issues/3099

rmannibucau:
@jack-berg AFAIK it has the same bug so an option can be to merge this one then back this one by the new one with the fix (todo) once released?

rmannibucau:
copied the test in the new module: https://github.com/open-telemetry/opentelemetry-java/pull/8783

rmannibucau:
can we get it in even if the extension supports it now? (for "legacy" dep support)?

PR #3091, opened 2026-09-02
Add docs for experimental-cassandra target system for jmx-scraper

jkoronaAtCisco:
**Description:**

Adds docs info about `experimental-cassandra` as a target system for the JMX scraper, inheriting the aligned Cassandra metric definitions from instrumentation rather than the legacy definitions embedded in jmx-scraper.

This completes the downstream half of the Cassandra JMX alignment: the definitions and the code-based compaction-progress handler were added upstream in opentelemetry-java-instrumentation [#19080](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/19080) and [#19290](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/19290), and shipped in the 2.31.1-alpha release that this repo now pins. Selecting `experimental-cassandra` resolves `jmx/rules/experimental-cassandra.yaml` directly from the instrumentation library jar — no local definitions are added here.


SylvainJuge:
I don't think we need to explicitly test `experimental-cassandra` here as it's already extensively tested in instrumentation.

Also, with https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/19783 the `experimental-cassandra` is being renamed to `cassandra`, I hope this lands in 2.32.0 instrumentation.

Anyway, we can still document the current `experimental-cassandra` target system, but this part of the documentation will have to be revised after that.

jkoronaAtCisco:
> I don't think we need to explicitly test `experimental-cassandra` here as it's already extensively tested in instrumentation.

Good point, I removed the test.

> Also, with [open-telemetry/opentelemetry-java-instrumentation#19783](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/19783) the `experimental-cassandra` is being renamed to `cassandra`, I hope this lands in 2.32.0 instrumentation.
> 
> Anyway, we can still document the current `experimental-cassandra` target system, but this part of the documentation will have to be revised after that.

Thanks @SylvainJuge, I wasn't aware of that effort.

SylvainJuge (approved):

jaydeluca (approved):

